PR #22, opened 2018-05-28
Py3 cleanup (-pandas, -unicodecsv)

uhhhuh:
Move away from pandas and unicodecsv to xlrd and csv respectively


pudo:
good work :)

PR #24, opened 2018-10-17
 Fix for GB HMT Sanctions crawler. 

paterry:
The stringify() call was setting '' values to None. This caused keys with '' values to be dropped further up the pipeline, resulting in KeyErrors for the pop() calls in the parse_entry() method.


pudo:
Urgh, that's probably messed up. Thanks for the fix. Are the empty values, in turn, messing anything else up?

PR #47, opened 2019-10-07
Update HMT Sanctions Data URL

Richard-Green-UK:
Update GB HMT Sanctions Data URL from the retired S3 storage to the new Azure Blob storage


pudo:
Verified this is a real thing: 

http://sanctionsnews.bakermckenzie.com/ofsi-publish-new-links-to-the-consolidated-list/

Thanks for the contribution, merging!

PR #59, opened 2020-10-27
Handle us_ofac data with unexpected feature or relation type IDs

dschulz-pnnl:
Encountered unexpected feature type ID 345, which caused the whole load to stop. This fix logs unexpected feature type or relation IDs in the same manner as REG_ID and then continues processing.


pxg:
Great work! I ran this and pushed a new dataset into our Aleph instance to test yesterday (5th November 2020), our original us_ofac2 dataset had 8,446 items, the new one has 8,880 items.

I also searched for [new additions on the sanctions list](https://home.treasury.gov/policy-issues/financial-sanctions/recent-actions/20201029) who were previously missing. "BAHREINI" now returns search results from Aleph on the new dataset.

pxg (approved):

PR #60, opened 2020-11-19
add eu_fsf crawler

thimios:
Added a crawler for Financial Sanctions Files (FSF) published by EU Commision


pudo:
Ok, this is really good. Thanks!

Issue #62, opened 2020-11-20
French sanction list


ilanoh:
Hello,

How can I contribute to integrate the french sanction list ?

It can be downloaded from this url in a zip that contains an excel file.

https://www.tresor.economie.gouv.fr/services-aux-entreprises/sanctions-economiques/tout-savoir-sur-les-personnes-et-entites-sanctionnees

sunu:
Hi @ilanoh, that looks like a useful source to scrape. We integrate individual sources by writing scrapers using the [Memorious](https://memorious.readthedocs.io/en/latest/) toolkit. Would you be interested in writing a scraper for this?

ilanoh:
Hello @sunu, Yes, I will check the documentation for Memorious, I never done it before, if I'm stuck I will come leave a message here.

sunu:
@ilanoh Awesome! If you need examples, take a look at these directories -- https://github.com/alephdata/opensanctions/tree/master/opensanctions/crawlers for the Python part of crawlers and https://github.com/alephdata/opensanctions/tree/master/opensanctions/config for the yaml configurations.

ang-st:
I'm also interested :)

@sunu I have already a scraper (not with memorius), the things is this list is published as a **xlsx** file (!). 
Is it ok to add dependencies like [openpyxl](https://openpyxl.readthedocs.io/en/stable/) ?  
I initially used pandas but we could stretch this down to  something with a lower footprint

sunu:
@ang-st We generally use `xlrd` for that. Here's an example from the Austrialian sanctions scraper: https://github.com/alephdata/opensanctions/blob/master/opensanctions/crawlers/au_dfat_sanctions.py#L70

ang-st:
@sunu thanks for the pointer, we'll start from there

sunu:
`xlrd` dropped `xlsx` support in the latest release. We'll have to use `openpyxl` as you suggested earlier.

ang-st:
Thanks for the info.

The "Gel des avoirs" listing will offer a JSON and XML API starting january
see [announcement](https://www.tresor.economie.gouv.fr/Articles/2020/12/07/refonte-du-registre-national-des-gels)

@TheophileDiot and I already checked data format and we will be able to update the P/R accordingly once the api is live

ilanoh:
@ang-st I have the json and xml file already

How can I share them ?

pudo:
[...]

PR #64, opened 2020-11-30
Add new feature France Freezing of Assets

TheophileDiot:
_**What is French's Freezing of Assets ?**_
Pursuant to Article L. 561-2 of the French Monetary and Financial Code, persons subject to **the fight against money laundering and terrorist financing** are required to implement without delay the asset freezes imposed by France. In addition, European regulations freezing EU own-account assets or resulting from United Nations resolutions are applicable by all natural and legal persons.

_**Why it's important to fetch this list ?**_
- It helps the fight against terrorist financing and money laundering.
- It prevent sanctioned people to have a banking account, etc...

**Changes :**

- Add **fr_freezing_of_assets.py** to **crawlers/**
- Add **fr_freezing_of_assets.yml** to **config/**
- Update **docs/_data/source.yml**

See also : #62, [French Freezing of Assets](https://www.tresor.economie.gouv.fr/services-aux-entreprises/sanctions-economiques/tout-savoir-sur-les-personnes-et-entites-sanctionnees)

### DIFF
--- a/docs/_data/sources.yml
+++ b/docs/_data/sources.yml
@@ -6,8 +6,8 @@
   category: sanctions
   publisher:
     title: United Nations Security Council
-    url: 'https://www.un.org/en/sc/'
-  url: 'https://www.un.org/sc/suborg/en/sanctions/un-sc-consolidated-list'
+    url: "https://www.un.org/en/sc/"
+  url: "https://www.un.org/sc/suborg/en/sanctions/un-sc-consolidated-list"
   data:
     url: https://scsanctions.un.org/resources/xml/en/consolidated.xml
     format: XML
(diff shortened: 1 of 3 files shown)


pudo:
One thing I might do, @TheophileDiot, is to remove the use of "None" as a string for missing values. Would this cause any issues on your end? In `followthemoney` we have `proxy.has()` which misbehaves when it's given such null-ish values.

TheophileDiot:
> One thing I might do, @TheophileDiot, is to remove the use of "None" as a string for missing values. Would this cause any issues on your end? In `followthemoney` we have `proxy.has()` which misbehaves when it's given such null-ish values. 

No problem, I left this in case you were using something to handle missing values differently.

PR #77, opened 2021-02-02
Skip notes on us_cia_world_leaders country pages

ChrisjanWust:
On an us_cia_world_leaders country page, people names are contained in `<p>` elements. However, some of these country pages, for example https://www.cia.gov/resources/government/vietnam/, also contain notes in `<p>` elements. These notes are the first element in the "content" key of "free_form_content", without any preceding headings.

The change simply adds another condition on which to skip the text inside.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ __pycache__/
 *.sqlite3
 .env
 .vscode
+.idea/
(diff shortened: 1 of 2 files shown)


pudo:
Oh very cool, I'd missed that!

PR #96, opened 2021-07-13
docker compose build --pull does not work

fastbone:
fixed it to docker-compose


pudo:
Thanks very much for the PR! For now it seems both syntaxes are supported, but they seem to slowly transitioning towards `docker compose`. Let's do `docker-compose` for now and switch later.

Issue #212, opened 2022-05-31
Incorrect assumption in us_ofac crawler? 


kmwang:
Recently we found a data quality issue in US OFAC datasets.

It seems that OpenSanctions data crawler assumes for all NamePartValue in a DocumentedName, each NamePartGroup only appears once at most. However this might not be the case. Take ofac-32179 as the example:

```
            <DocumentedName ID="32179" FixedRef="24474" DocNameStatusID="1">
              <DocumentedNamePart>
                <NamePartValue NamePartGroupID="63378" ScriptID="215" ScriptStatusID="1" Acronym="false">Tehrani</NamePartValue>
              </DocumentedNamePart>
              <DocumentedNamePart>
                <NamePartValue NamePartGroupID="63379" ScriptID="215" ScriptStatusID="1" Acronym="false">Sayyed</NamePartValue>
              </DocumentedNamePart>
              <DocumentedNamePart>
                <NamePartValue NamePartGroupID="63380" ScriptID="215" ScriptStatusID="1" Acronym="false">Mohammad</NamePartValue>
              </DocumentedNamePart>
              <DocumentedNamePart>
                <NamePartValue NamePartGroupID="63381" ScriptID="215" ScriptStatusID="1" Acronym="false">Ali</NamePartValue>
              </DocumentedNamePart>
              <DocumentedNamePart>
                <NamePartValue NamePartGroupID="63382" ScriptID="215" ScriptStatusID="1" Acronym="false">Haddadnezhad</NamePartValue>
              </DocumentedNamePart>
            </DocumentedName>
```
If you further refer to NamePartGroups, you could find 63380, 63381, and 63382 all belong to the same NamePartGroup - `Middle Name`. I think the middle name (in the final output) should be a concatenated string based on 63380, 63381 and 63382 instead of ONE of them (last write wins).

ref: https://github.com/opensanctions/opensanctions/blob/main/opensanctions/crawlers/us_ofac.py#L176-L178

```
          <NamePartGroups>
            <MasterNamePartGroup>
              <NamePartGroup ID="63378" NamePartTypeID="1520" />
            </MasterNamePartGroup>
            <MasterNamePartGroup>
              <NamePartGroup ID="63379" NamePartTypeID="1521" />
            </MasterNamePartGroup>
            <MasterNamePartGroup>
              <NamePartGroup ID="63380" NamePartTypeID="1522" />
            </MasterNamePartGroup>
            <MasterNamePartGroup>
[...]

PR #213, opened 2022-06-01
Fix opensanctions/opensanctions#212

kmwang:


pudo:
Looks good, thank you!

Issue #230, opened 2022-08-22
request for Dutch sanction list


janjjansen:
there are more national sanction lists that could be very useful.
I see belgian sanction list: be_fod_sanctions, and french sanction list: fr_tresor_gels_avoir

There is also a Dutch sanction list https://www.government.nl/documents/reports/2016/01/15/national-terrorism-list
in ods format. (https://www.government.nl/binaries/government/documenten/reports/2016/01/15/national-terrorism-list/eng-terrorismelijst.ods).

pudo:
That would be a cool addition, and the data looks simple but useable. I've written an ODS parser before here: 

https://github.com/alephdata/ingest-file/blob/main/ingestors/tabular/ods.py#L24-L73

Maybe we could use that library (`odf`) and even make an iterate-over-ODS-rows helper?

pudo:
Fixed in #234.

pudo:
> Can you not have sanctions on a person? It must be an entity?

No, a person is just a special sub-type of LegalEntity (that's just a broad term for people, companies etc.)

pudo:
It's live here: https://www.opensanctions.org/datasets/nl_terrorism_list/

PR #233, opened 2022-08-25
Draft: Source/nl national terrorism sanction

janjjansen:
This is a first try but i can't get it to work. 

See #230


pudo:
I'll trade you finishing it up for the FBI most wanted? ;)

PR #247, opened 2022-11-10
Added fbi most wanted

brammittendorff:


pudo:
This looks absolutely perfect! I may check the final loop on the table to see if there's a cleaner way of parsing things apart than splitting on `\n`, but that's roughly all my feedback on this. I'll merge it in now :)

Issue #248, opened 2022-11-11
Add "GB National Crime Agency Most Wanted"


brammittendorff:
Why?
We must implement the National Crime Agency Most Wanted list to enrich our Person database.

Resources:
```https://www.nationalcrimeagency.gov.uk/most-wanted-search```

brammittendorff:
Will start this one soon :smile_cat:

pudo:
I was just looking at the sheet and thinking “that would be a good one”….

pudo:
Very cool! Welcome to the party, @uberfresh, there's drinks and snacks and plenty of weird criminal figures to be scraped :)

uberfresh:
> Very cool! Welcome to the party, @uberfresh, there's drinks and snacks and plenty of weird criminal figures to be scraped :)

I'm very excited :laughing:

pudo:
Thank you, @uberfresh. Very nice work!

uberfresh:
No problem :smile:  @pudo

PR #250, opened 2022-11-23
Feature added gb nca most wanted

uberfresh:


uberfresh:
@pudo Hey can you review it ? :smiley: 
Issue: https://github.com/opensanctions/opensanctions/issues/248

pudo:
Looks really good, I'd be happy to merge. the most brittle thing is probably the `fix_label` function - is there maybe a more lightweight solution by selecting a more specific child element, or running it through a standardized func like `from normality import slugify`?

uberfresh:
> from normality import slugify

Hi @pudo 

These are the most specific child elements that change on every criminal page. So I can't select more specific but I have added 
`from normality import slugify` depending on your suggestion.

pudo:
Great! Let's do it :)

PR #262, opened 2023-01-24
feature_added-nl-most-wanted

ColinWaves:
Hey, 
I work for PESCHECK and I added a crawler for the most wanted list of the dutch police. If you could review it that would be amazing :)

In this PR persons are created with the values:
firstName, lastName,  name (wich includes alias from the website). When available the values birthDate, gender, nationality will also be added.
I set topics to 'crime'

Thanks!


pudo:
Heya! I had a bit of an edit of the code base, mainly to improve the handling of pagination of the listing page and to deal with the facts table without the counting indexes. Have a look if you like, keen for feedback!

pudo:
https://www.opensanctions.org/datasets/nl_most_wanted/

PR #264, opened 2023-01-29
Update jp_mof_sanctions.yml

etlel:
valid description of ministry


pudo:
Thanks, what a weird error we've had here!

PR #269, opened 2023-01-31
Update ch_seco_sanctions.py

etlel:
extract more metadata from generic tag


pudo:
This is really cool. Didn't even know you could use the walrus like that, btw.

PR #276, opened 2023-02-14
Edit crawler for last revision of ru_acf_bribetakers

etlel:


pudo:
Hey! Thanks for that PR. Maybe two notes before I test it out: 

a) The ACF list has some libellous content, which I've been trying to filter out. Need to make sure that filter still works against new versions. 
b) There seems to be an actual JSON API endpoint here now: https://sanctions.acf.international/ - maybe better to run against that than against the site which seems to have file names randomly generated by some CMS.

pudo:
Wow, thanks so much for doing that migration! I'll merge and review it now - but it looks very good!

PR #296, opened 2023-05-15
Add crawler for https://tender.gov.md/en/lista-de-interdictie

jbothma:
- Capturing owners using ownership
  - Capturing date as the sanction decision date
- Capturing "administrators" and "leaders" as members

Questions

- is it sufficient to assume an owner is a person unless their name contains s.r.l?
- Should we assume these are owners too, or only "administrators"?
- Is "administrator" a useful role label? I'm imagining they are the CEO except perhaps when there are multiple, in which case we can't make any assumption.
- should I generate a sanction ID like `md-interdictie-9-23` from `9/23 of 3 september 2023`?

### DIFF
--- a/opensanctions/crawlers/md_interdictie.py
+++ b/opensanctions/crawlers/md_interdictie.py
@@ -74,6 +74,7 @@ def crawl(context: Context):
         name = data.pop("denumirea-si-forma-de-organizare-a-operatorului-economic")
         entity.id = context.make_id(name, COUNTRY)
         entity.add("name", name)
+        entity.add("topics", "debarment")
         
         addr_string = data.pop("adresa-si-datele-de-contact-ale-operatorului-economic")
         address = h.make_address(context, full=addr_string, country=COUNTRY)
--- a/opensanctions/crawlers/md_interdictie.py
+++ b/opensanctions/crawlers/md_interdictie.py
@@ -1,9 +1,12 @@
 from lxml import html
 from normality import slugify, collapse_spaces
 from pantomime.types import HTML
+import re
 
 from opensanctions.core import Context
 from opensanctions import helpers as h
+
+REGEX_DELAY = re.compile(".+(\d{2}[\.\/]\d{2}[\.\/]\d{4}$)")
 
 IDX_ORG_NAME = 1
 IDX_ORG_ADDRESS = 2
@@ -28,7 +31,7 @@ def crawl(context: Context):
     for row in table.findall(".//tr"):
         if headers is None:
             headers = [c.text_content() for c in row.findall("./th")]
-            print(headers)
+
             # Assert that column order is as expected:
             # fail explicitly upon possible breaking change
             assert "Denumirea şi forma" in headers[IDX_ORG_NAME]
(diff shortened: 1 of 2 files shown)


opensanctions/crawlers/md_interdictie.py
```diff
@@ -77,27 +104,20 @@ def crawl(context: Context):
...
-        # Ownerships
-
-        # Memberships
-
-        
+        control_entities = parse_control(context, entity, decision_date, cells[IDX_ADMINS_FOUNDERS].text_content().strip())        
 
         context.emit(entity, target=True)
         context.emit(sanction)
+        for control_entity in control_entities:
```
pudo, line 119: Why not emit straight in `parse_control`?

opensanctions/crawlers/md_interdictie.py
```diff
@@ -115,15 +135,17 @@ def parse_date(text):
...
-        return match.group(1)
+        return match.group(1), parse_date(match.group(2))
     else:
-        context.log.warn(f'Failed to parse saction number from "{ text }"')
+        context.log.warn(f'Failed to parse saction number and date from "{ text }"')
+        return None, None
 
 
 def parse_address(context, text):
+    """Currently just the plain string as full address"""
```
pudo, line 148: If we do this, it'll probably live upstream (i.e. use libpostal). I'd just get rid of the function.

opensanctions/crawlers/md_interdictie.py
```diff
@@ -206,5 +204,38 @@ def parse_control(context: Context, text: str):
...
+
+
+def make_ownerships(context, company: Entity, date, owners: [str]) -> [Entity]:
+    for name in owners:
+        name = name.strip()
+        match = re.match("^([^\d\.\()]+)(\(?(\d+\.?\d*) ?%\)?)?$", name)
+        if match:
+            print(match.groups())
+            name = match.group(1)
+            person = context.make("Person")
```
pudo, line 224: Maybe this is a use case for `LegalEntity` to describe the fact that we're not 100% sure it's a person or organization?

opensanctions/crawlers/md_interdictie.py
```diff
@@ -37,6 +46,24 @@
...
+# Fondator/Administrator – 
+# or
+# Fondator - TOLMAŢCHI VALERI Administrator - TOLMAŢCHI VALERI
+# or
+# Lista conducătorilor - Galin Anatolie, Lista fondatorilor - Galin Anatolie
+# (List of leaders)
+# or
+# Conducători -
+# (Leaders)
+ROLES = {
```
pudo, line 58: FWIW, we're also getting the BO info from `md_companies`, so I'm not 100% sure we should go through the effort of parsing the ownership info here.

pudo (commented):
Some notes. One question: why not make the rows of the table into a dict with slug version of the column names? The index thing feels more likely to break if column names shift.

opensanctions/crawlers/md_interdictie.py
```diff
@@ -0,0 +1,241 @@
...
+        entity.add("name", name)
+        
+        address = parse_address(context, cells[IDX_ORG_ADDRESS].text_content())
+        h.apply_address(context, entity, address)
+
+        delay_until_date = parse_delay(context, cells[IDX_DELAY_UNTIL].text_content().strip())
+        start_date = delay_until_date or parse_date(cells[IDX_REGISTRATION_DATE].text_content().strip())
+        
+        sanction_num, decision_date = parse_sanction_decision(context, cells[IDX_DECISION_NUM_DATE].text_content().strip())
+        sanction = h.make_sanction(context, entity, sanction_num)
```
jbothma, line 108: use appropriate topic instead of sanction. Technically this isn't a sanction in international terms. It's just banned from public contracting.

opensanctions/crawlers/md_interdictie.py
jbothma: Just to be a bit more functional. On the other hand there's a convention of crawl_ also emitting and make_ returning. In this case it seems to make sense to change this to crawl_control and emit there

jbothma:
> Some notes. One question: why not make the rows of the table into a dict with slug version of the column names? The index thing feels more likely to break if column names shift.

The thinking was to have shorter, english keys rather than these super long romanian ones. But the dict approach is certainly more concise and just as safe.

opensanctions/crawlers/md_interdictie.py
jbothma: we said we'd keep the ownerships since they're done.

I added memberships for the rest but not adding their role now, since the way I'm scraping them doesn't leave me with the romanian term - they all become UNKNOWN or ADMINISTRATOR when they might be leader, etc.

Is that cool?

opensanctions/crawlers/md_interdictie.py
jbothma: was adding the debarment topic to the entity like we do with other instances of "debarment" sufficient? I thought I'd be replacing the sanction but I see the other debarments have a "sanction" too.

PR #309, opened 2023-06-12
Add common mappings for current issues

jbothma:


pudo:
This looks perfect, thanks so much for diving into it!

PR #311, opened 2023-06-19
Update Kazakh terrorism sanctions crawler to new source URL

jbothma:
Also use note instead of num for org id because num appears not to be stable in the new XML (was it n the old?)

TODO:

- [x] parse sanction start date from `включен от 15.12.2016`


pudo:
Really nicely done!

PR #327, opened 2023-07-20
A position entity helper

jbothma:
I think the code for turning an rupep position "Ministry of finance (the minister)" into "Minister of finance" should live closer to the crawler, since each source will have its own fun way of writing it.

It would have been nice if we could use the name and country to generate consistent Ids for different sources, but I'm actually not sure we should be producing consistent IDs, because we don't want entities from different sources to overwrite one another when one has more detail, perhaps position inception date, than another.

Whatever does that merging will need to look at a dataset of known positions or only upsert the properties it offers, and not clobber the others. 

Maybe we can have automated deduplication that says two positions are the same entity if
- they have the same name, e.g. "Minister of finance"
- same country
- same subnationalArea, null or otherwise
- same inception and dissolution date, null or otherwise
- same organization (by property comparison similar to this?)
- same wikidata id IFF they both have wikidata id, otherwise ignore for comparison

?

Merging that then ought to happen but isn't happening can then be fixed with better normalisation of names and subnationalArea strings.

Would that mean introducing another resolver.json sort of file for positions, read and replaced by the automation?


pudo:
This looks good to me, ready to merge. I wonder if we should also add a few more (optional) arguments which also make an Occupancy object? Or make a second helper which takes the position and the person entity, and then makes an occupancy?

jbothma:
Should this perhaps  outright use the wikidata ID as the ID if it has one, kind of like Person entities do by another mechanism?

> I wonder if we should also add a few more (optional) arguments which also make an Occupancy object? Or make a second helper which takes the position and the person entity, and then makes an occupancy?

I struggle a bit to reason about helpers like this before I try and create the entities in anger and see the common code I want to factor out. I'm happy to have a go and change later, but I'm not sure we're winning much over just digging in and writing the code actually creating the entities then seeing what's common.

jbothma:
for the moment I think a separate helper for occupancies is still best as a first round, but something around [this logic](https://github.com/opensanctions/opensanctions/pull/331/files#diff-91345f5c0dcfabcb931ad013c42878366b7247271d1997a59d42ef288271b29bR90-R95) will weigh in on where this goes as we make a decision on position classification and tagging

pudo:
I think it would be fun to merge this in soon, because after that we can rename `zavod.parse` to `zavod.helpers` and update all the crawler scripts. One feedback I still have on this is that we don't use camelCase python variable names anywhere else, so I'd advocate for making the arguments to this method into snake_case as well. Would that work?

jbothma:
That was silly, sorry.

Not testing language setting yet, but should be ready to merge now if you're happy for me to come back to that. Or I can add it today.

PR #331, opened 2023-07-24
Add US Congress crawler

jbothma:
Really early draft just to start playing with position and occupancy creation and discuss where different responsibilities should live

- The terms list seems to be the actual terms, more than the start and end date of them being a member spcifically. So leaving early isn't captured. Are we ok with this as an approximation for the occupancy start/end date?
-  Is it best to leave it just to reach the 1000 request/hr limit and fail with an error, and just let the cache allow it to complete after 3 runs?
    ```
    2023-07-24 07:23:26 [error    ] 429 Client Error: Too Many Requests for url: https://api.congress.gov/v3/member/T000260 
    [us_congress] dataset=us_congress response=<Response [429]> url=https://api.congress.gov/v3/member/T000260
    ```
- The current approach is going to keep only the terms which are within the inclusion period. But aren't more terms perhaps interesting if at least one term was relevant, to convey some degree of PEP-ness?

---

To do

- [ ] add other positions like president pro tempore from leadership field


opensanctions/crawlers/us_congress.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def fetch(context: Context, offset):
+    query = {"limit": LIMIT, "offset": offset}
+    url = f"{ context.source.data.url }?{ urlencode(query) }"
+    headers = {"x-api-key": API_KEY}
+    path = context.fetch_resource(f"members-{offset}.json", url, headers=headers)
+    context.export_resource(path, title=context.SOURCE_TITLE + f"offset {offset}")
+    with open(path, "r") as fh:
+        return json.load(fh)["members"], offset + LIMIT
```
jbothma, line 105: is this useful/worth it when we're actually just using it for getting IDs then making /member/id requests for the actual data? when do we bother exporting the resource and when don't we?

pudo: It's a bit fuzzy but the reason for keeping the exported resources is that we can reasonably write a script later to re-generate the data from it (or run a modified version of the crawler on historic data). For example, for the US sanctions list, we could relatively easy go and parse each version of the data released in the last 2 years to get fully accurate time stamps on entities. For a more complex crawler like this, it's really unlikely we'll ever be able to do time travel. We also don't want to show the 600+ resources in the dataset page on OpenSanctions. So I'd skip it.

pudo:
Regarding the "pro tempore": do we need this? To me this falls on the "pol sci oddity/interest" side of the utility divide, I don't think "pro tempore" senators will be subject to different AML requirements.

opensanctions/metadata/us_congress.yml
```diff
@@ -0,0 +1,29 @@
...
+        topics:
+          - role.pep
+          # - role.scope.nat
+          # - role.gov.legis
+      - match: Senate
+        name: Member of the United States Senate
+        topics:
+          - role.pep
+          # - role.scope.nat
+          # - role.gov.legis
```
jbothma, line 29: this is cute but pepdb has 121664 positions so this approach won't scale to that

pudo: Not with `yaml.safe_load` for sure :-P

opensanctions/crawlers/us_congress.py
```diff
@@ -0,0 +1,122 @@
...
+        parts = [person.id, position.id, start_date, end_date]
+        occupancy.id = context.make_id(*parts)
+        occupancy.add("holder", person)
+        occupancy.add("post", position)
+        occupancy.add("startDate", start_date)
+        occupancy.add("endDate", end_date)
+        if no_end_implies_current and not end_date:
+            status = "Current"
+        else:
+            status = "Ended"
```
jbothma, line 50: for this source this is fine, but the logic should actually be that it's only ended if there's an end date, and anything else is Unknown or Unsure or whatever we settle on

PR #333, opened 2023-07-25
Move exporters to zavod

jbothma:


pudo:
One note regarding the porting of the exporters: there's two types of metadata export right now - per dataset and global. 

The global exporter is really just a thing that collects the dataset indexes and combines them. It doesn't need to live in `zavod.exporters` and could be more appropriately stored in `zavod.tools`, a little module I've started for operational helper functions that can get exposed via the CLI.

legacy/opensanctions/tests/conftest.py
```diff
@@ -0,0 +1,20 @@
...
+from tempfile import mkdtemp, mkstemp
+
+from zavod import settings
+from zavod.meta import load_dataset_from_path, Dataset
+
+settings.DATA_PATH = Path(mkdtemp()).resolve()
+settings.RESOLVER_PATH = settings.DATA_PATH / "resolver.ijson"
+settings.ARCHIVE_BUCKET = None
+settings.CACHE_DATABASE_URI = None
+FIXTURES_PATH = Path(__file__).parent.parent.parent.parent / "zavod" / "zavod" / "tests" / "fixtures"
```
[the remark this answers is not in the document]
pudo: by soon, you mean in this PR, right? ;)

zavod/zavod/export/__init__.py
```diff
@@ -3,17 +3,16 @@
...
-from opensanctions.exporters.common import Exporter
-from opensanctions.exporters.ftm import FtMExporter
-from opensanctions.exporters.nested import NestedJSONExporter
-from opensanctions.exporters.names import NamesExporter
-from opensanctions.exporters.simplecsv import SimpleCSVExporter
-from opensanctions.exporters.senzing import SenzingExporter
-from opensanctions.exporters.statistics import StatisticsExporter
-from opensanctions.exporters.metadata import export_metadata, dataset_to_index
-from opensanctions.util import write_json
+from zavod.meta import Dataset, get_catalog
```
jbothma, line 6: I'm not sure I understand the cache thing yet - need to check this

zavod/zavod/tests/tools/test_export_index.py
```diff
@@ -0,0 +1,37 @@
...
+    dataset2 = load_dataset_from_path(DATASET_2_YML)
+    run_dataset(dataset2)
+    export(dataset2.name)
+
+    # Clear catalog as if this is a fresh process separate from the earlier exports
+    get_catalog.cache_clear()
+
+    collection = load_dataset_from_path(COLLECTION_YML)
+    export(collection.name)
+    export_index(collection)
```
jbothma, line 29: Is this horrid or reasonable?

My thinking here is that global index is particularly interesting when it's run for a collection, and a collection with more than one dataset. For it to find the datasets, they need to at least have index files in their data dirs. And I want issues too, so I ended up just running them and full-blown exporting them.

Earlier iterations tried to be more precise about what the prerequisites are but now I just want something working.

zavod/zavod/tests/test_exporters.py
```diff
@@ -0,0 +1,70 @@
...
+        assert len(list(read_entities(ftm))) == 11
+
+    with open(dataset_path / "index.json") as index_file:
+        index = load(index_file)
+        assert index["name"] == vdataset.name
+        assert index["entity_count"] == 11
+        assert index["target_count"] == 8
+        resources = {r["name"] for r in index["resources"]}
+        for r in expected_resources:
+            assert r in resources
```
jbothma, line 37: how clean do you want to go with the exporter tests? would I be wasting time if I do a method per class, each running the dataset and doing detailed assertions there?

Or is it reasonable at this stage just to make functions that test these files in as much detail as we like in this one test method?

zavod/zavod/tests/tools/test_export_index.py
pudo: I think this is really good, we should have way more `get_catalog.cache_clear()` in places :)

zavod/zavod/tests/test_exporters.py
pudo: I do think that having one overall test and one test that makes assertions on each format would be a good idea. This is the area of testing where we're most likely to catch customer-facing issues, so it's worth investing into.

zavod/zavod/tests/tools/test_export_index.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def test_export_index(vdataset: Dataset):
+    # Create dataset index files
+    context1 = Context(vdataset)
+    run_dataset(vdataset)
+    export(vdataset.name)
+
+    dataset2 = load_dataset_from_path(DATASET_2_YML)
+    run_dataset(dataset2)
```
jbothma, line 21: make testdataset1 and testdataset2 for consistency

zavod/zavod/tests/tools/test_export_index.py
```diff
@@ -0,0 +1,37 @@
...
+    export(collection.name)
+    export_index(collection)
+
+    with open(settings.DATA_PATH / "datasets" / "index.json") as index_file:
+        index = load(index_file)
+        datasets = {r["name"] for r in index["datasets"]}
+        assert "validation" in datasets
+        assert "test_dataset_2" in datasets
+
+    assert 1 == 2
```
jbothma, line 37: I still want to look at the warnings and make sure the test is valid

zavod/zavod/exporters/__init__.py
```diff
@@ -53,27 +52,28 @@ def export_data(context: Context, view: View):
...
         export_data(context, view)
 
-        # Export list of data issues from crawl stage
-        issues_path = context.get_resource_path("issues.json")
-        context.log.info("Writing dataset issues list", path=issues_path)
-        with open(issues_path, "wb") as fh:
-            issues = list(context.issues.all())
-            data = {"issues": issues}
-            write_json(data, fh)
+        write_issues(context)
```
jbothma, line 72: add test

zavod/zavod/tests/test_exporters.py
```diff
@@ -0,0 +1,70 @@
...
+from zavod.meta import Dataset
+from zavod.runner import run_dataset
+from zavod import settings
+from followthemoney.cli.util import read_entities
+from json import load, loads
+from csv import DictReader
+
+def test_export(vdataset: Dataset):
+    run_dataset(vdataset)
+    export(vdataset.name)
```
jbothma, line 13: also test export of a collection

jbothma:
make sure export timestamps are to the second, and date fields can be prefix dates

PR #340, opened 2023-08-02
Backfill tests

pudo:


zavod/zavod/tests/runtime/test_timestamps.py
```diff
@@ -26,3 +29,35 @@ def test_timestamps(vdataset: Dataset):
...
+    copyfile(
+        settings.DATA_PATH / "datasets" / vdataset.name / "statements.pack",
+        archive_path / "statements.pack",
+    )
+
+    settings.RUN_TIME = settings.RUN_TIME + timedelta(days=1)
+    settings.RUN_TIME_ISO = settings.RUN_TIME.isoformat(sep="T", timespec="seconds")
+    settings.RUN_DATE = settings.RUN_TIME.date().isoformat()
+    second_time = settings.RUN_TIME_ISO
+    run_dataset(vdataset)
```
jbothma, line 54: by dong this, we're

1. creating a context which creates a backfilled index (context, timestamps.build)
2. emitting new statements (nomenklatura entity merge)

jbothma: does this stuff make sense when we're testing this at the TimestampIndex level? if we're not asserting that last_seen is different, there isn't really any point in running the dataset a second time, is there?

zavod/zavod/archive.py
```diff
@@ -24,22 +37,96 @@
...
 @cache
-def get_archive_bucket() -> Optional[Bucket]:
-    if settings.ARCHIVE_BUCKET is None:
-        log.warn("No backfill bucket configured")
+def get_archive_backend() -> Optional[Backend]:
+    if settings.ARCHIVE_BACKEND is None:
+        log.info("No backfill backend configured.")
+        return None
+    try:
+        return cast(Backend, backends[settings.ARCHIVE_BACKEND]())
```
jbothma, line 118: I think this cast means we don't actually get type checking between the backends, but at least the protocol provides a bit of documentation. Not sure how to do this better right now. Nominal typing might be nicer in future

PR #346, opened 2023-08-07
Configurable exports

jbothma:


jbothma:
add tests for

- nonempty exporter config
- invalid exporter config

rename key to "exports"

PR #356, opened 2023-08-16
Add occupancy entities to cu_inventario_legislatura

jbothma:


datasets/cu/inventario_legislatura/crawler.py
```diff
@@ -61,5 +67,15 @@ def crawl(context: Context):
         entity.add("nationality", "cu")
         entity.add("topics", "role.pep")
 
+        occupancy = h.make_occupancy(
+            context,
+            entity,
+            position,
+            start_date=inception_date,
+            end_date=dissolution_date,
```
jbothma, line 75: make_occupancy will change these from current to ended in 2028

but @pudo can we assume this dataset is kept up to date and people who are no longer members are removed from the list? At least if they add an "ended" column we should get a warning from the audit.

pudo: No I think the page we're crawling is super specific to this congress. The reason for crawling it is also a bit diplomatic (they're friends). Maybe after that we should have a script for the parliament's web site?

jbothma:
One option then is to explicitly say their status is "unknown", or do so
say 1 year after initial publication?

jbothma:
@pudo how nasty would this be for this crawler?

```python

if settings.START_DATE > "2024-08-17":
    occupancy_status = OccupancyStatus.UNKNOWN
    context.log.warning("It's time to check if this page is getting updated")
else:
    occupancy_status = OccupancyStatus.CURRENT
```

pudo:
Oh, I actually really like this. I kind of wonder if this could be done via reading `context.dataset.coverage.end` and checking if the current date is much newer - that would make it generalisable?

jbothma:
I like that. Will do.

Issue #335, opened 2023-07-29
Sanctioned companies in Liechtenstein


brawer:
Would you accept a pull request for crawling this list? The sanctioned companies are banned from sending workers to Liechtenstein because of past violations of local law. The companies on this list might also be banned from public procurement but I am not a lawyer.

https://www.llv.li/de/unternehmen/berichtspflichten-bewilligungen/grenzueberschreitende-dienstleistungen/entsendung-von-arbeitnehmern/sanktionsliste-gemaess-art.-7-abs.-4-entsg

pudo:
Hi Sascha, thanks so much for offering to do a PR and this feels like an almost perfect first dataset for someone to take on. (The Swiss PDF will be a bit more of a hassle :/). One note on this: we are in the process of revamping our crawler system a bit at the moment, and moving things from `legacy/opensanctions/crawlers` to `datasets/`. The crawler-writing documentation is also likely to start moving from https://www.opensanctions.org/docs/opensource/crawlers/ to https://zavod.opensanctions.org/ in the next couple of days. Hope it's not too disorienting, it should be in a better place by next weekend at the latest.

PR #363, opened 2023-08-23
Add crawler for Liechtenstein Posted Workers Act sanctions

brawer:
Fixes https://github.com/opensanctions/opensanctions/issues/335


brawer:
Ah, in that case, I think adding it to the debarment collection makes sense. How to do this?

PR #365, opened 2023-08-24
Add Abgeordenetenwatch crawler for policitians and their mandates

aivuk:
At the moment just collects:

 * Politician fullname
 * Parliament period for politician mandate 
 * State from the mandate

### DIFF
--- a/datasets/de/abgeordnetenwatch/crawler.py
+++ b/datasets/de/abgeordnetenwatch/crawler.py
@@ -9,29 +9,6 @@ from zavod import helpers as h
 def crawl_politician(context: Context, politician_api_url: str):
     api_response = context.fetch_json(politician_api_url, cache_days=30)
     politician = api_response.pop("data")
-    context.audit_data(
-        politician,
-        ignore=[
-            "id",
-            "entity_type",
-            "api_url",
-            "abgeordnetenwatch_url",
-            "first_name",
-            "last_name",
-            "birth_name",
-            "sex",
-            "party",
-            "party_past",
-            "education",
-            "residence",
-            "occupation",
-            "statistic_questions",
-            "statistic_questions_answered",
-            "ext_id_bundestagsverwaltung",
-            "qid_wikidata",
-            "field_title",
-        ],
-    )
     return politician
 
 
@@ -60,20 +37,6 @@ def crawl(context: Context):
                 num_batches = math.ceil(batch_total_results / 100)
 
         for mandate in api_response.get("data"):
-            context.audit_data(
-                mandate,
-                ignore=[
-                    "id",
-                    "entity_type",
-                    "api_url",
-                    "id_external_administration",
-                    "id_external_administration_description",
-                    "type",
-                    "info",
-                    "electoral_data",
-                    "fraction_membership",
-                ],
-            )
             politician = mandate.pop("politician")
 
             # Get state from parliament info
@@ -114,6 +77,49 @@ def crawl(context: Context):
                 context.emit(person, target=True)
                 context.emit(position)
                 context.emit(occupancy)
+
+            # Audit mandate
+            context.audit_data(
+                mandate,
+                ignore=[
+                    "id",
+                    "entity_type",
+                    "api_url",
+                    "label",
+                    "id_external_administration",
+                    "id_external_administration_description",
+                    "type",
+                    "info",
+                    "electoral_data",
+                    "fraction_membership",
+                ],
+            )
+
+            # Audit
+            context.audit_data(
+                politician_data,
+                ignore=[
+                    "id",
+                    "entity_type",
+                    "label",
+                    "api_url",
+                    "abgeordnetenwatch_url",
+                    "first_name",
+                    "last_name",
+                    "birth_name",
+                    "sex",
+                    "party",
+                    "party_past",
+                    "education",
+                    "residence",
+                    "occupation",
+                    "statistic_questions",
+                    "statistic_questions_answered",
+                    "ext_id_bundestagsverwaltung",
+                    "qid_wikidata",
+                    "field_title",
+                ],
+            )
 
         # Increment batch number
         bi += 1
--- a/datasets/de/abgeordnetenwatch/crawler.py
+++ b/datasets/de/abgeordnetenwatch/crawler.py
@@ -9,12 +9,11 @@ from zavod import helpers as h
 def crawl(context: Context):
     # Fetch the source data URL specified in the metadata to a local path:
     api_response = context.fetch_json(context.dataset.data.url)
-    # TODO: Investigate why first query api_response["meta"]["result"]["total"] is 57726
-    # total_results = api_response["meta"]["result"]["total"]
-    total_results = 2733
+    total_results = api_response["meta"]["result"]["total"]
     num_batches = math.ceil(total_results / 100)
 
-    for bi in range(num_batches):
+    bi = 0
+    while bi < num_batches:
         context.log.info(
             "Get Politicians from range {} to {}".format(bi * 100, (bi + 1) * 100)
         )
@@ -22,16 +21,25 @@ def crawl(context: Context):
             api_response = context.fetch_json(
                 context.dataset.data.url, params={"range_start": bi * 100}
             )
+            batch_total_results = api_response["meta"]["result"]["total"]
+            if batch_total_results != total_results:
+                num_batches = math.ceil(batch_total_results / 100)
 
         for candidate in api_response.get("data"):
             parliament_info = candidate["parliament_period"]["label"].split(" ")
+            state = parliament_info[0]
+            inception_date = parliament_info[1]
+            dissolution_date = parliament_info[3]
+
             position = h.make_position(
                 context,
-                name="parliament_period",
+                name="Member of Parliament ({} - {}) - Germany/{}".format(
+                    inception_date, dissolution_date, state
+                ),
                 country="Germany",
-                subnational_area=parliament_info[0],
-                inception_date=parliament_info[1],
-                dissolution_date=parliament_info[3],
+                subnational_area=state,
+                inception_date=inception_date,
+                dissolution_date=dissolution_date,
             )
 
             person = context.make("Person")
@@ -40,3 +48,6 @@ def crawl(context: Context):
             h.make_occupancy(context, person, position, False)
             context.emit(person)
             context.emit(position)
+
+        # Increment batch number
+        bi += 1
--- a/datasets/de/abgeordnetenwatch/crawler.py
+++ b/datasets/de/abgeordnetenwatch/crawler.py
@@ -45,9 +45,13 @@ def crawl(context: Context):
             person = context.make("Person")
             person.id = context.make_id(candidate["politician"]["id"])
             h.apply_name(person, full=candidate["politician"]["label"])
-            h.make_occupancy(context, person, position, False)
-            context.emit(person)
-            context.emit(position)
+            occupancy = h.make_occupancy(
+                context, person, position, False, start_date=candidate["start_date"]
+            )
+            if occupancy:
+                context.emit(person, target=True)
+                context.emit(position)
+                context.emit(occupancy)
 
         # Increment batch number
         bi += 1
--- a/datasets/de/abgeordnetenwatch/crawler.py
+++ b/datasets/de/abgeordnetenwatch/crawler.py
@@ -61,6 +61,10 @@ def crawl(context: Context):
             birthYear = politician_data.pop("year_of_birth")
             person.add("birthDate", birthYear)
 
+            # Get WikiData ID
+            politician_wikidata_id = politician_data.pop("qid_wikidata")
+            person.add("wikidataId", politician_wikidata_id)
+
             # Apply politician fullname to Person entity
             h.apply_name(person, full=politician_fullname)
 
@@ -116,7 +120,6 @@ def crawl(context: Context):
                     "statistic_questions",
                     "statistic_questions_answered",
                     "ext_id_bundestagsverwaltung",
-                    "qid_wikidata",
                     "field_title",
                 ],
             )
--- a/datasets/de/abgeordnetenwatch/crawler.py
+++ b/datasets/de/abgeordnetenwatch/crawler.py
@@ -8,7 +8,7 @@ from zavod import helpers as h
 
 def crawl(context: Context):
     # Fetch the source data URL specified in the metadata to a local path:
-    api_response = context.fetch_json(context.dataset.data.url)
+    api_response = context.fetch_json(context.dataset.data.url, cache_days=30)
     total_results = api_response["meta"]["result"]["total"]
     num_batches = math.ceil(total_results / 100)
 
@@ -19,7 +19,9 @@ def crawl(context: Context):
         )
         if bi > 0:
             api_response = context.fetch_json(
-                context.dataset.data.url, params={"range_start": bi * 100}
+                context.dataset.data.url,
+                params={"range_start": bi * 100},
+                cache_days=30,
             )
             batch_total_results = api_response["meta"]["result"]["total"]
             if batch_total_results != total_results:
@@ -33,9 +35,7 @@ def crawl(context: Context):
 
             position = h.make_position(
                 context,
-                name="Member of Parliament ({} - {}) - Germany/{}".format(
-                    inception_date, dissolution_date, state
-                ),
+                name="Member of Parliament",
                 country="Germany",
                 subnational_area=state,
                 inception_date=inception_date,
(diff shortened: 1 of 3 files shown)


datasets/de/abgeordnetenwatch/crawler.py
```diff
@@ -0,0 +1,53 @@
...
+                inception_date=inception_date,
+                dissolution_date=dissolution_date,
+            )
+
+            person = context.make("Person")
+            person.id = context.make_id(candidate["politician"]["id"])
+            h.apply_name(person, full=candidate["politician"]["label"])
+            h.make_occupancy(context, person, position, False)
+            context.emit(person)
+            context.emit(position)
```
jbothma, line 50: Aha - the idea is that make_occupancy() will only return an occupancy entity if the position is within the allowable time period. Currently if it ended more than 5 years ago it won't return an Occupancy entity.

So you should check if an occupancy was returned, and only emit the person, position and occupancy if an occupancy was created.

aivuk: Thanks! Fixed on my last push and now the targets file contains both politicians and their positions

pudo:
Do we have any date of birth/place of birth info available in the source?

aivuk:
> Do we have any date of birth/place of birth info available in the source?

Yes, in another endpoint as https://www.abgeordnetenwatch.de/api/v2/politicians/180724 (the 180724 being the id already returned on the candidacies-mandates endpoint

aivuk:
@pudo I started coding the crawler to do an additional request to grab the birthday information but I have a question? What is the best way to add this additional source url in the yaml file?

pudo:
I don't think you need it in the YAML. If you can derive a URL from the first response and fetch that using `context.fetch_json(..., cache_days=30)` that should be perfectly OK.

aivuk:
> I don't think you need it in the YAML. If you can derive a URL from the first response and fetch that using `context.fetch_json(..., cache_days=30)` that should be perfectly OK.

ok, just did it as you said, makes sense. I was not sure if it was necessary to provide the full data provenance information in the yaml.

datasets/de/abgeordnetenwatch/crawler.py
```diff
@@ -0,0 +1,69 @@
...
+            inception_date = parliament_info[1]
+            dissolution_date = parliament_info[3]
+
+            position = h.make_position(
+                context,
+                name="Member of Parliament",
+                country="Germany",
+                subnational_area=state,
+                inception_date=inception_date,
+                dissolution_date=dissolution_date,
```
jbothma, line 42: It's best to remove these parameters. These parameters and properties are intended to represent the start of the existince of the position, more than the parliamentary term. So if you want to encode that you could be a member of landtag X from 1975, you could use this.  I'll try and clarify the docs to this effect.

We don't distinguish parliamentary terms right now, but we do capture when a person enters and leaves a position on the occupancy, which it looks like you've got at a glance.

datasets/de/abgeordnetenwatch/crawler.py
```diff
@@ -0,0 +1,69 @@
...
+            api_response = context.fetch_json(
+                candidate["politician"]["api_url"], cache_days=30
+            )
+            birthYear = api_response["data"]["year_of_birth"]
+            person.add("birthDate", birthYear)
+
+            # Get fullname
+            h.apply_name(person, full=candidate["politician"]["label"])
+            occupancy = h.make_occupancy(
+                context, person, position, False, start_date=candidate["start_date"]
```
jbothma, line 61: Consider using candidate.pop to access its members, and then context.audit(candidate) so that we automatically get warnings if new or unexpected fields show up, e.g. if they add end_dates. https://zavod.opensanctions.org/patterns/#detect-unhandled-data

aivuk: Just did it both suggestions. I'm not sure if I understood it correctly how the context.audit_data works because now I'm having a log of warnings from "Unexpected data found".

jbothma:
You can pass a list of field names which you want audit_data to explicitly
ignore.

If after that you still get warnings, maybe now, maybe in 6 months time, we
can see the data is different to what we expected and can either use the
new field, or explicitly ignore it too.

Warnings and errors end up at https://www.opensanctions.org/issues/

jbothma:
Try deleting the state directory in your data directory - probably
data/state I think.

Then export again.

I think there's a more official way but basically that's how I ensure the
export data is always completely regenerated.

In production this isn't an issue because it's always a fresh container.

datasets/de/abgeordnetenwatch/crawler.py
jbothma: Aha - I wasn't very clear.

Use `context.audit` once you've popped everything you feel you need from the dict. At that point you can see what fields remain and decide one by one whether you want its data.

e.g. after the following change I see there's an abgeordnetenwatch_url which would go nicely in Person.sourceUrl:

```diff
--- a/datasets/de/abgeordnetenwatch/crawler.py
+++ b/datasets/de/abgeordnetenwatch/crawler.py
@@ -9,7 +9,6 @@ from zavod import helpers as h
 def get_politician_data(context: Context, politician_api_url: str):
     api_response = context.fetch_json(politician_api_url, cache_days=30)
     politician = api_response.pop("data")
-    context.audit_data(politician)
     return politician
 
 
@@ -38,7 +37,6 @@ def crawl(context: Context):
                 num_batches = math.ceil(batch_total_results / 100)
 
         for mandate in api_response.get("data"):
-            context.audit_data(mandate)
             politician = mandate.pop("politician")
 
             # Get state from parliament info
@@ -57,6 +55,8 @@ def crawl(context: Context):
             person.id = context.make_id(politician.pop("id"))
             politician_fullname = politician.pop("label")
 
+            # abgeordnetenwatch_url -> sourceUrl
+
             # Get the politician birthday date
             context.log.info("Get Politician {} birth year".format(politician_fullname))
             politician_data = get_politician_data(context, politician.pop("api_url"))
@@ -75,5 +75,17 @@ def crawl(context: Context):
                 context.emit(position)
                 context.emit(occupancy)
 
+            context.audit_data(mandate, ignore=[
+                "id",
+                "entity_type",
+                "label",
+                "api_url",
+                "type",
+                "electoral_data",
+                "electoral_list",
+                "fraction_membership",
+            ])
+            context.audit_data(politician, ignore=["entity_type"])
+
```

datasets/de/abgeordnetenwatch/crawler.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def crawl(context: Context):
+    # Fetch the source data URL specified in the metadata to a local path:
+    api_response = context.fetch_json(context.dataset.data.url, cache_days=30)
+    total_results = api_response.pop("meta").pop("result").pop("total")
+    num_batches = math.ceil(total_results / 100)
+
+    bi = 0
+    while bi < num_batches:
```
jbothma, line 23: no need to change this now, but for future, our pattern is usually

def crawl(context): # loop over the bulk data
    for thing in data:
        crawl_thing(context, thing)

def crawl_think(...): # crawl _and_ emit the thing's data

that keeps the methods a bit smaller

datasets/de/abgeordnetenwatch/crawler.py
```diff
@@ -0,0 +1,79 @@
...
+            birthYear = politician_data.pop("year_of_birth")
+            person.add("birthDate", birthYear)
+
+            # Apply politician fullname to Person entity
+            h.apply_name(person, full=politician_fullname)
+
+            # Create occupancy
+            occupancy = h.make_occupancy(
+                context, person, position, False, start_date=mandate.pop("start_date")
+            )
```
jbothma, line 72: If the source is making a reasonable effort to add end_dates or remove mandates when the person is not a member any more, you can provide the value `True` to [no_end_implies_current]( https://zavod.opensanctions.org/helpers/#zavod.helpers.make_occupancy:~:text=required-,no_end_implies_current,-bool) and add mandate.pop("end_date"). Then the occupancies without an end date get the status `current`.

datasets/de/abgeordnetenwatch/crawler.py
```diff
@@ -0,0 +1,79 @@
...
+            politician = mandate.pop("politician")
+
+            # Get state from parliament info
+            parliament_info = mandate.pop("parliament_period").pop("label").split(" ")
+            state = parliament_info[0]
+
+            # Create position
+            position = h.make_position(
+                context,
+                name="Member of Parliament",
```
jbothma, line 51: If I make name f"Member of the {state} Parliament", I get nice positions like this:

```json
    "positionOccupancies": [
      {
        "id": "de-aw-193679c9e97d9b0d555fa03321246b50828f5b87",
        "caption": "Occupancy",
        "schema": "Occupancy",
        "properties": {
          "post": [
            {
              "id": "de-aw-5270d360f089c5c49848b20c805f66d0caaf35ca",
              "caption": "Member of the Berlin Parliament",
              "schema": "Position",
              "properties": {
                "subnationalArea": [
                  "Berlin"
                ],
                "name": [
                  "Member of the Berlin Parliament"
                ],
                "country": [
                  "de"
                ]
              },
              "referents": [],
              "datasets": [
                "de_abgeordnetenwatch"
              ],
              "first_seen": "2023-08-25T11:39:04",
              "last_seen": "2023-08-25T11:39:04",
              "last_change": "2023-08-25T11:39:04",
              "target": false
            }
          ],
          "startDate": [
            "2023-05-03"
          ],
          "status": [
            "current"
          ],
          "holder": [
            "de-aw-0407b462289daa638b2be909d14ef07f4e4bb35a"
          ]
        },
```

Does this API contain information about the parliaments, so you can generate names like `Member of the Abgeordnetenhaus of Berlin` and `Member of the Landtag of Baden-Württemberg`?

Otherwise just including the state name like I did is fine.

datasets/de/abgeordnetenwatch/crawler.py
aivuk: Just added the end_date and change no_end_implies_current, thanks!

datasets/de/abgeordnetenwatch/crawler.py
aivuk: I couldn't find this extra information even in another endpoint, I changed to  "Member of the {state} Parliament"

datasets/de/abgeordnetenwatch/crawler.py
aivuk: just changed it to crawl_politician.

datasets/de/abgeordnetenwatch/crawler.py
aivuk: Thanks for clarifying, just put all the fields that I'm not using with pop on ignore and now running the crawler I don't have any unexpected field warning anymore.

datasets/de/abgeordnetenwatch/crawler.py
```diff
@@ -0,0 +1,125 @@
...
+            # Get state from parliament info
+            parliament_info = mandate.pop("parliament_period").pop("label").split(" ")
+            state = parliament_info[0]
+
+            # Create position
+            position = h.make_position(
+                context,
+                name="Member of the {} Parliament".format(state),
+                country="Germany",
+                subnational_area=state,
```
jbothma, line 51: It looks like this data includes members of the Bundestag and members of the EU parliament so we can't give all of them a subnational_area, and we can't assume parliament_info is a state.

Let's exclude the EU parliament - we already have them via the official source so duplication isn't necessary.

Let's keep the Bundestag. Only provide a value to subnational_area when the position is for a state. You can provide None for the Bundestag.

I think it's worth making a request to the parliamentary_period api_url and via that to the parliament api_url and get the `label_external_long`

Then if you drop ` Parliament` from the position name, you'll get nice standard positions like "Member of the Bundestag" or "Member of the Bürgerschaft Bremen" automatically

If you make sure those requests are cached you'll hit the cache a lot.

datasets/de/abgeordnetenwatch/crawler.py
```diff
@@ -0,0 +1,125 @@
...
+                    "sex",
+                    "party",
+                    "party_past",
+                    "education",
+                    "residence",
+                    "occupation",
+                    "statistic_questions",
+                    "statistic_questions_answered",
+                    "ext_id_bundestagsverwaltung",
+                    "qid_wikidata",
```
jbothma, line 119: ah wikidata QID is a high value field for us - if you put that in the `wikidataId` property, we won't need to de-duplicate entities that already exist in the system with such an ID.

datasets/de/abgeordnetenwatch/de_abgeordnetenwatch.yml
```diff
@@ -0,0 +1,21 @@
...
+  On abgeordnetenwatch.de (which translates as “member of parliament watch”)
+  users can find all their representatives and ask them public questions, read
+  our investigative researchs and support our demands towards politics for more
+  transparency.
+entry_point: crawler
+prefix: de-aw
+publisher:
+    name: Parlamentwatch e.V.
+    organization: Parlamentwatch e.V.
+    authority: Deutscher Bundestag
```
jbothma, line 14: I don't see good documentation for this field other than the tutorial so I suspect this is for when the publisher is a unit within an official authority, which is not the case here. Or have you seen documentation to the contrary? I'd suggest we delete it for now while I sort out the metadata documentation.

```suggestion
```

jbothma (changes requested):

datasets/de/abgeordnetenwatch/de_abgeordnetenwatch.yml
```diff
@@ -0,0 +1,21 @@
+name: de_abgeordnetenwatch
+title: "abgeordnetenwatch"
+url: https://www.abgeordnetenwatch.de
+description: >
+  On abgeordnetenwatch.de (which translates as “member of parliament watch”)
+  users can find all their representatives and ask them public questions, read
+  our investigative researchs and support our demands towards politics for more
+  transparency.
```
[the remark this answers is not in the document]
aivuk: Just did it

datasets/de/abgeordnetenwatch/de_abgeordnetenwatch.yml
aivuk: > Or have you seen documentation to the contrary?

No, I just assumed that the publisher would be the responsible to the data source url. Your suggestion is to delete the full section `publisher` or just `authority`. I got confused by the comment and github 'Suggested change'.

datasets/de/abgeordnetenwatch/crawler.py
aivuk: Done in my last push

datasets/de/abgeordnetenwatch/de_abgeordnetenwatch.yml
jbothma: Just authority. The rest of your publisher is good

datasets/de/abgeordnetenwatch/crawler.py
```diff
@@ -0,0 +1,125 @@
...
+            h.apply_name(person, full=politician_fullname)
+
+            # Create occupancy
+            occupancy = h.make_occupancy(
+                context,
+                person,
+                position,
+                True,
+                start_date=mandate.pop("start_date"),
+                end_date=mandate.pop("end_date"),
```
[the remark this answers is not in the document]
aivuk: Just pushed an update getting the start data from the parliament api. I have a question about the data audit. Currently I'm, doing it in the end of the main crawler, but considering that I'm doing requests for 4 different endpoint, would it make sense to add the audi for each individual api endpoint crawler?

datasets/de/abgeordnetenwatch/crawler.py
aivuk: I'm getting the `label_external_long` as you suggested now and using subnational_are as None when the parliament is from the Bundestag. I'm also not getting the EU parliament anymore.

datasets/de/abgeordnetenwatch/crawler.py
```diff
@@ -0,0 +1,125 @@
...
+            context.audit_data(
+                mandate,
+                ignore=[
+                    "id",
+                    "entity_type",
+                    "api_url",
+                    "label",
+                    "id_external_administration",
+                    "id_external_administration_description",
+                    "type",
```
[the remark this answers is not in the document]
aivuk: I think this is the only point that it's not so clear to me. How would I filter only the mandates were the politician was elected?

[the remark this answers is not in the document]
jbothma:
```
curl https://www.abgeordnetenwatch.de/api/v2/candidacies-mandates|jq .|less
```
includes results like
Screenshot_2023-09-01_13-43-39

but if I understand correctly, that doesn't mean the person was elected - only that they stood for election.
According to the docs, 
```
curl https://www.abgeordnetenwatch.de/api/v2/candidacies-mandates?type=mandate|jq .|less
```
should give you only Mandat which I think is when they actually were elected to the position
Screenshot_2023-09-01_13-45-01

jbothma (commented):
Just some tweaks I'll make when merging

datasets/de/abgeordnetenwatch/crawler.py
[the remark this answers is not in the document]
aivuk: thanks, I got it now

PR #369, opened 2023-08-25
Liechtenstein Posted Workers Act: Emit full name of sanctioned persons

brawer:
After this change, [this list](https://www.opensanctions.org/search/?scope=li_entsg&schema=Person) should display full person names. Arguably this is a presentation issue, but other crawlers emit full names in addition to first and last names, so it seems best to be consistent across data sources.


pudo:
There's also a helper called `h.apply_name` which will do everything needed: set firstName, lastName and mint a full name.

brawer:
Thank you! Please take another look.

PR #373, opened 2023-08-25
Distrust future occupancy end dates if they are beyond the dataset coverage end

jbothma:
I'm not totally sure about this yet - this results in `last_change` becoming fixed to a past date once `coverage.end` is reached, even though it might have changed. But perhaps if `make_occupancy` emitted a warning it's an ok compromise for now?

I think I'd feel better about a dedicated field in the metadata like `coverage.verification_due` or something, or hardcoding the date in the crawler until we see more instances where this is relevant.


pudo (approved):
All makes sense to me!

Issue #364, opened 2023-08-24
Cayman PEPs


jbothma:
Ripe for automation:

- [Members of the House (legislativeassembly.ky) elected](http://www.legislativeassembly.ky/portal/page/portal/lglhome/members/elected-2021-2025) and the [previous term](http://www.legislativeassembly.ky/portal/page/portal/lglhome/members/previousmembers/2017-2021-Members)
- Under the Judicial Admin section has the various Judges: [Cayman Islands Judicial & Legal Website | An official website of the Cayman Islands Government – Cayman Islands Judicial & Legal Information Website](https://www.judicial.ky/judicial-administration)

Perhaps less ready for automation:

- Board of Directors for the regulator which is a statutory Authority [The Board of the Cayman Island Monetary Authority â€“ CIMA](https://www.cima.ky/the-board)
- List of government statutory authority and Government Companies: [Statutory Authorities | Cayman Islands Government](https://www.gov.ky/directory/statutory-authorities-and-government-companies)

aivuk:
Questions:

> [Members of the House (legislativeassembly.ky) elected](http://www.legislativeassembly.ky/portal/page/portal/lglhome/members/elected-2021-2025) and the [previous term](http://www.legislativeassembly.ky/portal/page/portal/lglhome/members/previousmembers/2017-2021-Members)

We have only the following information:
  * Name 
  * Position (e.g. Premier)
  * Full position (e.g. Premier and Minister for Sustainability and Climate Resiliency and Finance & Economic Development Member of Parliament for Newlands)

you want for all positions? Speaker of Parliament, Premier, Deputy Premier, Cabinet Ministers, Ex Officio Members, and the scrolling down on the first link Elected Members > Government Backbench Members, Opposition Members and Independent Opposition Member. 

And we would add the Occupancy for the positions above terms.

>  Under the Judicial Admin section has the various Judges: [Cayman Islands Judicial & Legal Website | An official website of the Cayman Islands Government – Cayman Islands Judicial & Legal Information Website](https://www.judicial.ky/judicial-administration)

Are you interested only in the judges on https://www.judicial.ky/judicial-administration/judges, both current as retired ones? Which information to get? The only information that everyone have is the full name. Some have the nationality, some have the birth year, but there is no standard in how this information is displayed. Some are in a block of text and some are in a bullet list.

[...]

PR #382, opened 2023-09-03
Add crawler for Cayman Islands Legislative Assembly

aivuk:
Related issue: #364

The following crawler gets the elected members from the Cayman Island Legislative Assembly. The first attempt was to parse the html using xmltree, but the html is malformed as you can see in the snippet below on the `<h3><li>` mixup instead of `<li><h3>`:

```
                        <li>
                            <h3>Hon. Andre M. Ebanks, MP,</h3> Minister for Financial Services &amp; Commerce and Minister for Investment, Innovation &amp; Social Development and Member of Parliament for West Bay South
                        </li>
                        <h3>
                            <li>Hon. Sabrina Turner, MP,
                        </li></h3> Minister for Health &amp; Wellness and Member of Parliament for Prospect
                        
                        <li>
                            <h3>Hon. Johany S. "Jay" Ebanks, MP,</h3> Minister for Planning, Agriculture, Housing, &amp; Infrastructure and Member of Parliament for North Side
                        </li>
```
 
The code convert the html page to text using the method .text_content(), parse its content and deals with details as:

 * there is almost a pattern in the positions of a member. Sometimes the different positions are separated by 'and', sometimes by space as "Premier and Minister for Sustainability and Climate Resiliency and Finance & Economic Development Member of Parliament for Newlands''.
 * most of the Minister information with multiple areas have a & as separator "Minister for Tourism & Ports", but there are exceptions as "Minister for Sustainability and Climate Resiliency"
 *  removes some additional information from the member's names as 'MP', 'Hon. ', etc

### DIFF
--- a/datasets/ky/legislativeassembly/crawler.py
+++ b/datasets/ky/legislativeassembly/crawler.py
@@ -61,7 +61,8 @@ def crawl(context: Context):
                 # position if it exists.
                 if i == len(words) - 1:
                     new_pos += [w]
-                p += [" ".join(new_pos).removesuffix(" and").strip()]
+                full_new_pos = " ".join(new_pos).removesuffix(" and").strip()
+                p += [full_new_pos]
                 new_pos = [w]
             else:
                 # Add word to position description
@@ -106,7 +107,7 @@ def crawl(context: Context):
         for position in person["positions"]:
             pos = h.make_position(
                 context,
-                name=position,
+                name=position if "Member" not in position else "Member of Parliament",
                 country="Cayman Islands",
             )
             occupancy = h.make_occupancy(
--- a/datasets/ky/legislativeassembly/crawler.py
+++ b/datasets/ky/legislativeassembly/crawler.py
@@ -23,12 +23,24 @@ def crawl(context: Context):
 
     # After the following part we are next to the legislative positions
     initial_pad = page_text.find("You are here")
+    if initial_pad < 0:
+        context.log.warning(
+            "Couldn't find the initial block of content in the page. Maybe the page HTML changed."
+        )
     page_text = page_text[initial_pad:]
 
     # Find the indexes in the text were each position is written
     positions_idx = []
     for pos in positions:
-        positions_idx += [(pos, page_text.find(pos))]
+        position_in_page = page_text.find(pos)
+        if position_in_page < 0:
+            context.log.warning(
+                "Couldn't find the block of content for '{}' in the page. Maybe the page HTML changed.".format(
+                    pos
+                )
+            )
+        else:
+            positions_idx += [(pos, position_in_page)]
 
     def get_positions(s):
         """
@@ -120,6 +132,7 @@ def crawl(context: Context):
             )
             person_positions += [(pos, occupancy)]
 
+        person_proxy.add("sourceUrl", context.dataset.data.url)
         context.emit(person_proxy, target=True)
         for pos in person_positions:
             context.emit(pos[0])
--- a/datasets/ky/legislativeassembly/crawler.py
+++ b/datasets/ky/legislativeassembly/crawler.py
@@ -1,0 +1,128 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from zavod import Context
+from zavod import helpers as h
+
+
+def crawl(context: Context):
+    api_response = context.fetch_html(context.dataset.data.url, cache_days=30)
+
+    # The list of positions sections that we are interested
+    positions = [
+        "Speaker of Parliament",
+        "Premier",
+        "Deputy Premier",
+        "Cabinet Ministers",
+        "Ex Officio",
+    ]
+
+    # The HTML is malformed and it's easier to parse the text content
+    # instead of the HTML
+    page_text = api_response.text_content()
+
+    # After the following part we are next to the legislative positions
+    initial_pad = page_text.find("You are here")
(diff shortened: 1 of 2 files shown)


datasets/ky/legislativeassembly/crawler.py
```diff
@@ -0,0 +1,128 @@
...
+
+                persons += [new_person]
+
+    for person in persons:
+        person_proxy = context.make("Person")
+        h.apply_name(person_proxy, full=person["name"])
+        # TODO: Make a better ID
+        person_proxy.id = "legislativeassembly-{}".format(
+            person["name"].replace(" ", "_")
+        )
```
jbothma, line 106: You could use https://zavod.opensanctions.org/context/#zavod.context.Context.make_id which should automatically generate `ky-la`-prefixed valid IDs for you.

These warnings are due to capitals in the current IDs 
```
2023-09-04 12:38:09 [warning  ] Rejected property value        [zavod.entity] dataset=ky_legislativeassembly entity_id=ky-la-fa6e409ac684162830bbc491b591a1666feeaef3 prop=holder value=legislativeassembly-Wayne_Panton
```

datasets/ky/legislativeassembly/ky_legislativeassembly.yml
```diff
@@ -0,0 +1,25 @@
...
+    name: The Cayman Islands Legislative Assembly 
+    description: |
+      The House of Parliament of the Cayman Islands is a unicameral Legislature
+      comprising 21 Members, 19 of whom are the Elected Representatives for the
+      Islands' 19 constituencies: one each from West Bay Central, West Bay North,
+      West Bay South, West Bay West, George Town Central, George Town East, George
+      Town North, George Town South, George Town West, Prospect, Red Bay, Bodden
+      Town East, Bodden Town West, Newlands, Savannah, North Side, East End, Cayman
+      Brac East, Cayman Brac West & Little Cayman.
+    acronym: ky-legislative
```
jbothma, line 20: You can drop acronym - I think the acronymn would more accurately be CILA but I can't see that being a common term actually used by anyone and adding value

jbothma:
very nice!

What might be nice, if it's not a lot of work, is if the Member of parliament... position was only named `Member of parliament` without the constituency, just to limit the explosion of position names we'll be generating. The constituency name isn't important here.

Also add the source page as sourceUrl then users can easily go and see the original page if they like.

Finally, I'm not really sure how this will fail, but I wonder if there's a check you can add so that if parsing fails partially, we'd get alerted by warning logs. Complete failure would become a warning because it would generate no entities. But if some entities are generated, and then some weird text gets interpreted as positions, that's something we should try to actively guard against and rather er on the side of emitting context.log.warning(s) and outputting less data, than outputting broken data.

aivuk:
> What might be nice, if it's not a lot of work, is if the Member of parliament... position was only named `Member of parliament` without the constituency, just to limit the explosion of position names we'll be generating. The constituency name isn't important here.

I did change as you requested.
 
> Also add the source page as sourceUrl then users can easily go and see the original page if they like.

I added this to the persons entities. Do you want me to add to all entities (position and occupancies)?

> Finally, I'm not really sure how this will fail, but I wonder if there's a check you can add so that if parsing fails partially, we'd get alerted by warning logs. Complete failure would become a warning because it would generate no entities. But if some entities are generated, and then some weird text gets interpreted as positions, that's something we should try to actively guard against and rather er on the side of emitting context.log.warning(s) and outputting less data, than outputting broken data.

I added some sanity checks, if the initial marker that I'm using in the page is not found (the text "You are here") or if some of the positions blocks ("Premier", "Deputy Premier", etc). For checking if a weird text was interpreted as position I can only think in doing some pattern check, like testing if the position text is some of the list `positions` or if it's starts with "Member " or "Minister for ". I don't know if it's worthwhile  to create regex to test for valid "Minister for", what do you think?

jbothma:
This looks great.

Last think I spotted before merging, in the page I count 21 people but only 9 targets in the data.

datasets/ky/legislativeassembly/ky_legislativeassembly.yml
```diff
@@ -0,0 +1,25 @@
+name: ky_legislativeassembly
+title: "legislativeassembly"
```
jbothma, line 2: Cayman Islands Legislative Assembly

jbothma: woops, this was meant to be a "suggestion"

aivuk:
> This looks great.
> 
> Last think I spotted before merging, in the page I count 21 people but only 9 targets in the data.

Because you asked to not get them all here https://github.com/opensanctions/opensanctions/issues/364#issuecomment-1697244802 Or maybe I misunderstood you?

> no need to capture back bench, ex officio, oppostion, independent, etc

datasets/ky/legislativeassembly/ky_legislativeassembly.yml
aivuk: done

aivuk:
> Oh sorry, I meant we don't need to be able to distinguish whether someone is "backbench MP" from "ex oficio MP" - we're happy just calling them all "Member of parliament" if that's the title they have beside their name.
> 
> I totally see why you understood what you did from what I said.

ok, more work that I was expecting, but now in my last commit I'm getting all 21 persons from the legislative page. I also simplified the positions from "Member of Parliament for ..." to just "Member of Parliament".

PR #386, opened 2023-09-06
Occupify wd peps

jbothma:
This is looking a bit weird because it now really brings focus to a single position (member of naitonal assembly, status Unknown) for Thabo Mbeki who was president of South Africa about 20 years ago.

We keep the position because we have a date of birth and no end date disqualifying the position (same as old wd_pep logic) and mark him as PEP.

the wikidata enricher then adds all his positions, I think (I should check its logic) so when you look at his positions property, it doesn't look odd. But it will look odd if he only has a single position, and he clearly hasn't held it in ages.

Once we extend national positions AFTER_OFFICE to 20 years, his presidency will also be there and things will look much more sane.

Screenshot_2023-09-07_22-39-40

### DIFF
--- a/datasets/eu/cor_members/crawler.py
+++ b/datasets/eu/cor_members/crawler.py
@@ -41,7 +41,7 @@ def crawl_person(context: Context, item: Dict[str, Any]) -> None:
     )
     if occupancy is not None:
         for value in item.pop("memberStatuses", []):
-            status = STATUS[value]
+            status = STATUS[value].value
             occupancy.add("status", status)
         context.emit(position)
         context.emit(occupancy)
--- a/datasets/eu/cor_members/crawler.py
+++ b/datasets/eu/cor_members/crawler.py
@@ -2,7 +2,7 @@ from typing import Any, Dict
 
 from zavod import Context
 from zavod import helpers as h
-from zavod.helpers.positions import OccupancyStatus
+from zavod.logic.pep import OccupancyStatus
 
 FORMATS = ("%d/%m/%Y",)
 
--- a/zavod/zavod/logic/pep.py
+++ b/zavod/zavod/logic/pep.py
@@ -1,0 +1,69 @@
+from enum import Enum
+from typing import Optional
+from datetime import datetime
+
+from zavod.context import Context
+from zavod import settings
+from zavod import helpers as h
+
+YEAR = 365  # days
+AFTER_OFFICE = 5 * YEAR
+AFTER_DEATH = 5 * YEAR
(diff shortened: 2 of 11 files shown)


zavod/zavod/logic/pep.py
```diff
@@ -0,0 +1,71 @@
+from enum import Enum
+from typing import Optional
+from datetime import datetime
+
+from zavod.context import Context
+from zavod import settings
+from zavod import helpers as h
```
jbothma, line 7: I don't think this loop is nice - and I don't think anything else is using backdate so I think it's best moved here til something else actually needs it.

pudo: What would be needed for us to be able to make different cut-off dates for subnational and national PEPs in the `wd_peps` importer? I'd love to go 20 years on national and international peps again.

jbothma: one more argument - scope. I was tempted to tackle that in this PR thought I'd ship this first and then do it.

jbothma:
@pudo a consequence of the current implementation is that all wikidata positions without an end date will have status "unknown" which maybe more explicitly casts doubt about the accuracy of the data than we intent. A lot of the wikidata data is up to date. That's also why I labeled it "Unsure" in some of the mockups.

One option is to see if we can pass through an edit date, and mark things edited in the last year, say, as current. 

But I think that complicates understanding of the status property and it's maybe better to just document a clear methodology for how we decide whether blank end dates mean current or unknown for a given source and now document how the occupancy data should be used by data users.

Happy with the latter?

zavod/zavod/logic/pep.py
jbothma: @pudo would you keep any national PEP from any source for 20 years? My thinking was basically that any source that's confident about the scope of the position can pass it along, and any PEP where we're not sure about the scope is only kept according to the current rules.

datasets/eu/cor_members/crawler.py
```diff
@@ -2,7 +2,7 @@
 
 from zavod import Context
 from zavod import helpers as h
-from zavod.helpers.positions import OccupancyStatus
+from zavod.logic.pep import OccupancyStatus
```
jbothma, line 5: Let's just double-check what the output here looks like after these changes before merging

jbothma: Still two conflicting statuses, but uses value now instead of __repr__ so it's a pareto improvement.

Screenshot_2023-09-08_08-18-34

PR #412, opened 2023-09-27
Added docker compose v2

brammittendorff:
Docker Compose V2 (often referred to as "Compose V2") is a major version upgrade from the original Docker Compose (V1). Compose V2 was introduced as a technical preview in late 2020 and became generally available in 2021.

The significant change in Compose V2 is its integration into the Docker CLI. This means that instead of using the standalone `docker-compose` command, you can use `docker compose` (without the hyphen) as a subcommand of the Docker CLI. This change aimed to provide a more seamless and integrated experience for users.

So we converted the Makefile so that both users can use opensanctions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,7 @@
-DOCKER=
+SHELL := /bin/bash
+# Check if 'docker-compose' is available, if not, use 'docker compose'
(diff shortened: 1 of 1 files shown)


pudo:
Uh this is really nice, thank you!

PR #414, opened 2023-09-28
Create consistent position identifiers between us_cia_world_leaders and us_cia_world_factbook

jbothma:
Screenshot_2023-09-28_14-23-50


pudo:
This smells a bit but it gets the job done. I guess adding the key_prefix in there in the first place is the bug, but backing out of that is a bit difficult.

jbothma:
Very leaky :(

PR #436, opened 2023-10-13
New source ddtc debarments

jbothma:


jbothma:
just confirming it's ok to assume "violated us law" -> us entity and "sanctioned for china forced labour -> china entity

PR #439, opened 2023-10-17
New source ddtc enforcements

jbothma:


jbothma:
@pudo are there any useful topics I can add this to? Not sure violating export controls fits under crim.fin.

ok to add them to the crime collection anyway?

they're not actually all sanctioned (mostly not, it seems) and historical data is included, so I'm not sure the sanction topic or collection make sense.

is it useful to add the link to each PDF as sourceUrl to each sanction even if we can't label the links?

pudo:
I'd add them to `crime`, and maybe even add `crime.traffick` to the topics for each. And definitely yes on the sourceUrl PDFs :)

datasets/us/ddtc_enforcements/crawler.py
```diff
@@ -40,22 +42,45 @@
...
     entity.add("country", "us")
+    entity.add("topics", "crime.traffick")
 
     description = row.pop("description")["value"]
     sanction = h.make_sanction(context, entity, description)
     sanction.add("reason", description)
     sanction.add("listingDate", row.pop("year")["value"])
     sanction.add("program", PROGRAM)
 
+    sanction.add(
```
pudo, line 67: Maybe only the charging letter?

PR #452, opened 2023-11-02
new source: Colombian public servants

jbothma:


datasets/co/funcion_publica/co_funcion_publica.yml
```diff
@@ -0,0 +1,125 @@
+title: Colombian Administrative Department of Public Service PEP Declarations
+entry_point: crawler.py
+prefix: co-cedula
```
jbothma, line 3: nope, then this applies to positions etc too

pudo: Aaaargh. Sometimes a simple string function is better than the framework :)

pudo:
Looks great!

PR #457, opened 2023-11-09
Check positions in position db

jbothma:
part of https://github.com/opensanctions/operations/issues/180

Adds the categorise() method for crawlers to get categorisation information from the positions database to determine if a position is a PEP position and ought to be emitted. Positions that are not yet in the database are added and crawlers can supply an initial is_pep decision and topics if known.

Also adds the annotations enricher which fetches categorisation information from the positions database and emits any topics found for the position.


zavod/zavod/helpers/positions.py
```diff
@@ -24,10 +74,16 @@ def make_position(
...
+    is_pep: Optional[bool] = True,
+) -> Optional[Entity]:
+    """Creates and returns a Position entity if is_pep is True or if it's
+    confirmed to be a PEP position in the positions database. Otherwise returns
+    None.
+
+    Also ensures the position exists in the positions database for further
+    categorisation.
+
+    Topics in the positions database override topics supplied to this helper.
```
jbothma, line 86: actually need to keep the annotation topics separate so that the enricher can add them with the correct provenance

zavod/zavod/helpers/positions.py
```diff
@@ -64,11 +121,22 @@ def make_position(
...
+    if not annotation.is_pep:
+        context.log.info(f"Position {country} {name} is not a PEP position.")
+        return
+
     position.add("name", name, lang=lang)
     position.add("summary", summary, lang=lang)
     position.add("description", description, lang=lang)
     position.add("country", country)
-    position.add("topics", topics)
+    position.add("topics", annotation.topics)
```
jbothma, line 139: revert

PR #470, opened 2023-12-01
Adding crawler for disreputable and suspended companies from Brazil

bgmello:
The crawler extracts information about sanctions applied to companies and individuals that have been sanctioned by the Brazilian government and aren't allowed to participate in public procurement.

Data is fetched from the following official website:
https://portaldatransparencia.gov.br/download-de-dados/ceis

### DIFF
--- a/datasets/br/suspended_companies/br_ceis.yaml
+++ b/datasets/br/suspended_companies/br_ceis.yaml
@@ -1,0 +1,31 @@
+title: Brazil National Register of Disreputable and Suspended Companies
+entry_point: crawler.py
+prefix: brazil-ceis
+coverage:
+  frequency: daily
+deploy:
+  schedule: "0 */9 * * *"
+load_db_uri: ${OPENSANCTIONS_DATABASE_URI}
+summary: >
+  The National Register of Ineligible and Suspended Companies (CEIS) lists individuals and companies sanctioned from public procurement in Brazil.
+description: |
+  The CEIS is an essential part of Brazil's Portal da Transparência,
+  providing a comprehensive list of individuals and companies that have faced sanctions restricting
+  their ability to participate in public tenders or contract with the public administration.
+  These sanctions include temporary suspension, declaration of ineligibility,
+  and other penalties under various Brazilian laws like the Bidding Law and the Anti-Corruption Law.
+
+  The CEIS not only promotes transparency for citizens but also serves
+  as a crucial reference for public managers in governmental procurement processes, 
+  ensuring compliance at all federal levels. 
+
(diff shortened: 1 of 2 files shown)


datasets/br/suspended_companies/br_ceis.yaml
```diff
@@ -0,0 +1,31 @@
+title: Brazil National Register of Disreputable and Suspended Companies
+entry_point: crawler.py
+prefix: brazil-ceis
```
pudo, line 3: We want the resulting IDs to be as short as possible, so I'd go for `br-ceis`

datasets/br/suspended_companies/crawler.py
```diff
@@ -0,0 +1,111 @@
+from zavod import Context
+from zavod import helpers as h
+from zipfile import ZipFile
+import csv
+import io
+import re
+from typing import List
+
+SITE_URL = "https://portaldatransparencia.gov.br/download-de-dados/ceis"
```
pudo, line 9: I would use `context.data_url` instead (which is `data.url` in the YAML metadata!) because it means that if the metadata is outdated, the code breaks...

datasets/br/suspended_companies/crawler.py
```diff
@@ -0,0 +1,111 @@
...
+    Fetches the CSV file as a zip from the website decompresses it and parses it using the csv library
+    and returns the data as a list of dicts.
+
+    :param csv_url: The URL of the CSV file.
+    :param context: The context object.
+
+    :return: The data fetched from the website as a list of dicts.
+    """
+    # The header is necessary because the website blocks requests without a user agent.
+    response = context.fetch_response(csv_url, headers=HEADERS)
```
pudo, line 53: We also have a download function, but for 2MB this is great.

datasets/br/suspended_companies/crawler.py
```diff
@@ -0,0 +1,111 @@
...
+            entity.id = context.make_id(raw_entity['CPF OU CNPJ DO SANCIONADO'])
+            entity.add('name', raw_entity['NOME DO SANCIONADO'])
+            entity.add('country', 'br')
+            entity.add('taxNumber', raw_entity['CPF OU CNPJ DO SANCIONADO'])
+
+            context.emit(entity, target=True)
+
+        sanction = h.make_sanction(context, entity)
+        sanction.add("program", "Brazil disreputed and sanctioned companies")
+        sanction.add("country", "br")
```
pudo, line 94: You can also set `country` here by adding it to the `publisher` section in the metadata.

pudo (approved):
This is incredibly nicely done, and ready to merge. Thank you for taking the time to study the zavod docs carefully and using the framework as intended :)

PR #474, opened 2023-12-07
Brazilian PEPs

bgmello:
This PR introduces the Brazilian dataset of Politically Exposed Persons (PEPs), officially published by the Brazilian government. The dataset encompasses approximately 150,000 individuals who have occupied positions in the Brazilian government over the past five years. This timeframe aligns with the legal definition of a PEP in Brazil.

### DIFF
--- a/datasets/br/pep/br_pep.yaml
+++ b/datasets/br/pep/br_pep.yaml
@@ -1,0 +1,26 @@
+title: Brazil Politically Exposed Persons
+entry_point: crawler.py
+prefix: br-pep
+coverage:
+  frequency: daily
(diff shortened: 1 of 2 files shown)


jbothma:
very nice, thank you!

I made a couple of small tweaks - just caching the requests to make development a bit quicker, and changed to only add entities who are considered PEPs according to our our automated and possibly manual categorisation. 

I see there's a column defined in the data dictionary as "_End date of the period in which the person is considered politically exposed
_" but it's always 5 years after the end date of the position, so I've just stuck to our logic. That means we'll keep people for 5 years after holding the position, or longer if we categorise them as national or intergovernmental PEPs.

PR #480, opened 2023-12-12
Migrate pepdb to zavod framework and positions DB

jbothma:
- [x] countries are discovered and alerted if not yet listed in yaml
- [x] positions candidates are discovered and added to position db with no pep decision
- [x] position holders are found for positions marked as PEP positions and emitted if they meet our criteria
- [ ] a way to migrate ignore decisions from pepdb to positions db  

About 0.5% reduction in targets

Explanations for fewer entities:

- PEPDB had repetitions of holderships across multiple edits of the item. This emits the state of the item as returned by the query, which basically generally represents the correct final state of the item after those edits.
  - in addition to occupancies disappearing, we see a position like https://www.opensanctions.org/entities/Q104003163/ disappearing because the only holder according to wd died in 1920, and the holder we have now does not have that statement any more.
  - Q104528417 redirects to https://www.wikidata.org/wiki/Q64399432
- changes to positions which were previously returned in our queries but aren't any more:
  - potential remedy: change `?position wdt:P31 wd:Q4164871` to `wdt:P31*/wdt:279*` - that would identify `ambassador of jordan` but not `ambassador of jordan to france` because the latter doesn't have the country. We could try and get the country from the superclasses

- losing https://www.wikidata.org/wiki/Q104408266 because they had position https://www.wikidata.org/wiki/Q61928772 but now only have position https://www.wikidata.org/wiki/Q122346833
  - https://www.wikidata.org/wiki/Q122346833 should potentially also be a PEP position - it's a subnational legislator position

local run vs yesterday's prod file
```
jq . /home/jdb/projects/opensanctions/opensanctions/contrib/delta/data/delta/20231213/wd_peps.delta.json | grep -A4 REMOVED |grep schema|sort | uniq -c
  26594     "schema": "Occupancy",
   2059     "schema": "Person",
    168     "schema": "Position",
```


jbothma:
1. run crawler in main branch

> wikidata positions with decision and holders now exist marked is_pep=true

2. run crawler in this branch

> wikidata positions without holders now exist with no decision

```
 (main) $ psql postgres://saas:saas@localhost/saas
psql (15.4, server 13.7 (Debian 13.7-1.pgdg110+1))
Type "help" for help.

saas=# select is_pep, count(*) from position where dataset = 'wd_peps' group by is_pep;
 is_pep | count  
--------+--------
        | 111587
 t      |  14355
```

PR #482, opened 2023-12-15
Smoke test datasets

jbothma:
- [x] number of entities with a given country
- [x] number of entities with a given schema
- [x] number of countries
- [x] tests


datasets/ky/parliament/ky_parliament.yml
```diff
@@ -26,3 +26,33 @@ publisher:
...
     url: https://parliament.ky/members/
     format: HTML
+
+assertions:
+  - comparison: GT
+    threshold: 100
+    action: WARN
+    filter:
+      attribute: schema
+      value: Person
```
jbothma, line 36: to support totally different metrics like number of countries, we could rather do

```yaml
assertions:
  - metric: entity_count
    comparison: GT
    threshold: 100
    action: WARN
    filter:
      attribute: schema
      value: Person

  - metric: country_count
    comparison: GT
    threshold: 123
    action: WARN
```

datasets/_wikidata/peps/wd_peps.yml
```diff
@@ -46,6 +46,49 @@ data:
   url: https://data.opensanctions.org/contrib/pepdb/holders.csv
   format: CSV
 
+assertions:
+  - metric: entity_count
+    comparison: gt
+    threshold: 160000
+    filter:
+      attribute: schema
+      value: Person
```
pudo, line 55: Looking at this DSL, I totally see why it's very flexible and expressive. But there's a little devil inside of me that's wondering if this isn't a bit smarter than it needs to be, and if we couldn't have something like: 

```yaml
assertions:
  min_entities: 100
  max_entities: 10000
  min_schema:
      Company: 70
      Person: 30
```

This could internally compile down to the same `Assertion` thing, it just feels a little less like it wants to be a framework and more like a human-readable set of constraints....

pudo (commented):
I think the stats exporter is both the logical and also a very weird place for this to live. That shouldn't stop us from merging and deploying this, but it's something we should keep as a note to self for another day.

Left one note regarding the DSL and if we should simplify it a bit.

datasets/_wikidata/peps/wd_peps.yml
jbothma: yeah - as a human, parsing 5 lines to decide what the expectation is isn't as much fun as it is for a computer

jbothma:
> I think the stats exporter is both the logical and also a very weird place for this to live. That shouldn't stop us from merging and deploying this, but it's something we should keep as a note to self for another day.

Yeah - the meta package didn't quite feel like the right place for the evaluation logic, and since the current assertions are very dependent on the current statistics data structure I think it's important that it sits close by. But going beyond what we're already doing for statistics will be a good forcing function to find a better home for it.

pudo:
Long-term, I wonder if there should be a separate "test" or "evaluate" phase to the pipeline. It would build the store and basically run the stats exporter, then give a go-ahead or cancel on whether to do the actual export...

jbothma:
I think that would be a nice way to bring a go/no-go decision in. I'm not sure it should then necessarily have that much to do with the statistics exporter, but an intermediate step makes sense

pudo (approved):

PR #488, opened 2023-12-29
support customizing Log's time zone through environment variables

jinrenjie:
Hello, thank you very much for your contribution to this project!

When I was using this project, I found that I specified the time zone through the `TZ` environment variable, but the output log was still in the UTC time zone!


pudo:
Thanks, this is a good reminder that we need to start adding more complex time support to the whole stack, and also begin to deprecate `utcnow()` as it's beginning to be phased out with 3.12.

PR #491, opened 2024-01-04
AR parliament

anna-parfenova:
Rationale:
Argentina has recently seen a change in government and our information about their political personnel is outdated. One thing we need to scrape is https://www.hcdn.gob.ar/diputados/ - the list of members of parliament for the country. 

Changes:
- added basic metadata to describe the new dataset
- added crawler.py that crawles the parliament website and creates Person, Occupancy and Position for each member of parliament and emits it.

Test run:
Screenshot 2024-01-05 at 11 30 05

☝️THIS IS A DRAFT. After review:
- remove excessive logging in the crawler
- review and finalise the metadata file

### DIFF
--- a/datasets/ar/parliament/crawler.py
+++ b/datasets/ar/parliament/crawler.py
@@ -2,6 +2,7 @@ from lxml import html, etree
 from datetime import datetime
 from zavod import Context
 from zavod import helpers as h
+from zavod.logic.pep import categorise
 
 
 def crawl_person(context: Context, element) -> dict:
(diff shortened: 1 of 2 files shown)


datasets/ar/parliament/crawler.py
```diff
@@ -0,0 +1,158 @@
...
+        )
+        person.add("sourceUrl", person_data["url"])
+        person.add("political", person_data["block"])
+
+        # Create position and occupancy
+        position = h.make_position(
+            context,
+            name="Member of National Congress of Argentina",
+            country="ar",
+            subnational_area=person_data.get("district"),
```
jbothma, line 111: Am I right in thinking district is their consitituency, rather than their jurisdiction? In that case, I think it's best not to put district in subnational_area, since it's intended to describe their jurisdiction. I totally understand the confusion though.

jbothma: I'll do that in a quick push in a bit

jbothma:
Thank you for this!

PR #495, opened 2024-01-08
Make syncing positions optional and document

jbothma:


pudo (approved):
Looks good - my only question is if this should be enabled by default in order to ease people using it locally.

jbothma:
> should be enabled by default in order to ease people using it locally.

I erred towards production failing more obviously if it's missing the variable to enable syncing, versus ease of onboarding.

I think ZAVOD_RESOLVER_PATH is another env var new installs need before they can fully run crawlers locally https://github.com/opensanctions/opensanctions/issues/505

I don't have a strong preference though - happy to go either way here.

zavod/docs/usage.md
```diff
@@ -21,3 +21,22 @@ $ zavod publish  --latest datasets/_global/icij_offshoreleaks/icij_offshoreleaks
...
+```
+
+Then run the exporter with `--clear` to ensure the latest statements are included in the output:
+
+```bash
+$ zavod export --clear ...```
+
+## Environment variables
+
+Zavod can be configured using a range of environment variables defined in `zavod.settings`. These are most-easily set during development in a `.envrc` file by installing [direnv](https://direnv.net/).
```
jbothma, line 42: Move to https://github.com/opensanctions/opensanctions/blob/main/zavod/docs/install.md?plain=1#L60

PR #500, opened 2024-01-11
Initialize PEP crawler for Uruguay

atehe:
Adds

-  Person: id number, name
- Organization: name
- Positions: name, country, language

### DIFF
--- a/datasets/uy/pep/crawler.py
+++ b/datasets/uy/pep/crawler.py
@@ -1,0 +1,55 @@
+from zavod import Context
+from typing import Dict, List, Optional
+from pantomime.types import XLSX
+import openpyxl
+from normality import slugify
+from zavod import helpers as h
+from zavod.logic.pep import categorise
+
+
+def sheet_to_dicts(sheet):
+    headers: Optional[List[str]] = None
+    for row in sheet.rows:
+        cells = [c.value for c in row]
+
+        if headers is None and all(cells):
+            headers = [slugify(h) for h in cells]
+            continue
+
+        if headers:
+            row = dict(zip(headers, cells))
+            yield row
+
+
+def parse_sheet_row(context: Context, row: Dict[str, str]):
+    person = context.make("Person")
+
+    id_number = row.pop("c-i")
+    person_name = row.pop("nombre")
+    role = row.pop("cargo")
+
+    person.id = context.make_slug(id_number, prefix="ur-cedula")
+    person.add("idNumber", id_number)
+    person.add("name", person_name)
+
+    organization_name = row.pop("organismo")
+
+    organization = context.make("Organization")
+    organization.id = context.make_slug(organization_name, prefix="ur-cedula")
+    organization.add("name", organization_name)
+
+    position = h.make_position(
+        context, role, country="uy", organization=organization, lang="spa"
+    )
+
+    context.emit(person, target=True)
+    context.emit(position)
+
+
+def crawl(context: Context):
+    path = context.fetch_resource("uy_pep_list.xlsx", context.data_url)
+    context.export_resource(path, XLSX, title=context.SOURCE_TITLE)
+    rows = sheet_to_dicts(openpyxl.load_workbook(path, read_only=True).worksheets[0])
+
+    for row in rows:
+        parse_sheet_row(context, row)
--- a/datasets/uy/pep/crawler.py
+++ b/datasets/uy/pep/crawler.py
@@ -31,15 +31,7 @@ def parse_sheet_row(context: Context, row: Dict[str, str]):
     person.add("idNumber", id_number)
     person.add("name", person_name)
 
-    organization_name = row.pop("organismo")
-
-    organization = context.make("Organization")
-    organization.id = context.make_slug(organization_name, prefix="uy-cedula")
-    organization.add("name", organization_name)
-
-    position = h.make_position(
-        context, role, country="uy", organization=organization, lang="spa"
-    )
+    position = h.make_position(context, role, country="uy", lang="spa")
 
     context.emit(person, target=True)
     context.emit(position)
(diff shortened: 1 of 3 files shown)


datasets/uy/pep/crawler.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def parse_sheet_row(context: Context, row: Dict[str, str]):
+    person = context.make("Person")
+
+    id_number = row.pop("c-i")
+    person_name = row.pop("nombre")
+    role = row.pop("cargo")
+
+    person.id = context.make_slug(id_number, prefix="ur-cedula")
```
jbothma, line 31:
```suggestion
    person.id = context.make_slug(id_number, prefix="uy-cedula")
```

Let's base the prefix on the [country code](https://www.opensanctions.org/reference/#type.country)

datasets/uy/pep/crawler.py
```diff
@@ -0,0 +1,55 @@
...
+    person.add("idNumber", id_number)
+    person.add("name", person_name)
+
+    organization_name = row.pop("organismo")
+
+    organization = context.make("Organization")
+    organization.id = context.make_slug(organization_name, prefix="ur-cedula")
+    organization.add("name", organization_name)
```
jbothma, line 39: I see why you were tempted to add these, but we don't currently make organization entities for every PEP position  - did put it into the data model but we've preferred not to create the entities yet. Organization entities are currently largely used for sanctioned organizations, or those linked to people with some higher crime risk, so I think we want to think a bit carefully before we go all out on adding all possible organizations in governments.

Could you remove this for now?

datasets/uy/pep/crawler.py
```diff
@@ -0,0 +1,55 @@
...
+
+    position = h.make_position(
+        context, role, country="uy", organization=organization, lang="spa"
+    )
+
+    context.emit(person, target=True)
+    context.emit(position)
```
jbothma, line 46: Did you notice in targets.nested.json that the positions aren't there? Don't worry, the third milestone where you add Occupancy entites will link people to their positions, which will result in the positions being populated nested in the person objects.

datasets/uy/pep/ur_pep.yml
```diff
@@ -0,0 +1,14 @@
+name: uy_pep
+title: "Politically Exposed Persons (PEP)"
+url: https://www.gub.uy/
+description: placeholder
+entry_point: crawler.py
+prefix: ur-pep
```
jbothma, line 6: let's name this file uy_pep and make the prefix uy-pep so that it all agrees on the country prefix

datasets/uy/pep/crawler.py
```diff
@@ -0,0 +1,55 @@
...
+    role = row.pop("cargo")
+
+    person.id = context.make_slug(id_number, prefix="ur-cedula")
+    person.add("idNumber", id_number)
+    person.add("name", person_name)
+
+    organization_name = row.pop("organismo")
+
+    organization = context.make("Organization")
+    organization.id = context.make_slug(organization_name, prefix="ur-cedula")
```
jbothma, line 38:
```suggestion
    organization.id = context.make_slug(organization_name, prefix="uy-cedula")
```

datasets/uy/pep/crawler.py
atehe: Thanks for catcching that! I've changed the `ur` prefix to the correct country code, `uy` in the new commit

datasets/uy/pep/crawler.py
atehe: Yeah, it has been removed

datasets/uy/pep/uy_pep.yml
```diff
@@ -0,0 +1,14 @@
+name: uy_pep
+title: "Politically Exposed Persons (PEP)"
```
jbothma, line 2: I would prefix this with Uruguayan, but otherwise looking good so far!

atehe:
Hi, @jbothma , I've updated the crawler descriptions.
While doing that, I saw that the organization's data may be needed, as the PEP criteria include them.
I understand why we don't need to include them in the organization entity, but is there another way we could add them to the dataset?

jbothma:
Hi @atehe, so we do want the organization name in the position name generally. In an ideal world, we'd have something like "Director of the ministry of energy" or "Mayor of Peopleville". But to avoid getting to deep into trying to construct sensible labels in all kinds of languages, what I tend to do is make a position name like `f"{cargo}, {organismo}"`

Then I just google translate a few of those to see if it makes sense when I read it, or if it looks really weird. If it's really weird, I try and get a sense of whether it would blow the budget to put more effort into it.

atehe:
Cool! Adding the organization name to the position makes sense. I'll try it out and see how it goes

datasets/uy/pep/crawler.py
```diff
@@ -26,19 +27,34 @@ def parse_sheet_row(context: Context, row: Dict[str, str]):
...
+
+    categorisation = categorise(context, position)
+    if not categorisation.is_pep:
+        return
+
+    occupancy = h.make_occupancy(
+        context,
+        person,
+        position,
+    )
```
jbothma, line 48: Because this list includes people who are PEPs but not necessarily currently in this role any more, it looks like we need to set no_end_date_implies_current to false in the call to make_occupancy

> As defined by Article 20 of Law No. 19574, PEPs are understood to include individuals who, within the last five years have held or currently hold important public functions, both nationally and internationally.

jbothma: while you're here, could you pass the categorisation value to make_occupancy as well? I haven't merged that documentation yet but it can be handy later.

datasets/uy/pep/crawler.py
```diff
@@ -26,19 +27,34 @@ def parse_sheet_row(context: Context, row: Dict[str, str]):
...
 
     context.emit(person, target=True)
     context.emit(position)
+    context.emit(occupancy)
```
jbothma, line 52: could you do context.audit_data() on row after this? then we'll get a little warning if they add more columns, prompting us to consider including that in the crawler in future

jbothma (changes requested):
Thanks! This is just about there - just a couple of small comments then I'll merge!

PR #508, opened 2024-01-12
Crawler for Hong Kong Legislative Council members as PEPs

dhdaines:
Here is a crawler that loads the JSON from the backend for the HK legislative council and emits entities for members.

There is some work to be done to get all the information and fit it into the schema, and the metadata will need to be completed as well.

### DIFF
--- a/datasets/hk/legco/crawler.py
+++ b/datasets/hk/legco/crawler.py
@@ -3,30 +3,45 @@ Crawler for extracting names and constituencies of Hong Kong
 Legislative Council members.
 """
 
-from typing import Dict, Union, List
+from typing import Dict, Union, NamedTuple
 
 from zavod import Context
 from zavod import helpers as h
 from zavod.logic.pep import categorise
 
-MEMBER_URL_FORMAT = (
-    "https://app4.legco.gov.hk/mapi/{lang}/api/LASS/getMember?member_id={member_id}"
-)
+
+class SearchKeyData(NamedTuple):
+    """Components of "search_key" field in member list."""
+
+    name_zh_hant: str  # Name (traditional)
+    name_zh_hans: str  # Name (simplified)
+    name_en: str  # Name (English)
+    cons_name_hant: str  # Constituency name (traditional)
+    cons_name_hans: str  # Constituency name (simplified)
+    cons_name_en: str  # Constituency name (English)
+    cons_type_hant: str  # Constituency type (traditional)
+    cons_type_hans: str  # Constituency type (simplified)
+    cons_type_en: str  # Constituency type (English)
+
+
+def extract_search_key(key: str) -> SearchKeyData:
+    """Extract fields from search_key."""
+    return SearchKeyData(*key.split("|"))
 
 
 def crawl_member(
     context: Context,
     member: Dict[str, Union[str, int]],
-    pages: Dict[str, Dict[str, Union[str, int, List]]],
 ):
-    context.log.info(
-        "Adding {name} ({name_tc})".format(
-            name=pages["en"]["name"], name_tc=pages["tc"]["name"]
-        )
-    )
+    """Emit entities for a member of the Legislative Council from JSON data."""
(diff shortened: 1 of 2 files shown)


datasets/hk/legco/hk_legco.yml
```diff
@@ -0,0 +1,38 @@
...
+    Hong Kong SAR.  The Legislative Council is composed of 90 members,
+    20 of whom are popularly elected by geographical constituencies,
+    30 of whom are elected by specific functional constituencies
+    composed of organizations or individuals in specific segments of
+    Hong Kong society, and 40 of whom are elected by the 1500-member
+    Election Committee.
+
+    It includes the current members along with their names (in
+    English, Traditional and Simplified Chinese), contact information,
+    and educational qualifications.
```
jbothma, line 33: All of this can move to the dataset description.

Very nice in general though!

datasets/hk/legco/crawler.py
```diff
@@ -0,0 +1,186 @@
...
+    person = context.make("Person")
+    person.id = context.make_id(member.pop("member_id"), member.pop("search_key"))
+    context.log.info("Unique ID {person_id}".format(person_id=person.id))
+    # Add names in English and both Chinese writing systems
+    h.apply_name(person, full=pages.en.pop("name"), lang="en")
+    h.apply_name(person, full=pages.hant.pop("name"), lang="zh_hant")
+    h.apply_name(person, full=pages.hans.pop("name"), lang="zh_hans")
+    for email in pages.en.pop("email_address", []):
+        if email:
+            context.log.info("Email: {email}".format(email=email))
```
jbothma, line 52: you can change these to debug-level once you're close to done

datasets/hk/legco/crawler.py
```diff
@@ -0,0 +1,186 @@
...
+        person.add("title", title)
+    for qual in pages.en.pop("qualification", []):
+        if qual:
+            context.log.info("Education: {qual}".format(qual=qual))
+            person.add("education", qual)
+    for address in pages.en.pop("office_address", []):
+        if address:
+            context.log.info("Address: {address}".format(address=address))
+            address_entity = h.make_address(context, address)
+            h.apply_address(context, person, address_entity)
```
jbothma, line 75: We don't normally add education and addresses for PEPs, although now that it's done, it doesn't cause harm. 
I've made a start to documenting which fields are priority and which aren't in https://zavod.opensanctions.org/priorities/ but I realise nothing's linking to it yet so I can't expect anyone to find it :)

datasets/hk/legco/crawler.py
```diff
@@ -0,0 +1,186 @@
...
+    hant: Dict[str, Any]
+    hans: Dict[str, Any]
+
+
+def crawl_member_pages(context: Context, member_id: int) -> MemberPages:
+    # Retrieve member pages in English, traditional and simplified Chinese
+    pages = []
+    for lang in "en", "tc", "cn":
+        member_url = MEMBER_URL_FORMAT.format(lang=lang, member_id=member_id)
+        response = context.fetch_json(member_url, cache_days=7)
```
jbothma, line 31: I reckon you can change the caching to 1 day unless it's particularly heavy. It's quite fun to be able to have new people added/departing people dropped the next day

dhdaines: Makes sense!  Done.

datasets/hk/legco/crawler.py
dhdaines: done!

datasets/hk/legco/crawler.py
dhdaines: OK, I can leave them in - the same data appears to be in Wikidata as well.

datasets/hk/legco/hk_legco.yml
dhdaines: Ah, of course - I went through the metadata documentation again and I think it's correct now.

PR #509, opened 2024-01-13
Callout pyicu and plyvel as required system dependencies in documentation

### LINKED ISSUE #504: Call out system dependencies in install docs
The [installation documentation](https://github.com/opensanctions/opensanctions/blob/main/zavod/docs/install.md#python-virtual-environment) currently states:

> zavod has an optional dependency on PyICU, a library related to the transliteration of names in other alphabets to the latin character set. This library is not installed by default because its configuration can be tricky. Consider following [the PyICU documentation](https://pypi.org/project/PyICU/) to install this library and achieve better transliteration results.

However, from my understanding, `pyicu` and `plyvel` are [required](https://github.com/opensanctions/opensanctions/blob/main/zavod/setup.py#L23) and their installation steps are not straightforward[^1]. It may be beneficial to highlight this early on, if this is indeed the case. A suggestion for revision could be:

> zavod has dependecies on PyICU - a library related to the transliteration of names in other alphabets to the latin character set - and Plyvel - a fast and feature-rich Python interface to LevelDB. The installation and configuration of both libraries can be complex due to system dependencies. Consider following the [PyICU](https://pypi.org/project/PyICU/) and [Plyvel](https://plyvel.readthedocs.io/en/latest/installation.html) documentation for the installation of both libraries.

Happy to open a PR with the suggested change.

[^1]: I'm on an M2 Mac and the tricky parts were related to https://github.com/Homebrew/homebrew-core/issues/36958 and https://github.com/wbolster/plyvel/issues/114#issuecomment-1162769419

fjuniorr:
Closes #504

### DIFF
--- a/zavod/docs/install.md
+++ b/zavod/docs/install.md
@@ -53,8 +53,7 @@ $ zavod --help
 If you encounter any errors during the installation, please consider googling errors related to libraries used by `zavod` (e.g.: SQLAlchemy, Python-Levenshtein, click, etc.).
 
 !!! info "Please note"
-    `zavod` has an optional dependency on PyICU, a library related to the transliteration of names in other alphabets to the latin character set. This library is not installed by default because its configuration can be tricky. Consider following [the PyICU documentation](https://pypi.org/project/PyICU/) to install this library and achieve better transliteration results.
-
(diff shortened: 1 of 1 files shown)


pudo:
Thank you, this makes sense to admit :)

PR #503, opened 2024-01-11
Add data on PEPs in Chile

fjuniorr:
Closes #502

PR #512, opened 2024-01-15
Add data on PEPs in Chile

### LINKED ISSUE #502: Data on PEPs in Chile
(PEPs) from the official list of [declarations of financial interests](https://www.infoprobidad.cl) of Chilean public officials.

fjuniorr:
Closes #502 (continuing #503).

@jbothma you wrote that:

> We will create position names like CONCEJAL, MUNICIPALIDAD DE PENCO by concatenating values later, but categorising positions with all these combinations might be impractical, so consider maintaining an allow list or deny list in code. The answer might be that all these people are politically-exposed persons - we haven't dug into it, so consider that. Propose a draft list if that is the suggested approach.

Indeed not everyone in the original dataset is a PEP. One example are inspectors[^1] but there are several others. 

I can take a closer look when I'm filling the dataset metadata, but for now I could use some feedback about the approach for the allow/deny list.

I opted for explicitly classifying the 511 unique positions (without organizations) in `datasets/cl/info_probidad/cl_info_probidad.yml` as pep or not so that it's easy to flag for review if (more when actually) the domain of the positions changes and the classification must be reviewed.

[^1]: There is a published manual about [law 20.880](http://dpi.minsegpres.gob.cl/files/Manual-ley-de%20probidad.pdf) with information about who needs to provide the declarations. One example are inspectors

    > Sujetos obligados a efectuar declaraciones de intereses y patrimonio
    > 
    > [...]
    > 
    > i) Los funcionarios que cumplan funciones directas de fiscalización. Se entenderá que un funcionario cumple funciones directa de fiscalización cuando dentro de sus funciones permanentes contemplen actividades de inspección directa o le competa intervenir directamente en procedimientos administrativos sancionatorios que no correspondan a procedimientos disciplinarios internos (artículo 2°, N° 9) del Reglamento). 

    which are not considered PEPs neither by [OpenSanctions](https://www.opensanctions.org/docs/pep/methodology/) nor [Chile](https://www.uaf.cl/legislacion/politica.aspx).


datasets/cl/info_probidad/cl_info_probidad.yml
```diff
@@ -0,0 +1,1047 @@
...
+      - match: "GERENTE GENERAL"
+        is_pep: false
+      - match: "GERENTE GENERAL INTERINO"
+        is_pep: false
+      - match: "GOBERNADOR "
+        is_pep: false
+      - match: "GOBERNADOR REGIONAL"
+        is_pep: false
+      - match: "GOBERNADORA REGIONAL"
+        is_pep: false
```
jbothma, line 473: If these are governors, as in the heads of regional government, I would say they're PEPs

jbothma:
Thank you! This looks great.

I didn't link to [this PR](https://github.com/opensanctions/opensanctions/pull/495/files#diff-45db18ca905fcd70a57d44ad0efbfa57f31846aa93d0eafc5ea0c505a3590124R73) yet because it's not merged so I didn't want to make it part of the task yet, but then I forgot that it might help with this case which wasn't in the original advertised task - sorry!

What I would propose is that you keep the ones that are categorised is_pep: true and call categorise(...) on them with is_pep=True, while calling it with is_pep=None for the rest. It looks like there are some that might well be PEPs but you don't need to get too bogged down in that - it's just handy that you'll have a number of probable PEPs to see in your output data.

Please bear with me, this categorise stuff is newish and I'm still figuring out how to make it part of the process for other crawler authors!

datasets/cl/info_probidad/cl_info_probidad.yml
fjuniorr: Indeed they are (done in https://github.com/opensanctions/opensanctions/pull/512/commits/ebebc840651620c6b3262f48e6e137854f2db6d2).

From what I could gather they are distinct because only after 2021 the (regional) governors started to be elected directly by the citizen vote instead of being appointed by the president.

fjuniorr:
> What I would propose is that you keep the ones that are categorised is_pep: true and call categorise(...) on them with is_pep=True, while calling it with is_pep=None for the rest. It looks like there are some that might well be PEPs but you don't need to get too bogged down in that - it's just handy that you'll have a number of probable PEPs to see in your output data.

I've changed this in https://github.com/opensanctions/opensanctions/pull/512/commits/2b64a00f6f95e445f69e3ee638f0044cd11e60c5, but when we start creating the positions by concatenating `Cargo` and `Institucion` we will go to 2417 combinations and if I understand correctly they will all end up in the database. Wasn't the ideia to categorise only by `Cargo`? Or should I send the positions to the database without the organization?

datasets/cl/info_probidad/crawler.py
```diff
@@ -0,0 +1,48 @@
...
+    position = h.make_position(
+        context, position_name, country="cl", lang="spa"
+    )
+    position_lookup = context.lookup("positions", position_name)
+    
+    # all positions (ie. Cargo) should be explicitly classified as either pep or no-pep
+    if not position_lookup:
+        context.audit_data(position.to_dict())
+    
+    categorisation = categorise(context, position, is_pep = getattr(position_lookup, "is_pep", None))
```
jbothma, line 34: let's do the lookup on Cargo only, but then make position name the concatenated position, so that `Governor, Region x` and `Governor, Region y` are distinct positions

datasets/cl/info_probidad/crawler.py
```diff
@@ -0,0 +1,48 @@
...
+
+    position_name = row.pop("Cargo")
+    position = h.make_position(
+        context, position_name, country="cl", lang="spa"
+    )
+    position_lookup = context.lookup("positions", position_name)
+    
+    # all positions (ie. Cargo) should be explicitly classified as either pep or no-pep
+    if not position_lookup:
+        context.audit_data(position.to_dict())
```
jbothma, line 32: We usually do audit_data on input data, e.g. a dict representing a row in a table. So at the end of the crawler, you might want to do context.audit_data(row) to see if there are fields left that you could well use, otherwise add them to the `ignroe` argument to `audit_data`.

datasets/cl/info_probidad/crawler.py
fjuniorr: Got it! Done in https://github.com/opensanctions/opensanctions/pull/512/commits/34a947ed64f58418a79b89a1db2e859e5557d3d4

datasets/cl/info_probidad/crawler.py
fjuniorr: Do you think it makes sense to log as warnings if new values for `Cargo` are identified? I was doing it before because I though that the categorisation was going to happen in `datasets/cl/info_probidad/cl_info_probidad.yml`, but now I'm not so sure because the positions will end un the database anyway.

fjuniorr:
Got it now! Thanks for the explanation. I have one small [question](https://github.com/opensanctions/opensanctions/pull/512#discussion_r1453451865) but after then that this is ready for review.

In terms of the [data priorities](https://zavod.opensanctions.org/priorities/)  in the [csv](https://datos.cplt.cl/catalogos/infoprobidad/csvdeclaraciones) that we used as source there is not information about relatives or close associates, but they are likely to be available in the other serialization formats (they are timing out).

datasets/cl/info_probidad/crawler.py
jbothma: Yeah not a bad idea! I like how this is working now - so the list tells us whether we've taken a look at the role before, and the database tells us whether it's a PEP or not.

Once this is running for a few months we can see whether this changes often, if ever.

jbothma:
I think relatives and close associated would be very helpful. Not needed right now, but I'll add it to the backlog so we can dig into it when it's a priority.

Generally, we aim for crawlers not to take more than 6 hours, but ideally less than 4 hours of effrort. So if RCAs and/or financial interests which fit [our data model](https://www.opensanctions.org/reference/) can be included within that time, that's awesome.

PR #514, opened 2024-01-15
add all wikidata item labels for languages we can process

jbothma:


pudo:
We have a list of languages here which our stack has .... any chance of processing correctly: https://github.com/alephdata/followthemoney/blob/main/followthemoney/types/language.py#L27 - how do you feel about limiting the WD statements we import to these languages?

jbothma:
That makes sense!

jbothma:
> We're probably want this in the enricher, too ....

@pudo I added this approach to skipping text indicated to be in languages we don't handle yet https://github.com/opensanctions/nomenklatura/pull/136

the enricher is already putting all labels except the "best choice" in the alias property. what did you mean should also be in the enricher? The restriction to our language whitelist? If so, is this sufficient?

Would it also make sense to only add aliases which are not already in the name property, to avoid duplication between wd_pep name property and wikidata enricher alias property?

pudo:
@jbothma Just one note here: `wd_pep` is only one source of Q-ID entities, so the enricher may not need to add them to these but perhaps to the Q-ID entities coming from other source (`sanctions`, `wd_curated`, etc.).

jbothma:
so I saw three approaches

1 only add all labels to name in enricher for entities which don't come from wd_pep
2 only add names to aliases which are not already in the entity name property
3 drop this PR and always add all labels to name in enricher https://github.com/opensanctions/nomenklatura/pull/138
  - we should then probably add wikidataID in wd_peps too - is there any reason we didn't?
  - I don't recall - was there any reason we wanted wd_peps to add the labels and not wait for the enricher to do it?

3 looks like the cleanest option but I don't recall if there was a reason we preferred not to do it?

PR #519, opened 2024-01-16
Creates the crawler for the Brazil Central Bank Disqualified Persons List

bgmello:
This PR is related to the issue https://github.com/opensanctions/crawlers/issues/15


datasets/br/bcb_disqualified_persons/bcb_disqualified_persons.yaml
```diff
@@ -0,0 +1,27 @@
+title: Brazil's Central Bank General Register of Disqualified Persons  
+entry_point: crawler.py
+prefix: bcb-disqualified-persons
```
jbothma, line 3: Since this prefixes each key, we try to keep it short but unique. We also prefix with the 2-letter country code where it's country-specific.

```suggestion
prefix: br-qgi
```

datasets/br/bcb_disqualified_persons/crawler.py
```diff
@@ -0,0 +1,90 @@
...
+    sanction.add(
+        "program", "Brazil's Central Bank General Register of Disqualified Persons"
+    )
+    sanction.add("authority", "Brazil's Central Bank")
+
+    # The ID of the process
+    sanction.add(
+        "description",
+        "Administrative Sanctioning Process Number: {}".format(input_dict.pop("PAS")),
+    )
```
jbothma, line 64: Would it make sense to add the PAS id as the key argument to the `make_sanction`? That way each individual process would have its own sanction entity and the dates wouldn't get bagged together.

I noticed this by checking the number of rows in the table and trying to understand why the number of persons doesn't match the number of rows - it's because one person can be the subject of multiple investigations.

datasets/br/bcb_disqualified_persons/bcb_disqualified_persons.yaml
```diff
@@ -0,0 +1,27 @@
...
+  serves as a crucial resource for monitoring regulatory compliance in the financial sector.
+  It enumerates individuals who, due to penalties incurred in administrative sanction proceedings,
+  are temporarily prohibited from occupying key positions such as administrators
+  or board members in financial or consortium administration entities.
+  This measure ensures the integrity and reliability of personnel in critical financial roles,
+  reflecting the Central Bank's commitment to maintaining stability and transparency in the financial system.
+  The dataset is a part of broader efforts to uphold high standards in financial governance
+  and protect the interests of stakeholders in the financial industry.
+publisher:
+  name: BCB (Brazil's Central Bank)
```
jbothma, line 20: I tend to put the official name in the local language in full in the name field - `Banco Central do Brasil (BCB)` - then I translate that in the publisher.description field.

datasets/br/bcb_disqualified_persons/crawler.py
```diff
@@ -0,0 +1,90 @@
...
+                        - Prazo_final_penalidade
+    :param entity: Entity the sanction refers to.
+    :param context: The context object.
+    """
+
+    sanction = h.make_sanction(context, entity)
+    sanction.add(
+        "program", "Brazil's Central Bank General Register of Disqualified Persons"
+    )
+    sanction.add("authority", "Brazil's Central Bank")
```
jbothma, line 58: this seems to get populated automatically by make_sanction so you can skip it

jbothma (changes requested):
Very nice and clean, thanks!

Just a few comments that should be quick to change - it's a lot of things from my experience that would be nice to move to documentation soon

datasets/br/bcb_disqualified_persons/bcb_disqualified_persons.yaml
```diff
@@ -0,0 +1,27 @@
...
+  url: https://www.bcb.gov.br
+  official: true
+  country: "br"
+url: https://olinda.bcb.gov.br/olinda/servico/Gepad_QuadrosGeraisInternet/versao/v1/odata/QuadroGeralInabilitados?format=text/html
+data:
+  url: https://olinda.bcb.gov.br/olinda/servico/Gepad_QuadrosGeraisInternet/versao/v1/odata/QuadroGeralInabilitados?format=text/html
```
jbothma, line 26: I suspect it would work just as well without this, because it's probably using the accept header and ignoring the querystring. Then this querystring is a bit confusing given the code's using JSON.

```suggestion
  url: https://olinda.bcb.gov.br/olinda/servico/Gepad_QuadrosGeraisInternet/versao/v1/odata/QuadroGeralInabilitados
```

datasets/br/bcb_disqualified_persons/crawler.py
```diff
@@ -0,0 +1,90 @@
...
+    )
+
+    # The duration is always in years
+    sanction.add("duration", "{} year(s)".format(input_dict.pop("Prazo_em_anos")))
+
+    # The start and end dates are in the format YYYY-MM-DD
+    sanction.add("startDate", input_dict.pop("Inicio_do_cumprimento"))
+    sanction.add("endDate", input_dict.pop("Prazo_final_penalidade"))
+
+    return sanction
```
jbothma, line 73: Could you add Penalidade under Sanction:provisions with the three-letter language code for bazil?

datasets/br/bcb_disqualified_persons/bcb_disqualified_persons.yaml
bgmello: QGI is the acronym for Quadro Geral de Inabilitados (which is General Register of Disqualified Persons). In those cases, we use the local language acronym?

datasets/br/bcb_disqualified_persons/crawler.py
bgmello: I didn't find the key parameter in the Sanction schema https://followthemoney.tech/explorer/schemata/Sanction/

datasets/br/bcb_disqualified_persons/crawler.py
bgmello: I think I didn't fully understood what you mean here, do you want to add something like:

```
sanction.add("provisions", "PNL")
```

Where PNL is an acronym for "Penalidade"?

datasets/br/bcb_disqualified_persons/crawler.py
jbothma: I meant the key argument shown here https://zavod.opensanctions.org/helpers/#zavod.helpers.make_sanction
like `h.make_sanction(context, entity, key=pas)`

datasets/br/bcb_disqualified_persons/crawler.py
jbothma: ah no, sorry - I mean
```
sanction.add("provisions", input_dict.pop("Penalidade"), lang="por") - right now the value is always INABILITAÇÃO which I'm not sure google translate is translating correctly to "disability", but it seems to fit the "Scope of sanctions", right?

datasets/br/bcb_disqualified_persons/crawler.py
bgmello: I'll add that, should I keep the `description`? I added it because I couldn't find where to put the `PAS`, but now it might be irrelevant

datasets/br/bcb_disqualified_persons/crawler.py
bgmello: It's always INABILITAÇÃO because this is the "Quadro Geral de Inabilitados". 

The word inabilitação comes from the word habilitação (it comes from the same latin root for able) which I think would be best translated to license. So for instance Driver's License is "Habilitação para dirigir" (License to drive). The "in-" prefix means the negative, so in-abilitação means they "lost their license" (more in a figurative way) and they are not able to hold some positions in financial companies.

Does that mean the same as disability? I'm not sure what would be the correct English term for that.

jbothma:
Thanks! keep description - it's a nice explanatory touch

bgmello:
I think I resolved all changes @jbothma

Issue #525, opened 2024-01-18
Add PEPs from Venezuelan National Assembly


dhdaines:
As seen here: https://www.asambleanacional.gob.ve/diputados

PR #526, opened 2024-01-18
Crawler for the National Assembly of Venezuela

dhdaines:
Crawls members from the website extracting some info from personal pages where possible (could do better).

See: #525


jbothma (changes requested):
very nice! thank you!

I'd suggest changing the info logs about unique IDs to debugs, the exception handler to info-level, and then adding a couple of `match` [datapatches](https://github.com/opensanctions/datapatch) for the date warnings like this

```
2024-01-22 10:46:26 [warning  ] Rejected property value        [zavod.runtime.cleaning] dataset=ve_asamblea_nacional entity_id=ve-asamblea-nacional-69be6cde3d6e13b51c55c35594c54fef93a88877 prop=birthDate value=31 de enero de 1.981
```

like this https://github.com/opensanctions/opensanctions/blob/main/datasets/_global/icij_offshoreleaks/icij_offshoreleaks.yml#L497

The thinking here is that errors and warnings end up at https://www.opensanctions.org/issues/ and I don't see them fixing them soon. Still, it's nice that they're visible.

dhdaines:
> I'd suggest changing the info logs about unique IDs to debugs, the exception handler to info-level, and then adding a couple of `match` [datapatches](https://github.com/opensanctions/datapatch) for the date warnings 

Ah, okay, I think I understand - basically we can add the unmatched values as overrides to the `lookups` in the yaml file.  There is also one member whose information is all contained in `<span>` tags so lxml concatenates it together with no whitespace :( though I think I could also special-case that in the crawler code.

datasets/ve/asamblea_nacional/ve_asamblea_nacional.yml
```diff
@@ -0,0 +1,37 @@
...
+summary: >-
+  Members of the National Asssembly of Venezuela.
+description: |
+  This dataset lists the current members of the National Assembly
+  of the Bolivian Republic of Venezuela.
+
+  The data is collected from the official website of the National
+  Assembly.
+entry_point: crawler.py
+prefix: ve-asamblea-nacional
```
jbothma, line 13:
```suggestion
prefix: ve-asamblea
```

datasets/ve/asamblea_nacional/crawler.py
```diff
@@ -0,0 +1,205 @@
...
+        member_link = el.find(".//a")
+        if member_link is None:
+            context.log.error(f"No page found in element {el}")
+            continue
+        crawl_member(context, member_link)
+
+
+def crawl_member_list(context: Context) -> Iterator[ElementOrTree]:
+    """Iterate through pages in member list from the website."""
+    context.log.debug(f"Fetching front page from {context.data_url}")
```
jbothma, line 171:
```suggestion
    context.log.info(f"Fetching front page from {context.data_url}")
```

jbothma (approved):

PR #533, opened 2024-01-19
Add crawler for South Africa wanted persons list

fjuniorr:
Closes https://github.com/opensanctions/opensanctions/issues/532

@jbothma this is ready for review but I do have some questions:

- Should I emit persons with unknown name? From https://zavod.opensanctions.org/priorities/ I'm guessing no, but good to check.
- Should I emit individuals listed as suspects?


datasets/za/wanted/crawler.py
```diff
@@ -0,0 +1,60 @@
...
+        name, crime, status = match.groups()
+        last_name, first_name = name.split(",", maxsplit=1)
+        result["first_name"] = first_name.strip()
+        result["last_name"] = last_name.strip()
+        result["crime"] = crime.strip()
+        result["status"] = status.strip()
+        details_id = cell.xpath(".//a/@href")[0]
+        source_url = (
+            f'https://www.saps.gov.za/crimestop/wanted/{details_id}'
+        )
```
jbothma, line 26: for future reference, here's another way to do this

https://github.com/opensanctions/opensanctions/blob/main/datasets/co/funcion_publica/crawler.py#L199

datasets/za/wanted/crawler.py
```diff
@@ -0,0 +1,60 @@
...
+    breakpoint()
+
+    h.apply_name(
+        person,
+        first_name=data.get("first_name"),
+        last_name=data.get("last_name"),
+    )
+
+    person.add("sourceUrl", data.get("source_url"))
+    person.add("notes", f"{data.get('status')} - {data.get('crime')}")
```
jbothma, line 49: nice

jbothma (changes requested):
Good questions

> Should I emit persons with unknown name? From https://zavod.opensanctions.org/priorities/ I'm guessing no, but good to check.

Let's skip those without names, yes. Thanks!

> Should I emit individuals listed as suspects?

Yes - thanks - here's a similar example https://www.opensanctions.org/datasets/nl_most_wanted/

Could you also add this dataset to the crime collection? Then I can merge once you add the name skip.

https://github.com/opensanctions/opensanctions/blob/main/datasets/_collections/crime.yml#L27

datasets/za/wanted/crawler.py
fjuniorr: Thanks!

Didn't knew about `make_links_absolute`. Always good to keep things DRY. 

Done in https://github.com/opensanctions/opensanctions/pull/533/commits/29b2f955479072c393cf0ddecff19ec031353a8b

fjuniorr:
> Let's skip those without names, yes. Thanks!

It hasn't happened in the data yet, but should I emit if there is only a first name?

> Could you also add this dataset to the crime collection? Then I can merge once you add the name skip.

Done in https://github.com/opensanctions/opensanctions/pull/533/commits/6e76884e1cd4967234d1127f3cf6bd5d61dd1be7

datasets/za/wanted/crawler.py
```diff
@@ -0,0 +1,61 @@
...
+    
+    # each wanted person has a dedicated details page 
+    # which appears to be a unique identifier
+    id = parse_qs(urlparse(data["source_url"]).query)["bid"][0]
+    person.id = context.make_slug(id)
+
+    h.apply_name(
+        person,
+        first_name=data.get("first_name"),
+        last_name=data.get("last_name"),
```
jbothma, line 44: Sorry for the late request - could you do the logic in parse_cell inside crawl_cell (perhaps call it crawl_person)?

The thinking is just that it's kinda making a two step process what would be just as clear in one method, and that's closer to our usual style.

jbothma: this gives more of a one to one mapping between the source of the data and the destination.

fjuniorr: I've refactored `parse_cell` and `crawl_cell` into `crawl_person` on https://github.com/opensanctions/opensanctions/pull/533/commits/bf44b48987e110f070c394139fd3c6d7c8b2f4f9 and I think things look much clearer. Let me know if this looks good.

datasets/za/wanted/crawler.py
```diff
@@ -0,0 +1,62 @@
...
+        person,
+        first_name=first_name,
+        last_name=last_name,
+    )
+
+    person.add("sourceUrl", source_url)
+    person.add("notes", f"{status} - {crime}")
+
+    person.add("topics", "crime")
+    person.add("country", "za")
```
fjuniorr, line 51: > yeah I think if there's either a first or a last name, it's fair to emit them. I think to make this a bit less risky, we can add `za` to the `country` property.

@jbothma I've added `za` to the `country` property but I'm wondering what are the semantics in this case? I've looked at the [data dictionary](https://www.opensanctions.org/reference/#schema.Person) but still couldn't understand. Or this is simply to add information to the [deduplication](https://www.opensanctions.org/articles/2021-11-11-deduplication/) process?

fjuniorr:
> Could you also add this dataset to the crime collection? Then I can merge once you add the name skip.
> 
> https://github.com/opensanctions/opensanctions/blob/main/datasets/_collections/crime.yml#L27

Should I also add the dataset to the [wanted collection](https://github.com/opensanctions/opensanctions/blob/main/datasets/_collections/wanted.yml)?

I noticed that only [spliting on comma](https://github.com/opensanctions/opensanctions/pull/533/files#diff-14d619b586522c0ec2f03f12915f96ca422b6047650e215c46a17b09592f5f73R19) is generating problems in some names. I've added a warning but still need to think a bit one the best way to handle this. 

It looks like in some cases there is also a nickname, such as for [Charl, Johann Hougaard](https://www.saps.gov.za/crimestop/wanted/detail.php?bid=10057) ([source](https://www.facebook.com/IRSinvestigations/photos/a.800543706643705/3048077305223656/?type=3)) but this is not the general case. [Darryl, Evans](https://www.saps.gov.za/crimestop/wanted/detail.php?bid=11554) for example simply has an extra comma (eg. `Evans, Darryl, (Driving Under The Influence) Wanted`).

jbothma:
ah wanted sounds more appropriate, thanks!

In that case, I'd suggest not splitting on comma - just use the name in the order it's given, and put the whole thing in the name field.

Our name matching API handles different order of names well enough, I believe.

datasets/za/wanted/crawler.py
jbothma: yeah my reasoning here is both when we're doing deduplication, and if people are doing matching against their customers and someone from this list with a partial name comes up, they can more easily determine if it's a true match.

PR #546, opened 2024-01-25
Adding TCU disqualified crawler

bgmello:
This PR is solving the issue https://github.com/opensanctions/crawler-planning/issues/17


bgmello:
I couldn't find an official site with details on the dataset, but this [article](https://suporte.linkana.com/pt-BR/articles/3137029-lista-de-inabilitados-para-funcao-publica) does a great job of explaining it. For the `url`, I've added the TCU page link for cross-checking documents on the list because I didn't know if it should only be official url, if you want, I can change to the link to the article.

datasets/br/br_tcu_disqualified/crawler.py
```diff
@@ -0,0 +1,92 @@
...
+    Fetches data from the website, or raises an exception on failure.
+
+    :param context: The context object.
+    :return: List of dicts containing data.
+    """
+    response = context.fetch_json(context.data_url)
+    if "items" not in response:
+        context.log.error("Items not found in JSON")
+        return []
+    return response["items"]
```
jbothma, line 16: you can probably just fold this into crawl() - it's not being reused I don't think we're gaining from it being a separate function.

datasets/br/br_tcu_disqualified/crawler.py
```diff
@@ -0,0 +1,92 @@
...
+    The data is already in the format of a list of dicts, so we just need to create the entities.
+
+    :param context: The context object.
+    """
+    data = fetch_data(context)
+    for line in data:
+        entity = create_entity(line, context)
+        sanction = create_sanction(line, entity, context)
+        context.emit(entity, target=True)
+        context.emit(sanction)
```
jbothma, line 92: our pattern would normally be to create and emit both the entity and the sanction in one method called something like crawl_row or crawl_item especially when it's this simple and it all actually fits on one screen

datasets/br/br_tcu_disqualified/crawler.py
```diff
@@ -0,0 +1,92 @@
...
+                        - cpf
+                        - nome
+    :param context: The context object.
+    """
+
+    entity = context.make("Person")
+    # the CPF comes in formatted form (i.e. with punctuaction, XXX.XXX.XXX-XX)
+    # and it's common practice to remove this punctuation when saving in the database
+    raw_tax_number = input_dict["cpf"]
+    tax_number = raw_tax_number.replace(".", "").replace("-", "")
```
jbothma, line 34: should we create a clean_br_cpf function somewhere and use it in the bcb crawlers too? that could go in zavod.helpers.text to start with until @pudo suggests a better place

datasets/br/br_tcu_disqualified/br_tcu_disqualified.yaml
```diff
@@ -0,0 +1,24 @@
...
+coverage:
+  frequency: daily
+load_db_uri: ${OPENSANCTIONS_DATABASE_URI}
+summary: >
+  The inabilitados do TCU database catalogs individuals sanctioned by Brazil's Tribunal de Contas da União, prohibiting them from commissioned government roles.
+description: |
+  The TCU Disqualified Individuals compiles a list of individuals who have 
+  faced sanctions due to various offenses, resulting in their disqualification
+  from occupying positions of trust or commissioned roles in governmental institutions.
+  This database enforces compliance with Article 60 of Law 8.443/92,
```
jbothma, line 13:
```suggestion
  This database enables compliance with Article 60 of Law 8.443/92,
```

datasets/br/br_tcu_disqualified/br_tcu_disqualified.yaml
```diff
@@ -0,0 +1,24 @@
+title: Brazil's Court of Account Disqualified Individuals
```
jbothma, line 1:
```suggestion
title: Brazil's TCU list of individuals disqualified from public service
```

datasets/br/br_tcu_disqualified/crawler.py
```diff
@@ -0,0 +1,92 @@
...
+    sanction = h.make_sanction(context, entity, key=process_number)
+
+    # The ID of the process
+    sanction.add(
+        "description",
+        "Process Number: {}, acording to the deliberation number: {}".format(
+            process_number, input_dict.pop("deliberacao")
+        ),
+    )
+
```
jbothma, line 70: Let's also add a sanction program property explaining briefly what what this restriction means, referring to the relevant section in the law. A hardcoded string often works well for sanction program.

jbothma (changes requested):
Looks good overall! I added some comments to bring it more in line with our usual style. 

I think you can add a markdown link like suggesting people can read more at that article, but keep the dataset url property to the current one - I think it's good for that to be official.

> Disqualification from public service occurs after legal proceedings and judgment by the TCU.  

From that article, it looks like the `crime.fin` topic is more appropriate than debarment, right @bgmello? @pudo ?

datasets/br/br_tcu_disqualified/crawler.py
bgmello: I think it would be a good idea, they sometimes put the punctuation because it's easier to read with it

bgmello: Do you want me to create a PR for that? I'm guessing it's [here](https://github.com/opensanctions/opensanctions/tree/main/zavod/zavod/helpers), but should I create a separate file?

datasets/br/br_tcu_disqualified/crawler.py
bgmello: I added that, let me know if it's good

datasets/br/br_tcu_disqualified/crawler.py
jbothma: yeah - I think in the `text` module in there to start with, and then exposed under `helpers.` via the __init__ module in that directory

datasets/br/br_tcu_disqualified/crawler.py
jbothma: it's perfect

datasets/br/br_tcu_disqualified/crawler.py
jbothma: this PR or new PR are both good, thank you!

bgmello: I added to this PR, let me know if I did it correctly. I also added the one for CNPJ that will be needed for future crawlers.

zavod/zavod/helpers/text.py
```diff
@@ -107,3 +107,54 @@ def remove_bracketed(text: Optional[str]) -> Optional[str]:
...
+    Returns:
+        The cleaned CPF number.
+    """
+
+    # Remove formatting characters
+    cpf = cpf.replace(".", "").replace("-", "")
+
+    # If the CPF is not 11 digits long, it's invalid
+    if len(cpf) != 11:
+        return ""
```
jbothma, line 132: instead of losing data, let's return what we have, broken though it may be.

zavod/zavod/helpers/text.py
```diff
@@ -107,3 +107,54 @@ def remove_bracketed(text: Optional[str]) -> Optional[str]:
...
+
+    # Remove formatting characters
+    cnpj = cnpj.replace(".", "").replace("/", "").replace("-", "")
+
+    # Check if CNPJ is now only digits and 14 digits long
+    if len(cnpj) == 14 and cnpj.isdigit():
+        return cnpj
+
+    # Return an empty string if CNPJ is not valid
+    return ""
```
jbothma, line 160: same here

jbothma (changes requested):
Do you mind adding a a couple of super basic tests for these, and adding them to the other brazilian crawlers with these numbers too?

Sorry for the scope expansion - we can do a different branch next time this comes up :)

zavod/zavod/helpers/text.py
jbothma: this is important because a) for an id, we don't want invalid numbers to end up giving different entities the same id, and b) a human might still make sense of an invalid ID in an investigative setting.

bgmello: That makes sense, I'll fix that

bgmello:
Where should I add the tests?

jbothma:
this should work https://github.com/opensanctions/opensanctions/blob/main/zavod/zavod/tests/helpers/test_text.py

bgmello:
Let me know if it's all good. If yes, I'll apply the same functions to the other Brazilian crawlers.

jbothma:
Thanks for the tests! Sorry I was unclear - if it's invalid after cleaning, I think it's still fair to return the clean but invalid version (as opposed to the original potentially dirty version) - one might write a separate function `is_valid` when needed, which could be used to drop invalid values in future.

With that change you can go ahead. Thanks, this is going to be nice, and if we already get some auto-merged entities, very fun.

jbothma:
Sorry I messed you around a bit on this one - just hang on, I'm speaking to Friedrich about this now

jbothma:
ok, I think my head is screwed back on.

You had a good idea about validation, I just didn't immediately see the right way to use it. Friedrich also showed me a good place to put this.

1 - could you make a PR to rigour adding CNPJ and CPF classes replacing what you've added to helpers.text? https://github.com/opensanctions/rigour/blob/main/rigour/ids/stdnum_.py

2 - let's then use it something like this:

```python
from rigour.ids.stdnum import CPF
...

cpf = CPF.normalize(id)
if cpf is not None:
    entity.id = context.make_slug(cpf, prefix="br-cpf")
else:
    context.log.warn(...)
    # or
    entity.id = context.make_id(whatever, other, entity, identifying, values, available)
```
with a clause for cnpj in crawlers where it's approproate

bgmello:
Waiting for the PR https://github.com/opensanctions/rigour/pull/4

bgmello:
I removed the helpers functions and added the rigour normalization. I think it's ready to merge, let me know if we need to make any other changes.

datasets/br/br_tcu_disqualified/br_tcu_disqualified.yaml
```diff
@@ -0,0 +1,24 @@
+title: Brazil's TCU list of individuals disqualified from public service
```
jbothma, line 1: Remember the directory structure convention is br/tcu_disqualified/br_tcu_disqualified.yaml i.e. no country code in second level

datasets/br/br_tcu_disqualified/br_tcu_disqualified.yaml
```diff
@@ -0,0 +1,24 @@
...
+description: |
+  The TCU Disqualified Individuals compiles a list of individuals who have 
+  faced sanctions due to various offenses, resulting in their disqualification
+  from occupying positions of trust or commissioned roles in governmental institutions.
+  This database enables compliance with Article 60 of Law 8.443/92,
+  ensuring that sanctioned individuals are appropriately restricted
+  from holding certain influential and sensitive positions in the public sector.
+publisher:
+  name: Tribunal de Contas da União (TCU)
+  url: https://portal.tcu.gov.br/inicio/
```
jbothma, line 18: Please add the english version of their name in publisher.description

datasets/br/br_tcu_disqualified/crawler.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+
+    entity = context.make("Person")
+
+    # make sure the CPF does not have any punctuation
+    tax_number = CPF.normalize(input_dict["cpf"])
+
+    # if the tax number is None, it means it was invalid
+    if tax_number is None:
+        entity.id = context.make_id()
```
jbothma, line 26: context.make_id needs arguments to hash - I think the entity name is ideal

datasets/br/br_tcu_disqualified/crawler.py
```diff
@@ -0,0 +1,73 @@
+from zavod import Context, helpers as h
+from rigour.ids.stdnum import CPF
```
jbothma, line 2: it's not working for me at the moment - it looks like it should maybe be rigour.ids.stdnum_ - the underscore is probably added to avoid conflict

datasets/br/br_tcu_disqualified/br_tcu_disqualified.yaml
```diff
@@ -0,0 +1,24 @@
+title: Brazil's TCU list of individuals disqualified from public service
+entry_point: crawler.py
+prefix: br-tcu-disqualified
```
jbothma, line 3:
```suggestion
prefix: br-tcu-disq
```

Let's try and go for shorter prefixes - they prefix all the generated IDs - in this case the sanctions

datasets/br/br_tcu_disqualified/crawler.py
```diff
@@ -0,0 +1,73 @@
...
+    # if the tax number is None, it means it was invalid
+    if tax_number is None:
+        entity.id = context.make_id()
+    else:
+        entity.id = context.make_slug(tax_number, prefix="br-cpf")
+
+    entity.add("name", input_dict["nome"])
+    entity.add("taxNumber", tax_number)
+    entity.add("country", "br")
+    entity.add("topics", "debarment")
```
jbothma, line 33: I would use the corp.disqual topic here and add it to the crime collection

bgmello:
Fixed. I'll update the PR https://github.com/opensanctions/opensanctions/pull/549 when this one is merged.

datasets/br/tcu_disqualified/crawler.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+
+    entity = context.make("Person")
+
+    # make sure the CPF does not have any punctuation
+    tax_number = CPF.normalize(input_dict["cpf"])
+
+    # if the tax number is None, it means it was invalid
+    if tax_number is None:
+        entity.id = context.make_id(input_dict["cpf"]+input_dict["nome"])
```
jbothma, line 26:
```suggestion
        entity.id = context.make_id(input_dict["cpf"], input_dict["nome"])
```

PR #557, opened 2024-02-03
Adding crawler for state governors of Mexico

bgmello:
PR related to the issue https://github.com/opensanctions/crawler-planning/issues/21

For some reason it is not issuing the occupancy properly in the third iteration. It looks like is something in the `h.make_occupancy` function (it's returning None) but it doesn't log any error or warning. Can you help me troubleshoot this bug? @jbothma 

Screenshot 2024-02-03 at 18 42 55

### DIFF
--- a/datasets/mx/governors/crawler.py
+++ b/datasets/mx/governors/crawler.py
@@ -1,0 +1,70 @@
+import re
+
+from zavod import Context, helpers as h
+from zavod.logic.pep import categorise
+
+def crawl_item(input_html, context: Context):
+    """
+    Creates an entity, a position and a occupancy from the raw HTMLElement.
+
+    :param input_html: The div representing that governor
+    :param context: The context object.
+    """
+
+    # The easiest way to get the data from the HTML is surprisingly from the text of the class
+    # All the information in the web site is written there in the form:
+    # col-xs-12 col-sm-6 col-md-4 pb15 mix {name of the governor} [{state}] {start_date} {end_date}
+    # We will then use a regex to extract the name, state, start_date and end_date
+
+    regex_pattern = r'col-xs-12 col-sm-6 col-md-4 pb15 mix (.*?) \[(.*?)\] (\d{2}/\d{2}/\d{4}) (\d{2}/\d{2}/\d{4})'
+
+    match = re.match(regex_pattern, input_html.get('class'))
+
+    if match is None:
+        context.log.warning("Unable to extract information from HTML element")
+
+    name, state, start_date, end_date = match.group(1), match.group(2), match.group(3), match.group(4)
+
+    person = context.make("Person")
+    person.id = context.make_id(name)
+    person.add("name", name)
+
+    name_of_position = "Governor of " + state.title()
+    position = h.make_position(context, name_of_position, country="mx")
+    categorisation = categorise(context, position, is_pep=True)
+
+    occupancy = h.make_occupancy(
+            context,
+            person,
+            position,
+            no_end_implies_current=True,
+            start_date=h.parse_date(start_date, formats=["%d/%m/%Y"])[0],
+            end_date=h.parse_date(end_date, formats=["%d/%m/%Y"])[0],
+            categorisation=categorisation,
+        )
+
+    context.emit(person, target=True)
+    context.emit(position)
+    context.emit(occupancy)
+    
+
+def crawl(context: Context):
+    """
+    Entrypoint to the crawler.
+
+    The crawler works by first downloading the HTML of the page
+    and then iterating for each div that represents a Mexico State Governor,
+    to create the entities and positions.
+
+    :param context: The context object.
+    """
+
+    response = context.fetch_html(context.data_url)
+
+    if response is None:
+        context.log.error("Error while fetching the HTML")
+        return
+
+    path_to_cards = '//*[@class="containerMixitup"]/div/div'
+    for item in response.xpath(path_to_cards):
+        crawl_item(item, context)
--- a/datasets/mx/governors/crawler.py
+++ b/datasets/mx/governors/crawler.py
@@ -62,14 +62,13 @@ def crawl_item(input_html, context: Context):
     # Ing. Carlos Lozano de la Torre, where Ing. stands for Engineer
     # We will add two name propreties, one with the "full" name
     # and other with the clean name.
-    name = input_html.xpath("./div/div/div/div[2]/h4/a/text()")[0].strip()
-    clean_name = " ".join(name.split()[1:])
-
+    raw_title = input_html.xpath("./div/div/div/div[2]/h4/a/text()")[0].strip()
+    name = re.sub(r"^([A-Z][a-z]*\.)+ ", "", raw_title)
+    
     person = context.make("Person")
     person.id = context.make_id(name)
 
     person.add("name", name)
-    person.add("name", clean_name)
 
     birth_place, birth_date = extract_birth_place_and_date(link_governor_page, context)
 
--- a/datasets/mx/governors/crawler.py
+++ b/datasets/mx/governors/crawler.py
@@ -17,7 +17,7 @@ def extract_birth_place_and_date(link_governor_page, context: Context):
 
     response = context.fetch_html(link_governor_page)
 
-    path_to_information = '//*[@id="fContenido"]/div/div/div[2]/div/div/div/div[2]/div/div[1]/ul/li'
+    path_to_information = '//*[@id="fContenido"]//li'
 
     for element in response.xpath(path_to_information):
         text = ' '.join(element.itertext())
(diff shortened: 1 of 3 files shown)


jbothma:
Hi! This sounds like it's happening by design. Roughly speaking, if someone left a position more than 5 years ago, we don't consider them a PEP and we don't return an occupancy from make_occupancy to indicate that. https://zavod.opensanctions.org/peps/#only-emit-if-theyre-a-pep

[The criteria are a bit more complex than that](https://www.opensanctions.org/docs/pep/methodology/#types), but make_occupancy is there to take care of it. You just have to add a condition: only emit if make_occupancy returns an occupancy.

bgmello:
That makes sense, should we still emit the position and the person even though the occupancy is None? Or we ignore the entire item?

jbothma:
Nope - skip the person and position as well. 

Take a look at those docs - I probably forgot to share them with you before. But it'd be good to know if they're clear or need to be improved.

bgmello:
I didn't notice them before. Sorry about that. They look good.

One last question, they prefix the name with some abbreviations for example:

- Ing. (Ingeniero): Engineer
- Mtro. (Maestro): Master
- C.P. : ??

Should I remove them from the `name` property in `Person`? It's always the first word so it's not that hard. I can also add two name properties. What do you think would be the best?

jbothma:
It's really good to clean up the name to only be forname(s) and surname(s), yes! Thanks.

What I usually do is write a regular expression trying to do so in a way that initials won't get stripped off. Are these qualifications/titles or initials?

bgmello:
They are titles apparently:

 https://es.wikipedia.org/wiki/Esteban_Villegas_Villarreal 
 https://es.wikipedia.org/wiki/Ismael_Hernández_Deras
 
 M.C. -> Maestro en Ciencias -> Master of Science
 C.P. -> Contador Público -> Public Accountant
 
 Looks like when someone doesn't have any titles, they put "Sr." which is the same as "Mr.".
 
 But since I'm not completely sure of that, I think it would be best to keep the "full" name in some property just in case.

datasets/mx/governors/crawler.py
```diff
@@ -0,0 +1,124 @@
...
+        text = ' '.join(element.itertext())
+        if 'Lugar y fecha de nacimiento:' in text:
+            pattern = r'Lugar y fecha de nacimiento: ([^.]+\.) (\d{1,2}/\d{1,2}/\d{4})\.'
+            match = re.search(pattern, text)
+            if match:
+                birth_place = match.group(1)
+                birth_date = match.group(2)
+                return birth_place, birth_date
+
+    context.log.warning("Failed to identify birth place and date, link: {}".format(link_governor_page))
```
jbothma, line 32: does this currently log anything? Or is it currently always matching?

datasets/mx/governors/crawler.py
```diff
@@ -0,0 +1,124 @@
...
+                  .replace("https://www.conago.org.mx/miembros/detalle/", "")
+                  .replace("-", " ")
+                  .title())
+    person.add("name", clean_name)
+
+    birth_place, birth_date = extract_birth_place_and_date(link_governor_page, context)
+
+    if birth_place is not None:
+        person.add("birthPlace", birth_place)
+        person.add("birthDate", h.parse_date(birth_date, formats=["%d/%m/%Y"])[0])
```
jbothma, line 78: We usually just do this:
```suggestion
        person.add("birthDate", h.parse_date(birth_date, formats=["%d/%m/%Y"]))
```
Entity.add accepts lists.

datasets/mx/governors/crawler.py
```diff
@@ -0,0 +1,124 @@
...
+    response = context.fetch_html(context.data_url)
+
+    if response is None:
+        context.log.error("Error while fetching the HTML")
+        return
+
+    path_to_cards = '//*[@class="containerMixitup"]/div/div'
+    path_to_links = '//*[@class="containerMixitup"]/div/div/div/div/div/div[1]/a'
+    for (item, a_tag) in zip(response.xpath(path_to_cards), response.xpath(path_to_links)):
+        crawl_item(item, a_tag.get('href'), context)
```
jbothma, line 124: This is cool but if feels a little too clever and there's the slightest chance it misaligns data one day. I think a little safer and more completely obvious in how it works would simply to do the link xpath on the card element inside the loop

datasets/mx/governors/crawler.py
```diff
@@ -0,0 +1,124 @@
...
+    :param link_governor_page: Link to that governor page
+    :param context: The context object.
+    """
+
+    # The easiest way to get the data from the HTML is surprisingly from the text of the class
+    # All the information in the web site is written there in the form:
+    # col-xs-12 col-sm-6 col-md-4 pb15 mix {name of the governor} [{state}] {start_date} {end_date}
+    # We will then use a regex to extract the name, state, start_date and end_date
+
+    regex_pattern = r'col-xs-12 col-sm-6 col-md-4 pb15 mix (.*?) \[(.*?)\] (\d{2}/\d{2}/\d{4}) (\d{2}/\d{2}/\d{4})'
```
jbothma, line 50: Puttng all the data including dates formatted like this in the class is just wild. But let me not judge them. Does it all work if we do this, to tighten up the regex a little more and be sure we'll find out if they change things around?
```suggestion
    regex_pattern = r'^col-xs-12 col-sm-6 col-md-4 pb15 mix (.*?) \[(.*?)\] (\d{2}/\d{2}/\d{4}) (\d{2}/\d{2}/\d{4})$'
```

jbothma (changes requested):
Thank you!

Just a few small things - we don't want to ship with things that would be warnings permanently right away, but if there currently aren't any warnings coming out, it's very nice to get them when we stop finding the values we're looking for.

datasets/mx/governors/crawler.py
bgmello: It does log sometimes, there is some governors where this information is not present or is just the birth place or birth date. Apparently it's almost always on the older governors, the current ones have this information.

datasets/mx/governors/crawler.py
```diff
@@ -0,0 +1,124 @@
...
+
+    match = re.match(regex_pattern, input_html.get('class'))
+
+    if match is None:
+        context.log.warning("Unable to extract information from HTML element")
```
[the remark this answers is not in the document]
bgmello: Currently no, it was just to ensure that if they change the formatting we would know.

datasets/mx/governors/crawler.py
bgmello: I will do that

datasets/mx/governors/crawler.py
bgmello: Some examples:

- https://www.conago.org.mx/miembros/detalle/juan-jose-leon-rubio
- https://www.conago.org.mx/miembros/detalle/fernando-moreno-pena

jbothma: Ok - then I think we need to make this an info-level log so that it doesn't show up on the issues page - it's not something they're going to be fixing any time soon so it's not a transient issue.

datasets/mx/governors/crawler.py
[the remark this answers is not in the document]
jbothma: super nice, thanks

datasets/mx/governors/crawler.py
```diff
@@ -0,0 +1,124 @@
...
+    # Ing. Carlos Lozano de la Torre, where Ing. stands for Engineer
+    # We will add two name propreties, one with the "full" name
+    # and other with the clean name. The clean name
+    # can be retrived from the url to that governors page
+    person.add("name", name)
+    clean_name = (link_governor_page
+                  .replace("https://www.conago.org.mx/miembros/detalle/", "")
+                  .replace("-", " ")
+                  .title())
+    person.add("name", clean_name)
```
jbothma, line 73: Sorry, only noticed this now.
Instead of de-slugifying the name, I think a better approach is

1. take the name from the h4 in the card
2. strip `r"^([A-Z][a-z]*\\.)+ "` from it

Then we have the names capitalised and from the field we can expect quite safely to contain the name.

Adding it twice with and without ligatures feels more messy than the above.

bgmello: I can do like this where I can just strip the first word in the h4 tag. 

My concern is that if we only put the name that is in the title, when someone searches for the name of a governor in the open sanctions database it wouldn't match because the name property has the title, or this won't be a problem?

jbothma: So the regex I proposed should strip any abbreviated word from the start of the heading, so it should catch all the titles and nothing else. That would clean the name up and match nicely. But people with many names could well match in our API even if a title slips in. Still, I think it's easy enough to strip titles here.

Issue #72, opened 2021-01-13
EU EEAS Sanctions seems to be moved behind a login


ChrisjanWust:
The current url specified in the EU EEAS Sanctions config file, [https://data.europa.eu/euodp/es/data/dataset/consolidated-list-of-persons-groups-and-entities-subject-to-eu-financial-sanctions](https://data.europa.eu/euodp/es/data/dataset/consolidated-list-of-persons-groups-and-entities-subject-to-eu-financial-sanctions), is no longer available, returning a 404. It seems the datasets have been moved behind a login, as per [https://eeas.europa.eu/headquarters/headquarters-homepage_en/8442/Consolidated%20list%20of%20sanctions](https://eeas.europa.eu/headquarters/headquarters-homepage_en/8442/Consolidated%20list%20of%20sanctions).

ChrisjanWust:
The EU Open Data Portal has restored the old data location of the EU EEAS Sanctions, with a deprecated flag. The scraper is running successfully, but unsure how long the data will remain there.

pudo:
FWIW here's a link to the new version with my personal authentication token. The fact that my API key coincides with my name is probably a pure accident, because the EC could not possibly... 

https://webgate.ec.europa.eu/fsd/fsf/public/files/xmlFullSanctionsList_1_1/content?token=nlinfrie

oddobert:
In the new URL you can either get the XML or a CSV, has anyone made a new crawler? 
I am considering making one myself, however I am just atrting out with opernsanctions and I am likely going to spend considerable time understanding the API.

pudo:
I'm torn between a bottle of wine and doing the EU parser for the XML version tonight... But if you're gonna make a PR, let me know and the wine gets it :)

pudo:
I merged #60, which provides a really good starting point for this (and also resolves this issue)

PR #566, opened 2024-02-07
Skip entities not in the dataset

atehe:
Fixes [ Ukraine NAZK crawler](https://github.com/opensanctions/crawler-planning/issues/71)

@pudo  please take a look when available

### DIFF
--- a/datasets/ua/nabc_sanctions/crawler.py
+++ b/datasets/ua/nabc_sanctions/crawler.py
@@ -107,8 +107,6 @@ def json_listing(context: Context, url, name):
 def crawl_common(
     context: Context, entity: Entity, row: Dict[str, Any], is_sanctioned: bool
 ):
-    if is_sanctioned:
-        entity.add("topics", "sanction")
 
     country = row.pop("country", None)
     entity.add("country", COUNTRIES.get(country, country))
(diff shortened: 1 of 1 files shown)


pudo:
This looks good as a fix. Did you check that the missing entities are different than the ones in #72? (see my last comment)

atehe:
you mean the extra dataset in [this issue](https://github.com/opensanctions/crawler-planning/issues/72) right?

I  checked some of them manually, but didnt get any match.
I'm writing a script to go through them now and will let you know how it goes

pudo:
Brilliant! Don't throw that script away, it's going to basically implement #72 :-D

atehe:
Just ran the script and we gett only two matches of missing persons entity in the `/api/v6/shame/ ` endpoint.

atehe:
Awesome! you're right @jbothma. 
We get all except one company id found in the `/api/v6/company-warning/`.
I checked the same for person ids and found 2806/2854 of the missing person ids.

jbothma:
Awesome - thanks. I think let's use that to get related entities then, and info-level log the ones we can't get that way

atehe:
Alright...
Do i also add the companies and people in the other endpoints as  sanctioned entities as well?

pudo:
If they're one graph I would add them all to the dataset but not mark them with topic sanction

atehe:
Ok got it...
thank you

atehe:
@pudo I've added the other endpoints for persons and companies.
The crawler first checks them all to get a list of existing companies and  flags any one that is not in the list while scraping. 
The endpoints are cralwled similarly and the sanction role is added only to entities gotten from the main company and person endpoints

jbothma:
@atehe Did you check, do the company IDs on the different compoany endpoints intersect? And if so, do you get the same company back from the different endpoints? Or is it possible that you'd get two companies' details added to the dataset on the same ID, coming from different datasets?

atehe:
I just checked @jbothma and there is no intersections of the ids in the different endpoints.
I could add a tracker that checks if a company/person has been crawled to the crawler though or what do you think?

jbothma:
Yeah i think it's a good idea to warn and skip if they intersect. Thanks!

atehe:
@jbothma I've added a check to warn and skip persons/companies that have already been crawled

jbothma:
Thanks @atehe - After a night's rest I realised that check should raise instead of warn, otherwise we might still be emitting relationships to the wrong entity - something like this? https://github.com/opensanctions/opensanctions/commit/c16fc223877b385f316d154948b924ec0c525c5a#diff-172ca37747291b1feac175161689237ecc0af41cb77c5bdcd2349deeda02aa94R211

Also I see logs like 
```
[warning  ] Unexpected data found          [ua_nabc_sanctions] data={'reasoning_en': '"The company supplies metal-cutting machines, components and...
```

which is `context.audit` doing its job nicely. Could you add the reasons to the entity as notes?

@pudo should we still emit sanctions for the related non-sanctioned entities or skip that as well? Should they get the `sanction.linked topic`?

pudo:
Hm, `sanction.linked` will be assigned automatically - we don't need to do that in this dataset. I would definitely not emit a `Sanction` object, but perhaps we can find some other topic that captures the intent of these subcollections?

atehe:
Alright noted.

PR #567, opened 2024-02-07
adding Riigikogu members

bgmello:
Solving the issue https://github.com/opensanctions/crawler-planning/issues/12

### DIFF
--- a/datasets/ee/riigikogu/crawler.py
+++ b/datasets/ee/riigikogu/crawler.py
@@ -1,0 +1,102 @@
+from zavod import Context, helpers as h
+from zavod.logic.pep import categorise
+
+def get_members_url(context: Context) -> list:
+    """
+    Fetches the URLs for the pages of each member from the main page.
+
+    :param context: The context object.
+
+    :return: The URLs for the pages.
+    """
+
+    try:
+        main_website = context.fetch_html(context.data_url)
+
+    except:
+        context.log.error("Couldn't fetch main website")
+        return None
+
+    # this XPath corresponds to the a tags wit the links for a member page
+    xpath_to_a_tags = "//*[@id='main']/section/div[2]/div/ul/li/div/div/ul/li[1]/h3/a"
+
+    return [a_tag.get("href") for a_tag in main_website.xpath(xpath_to_a_tags)]
+
+def crawl_item(member_url: dict, context: Context):
+    """
+    Creates an entity, a position and a occupancy for each member,
+    we fetch the page of that member, extract the
+    relevant information and finally create
+    a Person, Position and Occupation.
+
+    :param member_url: The url for that member webpage
+    :param context: The context object.
+    """
+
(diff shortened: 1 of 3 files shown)


datasets/ee/riigikogu/ee_riigikogu.yaml
```diff
@@ -0,0 +1,26 @@
+title: Members of the Riigikogu
```
jbothma, line 1:
```suggestion
title: Estonia Members of the Riigikogu
```

datasets/ee/riigikogu/crawler.py
```diff
@@ -0,0 +1,102 @@
...
+    if occupancy is None:
+        return
+
+    context.emit(entity, target=True)
+    context.emit(position)
+    context.emit(occupancy)
+    
+
+def crawl(context: Context):
+    """
```
jbothma, line 87: I think this function is clear enough without a comment. No need to document the params and design of every function, especially `crawl`, every time. Especially with nice clear variable names like members_url and member_url

Now that I read the code, I see members_url is a list - so perhaps make that `member_urls`, then it's nice and clear what's going on even without comments

datasets/ee/riigikogu/crawler.py
```diff
@@ -0,0 +1,102 @@
...
+    :return: The URLs for the pages.
+    """
+
+    try:
+        main_website = context.fetch_html(context.data_url)
+
+    except:
+        context.log.error("Couldn't fetch main website")
+        return None
+
```
jbothma, line 19: no need for you to handle exceptions from http requests generally - an error in fetch_html will raise an exception handled by the crawler runner, resulting in an error log and alert on the website at https://www.opensanctions.org/issues/

datasets/ee/riigikogu/crawler.py
```diff
@@ -0,0 +1,102 @@
...
+    entity = context.make("Person")
+    entity.id = context.make_id(name)
+
+    try:
+        # the phone number apparently is the first contact in the list
+        # but to make sure we correctly get it, we based our xpath on the icon
+        phone_number = member_page_html.xpath('//*[@class="icon icon-tel"]/../text()')[0].replace(" ", "")
+        entity.add("phone", phone_number)
+    except:
+        context.log.warning("Couldn't find phone number")
```
jbothma, line 58: are we currently getting these warnings or is this just in case they change things in future?

jbothma (changes requested):
Nice! Just a few small things

datasets/ee/riigikogu/crawler.py
bgmello: No, currently everyone has a e-mail and phone number. It's just in case it changes in the future

jbothma: Lovely, thanks!

Issue #561, opened 2024-02-05
Add groups from UK Home Office PTO list


dhdaines:
As in https://github.com/opensanctions/crawler-planning/issues/14

PR #568, opened 2024-02-08
Add proscribed terrorist organizations from UK Terror Act

dhdaines:
Using the source legislation instead of the policy paper here.  I think this is all in order but please double-check as I haven't done a Sanction crawler yet!

Fixes: #561 https://github.com/opensanctions/crawler-planning/issues/14

### DIFF
--- a/datasets/gb/proscribed_orgs/crawler.py
+++ b/datasets/gb/proscribed_orgs/crawler.py
@@ -1,0 +1,130 @@
+"""
+Crawl Schedule 2 of the Terrorism Act 2000 and output entities for
+proscribed terrorist groups and organizations.
+"""
+
+import re
+from typing import List
+
+from zavod import Context, Entity
+from zavod import helpers as h
+from lxml.html import HtmlElement
+
+
+ADDITION_RE = re.compile(r"(added|inserted)\s+\(([^)]+)\)", re.IGNORECASE)
+INCLUDING1_RE = re.compile(r"(.*)\s*\(including\s+(.*)\)", re.IGNORECASE)
+INCLUDING2_RE = re.compile(r"(.*),?\s*including\s+(.*)", re.IGNORECASE)
+COMMA_AND_RE = re.compile(r",\s+|\s+and\s+", re.IGNORECASE)
+ALIAS_RE = re.compile(r"\s+\(([^)]+)\)")
+JUNK = " ,.;\t\r\n"
+DATE_FORMATS = [
+    "%d.%m.%Y"
+]
+
+
+def parse_comment(context: Context, text: str) -> List[str]:
+    """
+    Parse stack of  commentary from the act looking for the most
+    recent date when groups were added.
+    """
+    m = ADDITION_RE.search(text)
+    if m is None:
+        return []
+    return h.parse_date(m.group(2), DATE_FORMATS)
+
+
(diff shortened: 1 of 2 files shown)


datasets/gb/proscribed_orgs/crawler.py
```diff
@@ -0,0 +1,145 @@
...
+        for alias in aka[1:]:
+            context.log.info(f"Adding alias: {alias}")
+            h.apply_name(entity, alias, alias=True)
+    sanction = h.make_sanction(context, entity)
+    if add_date:
+        sanction.add("startDate", add_date)
+    # For some reason make_sanction uses the top-level URL and
+    # publisher.name, which are not exactly accurate.
+    sanction.add("authority", "UK Home Secretary")
+    sanction.add("sourceUrl", context.data_url)
```
jbothma, line 96: no need to add these - make_sanction already does.

jbothma (commented):
Thank you! Good find and super nice metadata.

The crawler code is excellent. That said, for future crawlers I'd suggest seeing if a simpler option is available. E.g. what I think could work in this case is iterating over children of `.main-content-container` from _List of proscribed international terrorist groups_ and aggregating content till you see the next h3 or end. Your parsing code is super clean and I think very understandable, but the alternative might read even more easily and you could stuff the narrative for each entity available on the more accessible page into `Entity.notes`

@pudo I'm not 100% sure - these don't look like they're standard financial sanctions so the current `crime.terror` seems good. In that case the `crime` collection is best, right?

datasets/gb/proscribed_orgs/crawler.py
dhdaines: Hi! There's a comment in there about this actually - the issue is that it takes the publisher as the authority, and the top-level URL as the source, which in this case is wrong, I think - the publisher is `legislation.co.uk` but the *authority* would be the Home Secretary.  Likewise, the descriptive top-level URL (the policy paper) is not the source URL for the data.

But I might not have written the metadata right - what would the best way to do this be?

dhdaines:
> The crawler code is excellent. That said, for future crawlers I'd suggest seeing if a simpler option is available. E.g. what I think could work in this case is iterating over children of `.main-content-container` from _List of proscribed international terrorist groups_ and aggregating content till you see the next h3 or end

Yes, I had some hesitation about this - when I saw that the source of the data was in the text of the law, I figured it might be better to go to the more authoritative source.  The parsing of the names is a bit easier with the law text, and we also get the exact start dates instead of just the month.

But on the other hand the policy paper does have the longer descriptions.

datasets/gb/proscribed_orgs/crawler.py
jbothma: That's awkward - it's right there :) Sorry I missed the comment.

So Entity.add will add these values and leave the incorrect ones. I think these two URLs for this is quite handy actually, but I see there's an `Entity.set` you could use for `authority` - it's also used in eu_sanctions_map in this way.

jbothma:
```suggestion
    sanction.set("authority", "UK Home Secretary")
    sanction.add("sourceUrl", context.data_url)
```

pudo:
@jbothma Agree regarding the classifications, `crime.terror` topic and `crime` collection. Thanks for reflecting on this.

PR #571, opened 2024-02-09
Adding Austrain PEPs

bgmello:
Related to issue https://github.com/opensanctions/crawler-planning/issues/66

### DIFF
--- a/datasets/at/pep/at_pep.yaml
+++ b/datasets/at/pep/at_pep.yaml
@@ -1,0 +1,29 @@
+title: Austria's Politicians 
+entry_point: crawler.py
+prefix: at-pep
+coverage:
+  frequency: daily
+load_db_uri: ${OPENSANCTIONS_DATABASE_URI}
+summary: >
+  List of politicians and others officials from Austria.
+description: |
+  This dataset includes continuously updated dossiers of
+  all members of the National Council, members of the Federal
+  Government and the Bundesrat as well as all state governments
+  and state parliaments.  Other dossiers concern
+  the Federal President, the general secretaries of the
+  ministries, the Ombudsmen and many more.
+publisher:
+  name: Meine Abgeordneten
+  description: |
+   The “My Members of Parliament” platform is
+   a private initiative that was founded in 2011 by Respect.net
+   and operated together with WOTI World of Tomorrow Institute
+   (formerly Act.Now) from 2017 to 2021
+  url: https://www.meineabgeordneten.at/
+  official: false
+  country: "at"
+url: "https://www.meineabgeordneten.at/Abgeordnete"
+data:
+  url: https://www.meineabgeordneten.at/Abgeordnete
+  format: HTML
--- a/datasets/at/pep/crawler.py
+++ b/datasets/at/pep/crawler.py
@@ -1,0 +1,68 @@
+from dateutil.parser import parse
+
+from zavod import Context, helpers as h
+from zavod.logic.pep import categorise
+
+
+def parse_german_date(date_string):
+    return parse(date_string, dayfirst=True, locale='de')
+
+def crawl_item(url_info_page: str, context: Context):
+
+    info_page = context.fetch_html(url_info_page)
+
+    first_name = info_page.xpath("//span[@itemprop='http://schema.org/givenName']/text()")[0]
+    last_name = info_page.xpath("//span[@itemprop='http://schema.org/familyName']/text()")[0]
+
+    person = context.make("Person")
+    person.id = context.make_id(first_name, last_name)
+
+    person.add("firstName", first_name)
+    person.add("lastName", last_name)
+
+    person.add("website", url_info_page)
+
+    try:
+        birth_date_in_german = info_page.xpath("//span[@itemprop='birthDate']/text()")[0]
+        person.add("birthDate", parse_german_date(birth_date_in_german))
+    except:
+        pass
+
+    try:
+        person.add("email", info_page.xpath("//a[@itemprop='http://schema.org/email']/text()")[0])
+    except:
+        pass
+
+    try:
+        person.add("phone", info_page.xpath("//a[@itemprop='http://schema.org/telephone']/text()")[0]).replace(' ', '')
+    except:
+        pass
+
+    position_name = info_page.xpath('//*[@id="app"]/main/div/div[4]/div[1]/div/div/div[2]/p/text()')[0]
+
+    position = h.make_position(context, position_name, country="at")
+    categorisation = categorise(context, position, is_pep=True)
+
+    occupancy = h.make_occupancy(
+            context,
+            person,
+            position,
+            no_end_implies_current=True,
+            categorisation=categorisation,
+        )
+
+    if occupancy is None:
+        context.emit(person, target=True)
+        context.emit(position)
+        context.emit(occupancy)
+    
+
+def crawl(context: Context):
+
+    response = context.fetch_html(context.data_url)
+
+    # XPath to the url for the pages of each politician
+    xpath_politician_page = '//*[@id="app"]/main/div[2]/div[1]/div/div[2]/a/@href'
+
+    for item in response.xpath(xpath_politician_page):
+        crawl_item(item, context)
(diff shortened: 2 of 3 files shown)


datasets/at/pep/at_pep.yaml
```diff
@@ -0,0 +1,29 @@
...
+load_db_uri: ${OPENSANCTIONS_DATABASE_URI}
+summary: >
+  List of politicians and others officials from Austria.
+description: |
+  This dataset includes continuously updated dossiers of
+  all members of the National Council, members of the Federal
+  Government and the Bundesrat as well as all state governments
+  and state parliaments.  Other dossiers concern
+  the Federal President, the general secretaries of the
+  ministries, the Ombudsmen and many more.
```
jbothma, line 15:
```suggestion
  This dataset identifies
  all members of the National Council, members of the Federal
  Government and the Bundesrat as well as all state governments
  and state parliaments.  Other significant posts in this dataset include
  the Federal President, the general secretaries of the
  ministries, the Ombudsmen and many more.
```

The source website has dossiers, or perhaps dossier is a bad google translation from the German - we just have some cursory information here. So I don't think `dossier` quite fits.

datasets/at/pep/crawler.py
```diff
@@ -0,0 +1,68 @@
...
+        context.emit(position)
+        context.emit(occupancy)
+    
+
+def crawl(context: Context):
+
+    response = context.fetch_html(context.data_url)
+
+    # XPath to the url for the pages of each politician
+    xpath_politician_page = '//*[@id="app"]/main/div[2]/div[1]/div/div[2]/a/@href'
```
jbothma, line 65: Try to write a selector what makes as little assumption about the HTML as possible, while not accidentally emitting the wrong thing as if it's fact.

Something to bear in mind is that if a crawler emits zero entities, we will get an error message at https://www.opensanctions.org/issues/ and take a look. So 

In this case, your xpath could be `'//*[contains(@class, "abgeordneter")]'` which is as close as we get with `lxml` to doing the CSS selector `.abgeordneter`, and it's pretty cool to be able to select "MP" elements.

datasets/at/pep/crawler.py
```diff
@@ -0,0 +1,68 @@
...
+    first_name = info_page.xpath("//span[@itemprop='http://schema.org/givenName']/text()")[0]
+    last_name = info_page.xpath("//span[@itemprop='http://schema.org/familyName']/text()")[0]
+
+    person = context.make("Person")
+    person.id = context.make_id(first_name, last_name)
+
+    person.add("firstName", first_name)
+    person.add("lastName", last_name)
+
+    person.add("website", url_info_page)
```
jbothma, line 23: I think this fits better in the `sourceUrl` property, while `website` is more suited to their own website, if they had one.

datasets/at/pep/crawler.py
```diff
@@ -0,0 +1,68 @@
...
+        person.add("email", info_page.xpath("//a[@itemprop='http://schema.org/email']/text()")[0])
+    except:
+        pass
+
+    try:
+        person.add("phone", info_page.xpath("//a[@itemprop='http://schema.org/telephone']/text()")[0]).replace(' ', '')
+    except:
+        pass
+
+    position_name = info_page.xpath('//*[@id="app"]/main/div/div[4]/div[1]/div/div/div[2]/p/text()')[0]
```
jbothma, line 41: I think a less brittle way of selecting this would be

1. select the parent of the element containing the name
2. select the descendant of that parent which is a p tag with the bold class

index-based selectors (e.g. `div[4]`) can easily fail in surprising ways and either not select anything, or select something you didn't mean to, which is worse.

-----

Just for comparison, what would also be better than the index-based xpath, but more complicated and I don't think is the right thing for this iteration of this source, is looping over `.functionsText` descendants of `#mandate` for the position names? The date seems to be in the previous sibling, and you can probably add a check whether it contains .aktiv or .inaktiv

The benefit of this approach isn't just more detail about the position, it seems to also tap a bit more into a semantic selection of the element and a bit more safety in pulling out and emitting the position name

datasets/at/pep/crawler.py
```diff
@@ -0,0 +1,68 @@
...
+    person.add("firstName", first_name)
+    person.add("lastName", last_name)
+
+    person.add("website", url_info_page)
+
+    try:
+        birth_date_in_german = info_page.xpath("//span[@itemprop='birthDate']/text()")[0]
+        person.add("birthDate", parse_german_date(birth_date_in_german))
+    except:
+        pass
```
jbothma, line 29: There's a matra "use exceptions for exceptional circumstances" - I think a better pattern and more consistent with our codebase would be

```suggestion
    birth_date_in_german = info_page.find("//span[@itemprop='birthDate']/text()")
    if birth_date_in_german:
        person.add("birthDate", parse_german_date(birth_date_in_german))
```

lxml el.find returns only the first match. You only need el.xpath when you want to do `contains()` because it's a bit more powerful, for some reason. Most of the time `find` and `findall` are sufficient.

jbothma (changes requested):
Thanks!

Mostly comments on using a style a bit more like ours, and possibly a bit more predictable and maintainable

datasets/at/pep/crawler.py
bgmello: Nice, I didn't know about the `find` method.

datasets/at/pep/crawler.py
bgmello: Thanks for the feedback, I was just copying the chrome dev tools xpath, but I will try to use more generic paths in the other crawlers. Let me know if it looks good now.

datasets/at/pep/crawler.py
bgmello: I got an error while trying to run the crawl because of the find method:

Screenshot 2024-02-15 at 10 07 19

jbothma: Hi @bgmello - sorry about the delayed response.
It looks like you need `el.find(".//tagname...`
i.e. a dot before the // to start at the current element.

datasets/at/pep/crawler.py
```diff
@@ -0,0 +1,67 @@
...
+
+    birth_date_in_german = info_page.find("//span[@itemprop='birthDate']/text()")
+    if birth_date_in_german:
+        person.add("birthDate", parse_german_date(birth_date_in_german))
+
+    try:
+        person.add("email", info_page.xpath("//a[@itemprop='http://schema.org/email']/text()")[0])
+    except:
+        pass
+
```
jbothma, line 33: could you also change the other `try...except` instances to `if`s?

jbothma:
Thanks for the improvements @bgmello!

I was about to merge it when I spotted some things that I missed earlier, and I think are worth changing. Would you be up for adding in my changes to your branch (cherry pick or copy, whatever works for you) and picking up where I left off? See my changes at https://github.com/opensanctions/opensanctions/compare/e8f0796c87ce6d7b52d5611fc37dd766555e5932...f1c3da21c491092020ba03f8e8c5cdff111e685c

Some things I changed already:

- I think a better name is at_meine_abgordneten similarly to the german abgeordnetenwatch - since it describes that this is about legislators specifically and not really a broader PEP source
- it wasn't emitting anything - the condition was the wrong way around
- I'm nervous to change the system-wide encoding to parse the date, so I added a language month mapping like we've done in other crawlers
- the position data in the mandate list is really rich, and first I thought it wasn't worth the effort, but it seems to be quite parsable actually - so I started switching to extracting position information from there instead of the header, but there's more work to be done
- the bare list page URL seems to show only data for one party? Maybe the page became to heavy to default to all? Anyway I changed the URL to one which seems to show all parties. To be sure, I also added assertions on the number of people and positions discovered. Let's manually make sure the numbers are right when the crawler can do a complete crawl - I usually make the minimum about 10% under the current total (in case they can fluctuate) and the maximum about 20% over the current number. Sorry this feature isn't documented yet.

What still needs to be done:

- fix up date parsing - it runs for a while then fails - handle their ?'s meaning they don't know, and it doesn't mean current.
- strip party names from position names
- add the source links (so far I've only seen screenshot links) to the sourceUrl of the occupancy
- make sure the entity count assertions in metadata are close enough to the final numbers.

PR #577, opened 2024-02-13
Sanctioned entities from EU regulation 2023/2878 (decision 2023/2874)

dhdaines:
See https://github.com/opensanctions/crawler-planning/issues/74

Note one FIXME: not certain how to best use lookups to fix up a company name that was incorrectly extracted.

### DIFF
--- a/datasets/eu/reg_2878_sanctions/crawler.py
+++ b/datasets/eu/reg_2878_sanctions/crawler.py
@@ -16,11 +16,6 @@ def crawl_row(context: Context, row: Dict[str, str]):
     name = row.pop("Name").strip()
     country = row.pop("Country").strip()
 
-    # FIXME: Not sure of the correct way to automate this with
-    # lookups, is that documented anywhere? (using the "companies" key
-    # gives an exception that cannot be understood)
-    if name == "LLC CST":
-        name = "Zala Aero Group"
 
     context.log.info(f"Processing row ID {row_id}: {name}")
     entity = context.make(entity_type)
--- a/datasets/eu/reg_2878_sanctions/crawler.py
+++ b/datasets/eu/reg_2878_sanctions/crawler.py
@@ -1,0 +1,53 @@
+"""
+Add supplemental list of EU sanctions from annex in Regulation  2023/2878
+"""
+
+import csv
+from typing import Dict, Iterable
+
+import zavod.helpers as h
+from zavod import Context
+
+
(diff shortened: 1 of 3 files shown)


pudo:
Looks great, I also fixed the two values in the source sheet.

pudo (approved):

datasets/eu/reg_2878_sanctions/crawler.py
```diff
@@ -0,0 +1,53 @@
...
+    row_id = row.pop("List ID").strip(" \t.")
+    entity_type = row.pop("Type").strip()
+    name = row.pop("Name").strip()
+    country = row.pop("Country").strip()
+
+    # FIXME: Not sure of the correct way to automate this with
+    # lookups, is that documented anywhere? (using the "companies" key
+    # gives an exception that cannot be understood)
+    if name == "LLC CST":
+        name = "Zala Aero Group"
```
jbothma, line 23: I think in this case fixing in the spreadsheet made most sense. Matching on `Zala Aero Group` in the type.country lookup and returning Russia wouldn't have been ideal either, since Russia isn't actually mentioned in there.

Since this is manually extracted, the best fix was in the spreadsheet.

If we were scraping the data programmatically, I might write the code in such a way that my regex doesn't match on two sets of parens. So when the regex matches, I'd use the captured groups. If it doesn't match, I'd put the whole unmatched text into [context.lookup](https://github.com/opensanctions/opensanctions/blob/main/datasets/fr/tresor/crawler.py#L54) and [manually split the values I want from it in lookups](https://github.com/opensanctions/opensanctions/blob/main/datasets/fr/tresor/fr_tresor_gels_avoir.yml#L59)
```suggestion
```

datasets/eu/reg_2878_sanctions/eu_reg_2878_sanctions.yml
```diff
@@ -0,0 +1,57 @@
...
+  type.country:
+    lowercase: true
+    normalize: true
+    options:
+      - match:
+          - Autonomous Republic of Crimea, illegally annexed by Russia
+        value: ua-cri
+      - match:
+          - Zala Aero Group
+        value: ru
```
jbothma, line 57: Dropping this while it's fixed in the google sheet.

```suggestion
```

jbothma (commented):
awesome - thank you!

PR #589, opened 2024-02-21
Skip and log self-relations in RuPEP

jbothma:
self relations for people pretty clearly mistakes - mostly family or associate of self

Screenshot 2024-02-21 at 12 33 11

Companies about ownership/control of self - just need to confirm there isn't some meaning in this I'm missing

Screenshot 2024-02-21 at 12 33 45


pudo:
Wow this is a mess. But to be clear: if it's happening with the raw (unresolved) IDs of the entities, we should definitely skip it.

pudo (approved):

PR #590, opened 2024-02-21
Latvia list of sanction

atehe:


datasets/lv/sanction_list/lv_sanction_list.yml
```diff
@@ -0,0 +1,28 @@
...
+  It's important to note that the dataset only includes information on financial restrictions imposed on these individuals or entities and does not cover other types
+  of sanctions like restrictions on the circulation of goods.
+
+url: https://sankcijas.fid.gov.lv
+publisher:
+  name: Finanšu izlūkošanas dienests
+  description:
+    The Financial Intelligence Unit maintains information on the subjects of sanctions. The information is gotten from the [United Nations Security Council](https://www.un.org/securitycouncil/content/un-sc-consolidated-list)
+    and the [European Union(https://webgate.ec.europa.eu).
+    The Financial Intelligence Unit provide actual information without making any changes to the original source and is not responsible for any gaps or errors that occurs from the creators.
```
jbothma, line 23: I think you can use this description straight from https://fid.gov.lv/en instead - that describes the publisher better for me. You can prefix the lines with > to indicate it's a quote. And then use https://fid.gov.lv/en as the publisher link

> The Financial Intelligence Unit of Latvia is an independent authority and it's responsibility lies with prevention of money laundering in Latvia, with the goal of eliminating the possibility of using the Latvian financial system to launder money or finance terrorism.

datasets/lv/sanction_list/lv_sanction_list.yml
```diff
@@ -0,0 +1,28 @@
+name: lv_sanction_list
+title: Latvia List of Sanctions
+entry_point: crawler.py
+prefix: lv-sanction
+summary: List of Sanctions in Latvia
+description:
+  This dataset contains lists of individuals or entities subject to international or national sanctions as specified by laws in Latvia
+
+  International sanctions are imposed on subject in accordance to international laws as set by organizations such as the UN, EU or other international bodies.
+
```
jbothma, line 10: we're only doing the national sanctions, right? So I think let's change this to reflect that - no need even to mention that they also publish UN and EU sanctions

datasets/lv/sanction_list/lv_sanction_list.yml
```diff
@@ -0,0 +1,28 @@
+name: lv_sanction_list
+title: Latvia List of Sanctions
```
jbothma, line 2:
```suggestion
name: lv_fiu_sanctions
title: Latvia FIU Sanctions
```

could you update the file and directory names too?

datasets/lv/sanction_list/crawler.py
```diff
@@ -0,0 +1,116 @@
...
+    for node in doc.findall(".//Entity"):
+
+        entity_type = node.findtext(".//Type")
+
+        if entity_type.lower() == "fp":
+            entity = crawl_person(context, node)
+        elif entity_type.lower() == "jp":
+            entity = crawl_organization(context, node)
+        else:
+            raise RuntimeError("Cannot not infer entity type")
```
jbothma, line 106: did you see any docs describing these types or are you inferring from the data? it's probably a reasonable assumption but I'm just a tad nervous with 3 entries.

datasets/lv/sanction_list/crawler.py
```diff
@@ -0,0 +1,116 @@
...
+        address_entity.id = context.make_slug(
+            whole_address, prefix="lv-sanction-company-address"
+        )
+        address_entity.add("city", city)
+        address_entity.add("street", street)
+        address_entity.add("state", country)
+        address_entity.add("full", whole_address)
+        address_entity.add("remarks", address_remark)
+
+        entity.add("addressEntity", address_entity)
```
jbothma, line 87: you'll have to emit the address entity too.

datasets/lv/sanction_list/crawler.py
```diff
@@ -0,0 +1,116 @@
...
+        whole_address = address_node.findtext(".//AddressWhole")
+        address_remark = address_node.findtext(".//AddressRemark")
+        street = address_node.findtext(".//AddressStreet")
+        city = address_node.findtext(".//AddressCity")
+        country = address_node.findtext(".//AddressCountry")
+
+        address_entity = context.make("Address")
+        address_entity.id = context.make_slug(
+            whole_address, prefix="lv-sanction-company-address"
+        )
```
jbothma, line 80: could you use h.make_address for the address entity? then it'll make an id for you too.

datasets/lv/sanction_list/crawler.py
```diff
@@ -0,0 +1,116 @@
...
+
+def make_person_id(id: str) -> str:
+    return f"lv-sanction-person-{id}"
+
+
+def make_company_id(id: str) -> str:
+    return f"lv-sanction-company-{id}"
+
+
+def crawl_person(context: Context, node: ElementTree):
```
jbothma, line 15: let's add support for zero or more Alias nodes as well

jbothma (changes requested):
Looking good so far. I added some comments with improvements we can make.

Thank you!

datasets/lv/sanction_list/crawler.py
atehe: Lol, I understand.
I'm actually just infering.
How do you think we should proceed with it?

datasets/lv/sanction_list/crawler.py
atehe: Ok noted

datasets/lv/sanction_list/crawler.py
jbothma: ah I see in the notice,  fiziska persona and  juridiska persona
so we're good.

pudo:
@atehe since this will be part of the core sanctions set I've made another edit. Maybe it's useful to have a look at the latest commits here: https://github.com/opensanctions/opensanctions/commits/14db4a367c46c6ff184b8480b4480f8e0397f58b/datasets/lv/fiu_sanctions/crawler.py

atehe:
got it...
it is helpful

PR #614, opened 2024-03-02
French Senators

dhdaines:
Crawler for the members of the French Senate. See https://github.com/opensanctions/crawler-planning/issues/84

Note that to get occupancy dates we would need to download another 400k CSV file.  This is easy to do, not sure if there's a way to get it in the metadata though?  Let me know.

### DIFF
--- a/datasets/fr/senat/crawler.py
+++ b/datasets/fr/senat/crawler.py
@@ -96,3 +96,6 @@ def crawl(context: Context):
         decomment = (spam for spam in infh if spam[0] != "%")
         reader = csv.DictReader(decomment)
         crawl_csv(context, reader)
+
+    # NOTE: To get start dates we need this auxiliary CSV:
+    # https://data.senat.fr/data/senateurs/ODSEN_ELUSEN.csv
--- a/datasets/fr/senat/crawler.py
+++ b/datasets/fr/senat/crawler.py
@@ -1,0 +1,98 @@
+"""
+Crawler for active French senators.
+"""
+
+import csv
+from typing import Dict, Iterable
+
+from zavod import Context
+from zavod import helpers as h
+from zavod.logic.pep import categorise
+
+UNUSED_FIELDS = [
+    "Qualité",
+    "Groupe politique",
+    "Type d'app au grp politique",
+    "Commission permanente",
+    "Circonscription",
+    "Fonction au Bureau du Sénat",
+    "PCS INSEE",
+    "Catégorie professionnelle",
+    "Description de la profession",
+]
+DATE_FORMATS = [
+    "%Y-%m-%d %H:%M:%S.0",
+    "%Y-%m-%d %H:%M:%S",
+    "%Y-%m-%d",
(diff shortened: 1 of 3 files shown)


datasets/fr/senat/fr_senat.yaml
```diff
@@ -0,0 +1,56 @@
...
+  frequency: daily
+  start: 2024-03-01
+entry_point: crawler.py
+summary: >-
+  French Senators.
+description: |
+  This dataset lists the serving members of the French Senate.
+
+  The data is collected from the open data platform for the Senate,
+  which is the upper legislative chaber of the French government.  The
```
jbothma, line 14:
```suggestion
  which is the upper legislative chamber of the French government.  The
```

datasets/fr/senat/crawler.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def crawl_csv(context: Context, reader: Iterable[Dict[str, str]]):
+    """Process the CSV data"""
+    for row in reader:
+        crawl_row(context, row)
+
+
+def crawl(context: Context):
+    """Retrieve list of senators as CSV and emit POI entities for
```
jbothma, line 90:
```suggestion
    """Retrieve list of senators as CSV and emit PEP entities for
```
POI is actually its own distinct topic in OpenSanctions

datasets/fr/senat/crawler.py
```diff
@@ -0,0 +1,101 @@
...
+    if occupancy is not None:
+        context.emit(person, target=True)
+        context.emit(position)
+        context.emit(occupancy)
+
+
+def crawl_csv(context: Context, reader: Iterable[Dict[str, str]]):
+    """Process the CSV data"""
+    for row in reader:
+        crawl_row(context, row)
```
jbothma, line 86: We tend to just do this loop inside `crawl()` unless there's particularly interesting logic, so I think it could be inlined just to be concise.

datasets/fr/senat/crawler.py
```diff
@@ -0,0 +1,101 @@
...
+        topics=["gov.national", "gov.legislative"],
+    )
+    # is_pep=True because we expect all senators to be PEPs
+    categorisation = categorise(context, position, is_pep=True)
+    if not categorisation.is_pep:
+        # Unlikely that this would ever happen!
+        context.log.warning(f"Senator {given_name} {family_name} is not PEP")
+        return
+    person = context.make("Person")
+    person.id = context.make_id(senid)
```
jbothma, line 61: this could be context.make_slug(sen_id)

jbothma (commented):
I made the couple of tweaks in the comments while checking things locally - thanks!

datasets/fr/senat/crawler.py
dhdaines: Makes sense!  It doesn't really make the code clearer or more reusable to separate out `crawl_row` - will remember for the next one.

PR #615, opened 2024-03-02
Add Otkhozoria–Tatunashvili List

dhdaines:
Add sanctioned individuals from the OT list.  Attempt to provide coverage of different versions of their names (please verify this...)

Note that it isn't clear from the "demographic data" column whether the places listed are place of birth or current residence.  For this reason I haven't done anything with them.  The dates are clearly dates of birth though.

Names are not transliterated since the downstream entity matching framework will handle that.  Patronymics, of the sort "დავით **ჯემალის ძე** გურწიევი" (Davit **Jemalis dze** Gurtsievi) are extracted as they are not generally used in Georgian names, but could be useful for disambiguating (presumably they are used equivalently to Russian patronyms here)

See https://github.com/opensanctions/crawler-planning/issues/77

### DIFF
--- a/datasets/ge/ot_list/crawler.py
+++ b/datasets/ge/ot_list/crawler.py
@@ -6,7 +6,6 @@ the individuals on the Otkhozoria–Tatunashvili List.
 import re
 from typing import Dict
 
-from transliterate import translit
 from zavod import Context, Entity
 from zavod import helpers as h
 from lxml.html import HtmlElement
@@ -23,7 +22,6 @@ UNUSED_FIELDS = [
 ]
 DATE_FORMATS = ["%d.%m.%Y", "%Y"]
 PATROYNMIC = re.compile(r"\b(\S+)\s+ძე\s+")
-RUSSIANIZE = re.compile(r"([oe]v)i$")
 
 
 def extract_name(context: Context, person: Entity, name: str):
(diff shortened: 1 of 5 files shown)


zavod/setup.py
```diff
@@ -46,6 +46,7 @@
         "requests_oauthlib",
         "sqlalchemy[mypy]",
         "structlog",
+        "transliterate",
```
pudo, line 49: Let's please not include another transliteration mechanism - see other comments.

datasets/ge/ot_list/ge_ot_list.yml
```diff
@@ -0,0 +1,42 @@
...
+    statements.
+assertions:
+  min:
+    schema_entities:
+      Person: 33
+    country_entities:
+      ge: 1
+  max:
+    schema_entities:
+      Person: 33
```
pudo, line 42: Is this going to be a bit restrictive? Maybe we allow for up to 100?

datasets/ge/ot_list/crawler.py
```diff
@@ -0,0 +1,94 @@
...
+def extract_name(context: Context, person: Entity, name: str):
+    """Parse a Georgian name slightly and transliterate."""
+    patronym = None
+    m = PATROYNMIC.search(name)
+    if m is not None:
+        patronym = m.group(1)
+        name = name[:m.start()] + name[m.end():]
+        context.log.debug(f"Patroynmic: {m.group(1)}")
+    h.apply_name(person, full=name, patronymic=patronym, lang="geo")
+    transliterated = translit(name, "ka", reversed=True).title()
```
pudo, line 38: We don't need to transliterate names here. The rest of our stack is intended to work across unicode alphabets, including the name matchers. Georgian transliteration is a complete mess, so the results of this would also be questionable at best.

pudo (changes requested):
Thanks for submitting this crawler. It all looks pretty good, but I would prefer if we can get rid of the transliterated names and the extra depdendency. 

For the record, we have a transliterator that could be used: `normality.ascii_text` and `normality.latin_text`.

datasets/ge/ot_list/crawler.py
dhdaines: Ah ok - I was wondering if this was the case.  The hack to turn -ovi/-evi into -ov/-ev is unnecessary too I imagine?

Separating out the patronym, on the other hand, seems useful to me as Georgian names usually don't include it.

datasets/ge/ot_list/ge_ot_list.yml
dhdaines: Yes :) in theory it should be expanded (current Georgian government might not be very inclined to do so but that could change)

zavod/setup.py
dhdaines: Understood!

PR #621, opened 2024-03-05
Fix position descendant timeout

jbothma:
This unrolls the top level of the `?position wdt:P31/wdt:P279* position` assertion:

1. we find all direct subclasses of position
2. we find all wdt:P31/wdt:P279* of them
3. we find all wdt:P31 of position


jbothma:
Before merging this I wanna see how many fewer PEPs we'd have if we dropped the `/wdt:P279*` part of the constraint.

I think this is opening us up to lots of nonsensical things people are making descendants of position, and and I'm not sure if there's sufficient benefit.

PR #624, opened 2024-03-06
Adding Mexico Deputies crawler

bgmello:
Fixes opensanctions/crawler-planning#20

### DIFF
--- a/datasets/mx/deputies/crawler.py
+++ b/datasets/mx/deputies/crawler.py
@@ -1,0 +1,66 @@
+import re
+import requests
+
+from zavod import Context, helpers as h
+from zavod.logic.pep import categorise
+
+def crawl_item(input_dict: dict, context: Context):
+
+    entity = context.make("Person")
+    entity.id = context.make_id(input_dict["NombreCompleto"])
+
+    entity.add("lastName", input_dict.pop("Nombre"))
+    entity.add("firstName", input_dict.pop("PrimerApellido"))
+    entity.add("name", input_dict.pop("NombreCompleto"))
+
+    if input_dict["Telefono"] is not None:
+        entity.add("phone", "+52"+input_dict.pop("Telefono"))
+
+    entity.add("email", input_dict.pop("Correo"))
+
+    position = h.make_position(context, "Member of the Chamber of Deputies", country="mx")
+    categorisation = categorise(context, position, is_pep=True)
+
+    occupancy = h.make_occupancy(
+            context,
+            entity,
+            position,
+            True,
+            categorisation=categorisation,
+        )
+
+    if occupancy is None:
+        return
+
+    context.emit(entity, target=True)
+    context.emit(position)
+    context.emit(occupancy)
+    
+
+def crawl(context: Context):
+
+    HEADER = {
+    'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; rv:109.0) Gecko/20100101 Firefox/115.0',
+    'Accept': 'application/json, text/plain, */*',
+    'Accept-Language': 'en-US,en;q=0.5',
+    'Referer': 'https://web.diputados.gob.mx/',
+    'Content-Type': 'application/json',
+    'Origin': 'https://web.diputados.gob.mx',
+    'Connection': 'keep-alive',
+    'Sec-Fetch-Dest': 'empty',
+    'Sec-Fetch-Mode': 'cors',
+    'Sec-Fetch-Site': 'same-site',
+    }
+
+    json_data = {
+    'operationName': None,
+    'variables': {},
+    'query': '{\n  allDiputados {\n    Oid\n    Nombre\n    PrimerApellido\n    SegundoApellido\n    NombreCompleto\n    Estado\n    Partido\n    Distrito\n    Legislacion\n    PrimerApellido\n    CabeceraMunicipal\n    Suplente\n    id_dip\n    IdDiputado\n    Correo\n    Telefono\n    TipoEleccion\n    Licencia\n    __typename\n}\n}\n',
+    }
+
+    # We are going to query the graphql endpoint to retrive the data and then process each item
+    response = requests.post('https://micrositios.diputados.gob.mx:4001/graphql', headers=HEADER, json=json_data).json()
+
+    for item in response["data"]["allDiputados"]:
+        crawl_item(item, context)
+
--- a/datasets/mx/deputies/crawler.py
+++ b/datasets/mx/deputies/crawler.py
@@ -1,33 +1,39 @@
-import re
-import requests
-
 from zavod import Context, helpers as h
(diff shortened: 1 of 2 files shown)


datasets/mx/deputies/crawler.py
```diff
@@ -0,0 +1,66 @@
...
+    'operationName': None,
+    'variables': {},
+    'query': '{\n  allDiputados {\n    Oid\n    Nombre\n    PrimerApellido\n    SegundoApellido\n    NombreCompleto\n    Estado\n    Partido\n    Distrito\n    Legislacion\n    PrimerApellido\n    CabeceraMunicipal\n    Suplente\n    id_dip\n    IdDiputado\n    Correo\n    Telefono\n    TipoEleccion\n    Licencia\n    __typename\n}\n}\n',
+    }
+
+    # We are going to query the graphql endpoint to retrive the data and then process each item
+    response = requests.post('https://micrositios.diputados.gob.mx:4001/graphql', headers=HEADER, json=json_data).json()
+
+    for item in response["data"]["allDiputados"]:
+        crawl_item(item, context)
```
jbothma, line 65: Let's use the configured http session just for consistency, if nothing else.
Let's also raise_for_status() to discover http errors more easily.

```suggestion
    # We are going to query the graphql endpoint to retrive the data and then process each item
    ```suggestion
    response = context.http.post('https://micrositios.diputados.gob.mx:4001/graphql', headers=HEADER, json=json_data)
    response.raise_for_status()
    results = response.json()

    for item in results["data"]["allDiputados"]:
        crawl_item(item, context)
```

datasets/mx/deputies/crawler.py
```diff
@@ -0,0 +1,66 @@
...
+    'Connection': 'keep-alive',
+    'Sec-Fetch-Dest': 'empty',
+    'Sec-Fetch-Mode': 'cors',
+    'Sec-Fetch-Site': 'same-site',
+    }
+
+    json_data = {
+    'operationName': None,
+    'variables': {},
+    'query': '{\n  allDiputados {\n    Oid\n    Nombre\n    PrimerApellido\n    SegundoApellido\n    NombreCompleto\n    Estado\n    Partido\n    Distrito\n    Legislacion\n    PrimerApellido\n    CabeceraMunicipal\n    Suplente\n    id_dip\n    IdDiputado\n    Correo\n    Telefono\n    TipoEleccion\n    Licencia\n    __typename\n}\n}\n',
```
jbothma, line 58: this could be more readable as a multiline string
```python
'''This is a
multiline
string'''
```

datasets/mx/deputies/crawler.py
```diff
@@ -0,0 +1,66 @@
...
+    'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; rv:109.0) Gecko/20100101 Firefox/115.0',
+    'Accept': 'application/json, text/plain, */*',
+    'Accept-Language': 'en-US,en;q=0.5',
+    'Referer': 'https://web.diputados.gob.mx/',
+    'Content-Type': 'application/json',
+    'Origin': 'https://web.diputados.gob.mx',
+    'Connection': 'keep-alive',
+    'Sec-Fetch-Dest': 'empty',
+    'Sec-Fetch-Mode': 'cors',
+    'Sec-Fetch-Site': 'same-site',
```
jbothma, line 52: can we drop any of these headers without breaking things? I suspect accept language and the sec- headers aren't needed. also connection.

datasets/mx/deputies/crawler.py
```diff
@@ -0,0 +1,66 @@
...
+from zavod.logic.pep import categorise
+
+def crawl_item(input_dict: dict, context: Context):
+
+    entity = context.make("Person")
+    entity.id = context.make_id(input_dict["NombreCompleto"])
+
+    entity.add("lastName", input_dict.pop("Nombre"))
+    entity.add("firstName", input_dict.pop("PrimerApellido"))
+    entity.add("name", input_dict.pop("NombreCompleto"))
```
jbothma, line 14: the apply_name helper is the preferred way to add multiple names

datasets/mx/deputies/crawler.py
```diff
@@ -0,0 +1,66 @@
...
+            True,
+            categorisation=categorisation,
+        )
+
+    if occupancy is None:
+        return
+
+    context.emit(entity, target=True)
+    context.emit(position)
+    context.emit(occupancy)
```
jbothma, line 37:
```suggestion
    context.emit(occupancy)
    context.audit_data(input_dict)
```

adding audit_data shows a few more fields we can use:
- SegundoApellido 
- Partido -> Person:politicial

the rest you can add to the audit ignore argument

datasets/mx/deputies/crawler.py
jbothma: watch out - I think Nombre is first name.
Also I think the best is to concatenate Primer and Segundo Appelido and put both into lastName since I don't think w can assume the order of patrymic and matronymic

jbothma (changes requested):
Nice job! thank you. A few small things then it's done.

datasets/mx/deputies/crawler.py
bgmello: In fact we don't even need to send any headers. I copied the curl command from the website and it was already with this headers, but looks like they are not necessary

datasets/mx/deputies/crawler.py
bgmello: I added the Partido to `political` and `SegundoApellido` in the `lastName`

datasets/mx/deputies/crawler.py
jbothma: much better. thanks! It's worth noting that in python multiline strings, the indentation would be included in the string, but it doesn't hurt in this case.

datasets/mx/deputies/crawler.py
jbothma: nice. Try and strip out anything from the code that isn't necessary. That makes identifying issues and maintenance in future easier.

datasets/mx/deputies/crawler.py
```diff
@@ -0,0 +1,84 @@
...
+    )
+
+    if input_dict["Telefono"] is not None:
+        entity.add("phone", "+52" + input_dict.pop("Telefono"))
+
+    entity.add("email", input_dict.pop("Correo"))
+    entity.add("political", input_dict.pop("Partido"))
+
+    position = h.make_position(
+        context, "Member of the Chamber of Deputies", country="mx"
```
jbothma, line 26:
```suggestion
        context, "Member of the Chamber of Deputies of Mexico", country="mx"
```

jbothma: we tend to use the labels from https://www.wikidata.org/wiki/Wikidata:WikiProject_every_politician when they make sense

PR #628, opened 2024-03-07
adding lithuania illegal gambling list crawler

bgmello:
Fixes opensanctions/crawler-planning#78

### DIFF
--- a/datasets/lt/illegal_gambling/crawler.py
+++ b/datasets/lt/illegal_gambling/crawler.py
@@ -1,0 +1,35 @@
+import re 
+
+from zavod import Context
+
+def crawl_item(item, context: Context):
+
+    name = item.findtext(".//td[2]")
+    domain = item.findtext(".//td[3]")
+    contacts = item.findtext(".//td[4]")
+
+    if not domain:
+        return
+
+    entity = context.make("Organization")
+    entity.id = context.make_slug(domain, prefix="lt-illegal-gambling")
+
+    entity.add("name", name)
+    entity.add("website", domain)
+
+    # We find all emails in the contacts field and add them to the entity
+    emails = re.findall(r"[\w\.-]+@[\w\.-]+", contacts)
+    for email in emails:
+        entity.add("email", email)
+
+    entity.add("topics", "crime")
+
+    context.emit(entity, target=True)
+
+def crawl(context: Context):
+
+    response = context.fetch_html(context.data_url)
+
+    # We disconsider the two first rows because they are headers
+    for item in response.xpath('//*[@class="has-fixed-layout"]/tbody/tr')[2:]:
+        crawl_item(item, context)
--- a/datasets/lt/illegal_gambling/crawler.py
+++ b/datasets/lt/illegal_gambling/crawler.py
@@ -1,15 +1,48 @@
-import re 
+import re
+from typing import Generator, Dict, Tuple
 
+from normality import collapse_spaces
 from zavod import Context
 
+
+def parse_table(
+    table,
+    skiprows=2,
+    headers=["Number", "Name", "Domain", "Contacts", "Court information"],
+) -> Generator[Dict[str, str | Tuple[str]], None, None]:
+    '''
+    The first two rows of the table represent the headers, but they are not using the th tag and
+    so we are manually defining the column names.
+
+    Args:
+        table: The table element to parse
+        skiprows: The number of rows to skip before parsing the table
+        headers: The headers to use for the table columns
+    Returns:
+        A generator that yields a dictionary of the table columns and values. The keys are the
+        column names and the values are the column values. If the column contains a link, the
+        value is a tuple containing the link text and the link URL.
+    Raises:
+        AssertionError: If the number of headers and columns do not match. This indicates that the
+        table is malformed.
+    '''
+    for row in table.findall(".//tr")[skiprows:]:
+        cells = []
+        for el in row.findall("./td"):
+            a = el.find(".//a")
+            if a is None:
+                cells.append(collapse_spaces(el.text_content()))
+            else:
+                cells.append((collapse_spaces(a.text_content()), a.get("href")))
+
+        assert len(headers) == len(cells)
+        yield {hdr: c for hdr, c in zip(headers, cells)}
+
+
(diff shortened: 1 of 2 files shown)


datasets/lt/illegal_gambling/lt_illegal_gambling.yml
```diff
@@ -0,0 +1,25 @@
+title: Lithuania Illegal Gambling Operators 
+entry_point: crawler.py
+prefix: lt-illegal-gambling
+summary: >
+  List of illegal online gambling operators 
+description: |
+  In Lithuania, only proper licensed companies
```
jbothma, line 7:
```suggestion
  In Lithuania, only properly licensed companies
```

datasets/lt/illegal_gambling/lt_illegal_gambling.yml
```diff
@@ -0,0 +1,25 @@
+title: Lithuania Illegal Gambling Operators 
+entry_point: crawler.py
+prefix: lt-illegal-gambling
+summary: >
+  List of illegal online gambling operators
```
jbothma, line 5:
```suggestion
  List of online gambling operators regarded as illegal in Lithuania
```

datasets/lt/illegal_gambling/crawler.py
```diff
@@ -0,0 +1,35 @@
...
+
+    name = item.findtext(".//td[2]")
+    domain = item.findtext(".//td[3]")
+    contacts = item.findtext(".//td[4]")
+
+    if not domain:
+        return
+
+    entity = context.make("Organization")
+    entity.id = context.make_slug(domain, prefix="lt-illegal-gambling")
```
jbothma, line 15: dataset prefix is used automatically, isn't it?
```suggestion
    entity.id = context.make_slug(domain)
```

datasets/lt/illegal_gambling/crawler.py
```diff
@@ -0,0 +1,35 @@
...
+from zavod import Context
+
+def crawl_item(item, context: Context):
+
+    name = item.findtext(".//td[2]")
+    domain = item.findtext(".//td[3]")
+    contacts = item.findtext(".//td[4]")
+
+    if not domain:
+        return
```
jbothma, line 12: is this needed?

jbothma (changes requested):
Thanks! Looks good but I think let's improve in a few ways before merging

datasets/lt/illegal_gambling/crawler.py
```diff
@@ -0,0 +1,35 @@
+import re 
+
+from zavod import Context
+
+def crawl_item(item, context: Context):
+
+    name = item.findtext(".//td[2]")
+    domain = item.findtext(".//td[3]")
+    contacts = item.findtext(".//td[4]")
```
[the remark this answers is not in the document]
bgmello: I changed the function a little to work on this specific case. What I did was:

1. Add a `skiprows` parameter to skip the first 2 rows
2. Define the headers in the parameters, I did this because the table doesn't have the `th` tags, so it wasn't identifying the name of the columns

Let me know if it looks good like this or if I should change something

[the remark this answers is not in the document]
jbothma: yeah the lack of th means we have to assume how many columns there are.
The colspan and rowspan also seems more tricky to parse than is justified here.

So let's use your idea of hardcoding column names, but let's also assert we match the expected column names, so we detect when they change and update our hardcoded list https://github.com/opensanctions/opensanctions/pull/636/commits/0611e60deabb93502cd7d3254383af19cde29dac

datasets/lt/illegal_gambling/crawler.py
```diff
@@ -0,0 +1,75 @@
...
+from typing import Generator, Dict, Tuple
+
+from normality import collapse_spaces
+from zavod import Context
+
+
+def parse_table(
+    table,
+    skiprows=2,
+    headers=["Number", "Name", "Domain", "Contacts", "Court information"],
```
jbothma, line 11: I moved this hardcoded list to the top of the module, just a bit of best practice. We also tend to use snake_case, all lower case, for row keys normally - could you update to the same?

datasets/lt/illegal_gambling/crawler.py
```diff
@@ -0,0 +1,75 @@
...
+) -> Generator[Dict[str, str | Tuple[str]], None, None]:
```
jbothma, line 12: I've simplified it to not extract the URLs since that doesn't seem as useful as the full cell contents in this case - so just check if this type is still correct.

datasets/lt/illegal_gambling/crawler.py
```diff
@@ -0,0 +1,75 @@
...
+        assert len(headers) == len(cells)
+        yield {hdr: c for hdr, c in zip(headers, cells)}
+
+
+def crawl_item(item, context: Context):
+    name = item.pop("Name")
+    domain = item.pop("Domain")
+    contacts = item.pop("Contacts")
+
+    entity = context.make("Organization")
```
jbothma, line 47: I think it's safe to assume these are all companies so let's use that schema

jbothma:
Thanks for the updates - Just a few things then its ready

datasets/lt/illegal_gambling/crawler.py
bgmello: Did you make this commits in this branch? I didn't find them

jbothma: here https://github.com/opensanctions/opensanctions/commit/0611e60deabb93502cd7d3254383af19cde29dac - sorry that was unclear

bgmello: Do you want me to put these commits in this PR or you'll merge this branch with the https://github.com/opensanctions/opensanctions/pull/636?

jbothma: Please copy the changes or cherry-pick https://github.com/opensanctions/opensanctions/commit/0611e60deabb93502cd7d3254383af19cde29dac
Thanks!

bgmello: Done

Issue #91, opened 2021-06-06
Uniform name handling


pudo:
Names that are composed from component parts (e.g. first name, middle name, patronymic, last name) are currently concatenated in a haphazard manner. This should be made coherent and respect some basic cultural norms around names (e.g. Western Europe, Slavic, Arabic, Latino? Jeez this gets weird fast...)

PR #632, opened 2024-03-08
Pakistan Proscribed Persons crawler

atehe:
closes [Pakistan NACTA Proscribed Persons #91](https://github.com/opensanctions/crawler-planning/issues/91)

@jbothma, I wasn't sure which topic i should give the dataset so I used it a `crime.terror`.
Also,  I added an `addressEntity` because of district and province in the datasource but when i export it I see other address data in it.
Is this the expected behaviour?
Please have a look.

### DIFF
--- a/datasets/_collections/crime.yml
+++ b/datasets/_collections/crime.yml
@@ -45,4 +45,4 @@ children:
   - us_occ_enfact
   - us_sec_pause
   - za_wanted
-
+  - pk_proscribed_persons
--- a/datasets/_collections/crime.yml
+++ b/datasets/_collections/crime.yml
@@ -45,4 +45,3 @@ children:
   - us_occ_enfact
   - us_sec_pause
   - za_wanted
-  - pk_proscribed_persons
--- a/datasets/pk/proscribed_persons/crawler.py
+++ b/datasets/pk/proscribed_persons/crawler.py
@@ -25,7 +25,11 @@ def crawl_person(context: Context, row: dict):
     entity.add("topics", "sanction")
 
     address_entity = h.make_address(
-        context, state=province, region=district, full=f"{district}, {province}"
+        context,
+        state=province,
+        region=district,
+        full=f"{district}, {province}",
+        key=entity.id,
     )
     entity.add("addressEntity", address_entity)
 
--- a/datasets/pk/proscribed_persons/crawler.py
+++ b/datasets/pk/proscribed_persons/crawler.py
@@ -1,0 +1,33 @@
+from zavod import Context
+from zavod import helpers as h
+
+
+def crawl_person(context: Context, row: dict):
+    person_name = row.pop("name")
+    father_name = row.pop("fatherName")
+    cnic = row.pop("cnic")
+    province = row.pop("province")
+    district = row.pop("district")
+
+    entity = context.make("Person")
+    entity.id = context.make_slug(person_name, prefix="pk-proscribed")
+    entity.add("idNumber", cnic)
+    entity.add("name", person_name)
+    entity.add("fatherName", father_name)
+    entity.add("topics", "crime.terror")
+
+    address_entity = h.make_address(
+        context, state=province, region=district, country_code="pk"
+    )
+    entity.add("addressEntity", address_entity)
+    context.emit(address_entity)
+
+    context.emit(entity, target=True)
+    context.audit_data(row)
+
+
+def crawl(context: Context):
+    data = context.fetch_json(context.data_url, cache_days=1)
+
+    for record in data:
+        crawl_person(context, record)
--- a/datasets/pk/proscribed_persons/crawler.py
+++ b/datasets/pk/proscribed_persons/crawler.py
@@ -10,19 +10,32 @@ def crawl_person(context: Context, row: dict):
     district = row.pop("district")
 
     entity = context.make("Person")
-    entity.id = context.make_slug(person_name, prefix="pk-proscribed")
-    entity.add("idNumber", cnic)
+    entity.id = context.make_slug(
+        cnic or f"{person_name} - {district}, {province}", prefix="pk-cnic"
+    )
+    name_split = person_name.split("@")
+    if len(name_split) > 1:
+        person_name = name_split[0]
+        for alias in name_split[1:]:
+            entity.add("alias", alias)
+
     entity.add("name", person_name)
+    entity.add("idNumber", cnic)
     entity.add("fatherName", father_name)
-    entity.add("topics", "crime.terror")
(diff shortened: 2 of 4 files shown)


datasets/_collections/crime.yml
```diff
@@ -45,4 +45,4 @@ children:
   - us_occ_enfact
   - us_sec_pause
   - za_wanted
-
+  - pk_proscribed_persons
```
jbothma, line 48: let's put it in sanctions instead

jbothma:
Thanks for checking @atehe 
A bit of a google found me https://nacta.gov.pk/proscribed-persons/ which links to the same list. This would be a good URL for the top level url field in the metadata.
According to this, the consequences of being listed include bank accounts being frozen and passport embargo which sounds enough like full sanctions to use the `sanction` topic instead of `crime` to me.

datasets/pk/proscribed_persons/crawler.py
```diff
@@ -0,0 +1,33 @@
...
+
+    entity = context.make("Person")
+    entity.id = context.make_slug(person_name, prefix="pk-proscribed")
+    entity.add("idNumber", cnic)
+    entity.add("name", person_name)
+    entity.add("fatherName", father_name)
+    entity.add("topics", "crime.terror")
+
+    address_entity = h.make_address(
+        context, state=province, region=district, country_code="pk"
```
jbothma, line 20: can we assume pakistan? perhaps it's safest to skip the country_code unless there's something suggesting it will always be places in Pakistan

datasets/pk/proscribed_persons/crawler.py
```diff
@@ -0,0 +1,33 @@
...
+    cnic = row.pop("cnic")
+    province = row.pop("province")
+    district = row.pop("district")
+
+    entity = context.make("Person")
+    entity.id = context.make_slug(person_name, prefix="pk-proscribed")
+    entity.add("idNumber", cnic)
+    entity.add("name", person_name)
+    entity.add("fatherName", father_name)
+    entity.add("topics", "crime.terror")
```
jbothma, line 17: let's do sanction topic and sanction entity with `4th Schedule under the Anti Terrorism Act, 1997` as sanction program.

datasets/pk/proscribed_persons/crawler.py
```diff
@@ -0,0 +1,33 @@
+from zavod import Context
+from zavod import helpers as h
+
+
+def crawl_person(context: Context, row: dict):
+    person_name = row.pop("name")
```
jbothma, line 6: I noticed some @'s in some names - let's split on @, add the first element to name, and the rest as aliases

datasets/pk/proscribed_persons/crawler.py
```diff
@@ -0,0 +1,33 @@
...
+    father_name = row.pop("fatherName")
+    cnic = row.pop("cnic")
+    province = row.pop("province")
+    district = row.pop("district")
+
+    entity = context.make("Person")
+    entity.id = context.make_slug(person_name, prefix="pk-proscribed")
+    entity.add("idNumber", cnic)
```
jbothma, line 14: let's make id as make_slug(cnic, prefix="pk-cnic") - we have a similar pattern for "cedula" in some south american countries.

if cnic is currently not set for someone, let's use their name and address components together in make_id.

jbothma:
well-spotted on the address thing. They all end up with the same ID so they get merged. Just figuring out what the right thing is then I'll reply on that.

datasets/pk/proscribed_persons/crawler.py
jbothma: So one of the provinces I happened to google turns out to be a region shared by Iran, Afghanistan and Pakistan so I think sticking the pk country code on it isn't the right thing.

I asked Friedrich about this address and the merged id issue and (a) whether it's expected that such a limited address is expected to generate an emtpy `full`, and (b) whether this should be an address entity or just a string in the address property.

a. the decision was a bug, I reported it at https://github.com/pudo/addressformatting/issues/1 because we won't fix it right now
b. it's so limited so just combine `{region}, {state}` and stick it into the entity address property

datasets/pk/proscribed_persons/crawler.py
```diff
@@ -0,0 +1,50 @@
...
+
+    entity = context.make("Person")
+    entity.id = context.make_slug(
+        cnic or f"{person_name} - {district}, {province}", prefix="pk-cnic"
+    )
+    name_split = person_name.split("@")
+    if len(name_split) > 1:
+        person_name = name_split[0]
+        for alias in name_split[1:]:
+            entity.add("alias", alias)
```
jbothma, line 20: providing a list to entity.add is equivalent to this and a bit shorter

datasets/pk/proscribed_persons/crawler.py
```diff
@@ -0,0 +1,50 @@
...
+    person_name = row.pop("name")
+    father_name = row.pop("fatherName")
+    cnic = row.pop("cnic")
+    province = row.pop("province")
+    district = row.pop("district")
+
+    entity = context.make("Person")
+    entity.id = context.make_slug(
+        cnic or f"{person_name} - {district}, {province}", prefix="pk-cnic"
+    )
```
jbothma, line 15: nice approach, but let's not override the prefix when falling back to person_name and location. Let it default to the prefix in the metadata in that case.

The idea with pk-cnic is that it then becomes a very readable national ID schema so we don't want to mix in names and things in there.

datasets/pk/proscribed_persons/crawler.py
atehe: Ok I understand.
I'll stick the `region}, {state}` for now.

datasets/pk/proscribed_persons/crawler.py
```diff
@@ -0,0 +1,48 @@
...
+    district = row.pop("district")
+
+    entity = context.make("Person")
+
+    if cnic:
+        entity.id = context.make_slug(cnic, prefix="pk-cnic")
+    else:
+        entity.id = context.make_slug(
+            f"{person_name} - {district}, {province}", prefix="pk-proscribed"
+        )
```
jbothma, line 19:
```suggestion
        entity.id = context.make_slug(person_name, district, province)
```

jbothma:
Thank you! Nice work

FYI I noticed some IDs looked like pk-cnic-000000001 so I started writing a silly validator then remembered we have a validation library, so I pulled that in here https://github.com/opensanctions/opensanctions/commit/17baea1c014a851dfb06ed35d24de08448b7c2a5

PR #635, opened 2024-03-11
Added crawler for US HHS OIG list of excluded individuals/entities

KishanAdhikari11:


datasets/us/doh_debarred/crawler.py
```diff
@@ -0,0 +1,51 @@
...
+                entity.add("createdAt", parsedate(row.get("EXCLDATE")))
+                entity.add("topics", "debarment")
+                entity.add("country", "us")
+                context.emit(entity, target=True)
+            else:
+                comp = context.make("Company")
+                address = row.get('ADDRESS', '').strip()
+                if address:
+                    match = re.match(r'^([#\w\s/-]+)', address)
+                    first_line = match.group(1).strip() if match else address
```
jbothma, line 38: you can just use the address column as it is - no need to regex

datasets/us/doh_debarred/crawler.py
```diff
@@ -0,0 +1,51 @@
...
+                else:
+                    first_line = row["ZIP"]  # Use ZIP code as ID for empty address
+                comp.id = context.make_id(first_line)
```
jbothma, line 41:
```suggestion
                name = row.pop("BUSNAME")
                comp.id = context.make_id(name, first_line)
```

datasets/us/doh_debarred/crawler.py
KishanAdhikari11: ok. Also, there were some cases where the address row was empty (we are using address as id, so it can't be empty), In that case, what could be the solution?

datasets/us/doh_debarred/crawler.py
```diff
@@ -0,0 +1,51 @@
+from zavod import Context
+import csv
+import re
+
+def parsedate(date: str):
+    """Parse date from string."""
+    if date:
+        return f'{date[:4]}-{date[4:6]}-{date[6:]}'
+    return None
```
jbothma, line 9: nice. we'll change this to the `parse_date()` helper in milestone 3

datasets/us/doh_debarred/crawler.py
```diff
@@ -0,0 +1,51 @@
...
+                comp.add("name", row.get("BUSNAME"))
+                comp.add("address", address)
+                comp.add("createdAt", parsedate(row.get("EXCLDATE")))
+                comp.add("topics", "debarment")
+                comp.add("country", "us")
+                context.emit(comp, target=True)
+
+def crawl(context: Context) -> None:
+    """Entry point for crawling process."""
+    getdata(context)
```
jbothma, line 51: our usual pattern is to read the file, if it's an external resource, export the resource, then loop over the rows in `crawl()` and call a function named `crawl_item()` or something for each record int he resource.

So in milestone 3, let's move the loop here and rename getdata to crawl_item

datasets/us/doh_debarred/crawler.py
```diff
@@ -0,0 +1,51 @@
...
+                entity.add("name", name)
+                entity.add("notes", row.get("GENERAL"))
+                entity.add("birthDate", parsedate(row.get("DOB")))
+                entity.add("address", row.get("ADDRESS"))
+                entity.add("createdAt", parsedate(row.get("EXCLDATE")))
+                entity.add("topics", "debarment")
+                entity.add("country", "us")
+                context.emit(entity, target=True)
+            else:
+                comp = context.make("Company")
```
jbothma, line 34: if you name it entity, you can move the common fields like address, topics, notes and EXCLDATE out of the if use the same code for both types of entities

datasets/us/doh_debarred/crawler.py
```diff
@@ -0,0 +1,51 @@
...
+    with open(source_file, "r") as f:
+        reader = csv.DictReader(f)
+        for row in reader:
+            if row["LASTNAME"].strip() or row["FIRSTNAME"].strip():
+                entity = context.make("Person")
+                firstname = row["FIRSTNAME"]
+                midname = row["MIDNAME"]
+                lastname = row["LASTNAME"]
+                name = f"{firstname} {midname} {lastname}".strip()
+                entity.id = context.make_id(name)
```
jbothma, line 24:
```suggestion
                entity.id = context.make_slug(name, address)
```

datasets/us/doh_debarred/crawler.py
```diff
@@ -0,0 +1,51 @@
...
+                address = row.get('ADDRESS', '').strip()
+                if address:
+                    match = re.match(r'^([#\w\s/-]+)', address)
+                    first_line = match.group(1).strip() if match else address
+                else:
+                    first_line = row["ZIP"]  # Use ZIP code as ID for empty address
+                comp.id = context.make_id(first_line)
+                comp.add("name", row.get("BUSNAME"))
+                comp.add("address", address)
+                comp.add("createdAt", parsedate(row.get("EXCLDATE")))
```
jbothma, line 44: we'll put this into the startDate property of a Sanction entity instead of this property in milestone 3

datasets/us/doh_debarred/crawler.py
```diff
@@ -0,0 +1,51 @@
...
+                entity.add("country", "us")
+                context.emit(entity, target=True)
+            else:
+                comp = context.make("Company")
+                address = row.get('ADDRESS', '').strip()
+                if address:
+                    match = re.match(r'^([#\w\s/-]+)', address)
+                    first_line = match.group(1).strip() if match else address
+                else:
+                    first_line = row["ZIP"]  # Use ZIP code as ID for empty address
```
jbothma, line 40: nice solution

jbothma:
Looking great so far!

I've made a bunch of comments - most of them are more applicable to the third step than right now.

For the metadata, please expand particularly on what the consequences are of being listed in this dataset. our data users - banks, financial service providers, compliance checking companies, etc, need to understand what being listed here means - what should the person be prevented from doing now that they're in the list.

feel free to use markdown quotes in the description.

Also use the assertions part of the metadata to make sure you output as many entities as you expect.

PR #637, opened 2024-03-11
81 add esma register

SimonThordal:
https://registers.esma.europa.eu/publication/searchRegister?core=esma_registers_saris_new

This includes EU sanctioned entities. Gold mine!

### DIFF
--- a/datasets/eu/esma_saris/crawler.py
+++ b/datasets/eu/esma_saris/crawler.py
@@ -1,0 +1,8 @@
+from zavod import Context
+
+
+def crawl(context: Context) -> None:
+    context.fetch_html(context.dataset.url)
+    source_file = context.fetch_resource(
+        "source.csv", context.dataset.data.url, cookies={"criteria": "[]"}
+    )
--- a/datasets/eu/esma_saris/crawler.py
+++ b/datasets/eu/esma_saris/crawler.py
@@ -1,9 +1,71 @@
-from zavod import Context
+from zavod import Context, helpers
+import csv
 
 
 def crawl(context: Context) -> None:
-    print("Waiting for debugger attach")
-    context.fetch_html(context.dataset.url)
-    source_file = context.fetch_resource(
-        "source.csv", context.dataset.data.url, cookies={"criteria": "[]"}
+    # Set the initial cookie
+    data = {
+        "core": "esma_registers_saris_new",
+        "pagingSize": "10",
+        "start": 0,
+        "keyword": "",
+        "sortField": "effectiveFrom asc",
+        "criteria": [],
+        "wt": "json",
+    }
+    context.http.post(
+        "https://registers.esma.europa.eu/publication/searchRegister/doMainSearch",
+        json=data,
     )
+    source_file = context.fetch_resource(
+        "source.csv",
+        context.dataset.data.url,
+    )
+    with open(source_file, "r") as f:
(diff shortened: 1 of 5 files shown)


datasets/eu/esma_saris/crawler.py
```diff
@@ -0,0 +1,71 @@
...
+            if isin is None:
+                context.log.warn("No ISIN", row=row)
+                return
+            entity.id = f"isin-{isin}"
+            entity.add("isin", isin)
+            entity.add("name", row.pop("instrumentFullName", isin))
+
+            sanction = helpers.make_sanction(context, entity)
+            sanction.id = context.make_id(row.pop("id"))
+            sanction.add("program", "ESMA")
```
pudo, line 38: I don't think we need this: the sanction['authority'] is already going to contain the `publisher.name`.

datasets/eu/esma_saris/crawler.py
```diff
@@ -0,0 +1,71 @@
...
+            if row.get("effectiveTo") is not None:
+                sanction.add("endDate", row.pop("effectiveTo"))
+            else:
+                entity.add("topic", "sanction")
+            context.emit(sanction, target=False)
+
+            if row.get("issuer") is not None:
+                issuer = context.make("LegalEntity")
+                issuer.id = context.make_id(row.pop("issuer"))
+                issuer.add("leiCode", issuer)
```
pudo, line 54: If the issuer has a LEI, there's also a fake ID we could use: `lei-{lei}`. It's already being emitted by 3-4 other sources.

datasets/eu/esma_saris/crawler.py
```diff
@@ -0,0 +1,71 @@
...
+
+            sanction = helpers.make_sanction(context, entity)
+            sanction.id = context.make_id(row.pop("id"))
+            sanction.add("program", "ESMA")
+            sanction.add("provisions", row.pop("actionType"))
+            sanction.add("reason", row.pop("reasonsForTheAction"))
+            sanction.add("description", row.pop("comments"))
+            sanction.add("startDate", row.pop("effectiveFrom"))
+            sanction.add("country", row.pop("memberStateOfNotifyingCA"))
+            sanction.add("authority", row.pop("notifyingCA"))
```
pudo, line 44: Ah, so you may want to make this a `.set(`, replacing the previous value.

pudo (commented):
Looks really good!

datasets/eu/esma_saris/crawler.py
SimonThordal: Yeah, this is a bit of a weird one. As far as I can understand ([basis here](https://www.esma.europa.eu/publications-and-data/interactive-single-rulebook/mifid-ii/article-52-suspension-and-removal)) ESMA is sort of an "aggregator" of suspensions and removals.

Is the program the name of directive itself maybe? Or do we leave it blank

datasets/eu/esma_saris/crawler.py
SimonThordal: Looks like it only sometimes has a LEI and sometimes just has a 4 letter random string, so sometimes the `lei-` will work, sometimes it won't, so I made it conditional.

datasets/eu/esma_saris/crawler.py
SimonThordal: Thanks, fixed!

PR #639, opened 2024-03-13
French mayors

dhdaines:
Add mayors of French municipalities from the RNE

Closes https://github.com/opensanctions/crawler-planning/issues/85

### DIFF
--- a/datasets/fr/maires/crawler.py
+++ b/datasets/fr/maires/crawler.py
@@ -1,0 +1,85 @@
+"""
+Crawler for active French mayors.
+"""
+
+import csv
+from typing import Dict
+
+from zavod import Context
+from zavod import helpers as h
+from zavod.logic.pep import categorise
+
+UNUSED_FIELDS = [
+    "Code sexe",
+    "Code de la catégorie socio-professionnelle",
+    "Libellé de la catégorie socio-professionnelle",
+    "Date de début de la fonction",
+]
+DATE_FORMATS = ["%d/%m/%Y"]
+
+
+def crawl_row(
+    context: Context,
+    row: Dict[str, str],
+):
+    """Process one row of the CSV data"""
+    # Get infos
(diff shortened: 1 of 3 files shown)


datasets/fr/maires/crawler.py
```diff
@@ -0,0 +1,85 @@
...
+    )
+    context.log.debug(f"Unique ID {person.id}")
+    h.apply_name(person, given_name=given_name, last_name=family_name, lang="fra")
+    if birth_date:
+        person.add("birthDate", h.parse_date(birth_date, DATE_FORMATS))
+    position = h.make_position(
+        context,
+        name=f"Mayor of {municipality}",
+        country="fr",
+        subnational_area=subnational_area,
```
jbothma, line 56: if their jurisdiction is the municipality, that is what we'd like here - the idea is that subnational_area narrows the jurisdiction from the country_code to this area for subnational

datasets/fr/maires/crawler.py
```diff
@@ -0,0 +1,85 @@
...
+
+import csv
+from typing import Dict
+
+from zavod import Context
+from zavod import helpers as h
+from zavod.logic.pep import categorise
+
+UNUSED_FIELDS = [
+    "Code sexe",
```
jbothma, line 13: sex or gender is also nice to have, acknowledging that we're probably not treating it with as much nuance as we should, but the aim is just to reflect the source

jbothma (commented):
Superbe!

Merci!

I'll comment two tiny things but I'll make the changes and merge while I have a copy open.

datasets/fr/maires/crawler.py
dhdaines: Ah, of course! Yes also useful for disambiguating, I just didn't bother to look in the data dictionary to see what the right attribute name was.

PR #640, opened 2024-03-13
Add US DOH debarred crawler

KishanAdhikari11:


datasets/us/doh_debarred/crawler.py
```diff
@@ -0,0 +1,64 @@
...
+    context.emit(sanction)
+    context.emit(entity, target=True)
+    context.audit_data(row,
+                       ignore=[
+                           "SPECIALTY",
+                           "UPIN",
+                           "NPI",
+                           "CITY",
+                           "STATE",
+                           "REINDATE",
```
jbothma, line 52: I think this file only contains exclusions that are active, but let's make sure by asserting that REINDATE is `0` for each row.

datasets/us/doh_debarred/us_doh_debarred.yml
```diff
@@ -0,0 +1,55 @@
...
+coverage:
+  frequency: daily
+  start: "2024-03-13"
+summary: >
+  A list of debarred peoples and companies listed in US Department of Health and Human Service.
+description: |
+  This dataset provides information on Individuals and entities debarred from participation in Medicare,
+  Medicaid, and all other Federal health care programs for various reasons including conviction for Medicare 
+  or Medicaid fraud.
+  Debarments are imposed under the authority of sections 1128 and 1156 of the Social Security Act.
```
jbothma, line 13: let's use the word Excluded/exclusions rather than debarred/debarments to be consistent with their terminology. In the topics we're limited to the topics defined in our data model, but here it's best to align with their wording.

datasets/us/doh_debarred/crawler.py
```diff
@@ -0,0 +1,64 @@
...
+    entity.add("address", row.get("ADDRESS", None))
+    entity.add("topics", "debarment")
+    entity.add("country", "us")
+    sanction = h.make_sanction(context, entity)
+    sanction.add("startDate", h.parse_date(row.get("EXCLDATE"), FORMATS))
+    sanction.add("reason",row.get("EXCLTYPE"))
+    sanction.add("program","US HHS OIG List of Excluded Individuals/Entities ")
+    context.emit(sanction)
+    context.emit(entity, target=True)
+    context.audit_data(row,
```
jbothma, line 45: when I crawl the dataset, I get lots of warnings like this:

```
2024-03-14 11:03:32 [warning  ] Unexpected data found          [us_doh_debarred] data={'LASTNAME': 'ARTHUR', 'FIRSTNAME': 'WALLACE', 'MIDNAME': 'R', 'GENERAL': 'PHYSICIAN (MD, DO)', 'DOB': '19540813', 'ADDRESS': '2495 S KRAMERIA STREET', 'ZIP': '80222', 'EXCLTYPE': '1128b4', 'EXCLDATE': '20170119'} dataset=us_doh_debarred
```

we usually use `row.pop` rather than `row.get` or `row[...]` so that each field is removed from the dict as it's used. The idea is that we then "audit" the dict, expecting it to be empty. We normally only merge crawlers when they have no warnings, or if the warnings would be solved after merging, e.g. by adding some position definitions to our PEP position database, or if it's a temporary network issue.

So if we get the dict to be empty (or ignore unused columns as you have) we can merge the PR without warnings. If they later add a column to the CSV, we're automatically alerted because suddenly warnings would show up here https://www.opensanctions.org/issues/, which we check daily. We can then decide if we want to use or ignore the new data.

another benefit of `pop` over `get` is that pop will raise an exception if they key doesn't exist, alerting us to missing data, perhaps from a column header change.

datasets/us/doh_debarred/us_doh_debarred.yml
```diff
@@ -0,0 +1,55 @@
...
+assertions:
+  min:
+    schema_entities:
+      Person: 10000
+      Company: 1500
+    countries: 1
+  max:
+    schema_entities:
+      Person: 150000
+      Company: 10000
```
jbothma, line 55: looking at the generated statistics.json, it looks like you can bring the min within 10% and max within 20% of the 75800 Persons and 3284 Companies

datasets/us/doh_debarred/crawler.py
jbothma: Please make sure there aren't any more warnings because all the columns that aren't used are ignored. Please ask if there are still warnings you're not sure of.

jbothma (changes requested):
Thank you! 

Really nice job on the description.

Ideally push another commit to this pull request but no stress if you need to make a new branch.

datasets/us/doh_debarred/crawler.py
KishanAdhikari11: yes for each row REINDATE is 0.

datasets/us/doh_debarred/crawler.py
```diff
@@ -0,0 +1,64 @@
...
+    if row["LASTNAME"].strip() or row["FIRSTNAME"].strip():
+        entity = context.make("Person")
+        firstname = row["FIRSTNAME"]
+        midname = row["MIDNAME"]
+        lastname = row["LASTNAME"]
+        name = f"{firstname} {midname} {lastname}".strip()
+        address_data = get_address(row)
+        entity.id = context.make_slug(name, address_data)
+        entity.add("birthDate", h.parse_date(row.get("DOB"), FORMATS))
+        entity.add("name", name)
```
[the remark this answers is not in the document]
jbothma: Still needs the other name parts

datasets/us/doh_debarred/crawler.py
```diff
@@ -0,0 +1,71 @@
...
+        middle_name = row.pop("MIDNAME",None),
+        last_name = row.pop("LASTNAME",None)
+        )
+        
+        context.log.info(full_name)
+        address_data = get_first_line(row)
+        entity.id = context.make_slug(full_name, address_data)
+        h.apply_name(entity=entity,full=full_name,lang='eng')
+        entity.add("birthDate", h.parse_date(row.pop("DOB",None), FORMATS))
+        entity.add("position",f'{row.pop("GENERAL",None)} {row.pop("SPECIALTY",None)}')
```
jbothma, line 29: I think putting GENERAL in parens would work better than simply concatenating

`HEALTH CARE AIDE (INTER CARE FACILITY)`

datasets/us/doh_debarred/crawler.py
```diff
@@ -0,0 +1,71 @@
...
+    else:
+        entity = context.make("Company")
+        name = row.pop("BUSNAME",None)
+        address_data = get_first_line(row)
+        entity.id = context.make_slug(name, address_data)
+        entity.add("name", name)
+        entity.add("description",f'{row.pop("GENERAL",None)} {row.pop("SPECIALTY",None)}')
```
jbothma, line 36: I think adding as two distinct values would work best here, making sure blanks are None and not added as empty string.

datasets/us/doh_debarred/crawler.py
jbothma: and just the GENERAL value when SPECIALITY is blank

datasets/us/doh_debarred/crawler.py
jbothma: this should be an assertion in python so that we can detect when it changes

datasets/us/doh_debarred/us_doh_debarred.yml
```diff
@@ -0,0 +1,57 @@
+name: us_doh_debarred
+title:  US DOH Inspector General Exclusions
+entry_point: crawler.py
+coverage:
+  frequency: daily
+  start: "2024-03-13"
+summary: >
+  Peoples and companies excluded from Federally funded health care programs
```
jbothma, line 8:
```suggestion
  People and companies excluded from Federally funded health care programs
```

datasets/us/doh_debarred/us_doh_debarred.yml
```diff
@@ -0,0 +1,57 @@
...
+description: |
+  This dataset provides information on Individuals and entities excluded from participation in Medicare,
```
jbothma, line 10: let's wrap at 80 chars

datasets/us/doh_debarred/us_doh_debarred.yml
```diff
@@ -0,0 +1,57 @@
...
+  Medicaid, and all other Federal health care programs for various reasons including conviction for Medicare 
+  or Medicaid fraud.
+  Exclusions are imposed under the authority of sections 1128 and 1156 of the Social Security Act.
+  
+  From the OIG HHS page:
```
jbothma, line 15: Isn't it the other way around?

```suggestion
  From the HSS OIG page:
```

datasets/us/doh_debarred/us_doh_debarred.yml
```diff
@@ -0,0 +1,57 @@
...
+  >
+  > Individuals and entities who have been reinstated are removed from the LEIE.
+
+  For more information, refer to the [OIG's Exclusion Authorities FAQ page](https://oig.hhs.gov/faqs/exclusions-faq).
+
+publisher: 
+  name: U.S. Department of Health and Human Services Office of Inspector General (OIG)
+  description: |
+   OIG is the largest inspector general's office in the Federal Government responsible for oversight of the Department 
+   of Heallth and Human Services(HHS).The OIG team is made up of auditors, investigators, and experts in different areas.
```
jbothma, line 36:
```suggestion
   of Heallth and Human Services(HHS). The OIG team is made up of auditors, investigators, and experts in different areas. 
```

datasets/us/doh_debarred/crawler.py
```diff
@@ -0,0 +1,64 @@
...
+        entity.add("birthDate", h.parse_date(row.get("DOB"), FORMATS))
+        entity.add("name", name)
+    else:
+        entity = context.make("Company")
+        name = row.get("BUSNAME")
+        address_data = get_address(row)
+        entity.id = context.make_slug(name, address_data)
+        entity.add("name", name)
+    entity.add("notes", row.get("GENERAL",None))
+    entity.add("address", row.get("ADDRESS", None))
```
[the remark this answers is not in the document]
jbothma: we need all of address, city, state, zip and country - all parts of the address

datasets/us/doh_debarred/crawler.py
```diff
@@ -0,0 +1,71 @@
...
+        postal_code=row.pop("ZIP",None),
+        country_code="us"
+    )
+    h.apply_address(context,entity,address)
+    h.make_identification(
+        context,
+        entity,
+        key=row.pop("UPIN",None),
+        number=row.pop("NPI",None),
+    )
```
jbothma, line 51: no, for each of these two types, we need a distict identification entity, which indicates its type in the idenfication type property.

PR #642, opened 2024-03-14
CI for modified crawlers

jbothma:
- Lints all modified yaml files in `/datasets`
- Lints all modified python files in `/datasets`
- runs all datasets for modified yaml files in `/datasets`
- runs all datasets in the same directory as any modified files in `/datasets` (would run both yamls in a shared directory if the .py changed)

## Checklist

- [x] format the yamls caught in the linting
- [x] format the python caught in the linting

Post-merge

- [ ] Ensure dispatched when merged (`if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}`)

## Trying locally

```
act --job python --platform catthehacker/ubuntu:full-20.04 --container-architecture linux/amd64 --eventpath pull_request.json --workflows .github/workflows/build.yml --matrix python:3.11 pull_request
```

## Caveats

- with the matrix and dual push/pull_request runs, four crawlers hit a source at the same time. That's likely to be less of an issue if we add the retry/backoff logic as default.
- it looks like files modified elsewhere are considered part of the change set if master is merged into a PR branch, expanding the set of things that get linted and might trip up a PR.


bin/changed_datasets.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
pudo, line 1: I'm a weird, but does this maybe want to live in `contrib` ?

PR #644, opened 2024-03-14
adding finra enforcement actions crawler

bgmello:
Fixes opensanctions/crawler-planning#116

### DIFF
--- a/datasets/us/finra_enforcement_actions/crawler.py
+++ b/datasets/us/finra_enforcement_actions/crawler.py
@@ -5,7 +5,6 @@ from zavod import Context, helpers as h
 
 
 def clean_name(name):
-
     matches = {
         ", LLC": " LLC",
         ", L.L.C": " L.L.C",
@@ -66,7 +65,6 @@ def crawl_item(input_dict: dict, context: Context):
     # so we abort the crawl
     if "," in name:
         return
-
 
     entity = context.make(schema)
     entity.id = context.make_slug(name)
--- a/datasets/us/finra_enforcement_actions/crawler.py
+++ b/datasets/us/finra_enforcement_actions/crawler.py
@@ -1,0 +1,87 @@
+import re
+import requests
+from typing import Generator, Dict, Tuple, Optional
+from lxml.etree import _Element
+from normality import slugify, collapse_spaces
+from zavod import Context, helpers as h
+
+BASE_URL = "https://www.finra.org"
+
+def parse_table(
+    table: _Element,
+) -> Generator[Dict[str, Tuple[str, Optional[str]]], None, None]:
+    headers = None
+    for row in table.findall(".//tr"):
+        if headers is None:
+            headers = []
+            for el in row.findall("./th"):
+                headers.append(slugify(el.text_content()))
+            continue
+
+        cells = []
+        for el in row.findall("./td"):
+            a = el.find(".//a")
+            if a is None:
+                cells.append((collapse_spaces(el.text_content()), None))
+            else:
+                cells.append((collapse_spaces(a.text_content()), a.get("href")))
+
+        assert len(headers) == len(cells)
+        yield {hdr: c for hdr, c in zip(headers, cells)}, row
+
+def crawl_item(input_dict: dict, html_element: _Element, context: Context):
+
+    schema = 'LegalEntity'
+
+    # We try to find if it's a person or company using the icon class
+    if 'user' in html_element.find(".//i").get('class'):
+        schema = 'Person'
+        
+    if 'building' in html_element.find(".//i").get('class'):
+        schema = 'Company'
+
+    name = input_dict.pop("firms-individuals")[0]
+    case_summary = input_dict.pop("case-summary")[0]
+    source_url = BASE_URL + input_dict.pop("case-id")[0]
+    date = input_dict.pop("action-date-sort-ascending")[0]
+    document_type = input_dict.pop("document-type")[0]
+
+    entity = context.make(schema)
+    entity.id = context.make_slug(name)
+
+    entity.add("name", name)
+
+    sanction = h.make_sanction(context, entity)
+
+    sanction.add("description", case_summary)
+    sanction.add("date", h.parse_date(date, formats=["%m/%d/%Y"]))
+    sanction.add("publisher", "Document type: "+document_type)
+    sanction.add("publisherUrl", source_url)
+
+    context.emit(entity, target=True)
+    context.emit(sanction)
+
+    context.audit_data(input_dict)
+
+
+def crawl(context: Context):
+
+    # Each page only displays 15 rows at a time, so we need to loop until we find an empty table
+
+    base_url = context.data_url
+
+    page_num = 0
+
+    while True:
+        url = base_url + "?page=" + str(page_num)
+        response = context.fetch_html(url)
+        table = response.find(".//table")
+
+        if table is None:
+            break
+
+        for item, html_element in parse_table(table):
+            crawl_item(item, html_element, context)
+
+        page_num += 1
+
--- a/datasets/us/finra_enforcement_actions/crawler.py
+++ b/datasets/us/finra_enforcement_actions/crawler.py
@@ -63,7 +63,7 @@ def crawl_item(input_dict: dict, html_element: _Element, context: Context):
     context.emit(entity, target=True)
     context.emit(sanction)
 
-    context.audit_data(input_dict)
+    context.audit_data(input_dict, ignore=["document-type"])
 
 
 def crawl(context: Context):
--- a/datasets/us/finra_enforcement_actions/crawler.py
+++ b/datasets/us/finra_enforcement_actions/crawler.py
@@ -5,8 +5,6 @@ from lxml.etree import _Element
 from normality import slugify, collapse_spaces
 from zavod import Context, helpers as h
 
-BASE_URL = "https://www.finra.org"
-
 
 def parse_table(
     table: _Element,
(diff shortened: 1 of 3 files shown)


datasets/us/finra_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,87 @@
...
+    entity = context.make(schema)
+    entity.id = context.make_slug(name)
+
+    entity.add("name", name)
+
+    sanction = h.make_sanction(context, entity)
+
+    sanction.add("description", case_summary)
+    sanction.add("date", h.parse_date(date, formats=["%m/%d/%Y"]))
+    sanction.add("publisher", "Document type: "+document_type)
```
jbothma, line 58: I think you can drop document_type since case_summary tells us that and more.

datasets/us/finra_enforcement_actions/us_finra_enforcement_actions.yaml
```diff
@@ -0,0 +1,26 @@
...
+  a self-regulatory organization (SRO) that regulates
+  member brokerage firms and exchange markets. 
+publisher:
+  name: Financial Industry Regulatory Authority (FINRA)
+  url: https://www.finra.org/#/
+  official: false
+  country: "us"
+url: "https://www.finra.org/rules-guidance/oversight-enforcement/disciplinary-actions#:~:text=FINRA%20takes%20disciplinary%20actions%20against,the%20Municipal%20Securities%20Rulemaking%20Board."
+data:
+  url: https://www.finra.org/rules-guidance/oversight-enforcement/finra-disciplinary-actions?search=&firms=&individuals=&field_fda_case_id_txt=&field_core_official_dt%5Bmin%5D=&field_core_official_dt%5Bmax%5D=&field_fda_document_type_tax=All&token=f8W1uC3kpshSH5QzszZnszbMmQxZ77JDjlLYn16IuOw
```
jbothma, line 25: Does this cleaned up URL work just as well?

```suggestion
  url: https://www.finra.org/rules-guidance/oversight-enforcement/finra-disciplinary-actions
```

datasets/us/finra_enforcement_actions/us_finra_enforcement_actions.yaml
```diff
@@ -0,0 +1,26 @@
...
+  federal securities laws, rules, and regulations or
+  the rules of the Municipal Securities Rulemaking Board.
+  Financial Industry Regulatory Authority (FINRA)
+  is a private American corporation that acts as
+  a self-regulatory organization (SRO) that regulates
+  member brokerage firms and exchange markets. 
+publisher:
+  name: Financial Industry Regulatory Authority (FINRA)
+  url: https://www.finra.org/#/
+  official: false
```
jbothma, line 21: You can make this true on the basis of this line from their about page:

> FINRA FINANCIAL INDUSTRY REGULATORY AUTHORITY is authorized by Congress

datasets/us/finra_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,87 @@
...
+        schema = 'Company'
+
+    name = input_dict.pop("firms-individuals")[0]
+    case_summary = input_dict.pop("case-summary")[0]
+    source_url = BASE_URL + input_dict.pop("case-id")[0]
+    date = input_dict.pop("action-date-sort-ascending")[0]
+    document_type = input_dict.pop("document-type")[0]
+
+    entity = context.make(schema)
+    entity.id = context.make_slug(name)
```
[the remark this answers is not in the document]
jbothma: Yeah on the basis of

> is a government-authorized not-for-profit organization that oversees U.S. broker-dealers

you can add `us` to each entity

jbothma (changes requested):
Thanks! Some small changes then it's ready

datasets/us/finra_enforcement_actions/us_finra_enforcement_actions.yaml
bgmello: There is just a slight difference, in the cleaned up URL, it has 18288, and in the original one it has 18286 because it filters specifically for firms and individuals. I found one of the two not related to firms and individuals and it looks like this:
Screenshot 2024-03-19 at 10 05 20
Screenshot 2024-03-19 at 10 06 13
Screenshot 2024-03-19 at 10 06 18

Because there is no icon, it is throwing an error in page 36. What do you want me to do? Use the cleaned up URL and ignore the lines without an icon? Or use the original url?

datasets/us/finra_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,90 @@
...
+        context.log.info(page_num)
+        url = base_url + "?page=" + str(page_num)
+        response = context.fetch_html(url)
+        table = response.find(".//table")
+
+        if table is None:
+            break
+
+        for item, html_element in parse_table(table):
+            context.log.info("ok")
```
jbothma, line 87: I think we can remove these log lines - with `zavod --debug crawl` you can see the http requests

datasets/us/finra_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,90 @@
...
+
+    context.audit_data(input_dict)
+
+
+def crawl(context: Context):
+    # Each page only displays 15 rows at a time, so we need to loop until we find an empty table
+
+    base_url = context.data_url
+
+    page_num = 36
```
jbothma, line 75: we should start at 1, right?

jbothma:
Try merging the master branch in - it should enable retrying automatically so the 429 that failed your build workflow would be retried.

ruff has also found some linting issues with the python (see the lint workflow)

datasets/us/finra_enforcement_actions/us_finra_enforcement_actions.yaml
jbothma: Interesting! It doesn't work for me. I think let's use the cleaned up URL, and then skip if the firms-individuals cell doesn't contain a name? And if it has a name but no icon, you can leave it as LegalEntity

datasets/us/finra_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,90 @@
...
+    if "user" in html_element.find(".//i").get("class"):
+        schema = "Person"
+
+    if "building" in html_element.find(".//i").get("class"):
+        schema = "Company"
+
+    name = input_dict.pop("firms-individuals")[0]
+    case_summary = input_dict.pop("case-summary")[0]
+    case_id = input_dict.pop("case-id")[0]
+    source_url = BASE_URL + case_id
```
jbothma, line 47: try make_links_absolute in crawl() then I think you won't need t construct URLs from the base https://github.com/opensanctions/opensanctions/blob/main/datasets/at/meine_abgeordneten/crawler.py#L178

datasets/us/finra_enforcement_actions/us_finra_enforcement_actions.yaml
```diff
@@ -0,0 +1,26 @@
+title: FINRA Enforcement Actions
+entry_point: crawler.py
+prefix: us-finra-enfor
+coverage:
+  frequency: daily
+load_db_uri: ${OPENSANCTIONS_DATABASE_URI}
+summary: >
+  Firms and individuals that violated FINRA rules
+description: |
+  This dataset comprises firmas and individuals who FINRA have taken
```
jbothma, line 10:
```suggestion
  This dataset comprises firms and individuals who FINRA have taken
```

datasets/us/finra_enforcement_actions/crawler.py
jbothma: this is creating URLs like "https://www.finra.org2022077093901/", - do you mean to be using case-id[1]?

datasets/us/finra_enforcement_actions/us_finra_enforcement_actions.yaml
```diff
@@ -0,0 +1,26 @@
+title: FINRA Enforcement Actions
```
jbothma, line 1:
```suggestion
title: US FINRA Enforcement Actions
```

datasets/us/finra_enforcement_actions/crawler.py
bgmello: Yes, I was testing the error and forgot to change. Sorry

datasets/us/finra_enforcement_actions/crawler.py
bgmello: Nice, I didn't know about this function

bgmello:
The audit_data will keep raising warning because we are not using the document-type. Is it ok? Or should we save it somewhere?

bgmello:
Done

jbothma:
Thank you!

I was about to merge but found a couple of issues we hadn't spotted before:

Screenshot 2024-03-20 at 12 41 19

I think because of this, let's rather just use LegalEntity for all. But I think the icons might still be useful.

Could you see how many names have commas in them? I just came across "142866 William A. Shopoff, Stephen R. Shopoff" which seems to be a company and two individuals as plaintiffs. I suspect (hope) with companies, the commas will be part of their name, so we can use them all as-is. And with persons, the commas hopefully always separate individuals. But I worry that they might sometines write "Smith, John" instead of "John Smith". could you check if it's safe to split on comma and emit an entity for each name?

As usual, more to it than we think.

I suggest also making cache_days=1 in fetch() then dev/debugging will be faster.

jbothma:
Could you also add it to the crime collection?

bgmello:
Apparently there is the two cases:

- `,` in the name of a person
Screenshot 2024-03-20 at 15 38 29

- `,` separating two persons
Screenshot 2024-03-20 at 15 38 58


So I'm not sure how could we reliably split the names. Do you have any ideas?

bgmello:
This also happens in the opensanctions/crawler-planning#113. I took a look at that issue but notice this problem that I wasn't sure how to resolve:

Screenshot 2024-03-20 at 15 41 54

(`,` in the name of the company, `,` separating entities)

jbothma:
For the commas, I suggest

1. replace `, LLC` with ` LLC`, replace `, Inc.` with ` Inc.`
2. if after that there's still a comma in the name, look up the name in a `comma_names` [lookup](https://zavod.opensanctions.org/best_practices/datapatch_lookups/) like this:

```yaml
lookups:
  comma_names:
    options:
      - match: Vision Financial Markets LLC, Vision Brokerage Services LLC
        names:
          - Vision Financial Markets LLC
          - Vision Brokerage Services LLC
 ```
and make an entity for each name

and if you don't find a match, warn.

bgmello:
What I did in this last commit was apply the clean_name function and if there is still a comma, then we ignore the row because we cannot know for sure if it is just one entity or more. What do you think? Is it ok to do like this? Or is it better to emit all entities?

jbothma:
So I finished this up by spllitting what I can automatically (names that don't include and or &) and then adding manual mappings for the rest. After normalisation (lowercase) there were around 116 I think.

Github Copilot did an excellent job of doing the heavy lifting after a few examples. It could be cleaner, but I didn't want to put more time into it.

Another thing that helped was that it looks like companies and individuals are usually in different divs and spans. I added a newline to the span and later split on it to get one string for companies, and another for individuals.

PR #667, opened 2024-03-20
Add crawler for ESMA sanctions

SimonThordal:
Closes https://github.com/orgs/opensanctions/projects/2/views/1?pane=issue&itemId=54657779

### DIFF
--- a/datasets/eu/esma_sanctions/crawler.py
+++ b/datasets/eu/esma_sanctions/crawler.py
@@ -1,0 +1,67 @@
+from lxml import html
+
+from zavod import Context, helpers
+
+
+def get_json(context: Context, batch_size: int):
+    params = {
+        "q": "*",
+        "rows": batch_size,
+        "start": 0,
+        "wt": "json",
+    }
+    cont = True
+    while cont:
+        resp = context.fetch_json(context.data_url, params=params)
+        for doc in resp["response"]["docs"]:
+            yield doc
+        params["start"] += batch_size
+        cont = resp["response"]["numFound"] >= params["start"]
+
+
+def parse_name(name_markup: str) -> str:
+    """
+    Remove markup from name if present.
+    """
+    if name_markup is None:
+        return None
+    name = html.fromstring(name_markup).text_content().strip()
+    return name
+
+
+def crawl(context: Context) -> None:
+    for row in get_json(context, 1000):
+        id_hash = context.make_id(
+            row.pop("sn_entityEsmaID", None),
+            row.get("sn_entityLEI"),
+            row.get("sn_entityName"),
+            row.get("sn_otherEntityName"),
+            row.get("sn_otherEntityLEI"),
+        )
+        if id_hash is None or row.get("sn_sanctionEsmaID") is None:
+            context.log.warn(
+                "Response missing entity or sanction identifier.", response=row
+            )
+            continue
+        entity = context.make("Company")
+        if row.get("sn_entityLEI") is not None:
+            entity.id = f"lei-{row.get('sn_entityLEI')}"
+            entity.add("leiCode", row.pop("sn_entityLEI", None))
+        elif row.get("sn_otherEntityLEI") is not None:
+            entity.id = f"lei-{row.get('sn_otherEntityLEI')}"
+            entity.add("leiCode", row.pop("sn_otherEntityLEI", None))
+        else:
+            entity.id = id_hash
+        entity.add("name", parse_name(row.pop("sn_entityName", None)))
+        entity.add("name", parse_name(row.pop("sn_otherEntityName", None)))
+
+        sanction = helpers.make_sanction(context, entity)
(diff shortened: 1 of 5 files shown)


datasets/eu/esma_sanctions/eu_esma_sanctions.yaml
```diff
@@ -0,0 +1,25 @@
+title: "EU ESMA Sanctions"
+entry_point: crawler.py
+prefix: esma-sanctions
+coverage:
+  frequency: daily
+  start: 2024-03-12
+summary: |
+  List of sanctioned entities, as published by the European Securities and Markets Authority (ESMA).
+  The directives and regulations underpinning the sanctions can be found in the [Interactive Single Rulebook.](https://www.esma.europa.eu/publications-and-data/interactive-single-)rulebook
```
pudo, line 9: I think the summary field doesn't have markdown. This may want to move down to description.

datasets/eu/esma_sanctions/eu_esma_sanctions.yaml
```diff
@@ -0,0 +1,25 @@
+title: "EU ESMA Sanctions"
+entry_point: crawler.py
+prefix: esma-sanctions
+coverage:
+  frequency: daily
+  start: 2024-03-12
```
pudo, line 6: 20 :)

pudo: I think we need one in the yaml: `load_db_uri: ${OPENSANCTIONS_DATABASE_URI}`

datasets/eu/esma_sanctions/crawler.py
```diff
@@ -0,0 +1,75 @@
...
+    if name_markup is None:
+        return None
+    name = html.fromstring(name_markup).text_content().strip()
+    return name
+
+
+def crawl(context: Context) -> None:
+    for row in get_json(context, 1000):
+        if (
+            row.get("sn_entityName") is None
```
pudo, line 35: One alternative way to do this is to just throw all of these into `context.make_id` together, and see if the result of the whole thing is `None` :)

datasets/eu/esma_sanctions/crawler.py
```diff
@@ -0,0 +1,75 @@
...
+        sanction.add("program", row.pop("sn_sanctionLegalFrameworkName", None))
+        sanction.add("startDate", row.pop("sn_date", None))
+        sanction.add("endDate", row.pop("sn_expirationDate", None))
+        sanction.add("reason", row.pop("sn_text", None))
+        sanction.add("provisions", row.pop("sn_natureFullName", None))
+        sanction.set("authority", row.pop("sn_ncaCodeFullName", None))
+        sanction.set("country", row.pop("sn_countryName", None))
+        context.emit(entity, target=True)
+        sanction.add("entity", entity.id)
+        context.emit(sanction, target=False)
```
pudo, line 75: `target=False` is implicit, sanction.entity should be set in `h.make_sanction`

pudo (commented):
Nice going!

datasets/eu/esma_sanctions/eu_esma_sanctions.yaml
SimonThordal: Fixed and added :)

datasets/eu/esma_sanctions/crawler.py
SimonThordal: Makes sense, changed it

datasets/eu/esma_saris/crawler.py
```diff
@@ -31,31 +31,26 @@ def crawl(context: Context) -> None:
...
-            target = False
-            if row.get("effectiveTo") is not None:
-                sanction.add("endDate", row.pop("effectiveTo"))
-            else:
-                entity.add("topic", "sanction")
-                target = True
-            context.emit(sanction, target=target)
-
-            if row.get("issuer") is not None:
+            sanction.add("endDate", row.pop("effectiveTo"))
```
SimonThordal, line 41: I had some logic here that only added the `sanction` topic if the sanction hadn't ended, but looking back I guess that's wrong? The topic will always be `sanction` for this dataset, but if it has an end-date you'll know that the sanction no longer is enforced.

datasets/eu/esma_saris/crawler.py
```diff
@@ -31,31 +31,26 @@ def crawl(context: Context) -> None:
...
+            entity.add("topic", "sanction")
```
pudo, line 42: `topics`

pudo (changes requested):
Just a typo and a restore, then we're there!

datasets/eu/esma_saris/crawler.py
SimonThordal: OK, I restored it :)

pudo (approved):
Nice!

PR #673, opened 2024-03-21
Zavod validate step

jbothma:
Adds validator interface used in `zavod run` and `zavod validate`.

`zavod run` skips validation for collections.

Both commands exit with non-zero status if an aborting validator triggers abort.


zavod/zavod/verifiers/assertions.py
```diff
@@ -0,0 +1,66 @@
+from typing import Dict, Any, Optional, cast
+from zavod.context import Context
+
+from zavod.entity import Entity
+from zavod.meta.assertion import Assertion, Comparison, Metric
+from zavod.exporters.statistics import Statistics
```
jbothma, line 6: @pudo is this a bit too cheeky or is it reasonable to leave this like this til we actually want to mess with the numbers used by the assertions and debuild then?

zavod/setup.py
```diff
@@ -22,7 +22,7 @@
     zip_safe=False,
     install_requires=[
         "followthemoney == 3.5.*",
-        "nomenklatura[leveldb] == 3.10.*",
+        "nomenklatura[leveldb] @ git+https://github.com/opensanctions/nomenklatura.git@has-entity",
```
jbothma, line 25: Fix when nomenklatura's upgraded

zavod/zavod/validators/assertions.py
```diff
@@ -0,0 +1,74 @@
+from typing import Dict, Any, Optional, cast
+from zavod.context import Context
+
+from zavod.entity import Entity
+from zavod.meta.assertion import Assertion, Comparison, Metric
+from zavod.exporters.statistics import Statistics
```
jbothma, line 6: this feels a tad icky, but rather than move the Statistics object somewhere more specific, I think once we want more fancy assertions, we'll build a dedicated aggregator and leave the exporter alone.

zavod/zavod/cli.py
```diff
@@ -119,6 +135,7 @@ def run(
     try:
         clear_store(dataset)
         view = get_view(dataset, external=False)
+        validate_dataset(dataset, view)
```
pudo, line 138: I'm not sure this should be in here. It definitely wants to be in `run`, but maybe not in `export`?

zavod/zavod/validators/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+                if self.view.has_entity(other_id):
+                    continue
+                self.context.log.warning(
+                    f"{entity.id} property {prop.name} references missing id {other_id}"
+                )
+                self.fail = True
+
+
+# FollowTheMoney prevents direct self-references so we check 1 level deep
+class SelfReferenceValidator(BaseValidator):
```
pudo, line 28: I think we'll have fun with this an `wikidata`.

zavod/zavod/validators/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+
+def validate_dataset(dataset: Dataset, view: View) -> None:
+    try:
+        context = Context(dataset)
+        context.begin(clear=False)
+
+        validators = [validator(context, view) for validator in VALIDATORS]
+        for idx, entity in enumerate(view.entities()):
+            if idx > 0 and idx % 10000 == 0:
+                context.log.info("Verified %s entities..." % idx, dataset=dataset.name)
```
pudo, line 70: I'm interested in logging in this: can we make sure that `validate` has the issues log correctly populated and appends (clear above looks good) to it? I want warnings from the validators to show up on the issues page, of course :)

zavod/zavod/validators/assertions.py
```diff
@@ -0,0 +1,74 @@
...
+    def feed(self, entity: Entity) -> None:
+        self.stats.observe(entity)
+
+    def finish(self) -> None:
+        for assertion in self.context.dataset.assertions:
+            if not check_assertion(self.context, self.stats.as_dict(), assertion):
+                self.failed = True
+        
+        if self.failed:
+            raise RuntimeError("One or more assertions failed.")
```
pudo, line 74: Does this want to live in here? Or should all `BaseValidators` have a `abort` flag, and setting it will cancel the `export/publish` phases if run together?

pudo (commented):
I'm very excited about how this is implemented, just wonder if we should have more clearly defined notions of how we'd abort a botched validation. Throwing a RuntimeException works, but it's not the most elegant thing we could do :)

zavod/zavod/validators/assertions.py
pudo: no I like this a lot. Let's split it up early. The stats is part of our exposed interface so separating them makes a ton of sense to get ease of change for the validator.

jbothma: Do you mean I should implement statistics for the validator now, or use it as implemented here until we want a more powerful validator?

zavod/zavod/validators/__init__.py
jbothma: Not sure if I tested this in the best way - what felt safest was to check the archive dir in test_cli. This only works when the group setup (cli.cli) is run by the click test runner.

pudo (approved):
Looks good to me, unless you want to do the Statistics fork now? 

One thought: do we want to validate collections, too? Or just sources? Just in terms if how actionable the warnings are - it's all in the sources, no?

jbothma:
oh yes definitely makes sense not to run this on collections, so we can keep publishing even when one source is misbehaving. I fix.

PR #689, opened 2024-03-25
Add successor stub if we have an id for it

jbothma:


jbothma:
Gonna let it run through before merging

PR #693, opened 2024-03-26
Missing topics

jbothma:
@pudo particularly would like your eyes on ESMA topics

### DIFF
--- a/datasets/_wikidata/curated/wd_curated.yml
+++ b/datasets/_wikidata/curated/wd_curated.yml
@@ -24,3 +24,13 @@ url: https://docs.google.com/spreadsheets/d/1UyVYpg8jF6q4zRGGlIu_SaEe_QKqfhNcEHG
 data:
   url: https://docs.google.com/spreadsheets/d/e/2PACX-1vQsmLFolFMG4eKTx9rEoJEoZEYjOlO1Vorw_wCc4rCEZCbC_0eNHGjgg57WjBDghnsQuIFRsuUTNyfu/pub?gid=0&output=csv
   format: CSV
+
+assertions:
+  min:
+    schema_entities:
+      Person: 131
+      Organization: 2
+  max:
+    schema_entities:
+      Person: 135
+      Organization: 4
--- a/datasets/_wikidata/curated/wd_curated.yml
+++ b/datasets/_wikidata/curated/wd_curated.yml
@@ -20,7 +20,7 @@ publisher:
     sister projects including Wikipedia, Wikivoyage, Wiktionary, Wikisource, and others.
   url: https://www.wikidata.org
   official: false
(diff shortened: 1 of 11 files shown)


datasets/_wikidata/curated/wd_curated.yml
```diff
@@ -20,7 +20,17 @@ publisher:
     sister projects including Wikipedia, Wikivoyage, Wiktionary, Wikisource, and others.
   url: https://www.wikidata.org
   official: false
-url: https://docs.google.com/spreadsheets/d/1Yp51aX3tU1CezqaQZH-V2R_7P8gp82pQq2rzwumA2wU/edit#gid=0
+url: https://docs.google.com/spreadsheets/d/1UyVYpg8jF6q4zRGGlIu_SaEe_QKqfhNcEHG5KurL_r8/edit#gid=0
```
pudo, line 23: Thank you for doing this annoying thing.

datasets/eu/esma_saris/eu_esma_saris.yaml
```diff
@@ -38,3 +38,28 @@ url: https://registers.esma.europa.eu/publication/searchRegister?core=esma_regis
...
+          - "04. Take-over bid"
+          - "1_OLD_Market management arrangements"
+          - "2_OLD_Undisclosed price sensitive information (awaiting public release)"
+          - "3_OLD_Issuers failure to disclose periodic information on time"
+          - "5_OLD_Other noncompliance with rules of regulated market"
+          - "6_OLD_Filling of tender offers and related events"
+          - "7_OLD_Other disorderly trading conditions"
+          - "Removal"
+          - null
+        value: "corp.disqual"
```
pudo, line 61: I know this is anticlimactic, but perhaps we want to go for `null` here?

datasets/ru/billionaires/crawler.py
```diff
@@ -19,6 +19,8 @@ def crawl_row(context: Context, row: Dict[str, str]):
     entity.add("wikidataId", qid)
     topics = [t.strip() for t in row.get("topics", "").split(";")]
     topics = [t for t in topics if len(t)]
+    if len(topics) == 0:
+        topics.append("poi")
```
pudo, line 23: Why not in the source sheet?

pudo (changes requested):
I'm really conflicted about `corp.disqual` in the saris crawler... They just seem to be disabled for technical reasons. Disqual is the closest we have, but it's not consistent with the other uses of that topic. Maybe we skip topics+target on these (but still emit)?

datasets/eu/esma_saris/eu_esma_saris.yaml
jbothma: We agred on using `poi` rather than targets=False and topics=[]

datasets/ru/billionaires/crawler.py
jbothma: I kinda wanted to separate fallback from explicit topics but that makes sense

jbothma:
I also made ended ESMA suspensions/removals non-target

PR #697, opened 2024-03-27
adding US DEA Fugitives crawler

bgmello:
Fixes opensanctions/crawler-planning#98

### DIFF
--- a/datasets/us/dea_fugitives/crawler.py
+++ b/datasets/us/dea_fugitives/crawler.py
@@ -1,0 +1,69 @@
+from zavod import Context, helpers as h
+
+def parse_table(table):
+    """This function is used to parse the table of Labels and Descriptions
+    about the fugitive and return as a dict.
+    """
+
+    info_dict = {}
+
+    # The first row will always be the header (Label, Description)
+    # So we can skip it.
+    for row in table.findall(".//tr")[1:]:
+        label = row.findtext(".//td[1]")
+        description = row.findtext(".//td[2]")
+        info_dict[label] = description
+    
+    return info_dict
+
+
+def crawl_item(fugitive_url: str, context: Context):
+
+    response = context.fetch_html(fugitive_url)
+
+    name = response.findtext('.//h2[@class="fugitive__title"]')
+    info_dict = parse_table(response.find('.//table'))
+
+    entity = context.make("Person")
+    entity.id = context.make_id(name)
+
+    if "Sex" in info_dict:
+        if info_dict["Sex"] == "Male":
+            entity.add("gender", "male")
+        else:
+            entity.add("gender", "female")
+
+    if "Year of Birth" in info_dict:
+        entity.add("birthDate", info_dict["Year of Birth"])
+
+
+    sanction_description = ''.join([d.text_content() for d in response.findall('.//div[@class="meta"]') if "Wanted for the following" in d.text_content()])
+
+    if "NCIC #" in info_dict:
+        sanction = h.make_sanction(context, entity, key=info_dict["NCIC #"])
+    else:
+        sanction = h.make_sanction(context, entity)
+
+    sanction.add("description", sanction_description)
+    sanction.add("sourceUrl", fugitive_url)
+
+    context.emit(entity, target=True)
+    context.emit(sanction)
+
+
+def crawl(context: Context):
+    # Each page only displays 10 fugitives at a time, so we need to loop until we don't find any more fugitives
+
+    base_url = context.data_url
+
+    page_num = 0
+
+    while True:
+        url = base_url + "?page=" + str(page_num)
+        response = context.fetch_html(url, cache_days=1)
+        response.make_links_absolute(url)
+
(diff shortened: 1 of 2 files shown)


datasets/us/dea_fugitives/us_dea_fugitives.yaml
```diff
@@ -0,0 +1,26 @@
...
+description: |
+  The Drug Enforcement Administration's (DEA) most wanted list consists
+  of the world's most dangerous drug traffickers and criminal organizations.
+  The individuals on the list are wanted for a variety of offenses,
+  including criminal conspiracy, money laundering, trafficking of
+  illegal substances, and murder.  The most wanted criminals
+  are believed to be highly dangerous and pose a significant
+  risk to public safety.
+publisher:
+  name: US's Drug Enforcement Administration
```
jbothma, line 19:
```suggestion
  name: US Drug Enforcement Administration
```

datasets/us/dea_fugitives/crawler.py
```diff
@@ -0,0 +1,75 @@
...
+
+    while True:
+        url = base_url + "?page=" + str(page_num)
+        response = context.fetch_html(url, cache_days=1)
+        response.make_links_absolute(url)
+
+        for item in response.findall('.//div[@class="teaser "]/div/h3/a'):
+            crawl_item(item.get("href"), context)
+
+        page_num += 1
```
jbothma, line 75: Does this loop exit?

datasets/us/dea_fugitives/crawler.py
```diff
@@ -0,0 +1,75 @@
...
+def crawl(context: Context):
+    # Each page only displays 10 fugitives at a time, so we need to loop until we don't find any more fugitives
+
+    base_url = context.data_url
+
+    page_num = 0
+
+    while True:
+        url = base_url + "?page=" + str(page_num)
+        response = context.fetch_html(url, cache_days=1)
```
jbothma, line 69: This seems to get blocked very quickly - looks like you need some retry config https://zavod.opensanctions.org/metadata/#http-options

make sure it considers 403 a code it can retry on, and perhaps increase from the usual delay factor - it does look like the block is lifted a short while after being blocked

datasets/us/dea_fugitives/crawler.py
```diff
@@ -0,0 +1,75 @@
...
+    info_dict = parse_table(response.find(".//table"))
+
+    entity = context.make("Person")
+    entity.id = context.make_id(name)
+
+    if "Sex" in info_dict:
+        if info_dict["Sex"] == "Male":
+            entity.add("gender", "male")
+        else:
+            entity.add("gender", "female")
```
jbothma, line 34: what are the values they provide? I think it's better to see if we can automatically normalise and accept these values, and add [lookup mappings](https://zavod.opensanctions.org/best_practices/datapatch_lookups/) for those we can't

datasets/us/dea_fugitives/crawler.py
```diff
@@ -0,0 +1,75 @@
...
+        ]
+    )
+
+    if "NCIC #" in info_dict:
+        sanction = h.make_sanction(context, entity, key=info_dict["NCIC #"])
+    else:
+        sanction = h.make_sanction(context, entity)
+
+    sanction.add("description", sanction_description)
+    sanction.add("sourceUrl", fugitive_url)
```
jbothma, line 53: I think since this is crime, let's add the description to the person's description field instead, and the URL to the person. Rather than a sanction entity. I know it isn't super clearly defined. We tend to use sanction entities for proper financial sanctions, and things that are a bit more closely related to sanctions, e.g. debarments with interesting metadata, especially if they have start/end dates

datasets/us/dea_fugitives/crawler.py
jbothma: I suspect it will also be painful running this in CI, so lets' add `ci_test: false` to the metadata for the same reason.

jbothma (changes requested):
Nice!

datasets/us/dea_fugitives/crawler.py
jbothma: It looks like this is better-solved with a 1s sleep - could you cherry-pick this? https://github.com/opensanctions/opensanctions/commit/745c748a59ac07f4ee636af251cabc2f01e14cc0

datasets/us/dea_fugitives/crawler.py
jbothma: have a look how I addressed this in https://github.com/opensanctions/opensanctions/commit/745c748a59ac07f4ee636af251cabc2f01e14cc0 - we prefer mappings like this over python logic since it's clear once you know the approach and easy to add more cases.

bgmello: Nice, I will implement this pattern in the next crawlers

datasets/us/dea_fugitives/crawler.py
bgmello: Done

PR #705, opened 2024-03-28
adding Turkish parliament crawler

bgmello:
There is an issue when trying to fetch the page of each member, because the websites sets a cookie. I was able to run for some pages but after some time it fails. I suppose it's possible to do using selenium because the cookies are set automatically in the page. Any idea on how to solve this issue? @jbothma 

opensanctions/crawler-planning#64

### DIFF
--- a/datasets/tr/parliament/crawler.py
+++ b/datasets/tr/parliament/crawler.py
@@ -1,0 +1,102 @@
+import re
+from lxml import html
+
+from zavod import Context, helpers as h
+from zavod.logic.pep import categorise
+
+BASE_URL = "https://www.tbmm.gov.tr"
+HEADERS = {
+    "Host": "www.tbmm.gov.tr",
+    "User-Agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.4 Safari/605.1.15",
+    "Referer": "https://www.tbmm.gov.tr/milletvekili/liste",
+}
+
+COOKIES = {
+    "TS6d43ef95027": "08ffcef486ab2000c4eb97c00fb5974b83a784e35ed9c80238e4c675dd554055520927fb3bba9d1d080a7c4b5f113000116852f1d1ddd515bd20c2cd5da5d10b9841a4e7f20c8190e864e2d4f3cd4a8b050a7ee5e4dcb4172ae46e363fa26eee",
+    "TS8711347c077": "08ffcef486ab2800d4ef7985e124f76edc175ec090f123faab048a0a7c71cfaf27206a9959ae53431c8364f4d04ae53c0841a58d191720007248e9bf00583a51a1566711f8ce2ba9c82a914b23109318b6c0030123067126",
+    "TBMM": "!Q+kWfZ7tiLZBh2sMIgyhC/skcd/XaJIw5qljH+4VWyK4YYQkY8iQ2QWTBzto82JJ4OqiSFV+ngxYsQ==",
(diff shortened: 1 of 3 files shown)


datasets/tr/parliament/tr_parliament.yml
```diff
@@ -0,0 +1,37 @@
...
+description: |
+  The Grand National Assembly of Turkey is the unicameral Turkish legislature.
+  It is the sole body given the legislative prerogatives by the Turkish Constitution.
+  There are 600 members of parliament (deputies) who are elected for a five-year term
+  from 87 electoral districts which represent the 81 administrative provinces of Turkey
+  (Istanbul and Ankara are divided into three electoral districts whereas
+  İzmir and Bursa are divided into two each because of its large populations).
+publisher:
+  name: Türkiye Büyük Millet Meclisi
+  description: Grand Nationasl Assembly of Turkey
```
jbothma, line 18:
```suggestion
  description: Grand National Assembly of Turkey
```

datasets/tr/parliament/crawler.py
```diff
@@ -0,0 +1,105 @@
...
+def crawl_item(deputy_url: str, context: Context):
+    response = context.http.get(deputy_url, headers=HEADERS, cookies=COOKIES)
+
+    doc = html.fromstring(response.text)
+
+    name = doc.findtext('.//*[@id="content-title-type"]/span').replace(" - ")
+    place_year_of_birth = doc.findtext(
+        './/*[@class="col-md-12 profile-ozgecmis-div"]/span/div[1]'
+    ).split(",")[0]
+    place_of_birth, year_of_birth = place_year_of_birth.split(" – ")
```
jbothma, line 51: it looks like this is a bunch of free text, so I think it's safest not to bother trying to parse the place and date of birth for now.

jbothma (commented):
So http requests done via the context object already maintain cookies set in each request. The trick is to visit the page or do the thing that sets those cookies in the first place. e.g. https://github.com/opensanctions/opensanctions/blob/main/datasets/eu/esma_saris/crawler.py#L8

bgmello:
I tried sending a get to that url before the crawl but it stills doesn't work. I also tested in a Jupyter notebook, although it sets the cookie, it gives an error in the deputy page:

Screenshot 2024-04-04 at 08 42 40

jbothma:
to use the session, you need to do `ses.get` instead of `requests.get`

jbothma:
Some things you could try:

- see if the cookies that are sent in the working browser request are all set in the requests session when you make that request - perhaps some cookie is only set client-side?
- make sure you're sending your custom headers with the headers argument in all the requests
- you could make a request work temporarily by passing cookies which you got frm the browser, right? Can you tell if any are missing in the response from /liste? Can you tell where they come from if not that page?

jbothma:
when I directly open an MP's page in an incognito window it shows nothing, then I see a second request to the same page. The first didn't seend cookies because it's a new incognito session. The second sent the cookies that were sent in the first request. Perhaps if you make the request to the MP page twice, at least in cases where the expected content isn't found, it will work?

bgmello:
I tried doing the request twice to the same endpoint but it didn't work, I tried analyzing using burp suite and apparently it also needs to send requests to other three endpoints that are in the form `/TSPD/{hash}?type=11`, `/TSPD/{hash}?type=12`, `/TSPD/{hash}?type=13`, where hash apparently changes every time. I tried to verify if hash was some of the cookies but apparently it is not.
Screenshot 2024-04-04 at 17 34 50
(82 is the first request without cookies and 86 is when it actually sends the information)

jbothma:
@pudo what do you think of a time-boxed experiment using https://github.com/scrapinghub/splash to see if allowing the other requests to happen is an easy fix?

jbothma:
I've added an unblocking service which reliably loads the list page - I've used that to emit these people for now https://github.com/opensanctions/opensanctions/pull/808

I'm working with the support of this service to see if we can get the detail pages working to get more information about them.

Thanks for your work on this!

PR #717, opened 2024-04-01
Adding support for POST (and other) methods to context http helpers, …

dchaplinsky:
…rewiring also caching to use fingerprints for requests, switching context to orjson, rearranging imports to comply with mypy, adding tests for failures, posts and caching

### DIFF
--- a/zavod/zavod/context.py
+++ b/zavod/zavod/context.py
@@ -1,17 +1,18 @@
-import json
-from lxml import html, etree
+import hashlib
 from pathlib import Path
 from datetime import datetime
-from requests import Response
-from prefixdate import DatePrefix
 from functools import cached_property
 from typing import Any, Optional, Union, Dict, List, Tuple, Mapping
+import orjson
+from requests import Response
+from prefixdate import DatePrefix
+from lxml import html, etree
 from datapatch import LookupException, Result, Lookup
 from followthemoney.schema import Schema
 from followthemoney.util import make_entity_id
 from nomenklatura.cache import Cache
-from rigour.urls import build_url, ParamsType
 from nomenklatura.util import PathLike
+from rigour.urls import build_url, ParamsType
 from structlog.contextvars import clear_contextvars, bind_contextvars
 
 from zavod import settings
@@ -31,6 +32,7 @@ from zavod.util import join_slug
 
 _Auth = Optional[Tuple[str, str]]
 _Headers = Optional[Mapping[str, str]]
+_Body = Optional[Union[Dict, List[Tuple]]]
 
 
 class Context:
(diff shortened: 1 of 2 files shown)


zavod/zavod/context.py
```diff
@@ -316,19 +372,61 @@ def fetch_html(
...
+        Args:
+            url: The URL of the request.
+            auth: The authentication credentials.
+            method: The HTTP method of the request.
+            data: The data to be sent in the request body.
+        Returns:
+            A unique fingerprint for the request (url + hashed payload).
+        """
+
+        hsh = hashlib.sha1(
```
pudo, line 417: Can we use `from banal import hash_data` for this? It's used in other places of the framework and will handle things like sets.

pudo (commented):
Looks good to me! One lib call we might add!

zavod/zavod/context.py
dchaplinsky: Sure. Let me know if you want to switch orjson.loads back to json.loads too.

pudo: The thing coming out of `hash_data` is already a sha1 :)

dchaplinsky: I mean this one: https://github.com/opensanctions/opensanctions/pull/717/files#diff-087e34be4ac7c0656253b06a5737f1de4d1abab35f78df0f286828c60f618b9cL286

pudo: Ah no that makes sense!

zavod/zavod/context.py
```diff
@@ -190,26 +192,49 @@ def fetch_resource(
...
+        # This mimics the allow_redirects login found in requests.sessions
+        if method in ["GET", "OPTIONS"]:
+            kwargs["allow_redirects"] = True
+        elif method == "HEAD":
+            kwargs["allow_redirects"] = False
+        elif method in ["POST", "PUT", "PATCH", "DELETE"]:
+            # Deliberately noop for the sake of explicitness
+            pass
+        else:
+            raise ValueError(f"Unsupported HTTP method: {method}")
```
jbothma, line 232: is this needed? How did this come about? I ask because I think 307 and 308 response automatically reuse the same method, while 302 and 301 switch to GET? And now we're always providing a "allow_redirects" rather than only providing one when the crawler has specifically done so.

dchaplinsky: As I mentioned in the comment, this is exact defaults taken from the requests lib.
i.e if you do requests.get with no allow_redirects set, it will set it to False, if you do requests.head, it'll set it to False.
Check this out:
https://github.com/psf/requests/blob/main/src/requests/sessions.py#L593-L624

jbothma: Oh cool. I see you're adding it because session.request always defaults to True while session.head defaults to False.

PR #714, opened 2024-04-01
adding US SEC Harmed Investors Crawler

bgmello:
Fixes opensanctions/crawler-planning#113

PR #723, opened 2024-04-02
Helper to split names separated by commas

jbothma:
I couldn't get a general regex approach working so here's the uncool hacky version.

Basically if there's an and or & and a , it assumes it needs a human to split to avoid splitting incorrectly.

For reuse in https://github.com/opensanctions/opensanctions/pull/714/files#diff-455320ba180621715e0aae133aeadb928283b559d3f504883089756003d006baR26


pudo (approved):
This is getting deep, especially with the name convention lookup. But as long as we have a test that breaks if we forget about, let's try and see what happens.

zavod/zavod/helpers/names.py
```diff
@@ -1,9 +1,18 @@
...
+from typing import List, Optional, cast
 from normality import collapse_spaces
 from followthemoney.util import join_text
+import re
 
+from zavod.context import Context
 from zavod.entity import Entity
 
+REGEX_AND = re.compile(r"(\band\b|&|\+)", re.I)
+REGEX_JUST_A_NAME = re.compile(r"^[a-z]+, [a-z]+$", re.I)
```
SimonThordal, line 10: I think `[a-z]` will mean the regex will miss any non-ascii characters. AFAIK Pythons built-in `re` doesn't support character classes, but there is a library called `regex` that will do it: 
https://pypi.org/project/regex/

SimonThordal: Just took a 2nd look and `\w+` and the `re.U` works as well :)

jbothma: yeah - I used a-z because \w includes numbers and I've seen companies referenced as just a number. This is really quite a specific solution to US sort of lists. I'll improve the regex name to indicate intent a bit more clearly. Of course we can do better with this case but I need to not go deeper here

jbothma: actually matching too widely is the safer thing to do and we'll notice if it becomes unbearable from warnings

PR #730, opened 2024-04-03
Add dossier centre pep list

atehe:
closes  [Dossier Center PEPs #114](https://github.com/opensanctions/crawler-planning/issues/114)
@jbothma  I was thinking this could be a sanction dataset as it is mentioned in the homepage of the list but it is currently added to the PEP list due to it name.
Please have look and let me know your thoughts.


datasets/ru/dossier_center_pep/ru_dosseier_center_pep.yml
```diff
@@ -0,0 +1,21 @@
...
+description: |
+  This dataset publishes the list of possible organizers of the Kremlin organized crime group
+  and their probable accomplices with brief profiles.
+
+entry_point: crawler.py
+publisher:
+  name: Dossier Center
+  description: |
+    The Dossier Center is a non-profit project of Mikhail Khodorkovsky.  Its goal is to promote
+    the formation of a rule of law state and civil society in Russia.
```
jbothma, line 15: according to https://peps.dossier.center/about/ I would rather speak to it being `a group of independent civil activists`

datasets/ru/dossier_center_pep/crawler.py
```diff
@@ -0,0 +1,152 @@
...
+        doc,
+        '//div[@class="b-pr-section__label"][contains(.//text(), "Гражданство")]//following-sibling::div[@class="b-pr-section__value"]',
+    )
+    nationality = None
+    if "Кипр" in citizenship:
+        nationality = "cy"
+    elif "Российская" in citizenship or "РФ" in citizenship:
+        nationality = "ru"
+    else:
+        context.log.warn(f"Unknown nationality: {citizenship}")
```
jbothma, line 100: a common pattern we use for this is to just stick the value in citizenship without python logic on it, then use a `type.country` [lookup](https://zavod.opensanctions.org/best_practices/datapatch_lookups/) for translations if needed.

datasets/ru/dossier_center_pep/ru_dosseier_center_pep.yml
```diff
@@ -0,0 +1,21 @@
+name: ru_dossier_center_pep
```
jbothma, line 1: Let's add [load_db_uri ](https://zavod.opensanctions.org/metadata/#exports) - a nice effect this will have I think is that the notes will show up on their profiles on OpenSanctions

jbothma (commented):
Great job here.

I see the reasoning for PEP, but this case this source is more like https://www.opensanctions.org/datasets/md_rise_profiles/ where it's a bunch of people who have compiled info about a bunch of maybe dodgy people. So instead of PEP positions and occupancies, let's just create the Persons and slap a `poi` topic on them.

Let's also remove the `pep` from names and titles and things and rather refer to "Dossier Centre Persons of Interest"

datasets/ru/dossier_center_poi/crawler.py
```diff
@@ -0,0 +1,131 @@
...
+    person.add("sourceUrl", url)
+    person.add("topics", "poi")
+
+    if date_of_birth:
+        person.add("birthDate", to_date(date_of_birth))
+    person.add("birthPlace", place_of_birth)
+    person.add("nationality", citizenships)
+    person.add("notes", [reason_on_list, possible_violation])
+    person.add(
+        "summary", "Probable Accomplices" if accomplice else "Probable Organizers"
```
jbothma, line 121:
```suggestion
        "summary", "Probable Accomplice" if accomplice else "Probable Organizer"
```

jbothma:
Thanks! I'll fix the imports shown in the lint CI job when I merge, but do take a look at those next time :)

PR #738, opened 2024-04-05
Israli crypto scrape

SimonThordal:
Closes https://github.com/orgs/opensanctions/projects/2/views/1?pane=issue&itemId=58401354


datasets/il/mod_crypto/il_mod_crypto.yml
```diff
@@ -0,0 +1,28 @@
+title: Israel Sanctioned Crypto Wallets List
+entry_point: crawler.py
+prefix: il-nbctf
+coverage:
+  frequency: daily
```
jbothma, line 5: these google sheets ones we set frequency to `never` and then add a note about it in the description as in https://www.opensanctions.org/datasets/lt_illegal_websites/

datasets/il/mod_crypto/il_mod_crypto.yml
```diff
@@ -0,0 +1,28 @@
...
+  start: 2021-10-22
+deploy:
+  schedule: "0 */6 * * *"
```
jbothma, line 8: I think you can make it run weekly instead of 6-hourly since it only needs to pick up data model changes or if someone goes and updates the csv

```suggestion
deploy:
  schedule: "@weekly"
```

datasets/il/mod_crypto/crawler.py
```diff
@@ -0,0 +1,89 @@
...
+                name
+                for name in [
+                    row.get("English Name"),
+                    row.get("Name Hebrew"),
+                    row.get("Name Arabic"),
+                ]
+                if name != ""
+            ]
+            if len(names) > 0:
+                person = context.make("Person")
```
jbothma, line 22: should we initialise person to None or set it to None in an else clause or is the persona associated with all the accounts below their name? If it's the latter case, perhaps a comment to that effect is a good idea to show it's intentional.

jbothma (commented):
very nice! Just a couple of things noted then we can merge

datasets/il/mod_crypto/crawler.py
SimonThordal: Oooh, good point. I'll init to None, I think the person is only tied to the accounts in their row.

datasets/il/mod_crypto/il_mod_crypto.yml
SimonThordal: Thanks, done!

pudo:
Nice work!

PR #750, opened 2024-04-10
adding Malaysia financial consumer alert list

bgmello:
Fixes opensanctions/crawler-planning#96

### DIFF
--- a/datasets/_collections/incoming.yml
+++ b/datasets/_collections/incoming.yml
@@ -13,4 +13,5 @@ children:
   - us_finra_actions
   - tr_wanted
   - ru_dossier_center_poi
+  - my_consumer_alert_list
 ci_test: false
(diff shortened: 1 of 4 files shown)


jbothma:
Could you save the html file in the data dir and then load and parse the file from there? a la https://github.com/opensanctions/opensanctions/blob/main/datasets/lt/fiu_freezes/crawler.py#L13

and one of these to remind us in 3 months to update

```python
def crawl(context):
    if datetime.datetime.now() > datetime.datetime(2024, 9, 15):
        context.log.warn("Check if there's an update of the data behind the bot check.")
```

Remember for these we set coverage.frequency: never, coverage.end to the date you save the data, and add a note in the description that the data is only updated occasionally

bgmello:
Done

jbothma:
I'm just confirming the most appropriate topic for this

pudo:
If we can't quite agree to the topic, let's default to `poi` as you suggested @jbothma  :)

PR #752, opened 2024-04-11
adding US FRB enforcement actions

bgmello:
Fixes opensanctions/crawler-planning#106


bgmello:
Still pending due to several names with commas that the crawler where not able to parse (335)

jbothma (commented):
sorry, looks like I had review comments I didn't send yet

bgmello:
I did that and also considered the case where we have " and " and if we separate the string and all the parts are valid names, then we are able to split correctly. It went down from ~330 cases to ~230, which is a very good improvement I think. I think it will probably go down to around 100 if we consider the name of the countries, do you known any library that has this information? Or is it best to just write it as a global variable as I did for the US states?

jbothma:
You could try `code = registry.country.clean(country,fuzzy=True)`
and then if code is None, it didn't match one of https://www.opensanctions.org/reference/#type.country

bgmello:
Done, now we are down to around 190 cases. Know most cases are like the following:

- stark bank group ltd. and first american bank, both of fort dodge, iowa
- florida bank group inc. and florida bank, both of tampa, florida
- miners and merchants bancorp inc., grundy, virginia

jbothma:
Excellent - so I think the next step is lookups for those in the yaml file.

Do you have github actions? It does a really amazing job of that once you've done a couple of examples.

bgmello:
You mean GitHub copilot, right? The AI tool to predict code. I've used AWS code whisper, which is a free version from Amazon, but I'll try copilot, it's probably better

jbothma:
Lol yes I mean github copilot, sorry.

What I did with similar cases is

1. put the strings I want to map in the yaml file, one line at a time
2. prefix them all with `    - match: ` using multiline edit
3. one by one, add a line after, add 
```
      names:
        - 
```
and start writing the correct names
4. wait a moment after adding the next line, and see if the prediction is good

bgmello:
Done, indeed it was much faster than doing manually, copilot guessed it correctly in almost all cases

PR #756, opened 2024-04-12
adding malaysian securities comission aob sanctions

bgmello:
Fixes opensanctions/crawler-planning#95

### DIFF
--- a/datasets/my/aob_sanctions/crawler.py
+++ b/datasets/my/aob_sanctions/crawler.py
@@ -1,0 +1,93 @@
+from typing import Generator, Dict, cast
+from lxml.etree import _Element
+from normality import collapse_spaces, slugify
+
+from zavod import Context, helpers as h
+
+
+def parse_table(table: _Element) -> Generator[Dict[str, str], None, None]:
+    """
+    First we find the headers by searching the th tags, then we iterate over the tr tags (i.e. rows).
+
+    Returns:
+        A generator that yields a dictionary of the table columns and values. The keys are the
+        column names and the values are the column values.
+    Raises:
+        AssertionError: If the headers don't match what we expect.
+    """
+    headers = [th.text_content().strip() for th in table.findall(".//th")]
+    for row in table.findall(".//tr")[1:]:
+        if headers is None:
+            headers = []
+            for el in row.findall("./td"):
+                headers.append(slugify(el.text_content()))
+            continue
+
+        cells = []
+        for el in row.findall("./td"):
+            cells.append(collapse_spaces(el.text_content()))
+
+        # There are some incorrect rows with a smaller
+        # number of values, we'll just ignore those
+        if len(cells) != len(headers):
+            continue
+
+        # The table has a last row with all empty values
+        if all(c == "" for c in cells):
+            continue
+
+        yield {hdr: c for hdr, c in zip(headers, cells)}
+
+
+def crawl_item(input_dict: dict, context: Context):
+    name_col = "Auditor" if "Auditor" in input_dict else "Parties Involved"
+    date_col = (
+        "Date of Action" if "Date of Action" in input_dict else "Date of AOB's Action"
+    )
+    description_col = (
+        "Brief Description of Breach"
+        if "Brief Description of Breach" in input_dict
+        else "Brief Description of Misconduct"
+    )
+    reason_col = (
+        "Nature of Breach"
+        if "Nature of Breach" in input_dict
+        else "Nature of Misconduct"
+    )
+    name = input_dict.pop(name_col)
+
+    clean_names = context.lookup("clean_names", name)
+    clean_names = cast("List[str]", clean_names.names)
+
+    for clean_name in clean_names:
+        if not clean_name:
+            context.log.warning(f"Unable to parse name: {name}")
+            return
+
+        entity = context.make("Person")
+        entity.id = context.make_id(clean_name)
+
+        entity.add("name", clean_name)
+        sanction = h.make_sanction(context, entity)
+
+        sanction.add("description", input_dict.pop(description_col))
+        sanction.add("reason", input_dict.pop(reason_col))
+
+        sanction.add(
+            "date",
+            h.parse_date(input_dict.pop(date_col), formats=["%d %B %Y"]),
+        )
+
+        context.emit(entity, target=True)
+        context.emit(sanction)
+
+        context.audit_data(input_dict, ignore=["No.", "No", "Action Taken"])
+
+
+def crawl(context: Context):
+    response = context.fetch_html(context.data_url)
+
+    # There is a table for each year
+    for table in response.findall(".//table"):
+        for item in parse_table(table):
+            crawl_item(item, context)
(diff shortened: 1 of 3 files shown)


datasets/my/aob_sanctions/my_aob_sanctions.yaml
```diff
@@ -0,0 +1,230 @@
...
+  Auditors sanctioned by the Malaysian's Securities Comission Audit Oversight Board
+description: |
+  This dataset comprises information about auditors sanctioned by
+  the Malaysian's Securities Comission Audit Oversight Board,
+  The focus of AOB enforcement is whether the auditors comply
+  with the recognised auditing and ethical standards.
+  Such action from the AOB may not necessary imply the audited
+  financial statement does not give a true and fair view.
+  The sanctions are: 
+    - Directing the person in breach to comply with the
```
jbothma, line 18: This gets interpreted as markdown, so it would be great to get blank lines to split the text into paragraphs to make reading a bit easier.
```suggestion
description: |
  This dataset comprises information about auditors sanctioned by
  the Malaysian's Securities Comission Audit Oversight Board.
  
  The focus of AOB enforcement is whether the auditors comply
  with the recognised auditing and ethical standards.
  Such action from the AOB may not necessary imply the audited
  financial statement does not give a true and fair view.
  
  Potential sanctions are: 
  
    - Directing the person in breach to comply with the
```

datasets/my/aob_sanctions/my_aob_sanctions.yaml
```diff
@@ -0,0 +1,230 @@
+title: Malaysian's Securities Comission Sanctions
```
jbothma, line 1:
```suggestion
title: Malaysia Securities Commission AOB Enforcements
```

datasets/my/aob_sanctions/my_aob_sanctions.yaml
```diff
@@ -0,0 +1,230 @@
...
+entry_point: crawler.py
+prefix: my-aob
+coverage:
+  frequency: daily
+  start: 2024-04-12
+load_db_uri: ${OPENSANCTIONS_DATABASE_URI}
+summary: >
+  Auditors sanctioned by the Malaysian's Securities Comission Audit Oversight Board
```
jbothma, line 9:
```suggestion
  Enforcement actions against auditors by the Malaysian Securities Commission Audit Oversight Board
```

datasets/my/aob_sanctions/my_aob_sanctions.yaml
```diff
@@ -0,0 +1,230 @@
...
+description: |
+  This dataset comprises information about auditors sanctioned by
+  the Malaysian's Securities Comission Audit Oversight Board,
```
jbothma, line 12:
```suggestion
  the Malaysian Securities Commission Audit Oversight Board,
```

datasets/my/aob_sanctions/crawler.py
```diff
@@ -0,0 +1,93 @@
...
+    description_col = (
+        "Brief Description of Breach"
+        if "Brief Description of Breach" in input_dict
+        else "Brief Description of Misconduct"
+    )
+    reason_col = (
+        "Nature of Breach"
+        if "Nature of Breach" in input_dict
+        else "Nature of Misconduct"
+    )
```
jbothma, line 56: A pattern we've used before is to "translate" column headings using lookups, e.g. https://github.com/opensanctions/opensanctions/blob/main/datasets/jp/mof_sanctions/jp_mof_sanctions.yml#L95 - in this case I think it's ok as it is, but keep in mind if you do much more dev on this part, or when this comes up again

datasets/my/aob_sanctions/crawler.py
```diff
@@ -0,0 +1,93 @@
...
+    name = input_dict.pop(name_col)
+
+    clean_names = context.lookup("clean_names", name)
+    clean_names = cast("List[str]", clean_names.names)
```
jbothma, line 60: I see all your effort with the name lookup, but in this case, how wrong would `clean_name = h.multi_split(name, ["(", " of "])[0]` be?

```suggestion
    res = context.lookup("clean_names", name)
    clean_names = cast("List[str]", res.names)
```

our naming convention is `res`, as in `result`, I think. Not super clear, but at least not changing the type of a variable

datasets/my/aob_sanctions/crawler.py
```diff
@@ -0,0 +1,93 @@
...
+
+        sanction.add("description", input_dict.pop(description_col))
+        sanction.add("reason", input_dict.pop(reason_col))
+
+        sanction.add(
+            "date",
+            h.parse_date(input_dict.pop(date_col), formats=["%d %B %Y"]),
+        )
+
+        context.emit(entity, target=True)
```
jbothma, line 81: let's add the topic `crime.fin` and the country code for malaysia to the entity.

But don't resolve until we get confirmation from @pudo - I think we might need a more fine-grained mapping of sanctions to topics since some of these sanctions are just a slap on the wrist

datasets/my/aob_sanctions/crawler.py
```diff
@@ -0,0 +1,93 @@
...
+        context.emit(sanction)
+
+        context.audit_data(input_dict, ignore=["No.", "No", "Action Taken"])
```
jbothma, line 84: Please put action taken into sanction:provisions

jbothma (changes requested):
nice work! A few small changes, and a query for Friedrich, then we have a nice addition.

datasets/my/aob_sanctions/crawler.py
bgmello: I haven't noticed this pattern when I was doing it, indeed most of them fall into this categories. Do you want me to update it to use that? There are some instances where this will fail, but we could use a lookup for those, for example:

```
Mohd. Neezal bin Md. Noordin Partner of AljeffriDean who was the engagement partner in the audit of a public interest entity (PIE) for the financial year ended 31 March 2011.
```

This would return with the `Partner` in the name, but it is not correct.

datasets/my/aob_sanctions/crawler.py
bgmello: Done

datasets/my/aob_sanctions/crawler.py
bgmello: Done

pudo:
hey @bgmello. Since yesterday we have a new topic, `reg.warn` which feels relevant here. Let's add this and wrap it up :)

datasets/my/aob_sanctions/crawler.py
```diff
@@ -0,0 +1,90 @@
...
+            continue
+
+        cells = []
+        for el in row.findall("./td"):
+            cells.append(collapse_spaces(el.text_content()))
+
+        # There are some incorrect rows with a smaller
+        # number of values, we'll just ignore those
+        if len(cells) != len(headers):
+            continue
```
jbothma, line 33: nope, can't just ignore data - we'd be missing people who are banned. I'll add a fix

PR #762, opened 2024-04-15
Crawler for Hetq Armenia PEPs

dhdaines:
Some things need to be resolved before merging:

1. migrate data to internal storage
2. endDate is being omitted, unsure if this is a good idea or not
3. some people (presumably family relations without official positions) are in the data but not PEPs, not certain what to do with them
4. some position titles are absurdly long
5. we *could* get some more places of birth in Armenian but it may not be worth the trouble

### DIFF
--- a/datasets/am/hetq/crawler.py
+++ b/datasets/am/hetq/crawler.py
@@ -7,6 +7,8 @@ import re
 from typing import Dict, List, Literal, Any
 from zipfile import ZipFile
 
+from lxml.html import document_fromstring
+
 from zavod import Context
 from zavod import helpers as h
 from zavod.logic.pep import categorise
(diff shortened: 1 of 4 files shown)


dhdaines:
Linking to issue on crawlers project: https://github.com/opensanctions/crawler-planning/issues/8

datasets/am/hetq/crawler.py
```diff
@@ -0,0 +1,298 @@
...
+                assert pep[f"position_{lang}"] == lastPosition
+
+
+def crawl_relations(
+    context: Context, zipfh: ZipFile, peps: Dict[int, Dict[str, Any]], person_id: int
+):
+    """Read relation graph and create entities."""
+    if person_id not in peps or "entity" not in peps[person_id]:
+        context.log.debug(f"Skipping relations for {person_id}: is not PEP")
+        return
```
jbothma, line 215: let's keep them due to the decision to emit all as `poi`

datasets/am/hetq/crawler.py
```diff
@@ -0,0 +1,298 @@
...
+        node_to_person[n["id"]] = n["personID"]
+    for e in graph.get("edges", []):
+        source = node_to_person.get(e.get("source"))
+        if source is None:
+            context.log.warning(f"Unknown source node in edge: {e}")
+            continue
+        # Also unnecessary to follow all the edges, only ones exiting
+        # the PEP in question
+        if source != person_id:
+            continue
```
jbothma, line 240: not sure if I'm following, but if we have chains of relationships, let's emit them too

jbothma: this dataset might have some similarity to https://www.opensanctions.org/datasets/md_rise_profiles/

jbothma (commented):
> migrate data to internal storage

I'll do this when I merge

> endDate is being omitted, unsure if this is a good idea or not

Yeah I think the way you're treating dates is a good compromise

> some people (presumably family relations without official positions) are in the data but not PEPs, not certain what to do with them

Let's include all the people in the data with the poi topic, and when the PEP logic gives them an occupancy, emit that and the position too.

> some position titles are absurdly long

That's ok - we have various instances of this - don't think it's a train smash anywhere

> we could get some more places of birth in Armenian but it may not be worth the trouble

Don't bother - what you've done so far is amazing.

datasets/am/hetq/crawler.py
```diff
@@ -0,0 +1,298 @@
...
+    context.log.debug(f"position: {position_name} years: {data['years']}")
+    if not position_name:
+        context.log.warning(f"Skipping person {person_id} with no position")
+        return None
+    position = h.make_position(
+        context,
+        name=position_name,
+        country="am",
+    )
+    categorisation = categorise(context, position, is_pep=True)
```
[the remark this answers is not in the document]
dhdaines: Good! Some of them are ... elaborate!

datasets/am/hetq/crawler.py
```diff
@@ -0,0 +1,298 @@
...
+    if person_id not in peps or "entity" not in peps[person_id]:
+        context.log.debug(f"Skipping relations for {person_id}: is not PEP")
+        return
+    # Unnecessary to use the Armenian versions here since the data is the same
+    graph = json.loads(zipfh.read(f"hetq-data/relations/{person_id}-en.json"))
+    if not graph:
+        context.log.warning(
+            f"Graph from hetq-data/relations/{person_id}-en.json is empty"
+        )
+        return
```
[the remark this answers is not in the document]
dhdaines: Yeah, unless the data itself changes this is not transient, so I'll make it debug-level.

datasets/am/hetq/crawler.py
dhdaines: Ah - actually the log message here (and the variable name `peps`) is misleading.  We need an entity in order to make a relation, that's all.   So if we keep and emit all the Person entities then we'll do the right thing here by default.

datasets/am/hetq/crawler.py
dhdaines: a person in the dataset, that is

datasets/am/hetq/crawler.py
```diff
@@ -0,0 +1,298 @@
...
+        # end_date=str(max(data["years"]) + 1),
+        categorisation=categorisation,
+    )
+    context.log.debug(f"categorization {categorisation} occupancy {occupancy}")
+    if occupancy is not None:
+        context.emit(person, target=True)
+        context.emit(position)
+        context.emit(occupancy)
+        return person
+    context.log.info(f"No entities emitted for {person_id} ({name_en} / {name_hy})")
```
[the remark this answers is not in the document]
dhdaines: Ok, makes sense!

datasets/am/hetq/crawler.py
dhdaines: Oh, nope. This is not actually the case - people in the relation graph without official positions have a personID but don't actually show up (for obvious reasons in hindsight) in the front-page list of PEPs.  So I will create POI entities for them and also follow their links recursively.

dhdaines: Also - they all have profile pages, most of which are empty, but some of which are not (and have DOBs), for instance https://data.hetq.am/hy/profile/577 (Anna Hakobyan)

dhdaines: So, there will be a new data dump as unfortunately I did not download these profiles the first time around since they don't show up on the front-page list!

dhdaines:
Okay, should be ready now.  Note that as I mentioned earlier, I updated the data dump (it should be 88.5MB now) to include the profile pages for related POIs, so make sure to copy the latest version.  A very small number of these POIs might also be PEPs as they have a position on their profile page but this probably isn't worth the trouble (the only one I can find is https://data.hetq.am/hy/profile/7227)

dhdaines:
I notice that the Person entities that are just "poi" and not "role.pep" are exported (in `statements.pack`) but do not ultimltely show up in the exported data (`targets.nested.json`), I assume this is normal behaviour?

PR #768, opened 2024-04-16
fix: set full name when crawling mx senate

SimonThordal:


jbothma (approved):
thanks for the fix! looks good to me

PR #778, opened 2024-04-18
Croatian PEPs 

kdeden:
First run at the Croatian PEPs crawler: 

* Each record has one PEP and up to two affiliations -- position, start/end dates, and legal entity
* Data is modeled as `Person`s, `Position`s, `LegalEntity`s, and `Occupancy`s.
* Many Occupancies are excluded as they do not meet the criteria outlined [here](https://zavod.opensanctions.org/helpers/#zavod.helpers.make_occupancy). In this case positions are excluded too. 
*  I'd like your feedback about how I handled the data irregularities in `handle_affiliation_data`. (There are some bizzare cases where start and end dates are provided with no position name or legal entity name.)


kdeden:
@jbothma just waiting for approval to run workflows

kdeden:
@jbothma I've updated the linting to include Ruff.

datasets/hr/public_officials/crawler.py
```diff
@@ -0,0 +1,126 @@
...
+        k.removeprefix(prefix): data.pop(k) for k in key_names if k.startswith(prefix)
+    }
+
+
+def crawl(context: Context):
+    """Fetches the current CSV file and crawls each row, making persons, legal entities, and positions"""
+    file_path = context.fetch_resource("daily_csv_release", context.data_url)
+    context.export_resource(file_path, CSV, title=context.SOURCE_TITLE)
+    with open(file_path, encoding="utf-8-sig") as fh:
+        reader = csv.DictReader(fh, fieldnames=FIELD_NAMES, delimiter=";")
```
jbothma, line 115: nice fix for their column header repetition. Could you also assert that the original headers are as we currently assume they are, e.g. just read the first row and close the file again and assert expected == actual? So we make sure we detect when they change their headers, breaking our assumption of the header meaning?

datasets/hr/public_officials/crawler.py
```diff
@@ -0,0 +1,126 @@
...
+    end_date = data.pop("Datum kraja obnašanja dužnosti")
+    if any([legal_entity_name, position_name, start_date, end_date]):
+        legal_entity = make_legal_entity(context, legal_entity_name)
+        position_name = position_name or "Unspecified position"
+        position = h.make_position(
+            context,
+            position_name,
+            topics=None,
+            organization=legal_entity,
+        )
```
jbothma, line 61: We don't currently create entities for the Position:organization property.

What we normally do here is `f"{position_name}, {entity_name}"` if both are available, otherwise just the one.

If neither is available, we skip the position unless we can infer a specific position from the dataset (e.g. we know the source only contains members of parliament)

Sorry for the sidetrack

datasets/hr/public_officials/crawler.py
```diff
@@ -0,0 +1,126 @@
...
+    """
+    legal_entity_name = data.pop("Pravna osoba u kojoj obnaša dužnost")
+    position_name = data.pop("dužnost")
+    start_date = data.pop("Datum početka obnašanja dužnosti")
+    end_date = data.pop("Datum kraja obnašanja dužnosti")
+    if any([legal_entity_name, position_name, start_date, end_date]):
+        legal_entity = make_legal_entity(context, legal_entity_name)
+        position_name = position_name or "Unspecified position"
+        position = h.make_position(
+            context,
```
jbothma, line 57: Also give the position a country please

datasets/hr/public_officials/crawler.py
```diff
@@ -0,0 +1,126 @@
...
+    context.audit_data(data)
+
+
+def make_person(context: Context, first_name: str, last_name: str) -> Entity:
+    person = context.make("Person")
+    person.id = context.make_slug(first_name, last_name)
+    person.add("firstName", first_name)
+    person.add("lastName", last_name)
+    person.add("country", "LT")
+    person.add("topics", "role.pep")
```
jbothma, line 86: make_occupancy will also give the person the topic and country.

datasets/hr/public_officials/crawler.py
```diff
@@ -0,0 +1,126 @@
...
+            person = make_person(context, row.pop("Ime"), row.pop("Prezime"))
+
+            affiliation_1 = extract_dict_keys_by_prefix(row, FIELD_NAMES, "Primarna ")
+            handle_affiliation_data(context, person, affiliation_1)
+
+            affiliation_2 = extract_dict_keys_by_prefix(row, FIELD_NAMES, "Sekundarna ")
+            handle_affiliation_data(context, person, affiliation_2)
+
+            context.audit_data(row)
+            context.emit(person, target=True)
```
jbothma, line 126: We only want to emit people in PEP datasets that are actually PEPs, and drop them when they leave our "PEP period" - the period for which we consider someone a PEP after they've left the position.

A pattern we sometimes use is to make a list of position and occupancy entities if any occupancies are created for the person, and only in that case, emit the person (and the occupancies and positions)

That way we also only emit positions for which there are persons who hold that position.

Even if we start off defaulting to assume all these positions are PEP positions (categorise(is_pep=True)) which seems reasonable looking at these positions, we might later categorise some as non-PEP, or some local gov person might not enter any new positions for 5 years but still be in the source file.

datasets/hr/public_officials/crawler.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def make_legal_entity(context: Context, legal_entity_name: str) -> Optional[Entity]:
+    legal_entity: Optional[Entity] = None
+    if legal_entity_name:
+        legal_entity = context.make("LegalEntity")
+        legal_entity.id = context.make_slug(legal_entity_name)
+        legal_entity.add("name", legal_entity_name)
+        context.emit(legal_entity)
+    return legal_entity
```
jbothma, line 36: remove this please - not using the organization property - sorry

datasets/hr/public_officials/crawler.py
```diff
@@ -0,0 +1,126 @@
...
+    "Primarna Datum kraja obnašanja dužnosti",
+    "Sekundarna dužnost",  # second affiliation
+    "Sekundarna Pravna osoba u kojoj obnaša dužnost",
+    "Sekundarna Datum početka obnašanja dužnosti",
+    "Sekundarna Datum kraja obnašanja dužnosti",
+]
+
+
+def convert_date(date: Optional[str]) -> Optional[str]:
+    return datetime.strptime(date, "%d/%m/%Y").strftime("%Y-%m-%d") if date else None
```
jbothma, line 26: Could you use our https://zavod.opensanctions.org/helpers/#zavod.helpers.parse_date helper instead? Lots of examples in the code

jbothma (commented):
awesome job! Just a few changes that I think won't be too tricky then we have a nice new dataset!

datasets/hr/public_officials/crawler.py
kdeden: Sure thing, just to be clear - if the expected columns are present and in order, but there are additional columns, we don't want to raise an error, right? We want to `log.warn` and move on. And if the expected columns are not present or out of order, only in that case will we raise the AssertionError and stop execution.

BTW looks like there is need to re-open the file - when passing `fieldnames` the DictReader assumes that the first line is data and reads it in.

jbothma: > if the expected columns are present and in order, but there are additional columns, we don't want to raise an error, right?

correct

> BTW looks like there is need to re-open the file - when passing fieldnames the DictReader assumes that the first line is data and reads it in.

oh of course!

Thanks!

PR #790, opened 2024-04-23
Try fixing doc build

jbothma:
The docs workflow is failing when we don't pin the python version.

It looks like https://github.com/pypa/pip/issues/10851 but I don't see new mentions of this and I didn't try different pip versions. The latest release of pip was in Feb, and the latest release of the ubuntu-latest runner last week doesn't seem to have touched these things, so I don't understand.

The default version of python on ubuntu-latest seems to be 3.10 but it works when pinning to 3.10. Maybe pinning with setup-python uses a different installation with different pip?

Anyway, pinning to 3.11 works, so whatever.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -29,7 +29,7 @@ jobs:
           sudo apt-get install -y -qq libicu-dev
       - uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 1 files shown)


SimonThordal (commented):
lgtm :100:

PR #807, opened 2024-04-29
Add utility to crawl sites expecting full browsers

jbothma:
Example usage in https://github.com/opensanctions/opensanctions/pull/808 and https://github.com/opensanctions/opensanctions/pull/809

Placed in shed for now, where we have some other reusable tools that aren't core yet. I think it'd be nice to see this running for a month before getting too excited.

I was tempted to enable this with `context.fetch_html(..., unblock_at_cost=True)` to use the cache in the context transparently, but then some requests required zyte's actions, and we might need other services at some point so the context API would get quite messy. So I'm thinking it's better to figure out the services we need and can rely on first, then see if we can move it into the more commonly used context API.


zavod/zavod/shed/zyte_api.py
```diff
@@ -0,0 +1,110 @@
...
+        auth=(settings.ZYTE_API_KEY, ""),
+        json=zyte_data,
+    )
+    api_response.raise_for_status()
+    text = api_response.json()["browserHtml"]
+    assert text is not None
+    doc = html.fromstring(text)
+
+    if not unblock_validator(doc):
+        if retries > 0:
```
SimonThordal, line 86: Should there be a check that `previous_retries < retries`?

SimonThordal (commented):
Looks good :)
Maybe we should consider using dependency injection or configuration objects for `context.fetch_html` to avoid making multiple ways of fetching data. That could make crawlers more configurable and we could use Pythons own implementation of [backoff](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.retry) instead of rolling our own.

jbothma:
I hear you on using dependency injection. One of the things that made me think overloading the current interface isn't ideal is that you can't supply headers or POST data to the browserHTML API in Zyte, but context.fetch_html supports that. So I had guards raising if those were set, but it felt a bit yucky. Is that the way it's often done? Raise if an unsupported argument is provided?

jbothma:
Another reason I didn't want to change the existing API too much yet is that I can totally imagine a crawler needing a paid unblocking service for one page, like the index, but subsequent detail page or API requests might not need that, so in one context you could want to use two different HTTP clients - the unblocking one, and the plain one.

zavod/zavod/shed/zyte_api.py
SimonThordal: Ah, got it, you're reducing the number of retries per call.

SimonThordal:
> I hear you on using dependency injection. One of the things that made me think overloading the current interface isn't ideal is that you can't supply headers or POST data to the browserHTML API in Zyte, but context.fetch_html supports that. So I had guards raising if those were set, but it felt a bit yucky. Is that the way it's often done? Raise if an unsupported argument is provided?

Right, so part of the `fetch_html` signature would be invalid if used with Zyte. Would it make sense to keep it as a separate method as you do here and then use the context.http method to avoid re-implementing exponential backoff?

SimonThordal:
> Another reason I didn't want to change the existing API too much yet is that I can totally imagine a crawler needing a paid unblocking service for one page, like the index, but subsequent detail page or API requests might not need that, so in one context you could want to use two different HTTP clients - the unblocking one, and the plain one.

Good point. I suppose in that case you would want to be able to inject a client/session into the `fetch_html` call itself instead of the `Context` constructor. In that way you could use different clients per call.

SimonThordal (approved):

jbothma:
yeah that'll be a good idea - to pass in a properly-configured session.Thank you!

jbothma:
So at this stage I think it's fine to use the context.http session. If it's really important to have different retry functionality, that can be overridden by mounting a more specific HTTPAdapter.

I realised though that the homebrew retries is still needed for retrying if the unblocking validator wasn't satisfied.

zavod/zavod/shed/zyte_api.py
jbothma: ah both incrementing and decrementing is a bit redundant - I changed to just incrementing previous and checking previous_retries < retries - thanks

SimonThordal (approved):
Looks good to me :+1:

PR #820, opened 2024-05-02
adding danish PEPs crawler

bgmello:
Fixes opensanctions/crawler-planning#164

### DIFF
--- a/datasets/dk/pep/crawler.py
+++ b/datasets/dk/pep/crawler.py
@@ -8,7 +8,7 @@ from datetime import datetime
 from zavod import Context, helpers as h
 from zavod.logic.pep import categorise
 
-RESOURCEs = [
+RESOURCES = [
     "PEP_listen.xlsx",
     "PEP_listen_faeroeerne.xlsx",
     "PEP_listen_groenland.xlsx",
@@ -25,6 +25,13 @@ def header_names(cells):
         headers.append(slugify(cell, "_").lower())
     return headers
 
+def parse_date(date: str) -> datetime:
+    dates = h.parse_date(date, formats=["%d.%m.%Y"])
+
+    if dates:
+        return dates[0]
+
+    return None
 
 def parse_old_pep(
     sheet: openpyxl.worksheet.worksheet.Worksheet,
@@ -113,7 +120,7 @@ def crawl_current_pep_item(input_dict: dict, context: Context):
     h.apply_name(entity, first_name=first_name, last_name=last_name)
 
     entity.add(
-        "birthDate", h.parse_date(input_dict.pop("birth-date"), formats=["%d.%m.%Y"])
+        "birthDate", parse_date(input_dict.pop("birth-date"))
     )
 
     position = h.make_position(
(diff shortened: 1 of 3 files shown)


datasets/dk/pep/dk_pep.yaml
```diff
@@ -0,0 +1,25 @@
...
+  Official PEP list for Denmark, Faroe Islands and Greenland
+description: |
+  The Danish Financial Supervisory Authority maintains a list of
+  the persons who are PEPs in Denmark. The list is based on reports
+  from the authorities and organisations, which, according
+  to the PEP order, must report the names and birthdays of the
+  PEPs associated with them. 
+publisher:
+  name: Finanstilsynet
+  description: Danish Financial Supervisory Authority
```
pudo, line 18: Can we add a bit of a description here?

datasets/dk/pep/crawler.py
```diff
@@ -0,0 +1,191 @@
...
+def crawl_current_pep_item(input_dict: dict, context: Context):
+    first_name = input_dict.pop("first-name")
+    last_name = input_dict.pop("last-name")
+
+    entity = context.make("Person")
+    entity.id = context.make_slug(first_name, last_name)
+    h.apply_name(entity, first_name=first_name, last_name=last_name)
+
+    entity.add(
+        "birthDate", h.parse_date(input_dict.pop("birth-date"), formats=["%d.%m.%Y"])
```
pudo, line 116: We're doing a few of these parse calls with this format, perhaps extract into a function?

datasets/dk/pep/crawler.py
```diff
@@ -0,0 +1,191 @@
...
+        else:
+            context.log.warning(f"Couldn't parse row: {[c.value for c in row]}")
+
+
+def crawl_current_pep_item(input_dict: dict, context: Context):
+    first_name = input_dict.pop("first-name")
+    last_name = input_dict.pop("last-name")
+
+    entity = context.make("Person")
+    entity.id = context.make_slug(first_name, last_name)
```
pudo, line 112: do we have some sort of person ID?

datasets/dk/pep/crawler.py
```diff
@@ -0,0 +1,191 @@
...
+from openpyxl import load_workbook
+from typing import Generator
+from pantomime.types import XLSX
+from normality import stringify, slugify
+from datetime import datetime
+
+from zavod import Context, helpers as h
+from zavod.logic.pep import categorise
+
+RESOURCEs = [
```
pudo, line 11: `RESOURCES`

pudo (changes requested):
Looks really nice, just some nitpicks.

datasets/dk/pep/crawler.py
bgmello: We don't unfortunately, the data in the excel workbooks look like this:

Screenshot 2024-05-08 at 15 38 35

We have the names and the birth date to identify the PEP

jbothma:
I've pushed some improvements to https://github.com/opensanctions/opensanctions/pull/836 - will merge shortly

Thanks for this!

Going forward, keep an eye on the following:

- language codes must be three-letter https://www.loc.gov/standards/iso639-2/php/code_list.php
- Try and break long sentences and blocks of text in the metadata into short simple sentences and small paragraphs
- Remember to have some certainty that if column order changes, the data won't become nonsense - e.g. if we're not doing key based lookups on keys generated from the headers, perhaps an assertion of the column headers is in order.

PR #823, opened 2024-05-03
adding czech republic national sanctions list

bgmello:
Fixes opensanctions/crawler-planning#158

### DIFF
--- a/datasets/cz/national_sanctions/crawler.py
+++ b/datasets/cz/national_sanctions/crawler.py
@@ -1,0 +1,159 @@
+from typing import Generator, cast
+from zavod import Context, helpers as h
+import openpyxl
+from openpyxl import load_workbook
+from typing import Generator
+from pantomime.types import XLSX
+from normality import stringify, slugify
+from datetime import datetime
+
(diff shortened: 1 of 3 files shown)


pudo (approved):
Looks good to me. We need some library tooling for the months thing soon! :)

datasets/cz/national_sanctions/crawler.py
```diff
@@ -0,0 +1,158 @@
...
+    )
+    if res:
+        countries = cast("List[str]", res.names)
+    else:
+        context.log.warning("Country not identified", text=res)
+        countries = None
+
+    if first_name is None:
+        entity = context.make("LegalEntity")
+        entity.id = context.make_slug(last_name)
```
pudo, line 83: We do need a more detailed ID here. Do we have any sort of extra key material we can use?

bgmello: For Legal entities, the only other information that is not related to the sanction is the country, we can use that. What do you think?

bgmello:
I was impressed that apparently the python `datetime` library doesn't have any way to deal with months in other languages.

PR #832, opened 2024-05-09
adding French interest representatives

bgmello:
Fixes opensanctions/crawler-planning#165.

### DIFF
--- a/datasets/_collections/incoming.yml
+++ b/datasets/_collections/incoming.yml
@@ -10,4 +10,5 @@ description: |
   before being moved to a collection under default.
 children:
   - us_finra_actions
+  - fr_interest_representative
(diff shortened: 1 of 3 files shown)


bgmello:
There is some interesting data that I didn't use because I couldn't find a way to add it to any model. For example, there is the `affiliations`, which are associations linked to the interests represented to which the organization belongs. The data dictionary can be found [here](https://www.hatvp.fr/wordpress/wp-content/uploads/2024/01/descriptif_JSON_AGORA.xlsx)

jbothma:
Closing since we're not yet able to extract the PEPs easily from free text
https://github.com/opensanctions/crawler-planning/issues/165#issuecomment-2136556863

PR #861, opened 2024-05-16
adding China National People's Congress

bgmello:
Fixes opensanctions/crawler-planning#170

It's currently throwing an assertion error because it's not creating an entity for each line, but I couldn't find why.


datasets/cn/national_congress/cn_national_congress.yaml
```diff
@@ -0,0 +1,44 @@
...
+  all of the senior officials in Chinese politics.
+publisher:
+  name: 全国人民代表大会
+  description: National People's Congress
+  url: http://www.npc.gov.cn
+  official: true
+  country: "cn"
+url: http://www.npc.gov.cn
+data:
+  format: CSV
```
jbothma, line 28: Could you paste the data into could you paste the data into https://docs.google.com/spreadsheets/d/1WfvKmydpsyIQe-5V28meRx2miVNCxMAtbtM-taJDFNw/edit#gid=0  then download context.data_url in the crawler. Then we can easily update the data.

```suggestion
data:
  url: https://docs.google.com/spreadsheets/d/e/2PACX-1vRWjzoZCQ4SSJGimuIP_ESCS23ZO2HWO7ENuwNr_4yjiXigeCezqtVz5rUOmUj3AfHeLyAalUiBmG1D/pub?output=csv
  format: CSV
```

datasets/cn/national_congress/cn_national_congress.yaml
```diff
@@ -0,0 +1,44 @@
...
+  are allowed to hold seats in other bodies of government
+  simultaneously and the party and the NPC typically includes
+  all of the senior officials in Chinese politics.
+publisher:
+  name: 全国人民代表大会
+  description: National People's Congress
+  url: http://www.npc.gov.cn
+  official: true
+  country: "cn"
+url: http://www.npc.gov.cn
```
jbothma, line 26: Our source is actually the wikipedia page, isn't it? 
so that's the wikipedia page title, url, unofficial, and country code zz

datasets/cn/national_congress/cn_national_congress.yaml
```diff
@@ -0,0 +1,44 @@
...
+  Current members of the Chinese National People's Congress
+description: |
+  The National People's Congress (NPC) is the highest organ of
+  state power of the People's Republic of China. It is the only 
+  branch of the Chinese government, and is unicameral. The NPC
+  is elected for a term of five years, and the last election
+  was held in 2023. Delegates to the National People's Congress
+  are allowed to hold seats in other bodies of government
+  simultaneously and the party and the NPC typically includes
+  all of the senior officials in Chinese politics.
```
jbothma, line 19: Let's add a paragraph with wikipedia's license for reuse of this data to the description

datasets/cn/national_congress/cn_national_congress.yaml
```diff
@@ -0,0 +1,44 @@
+title: China National People's Congress
+entry_point: crawler.py
+prefix: cn-npc
+coverage:
+  frequency: never 
+  start: 2024-05-16
+  end: 2028-03-01
```
jbothma, line 7: same as start date since that's when we last updated it.

datasets/cn/national_congress/crawler.py
```diff
@@ -0,0 +1,51 @@
...
+    name = input_dict.pop("姓名/Name")
+    ethnicity = input_dict.pop("民族/Ethnicity")
+    gender = input_dict.pop("性别/Gender")
+    delegation = input_dict.pop("代表团/Delegation")
+
+    entity = context.make("Person")
+    entity.id = context.make_slug(name, ethnicity, gender, delegation)
+
+    entity.add("name", name, lang="chi")
+    entity.add("gender", context.lookup("gender", gender))
```
jbothma, line 19: if you name the lookup `type.gender` then the context will use it automatically when adding the value

```suggestion
    entity.add("gender", gender)
```

datasets/cn/national_congress/crawler.py
```diff
@@ -0,0 +1,51 @@
...
+
+    occupancy = h.make_occupancy(
+        context,
+        entity,
+        position,
+        True,
+        categorisation=categorisation,
+    )
+
+    position.add("subnationalArea", delegation, lang="chi")
```
jbothma, line 35: actually subnationalArea is for when the jurisdiction of the position is limited to a subnational area. In this case their jurisdiction is national, they just represent a subnational region of their delegation, so let's skip this property.

datasets/cn/national_congress/crawler.py
```diff
@@ -0,0 +1,51 @@
...
+def crawl_item(input_dict: dict, context: Context):
+    name = input_dict.pop("姓名/Name")
+    ethnicity = input_dict.pop("民族/Ethnicity")
+    gender = input_dict.pop("性别/Gender")
+    delegation = input_dict.pop("代表团/Delegation")
+
+    entity = context.make("Person")
+    entity.id = context.make_slug(name, ethnicity, gender, delegation)
+
+    entity.add("name", name, lang="chi")
```
jbothma, line 18: I think you can also add the position column value like "Chairman of Inner Mongolia Mengcao Ecological Group" to the Person:position property. We don't use it normally any more but in this case I think it's useful a description for identifying the person.

jbothma (commented):
Let's call this something like cn_wikidata_npc so that if we get an official source one day, we can use this name or similar for it.

bgmello:
I'm actually using the data from [NPC observer](https://npcobserver.com/2023/02/china-14th-npc-demographics/), because it was already in a [google sheets ](https://npcobserver.com/2023/02/china-14th-npc-demographics/). Do you want me to extract from the wikipedia or is this ok?

datasets/cn/national_congress/crawler.py
bgmello: Ok, in this case would it be ok to write something like "Chairman of 北京" (part in English and part in Chinese)? Because in the data the name of the delegations is in Chinese.

jbothma:
Oh right - I pasted one table into my google sheet from wikipedia and it seems to work fine https://docs.google.com/spreadsheets/d/1WfvKmydpsyIQe-5V28meRx2miVNCxMAtbtM-taJDFNw/edit#gid=0

I think it's worth using the wikipedia data since it contains dates of birth which help a lot to disambiguate when two people have the same name.

datasets/cn/national_congress/crawler.py
jbothma: Sorry, I forgot I was looking at a google translate of the page.

Use the values as they are in the table.

bgmello:
I added the data from wikipedia to the google sheets. I used the following code to extract the data:

```
import requests
import pandas as pd

r = requests.get('https://zh.wikipedia.org/wiki/第十四届全国人民代表大会代表名单')
dfs = pd.read_html(r.text)
final_df = []
columns = ['姓名', '政党', '性别', '民族', '出生日期', '职务', '备注']
for i, df in enumerate(dfs[1:36]):
    df.columns = df.columns.get_level_values(0)
    df = df.iloc[:, :7]
    df.columns = columns
    final_df.append(df)
    
final_df = pd.concat(final_df)
final_df.to_excel("china_national_people_congress.xlsx", index=False)
```

datasets/cn/national_congress/cn_national_congress.yaml
bgmello: Do you mean this part? "All text on this site is available under the terms of [a Creative Commons Attribution-Share Alike 4.0 License](https://creativecommons.org/licenses/by-sa/4.0/deed.zh)"

datasets/cn/national_congress/cn_national_congress.yaml
bgmello: Do you have any example on how we download the data from the google sheets using python? I tried using a `requests.get` but it didn't work

jbothma: did you see the URL in the code suggestion and the comment are different? The URL in the comment is the sheet editor, while the url in the suggestion serves a CSV file.

What we usually do is "publish to web" from the file menu and choose the csv option. If we need multiple sheets, we create a URL for each worksheet or alter the querystring based on the selected sheet. We then fetch the csv using context.fetch_resource as usual https://github.com/opensanctions/opensanctions/blob/main/datasets/_wikidata/oligarchs/crawler.py#L33

datasets/cn/national_congress/cn_national_congress.yaml
jbothma: yeah - so to cite it I'd say something like  "Data sourced from Wikipedia licensed [Creative Commons Attribution-Share Alike 4.0](https://creativecommons.org/licenses/by-sa/4.0/deed.zh)"

bgmello:
I'm having trouble with the assertion because there are some cases we don't have enough information to create a unique ID. For example, there are two deputies with the name Zhang Qiang, that are male and ethnicity Han, since we don't have any other relevant information about them like birth date, we create the same ID. What we do in this case?

jbothma:
well spotted!

It looks like one represents Jiangsu Province and one represents Jiangxi Province.

Let's include the province they represent in the ID

PR #897, opened 2024-05-29
use zf context manager instead of tmpfile

SimonThordal:
Probably not what causes the OOM, but I cannot replicate locally, so working a bit in the dark.

### DIFF
--- a/datasets/dk/cvr/crawler.py
+++ b/datasets/dk/cvr/crawler.py
@@ -1,19 +1,14 @@
-import os
 import zipfile
-from tempfile import TemporaryDirectory
 
 from zavod import Context
 from zavod.shed.bods import parse_bods_fh
 
 
 def crawl(context: Context) -> None:
-    with TemporaryDirectory() as tmpdir:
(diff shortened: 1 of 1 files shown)


pudo:
Ha, I made the opposite change in order to get rid of the memory leak - so it's definitely not the zip handling, I guess. Really curious what the issue might be :/

pudo:
I've done a bit of a log read and am pretty convinced that it's not actually the crawler but the exporter that fails. This likely points to some anomaly with the data, but I'm going to first just up the memory to see how high the spike goes if left to itself.

SimonThordal:
> I've done a bit of a log read and am pretty convinced that it's not actually the crawler but the exporter that fails. This likely points to some anomaly with the data, but I'm going to first just up the memory to see how high the spike goes if left to itself.

That makes sense considering the crawl itself did not spike locally.

pudo:
Ok so the data is in fact in some ways buggy, and OO is also considering dropping it. Let's not waste more time here.

PR #911, opened 2024-06-03
An experimental approach to datasets revisions and separate databases

dchaplinsky:
I'm not sure I know what I am doing, therefore I'm presenting it to you.


pudo:
Unfortunately, I think this is a bit too duplicative with the versions/last_seen/first_seen system we already have. Thanks for schemeing it out, though!

PR #922, opened 2024-06-06
Adding US DOS terrorist exclusion list

bgmello:
Fixes opensanctions/crawler-planning#112.


datasets/us/dos_terrorist_exclusion/us_dos_terrorist_exclusion.yml
```diff
@@ -0,0 +1,206 @@
...
+  url: https://www.state.gov/terrorist-exclusion-list/
+  format: HTML
+  lang: eng
+ci_test: false
+
+lookups:
+  aliases:
+    lowercase: true
+    options:
+      - match: Afghan Support Committee (a.k.a. Ahya ul Turas; a.k.a. Jamiat Ayat-ur-Rhas al Islamia; a.k.a. Jamiat Ihya ul Turath al Islamia; a.k.a. Lajnat el Masa Eidatul Afghania)
```
jbothma, line 42: I don't think you need lookups for this.
I think `h.multi_split(name, ["; a.k.a.", "(a.k.a", "(f.k.a.", "; f.k.a"])` would split it very nice and cleanly.

datasets/us/dos_terrorist_exclusion/crawler.py
```diff
@@ -0,0 +1,54 @@
...
+
+    sanction = h.make_sanction(context, entity)
+
+    context.emit(entity, target=True)
+    context.emit(sanction)
+
+
+def crawl(context: Context):
+
+    response = requests.get(context.data_url, headers=HEADERS)
```
jbothma, line 46: always use `context.fetch_*`.
If that really doesn't handle some special thing we need, use `context.html.get` so that we use the http session with our standard session setup

datasets/us/dos_terrorist_exclusion/crawler.py
```diff
@@ -0,0 +1,54 @@
+from zavod import Context, helpers as h
+import requests
+from typing import cast
+from lxml import html
+
+HEADERS = {
+    "User-Agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.4.1 Safari/605.1.15",
+}
```
jbothma, line 8: if this needs to be overwritten, do so using the metadata `http` config and see if you can make it work with `(zavod, opensanctions.org)` at the end so they can reach us if we're hurting them

datasets/us/dos_terrorist_exclusion/crawler.py
```diff
@@ -0,0 +1,54 @@
...
+
+    entity.id = context.make_id(name)
+    entity.add("name", name)
+
+    for alias in aliases:
+        entity.add("alias", alias)
+
+    entity.add("topics", "crime.terror")
+
+    sanction = h.make_sanction(context, entity)
```
jbothma, line 38: no need for a sanction entity if there isn't additional data about the ban like start/end date.

In this case the `program` could be `Section 411 of the USA PATRIOT ACT of 2001 (8 U.S.C. § 1182) Terrorist Exclusion List (TEL)` but that can go on the legal entity rather than the sanction entity since there's still no additional data for the sanction, making it overkill.

datasets/us/dos_terrorist_exclusion/us_dos_terrorist_exclusion.yml
```diff
@@ -0,0 +1,206 @@
...
+        names:
+          - GRAPO
+          - Grupo de Resistencia Anti-Fascista Premero De Octubre
+assertions:
+  min:
+    schema_entities:
+      LegalEntity: 56
+  max:
+    schema_entities:
+      LegalEntity: 56
```
jbothma, line 205: make these a little bit looser so we don't have to change this immediately when they add/remove someone.

> A good rule of thumb for datasets that change over time is minima 10% below the expected number to allow normal variation, unless there's a known hard minimum, and a maximum around 20% above the expectation to leave room to grow, if the number is expected to fluctuate.

https://zavod.opensanctions.org/metadata/#data-assertions

datasets/us/dos_terrorist_exclusion/crawler.py
bgmello: I was able to run with `fetch_html`, but if I remove this header it doesn't work. Do you mean adding `(zavod, opensanctions.org)` to the user agent? Like this?

```
Mozilla/5.0 (Macintosh; Intel ... (zavod, opensanctions.org)
```

jbothma: I actually meant like https://github.com/opensanctions/opensanctions/blob/main/datasets/_global/interpol/interpol_red_notices.yml#L42 but I see their website now has some kind of bot blocking that needs more than user agent to bypass, so I've switched to using zyte https://github.com/opensanctions/opensanctions/pull/932/commits/455eec80a1e68da6517b09765989a457892917c8

PR #925, opened 2024-06-08
Use GPT-4o to extract data from UN GA PDF

pudo:


datasets/jp/meti_eul/jp_meti_eul.yml
```diff
@@ -0,0 +1,33 @@
...
+    relationships, and to secure stable and efficient supply of energy and mineral
+    resources.
+  country: jp
+  url: https://www.meti.go.jp/
+  official: true
+url: https://www.meti.go.jp/english/policy/external_economy/trade_control/index.html#Endu
+data:
+  url: https://www.meti.go.jp/english/policy/external_economy/trade_control/index.html#Endu
+  format: PDF
+  lang: jpn
```
jbothma, line 27: Totally see the need for a zyte `fetch_resource`. In this case, a user agent string seems to satisfy akamai

```suggestion
http:
  user_agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/122.0.0.0 Safari/537.36 (zavod; opensanctions.org)
```

PR #936, opened 2024-06-11
Adding Iceland parliament crawler

bgmello:
Fixes opensanctions/crawler-planning#173

### DIFF
--- a/datasets/is/althingi/crawler.py
+++ b/datasets/is/althingi/crawler.py
@@ -45,7 +45,7 @@ def crawl_item(member_url: str, name: str, context: Context):
     for link in social_media_links:
         person.add("sourceUrl", link)
 
-    position = h.make_position(context, "Member of the Alþingi")
+    position = h.make_position(context, "Member of the Althing")
     categorisation = categorise(context, position, is_pep=True)
 
     occupancy = h.make_occupancy(
--- a/datasets/is/althingi/crawler.py
+++ b/datasets/is/althingi/crawler.py
@@ -46,6 +46,7 @@ def crawl_item(member_url: str, name: str, context: Context):
         person.add("sourceUrl", link)
 
     position = h.make_position(context, "Member of the Althing")
+    position.add("country", "is")
     categorisation = categorise(context, position, is_pep=True)
 
     occupancy = h.make_occupancy(
--- a/datasets/is/althingi/crawler.py
+++ b/datasets/is/althingi/crawler.py
@@ -27,13 +27,13 @@ def crawl_item(member_url: str, name: str, context: Context):
         email = None
 
     person = context.make("Person")
-    person.id = context.make_id(name)
+    person.id = context.make_id(name, birth_date)
 
     person.add("name", name)
     # The birth date has a suffix that we need to remove
     person.add(
         "birthDate",
-        h.parse_date(re.sub(SUFFIX_PATTERN, "", birth_date), formats=["%B %d, %Y"]),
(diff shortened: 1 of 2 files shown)


datasets/is/althingi/crawler.py
```diff
@@ -0,0 +1,74 @@
...
+        h.parse_date(re.sub(SUFFIX_PATTERN, "", birth_date), formats=["%B %d, %Y"]),
+    )
+    person.add("phone", telephone)
+    person.add("political", party)
+    person.add("email", email)
+
+    for link in social_media_links:
+        person.add("sourceUrl", link)
+
+    position = h.make_position(context, "Member of the Alþingi")
```
jbothma, line 48: Let's use [Member of the Althing](https://www.wikidata.org/wiki/Q21272959) as per 
https://www.wikidata.org/wiki/Wikidata:WikiProject_every_politician/Iceland

```suggestion
    position = h.make_position(context, "Member of the Althing")
```

jbothma, line 48: remember to add the country to the position

datasets/is/althingi/crawler.py
```diff
@@ -0,0 +1,74 @@
...
+    person.add(
+        "birthDate",
+        h.parse_date(re.sub(SUFFIX_PATTERN, "", birth_date), formats=["%B %d, %Y"]),
+    )
+    person.add("phone", telephone)
+    person.add("political", party)
+    person.add("email", email)
+
+    for link in social_media_links:
+        person.add("sourceUrl", link)
```
jbothma, line 46: Let's rather add the detail URL for the person as sourceUrl and skip the social media links. They're also not really sources.

datasets/is/althingi/crawler.py
```diff
@@ -0,0 +1,74 @@
...
+    ).text_content()
+    match = re.search(EMAIL_PATTERN, email_script)
+    if match:
+        local_part = match.group(1)
+        email = local_part + "@althingi.is"
+    else:
+        email = None
+
+    person = context.make("Person")
+    person.id = context.make_id(name)
```
jbothma, line 33: let's include their date of birth in the ID. Is that fairly commonly available?

datasets/is/althingi/crawler.py
```diff
@@ -0,0 +1,74 @@
...
+
+    person.add("name", name)
+    # The birth date has a suffix that we need to remove
+    person.add(
+        "birthDate",
+        h.parse_date(re.sub(SUFFIX_PATTERN, "", birth_date), formats=["%B %d, %Y"]),
+    )
```
jbothma, line 40: it looks like date isn't parsing correctly. Looking at the exported data, it's falling back to only the year.

jbothma (commented):
Nice and clean, thank you! Just a few small fixes before we merge.

datasets/is/althingi/crawler.py
bgmello: I added that way because it's how is written in the opensanctions [website](https://www.opensanctions.org/entities/evpo-3fe9afd515cd95a0ec2886365467a158563bcde5/)

jbothma: this can be done via make_position too :)

datasets/is/althingi/is_althingi.yaml
```diff
@@ -0,0 +1,35 @@
+title: Iceland parliament
+entry_point: crawler.py
+prefix: is-par
+coverage:
+  frequency: daily
+  start: 2024-06-11
```
jbothma, line 6:
```suggestion
  start: 2024-06-20
```

PR #963, opened 2024-06-19
Include entities with same ID not returned by index

jbothma:
e.g. NK-IDs for positive matches, QIDs, common ID schemes, etc.

needs test


jbothma:
The actual issue is that they shouldn't have the same ID. The target store should be loaded with a linker aware of the store, which says it should not be resolved.

PR #969, opened 2024-06-19
Export dataset.resolve field 

jbothma:


zavod/zavod/tests/exporters/test_exporters.py
```diff
@@ -182,6 +182,7 @@ def test_ftm_referents(testdataset1: Dataset):
     identifier = resolver.decide(
         "osv-john-doe", "osv-johnny-does", Judgement.POSITIVE, user="test"
     )
+    testdataset1.resolve = True
```
jbothma, line 185: not sure if this is just too yuck

pudo (changes requested):
Can we make sure that the default for absent `resolve` is true? We also probably don't want to include it in the JSON unless it's false.

jbothma:
Thanks, fixed.

zavod/zavod/tests/fixtures/testdataset1/testdataset1.yml
```diff
@@ -2,6 +2,7 @@ entry_point: testentrypoint1
 title: OpenSanctions Validation Dataset
 prefix: osv
 hidden: true
+resolve: false
```
pudo, line 5: This hits me as super surprising - that our main fixture is going to have one of the default features disabled. Does it not break any tests?

jbothma: yeah it definitely has a smell, and not a nice one.

Only two failed!

I need to go, but I'll fix when I'm back. This is not an urgent change

Issue #31, opened 2019-02-26
error building docker image 


aplekhov:
Hi, I'm having trouble building the docker image.  It's failing with the following PIP error when collecting pandas:
```
ModuleNotFoundError: No module named 'numpy'
```

Also, are the csvs files on opensanctions.org no longer being updated?  What's the best current way to to get the data?

Thanks.

alexmojaki:
I got some guidance and an update on https://alephdata.slack.com:

> At the moment, OpenSanctions is in embarrassing state of disarray, and we just haven’t had the time to fix it up a little, which shouldn’t be super hard.
> 
> We used it for a while to prototype some tools, that’s why it’s in a somewhat weird state. These tools are now much more mature, we just need to backport them into OpenSanctions.
> 
> The original idea was for OS to emit the types of entities we deal with in the rest of the Aleph toolchain
> 
> But when we converted the crawlers to that, we didn’t have a backend to store these entities
> 
> There’s now `balkhash`, doing exactly that.
>
> The project is definitely not forgotten, just been super busy on other ends

With help I made some contributions which are somewhat allowing me to create the CSVs on my own. See the updated README.

pudo:
Fixed.

PR #970, opened 2024-06-19
Add Germany Bundeskriminalam wanted crawler

atehe:
closes [Germany Bundeskriminalamt #31](https://github.com/opensanctions/crawler-planning/issues/31)

### DIFF
--- a/datasets/de/bka_wanted/crawler.py
+++ b/datasets/de/bka_wanted/crawler.py
@@ -1,0 +1,124 @@
+from zavod import Context
+from zavod import helpers as h
+from xml.etree import ElementTree
+from normality import collapse_spaces
+
+
+def get_element_text(
+    doc: ElementTree, xpath_value: str, to_remove=[], join_str=""
+) -> str:
+    """Extract text from from an xpath
+
+    Args:
+        doc (ElementTree): HTML Tree
+        xpath_value (str):  xpath to extract text from
+        to_remove (list, optional): string to remove in the extracted text.
+        join_str (str): String to and join the extracted text
+    """
+    element_tags = doc.xpath(xpath_value)
+
+    tag_list = []
+    for tag in element_tags:
+        try:
+            tag_list.append(tag.text_content())
+        except AttributeError:  #  node is already a text content
+            tag_list.append(tag)
+
+    element_text = join_str.join(tag_list)
+
+    for string in to_remove:
+        element_text = element_text.replace(string, "")
+
+    return collapse_spaces(element_text.strip())
+
+
+def crawl(context: Context):
+    base_url = context.data_url
+    additional_url = f"{base_url}?additionalHitsOnly=true"
+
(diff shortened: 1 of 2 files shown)


datasets/de/bka_wanted/de_bka_wanted.yml
```diff
@@ -0,0 +1,73 @@
+title: "Germany BKA Wanted Fugitives"
+entry_point: crawler.py
+prefix: us-bkwant
```
pudo, line 3: Let's make this de-bka

datasets/de/bka_wanted/crawler.py
```diff
@@ -0,0 +1,124 @@
...
+    alias = alias.split(",")
+
+    date_of_birth = get_element_text(
+        doc,
+        '//li[@class="c-listitem-detailinfos"][.//span[contains(.//text(),"Geburtsdatum")]]//span[contains(@class,"c-value-detailinfos")]',
+    )
+    date_of_birth = h.parse_date(date_of_birth, ["%d.%m.%Y"])
+    place_of_birth = get_element_text(
+        doc,
+        '//li[@class="c-listitem-detailinfos"][.//span[contains(.//text(),"Geburtsort")]]//span[contains(@class,"c-value-detailinfos")]',
```
pudo, line 79: perhaps we want a helper method to make these paths so that’s  this becomes more readable?

datasets/de/bka_wanted/crawler.py
```diff
@@ -0,0 +1,124 @@
+from zavod import Context
+from zavod import helpers as h
+from xml.etree import ElementTree
+from normality import collapse_spaces
+
+
+def get_element_text(
```
pudo, line 7: is this different from el.text_content() ?

pudo (commented):
Nice work, just the prefix seems like a showstopper

datasets/de/bka_wanted/crawler.py
atehe: Not very, 
It helps to get all the text content from xpaths that has multiple node match.
Example for an xpath `//p` , it extracts the text contents from all `<p>` and joins them together

atehe: Not very, 
It helps to get all the text content from xpaths that has multiple node match.
Example for an xpath `//p` , it extracts the text contents from all `<p>` and joins them together

atehe: Not very, 
It helps to get all the text content from xpaths that has multiple node match.
Example for an xpath `//p` , it extracts the text contents from all `<p>` and joins them together

datasets/de/bka_wanted/crawler.py
atehe: Do you mean something like a function that generates the xpath based on the text it contains?

PR #987, opened 2024-06-25
Adding Estonia Sanctions

bgmello:
Fixes opensanctions/crawler-planning#202


datasets/ee/international_sanctions/crawler.py
```diff
@@ -0,0 +1,70 @@
...
+    last_name, first_name = name.split(", ")
+
+    entity = context.make("Person")
+    entity.id = context.make_id(name)
+    h.apply_name(entity, first_name=first_name, last_name=last_name)
+    entity.add("topics", "sanction")
+
+    for alias in aliases:
+        entity.add("alias", alias)
+
```
pudo, line 49: Can we make a sanctions object in each of these that emits the title of the list in some form, and the source URL? Perhaps we should also include a `sourceUrl` with the entities themselves.

datasets/ee/international_sanctions/ee_international_sanctions.yaml
```diff
@@ -0,0 +1,35 @@
...
+description: |
+  This dataset contains information about individuals and entities
+  sanctioned by the government of Estonia in accordance with section
+  27 of the International Sanctions Act, the sanctions are imposed
+  by a regulation of the Government of the Republic on the proposal
+  of the Ministry of Foreign Affairs. The subjects of the sanctions
+  are designated by a directive of the minister in charge of the
+  policy sector.
+publisher:
+  name: Välisministeerium
```
pudo, line 19: We just added an `acronym` field here, which could be `VM` for this.

pudo (commented):
This is a lovely short one! I'd still add sanctions objects to distinguish the two lists, but other than that it's a go!

PR #988, opened 2024-06-26
adding philippines SEC advisories

bgmello:
The formatting is inconsistent on each year, so it's not very easy to identify what is the name of the entity:

Screenshot 2024-06-26 at 10 21 59
Screenshot 2024-06-26 at 10 22 15

I tried identifying if there was a bold tag and consider that the name and if there is not, then consider the entire text as the name, but I think there are cases where this might fail.


bgmello:
fixes opensanctions/crawler-planning#182

datasets/ph/sec-advisories/ph_sec_advisories.yaml
```diff
@@ -0,0 +1,52 @@
...
+summary: >
+  Individuals and entities with advisories issued by the Philippines Securities and Exchange Commission
+description: |
+  This dataset comprises information about advisories issued by the
+  Philippines Securities and Exchange Commission (SEC) against individuals
+  and entities. The advisories are issued to warn the public about individuals
+  and entities that are not authorized to solicit investments from the public.
+
+publisher:
+  name: Philippines Securities and Exchange Commission
```
pudo, line 17: one acronym :)

datasets/ph/sec-advisories/ph_sec_advisories.yaml
```diff
@@ -0,0 +1,52 @@
...
+    schema_entities:
+      LegalEntity: 500
+  max:
+    schema_entities:
+      LegalEntity: 600
+
+lookups:
+  names:
+    lowercase: true
+    options:
```
pudo, line 43: I actually like this quite a bit!

pudo (commented):
Looks pretty much done, just two nitpicks!

datasets/ph/sec-advisories/ph_sec_advisories.yaml
bgmello: What should I do with the cases we can't identify the name automatically? Do you want me to create a lookup for each one? 

The problem is there seems to be no pattern so I guess we can only do it manually.

For example, on the year 2017 there is some bullet points which are just the name of the company, others that don't have a name and others that have a name plus a explanation (XTRADE.COM is NOT Registered with SEC)
Screenshot 2024-07-04 at 08 47 58

pudo:
@bgmello I've played around a good bit with it in https://github.com/opensanctions/opensanctions/pull/1001 but ended up just brute forcing it with lookups. Do you have another 30 mins to finish the lookup table after I've merged this one?

bgmello:
Sure!

bgmello:
Done

pudo:
Where? :)

bgmello:
I created the PR https://github.com/opensanctions/opensanctions/pull/1002 but it's giving merge conflicts.

PR #991, opened 2024-06-27
adding India NSE debarred entities

bgmello:
Fixes opensanctions/crawler-planning#176


datasets/in/nse_debarred/crawler.py
```diff
@@ -0,0 +1,97 @@
...
+    pan = input_dict.pop("pan")
+
+    if name is None:
+        return
+
+    entity = context.make("LegalEntity")
+    entity.id = context.make_id(name, pan)
+    entity.add("name", name)
+    entity.add("taxNumber", pan)
+    entity.add("topics", "sanction")
```
jbothma, line 57:
```suggestion
    entity.add("topics", "reg.warn")
```

datasets/in/nse_debarred/crawler.py
```diff
@@ -0,0 +1,97 @@
...
+            if isinstance(value, datetime):
+                value = value.date()
+            record[header] = stringify(value)
+        if len(record) == 0:
+            continue
+        yield record
+
+
+def crawl_item(input_dict: dict, context: Context):
+    name = input_dict.pop("entity_individual_name")
```
jbothma, line 47: let's split on:
`Proprietor of`, `Proprietor`
watch out, in the first case the right side is the company. In the second case the left side is the company. It looks like the PAN and the ban relate to the person, not the company.

So I think we want the company emitted as well, but you don't have to create a sanction for it. Just an ownership relation between the person and the company.

datasets/in/nse_debarred/crawler.py
```diff
@@ -0,0 +1,97 @@
...
+    sanction = h.make_sanction(context, entity, key=input_dict.pop("nse_circular_no"))
+
+    sanction.add(
+        "date", h.parse_date(input_dict.pop("order_date"), formats=["%d-%b-%y"])
+    )
+    if input_dict.get("order_particulars"):
+        sanction.add(
+            "description", "Order Particulars: " + input_dict.pop("order_particulars")
+        )
+    sanction.add("duration", input_dict.pop("period"))
```
jbothma, line 69: it looks like if it contains the word Revoked, make topic `reg.action` and don't emit as target.

datasets/in/nse_debarred/crawler.py
```diff
@@ -0,0 +1,97 @@
...
+        return
+
+    entity = context.make("LegalEntity")
+    entity.id = context.make_id(name, pan)
+    entity.add("name", name)
+    entity.add("taxNumber", pan)
+    entity.add("topics", "sanction")
+    entity.add("idNumber", input_dict.pop("din_cin_of_entities_debarred"))
+
+    sanction = h.make_sanction(context, entity, key=input_dict.pop("nse_circular_no"))
```
jbothma, line 60: if it doesn't take more than 10 mins, it'd be nice to extract the circular URL and add it to sourceUrl

jbothma (changes requested):
nice, thanks! some fixes needed but shouldn't take too long

datasets/in/nse_debarred/crawler.py
```diff
@@ -0,0 +1,97 @@
...
+
+    if name is None:
+        return
+
+    entity = context.make("LegalEntity")
+    entity.id = context.make_id(name, pan)
+    entity.add("name", name)
+    entity.add("taxNumber", pan)
+    entity.add("topics", "sanction")
+    entity.add("idNumber", input_dict.pop("din_cin_of_entities_debarred"))
```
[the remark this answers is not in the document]
bgmello: Not all of them are blank:
Screenshot 2024-07-25 at 10 08 26

Should I add it to any property if it's not blank or just ignore this column?

datasets/in/nse_debarred/crawler.py
bgmello: From what I searched, I would need to use openpyxl to get the hyperlink. But when I try to use it I get the following error:

Screenshot 2024-08-01 at 11 53 30

datasets/in/nse_debarred/crawler.py
bgmello: Should I just do a simple split on those strings or use some kind of regex? Because there appears to be some variations:

- Mr. Naresh Pal (Proprietor of Equity Mantra)
- Mr. Shashikant Kumar (Owner of SM Global Research)
- M/s G N Traders (Proprietor – Mr. Kashif Siddique)
- M/s Ideal Equity (Proprietor: Anilkumar Talpada)
- GTR Solutions Financial Service(Prop.Nishant Chaturvedi)

datasets/in/nse_debarred/crawler.py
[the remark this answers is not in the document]
jbothma: Good catch, thanks.
Please put DIN in description with the prefix "Director Identification Number "
Please strip CIN from the CIN, split on space, and add all to registrationNumber

PR #996, opened 2024-07-02
Simplify the use of the linker

pudo:


zavod/zavod/dedupe.py
```diff
@@ -22,33 +21,26 @@
...
-        return Resolver()
-    return get_resolver()
+@cache
+def get_resolver() -> Resolver[Entity]:
+    """Load the deduplication resolver."""
+    return _get_resolver()
 
 
+@cache
 def get_dataset_linker(dataset: Dataset) -> Linker[Entity]:
```
SimonThordal, line 39: Would it make sense to hoist the `dataset.resolve` check out of this function? As far as I can tell we're loading the same resolver for all datasets, so there's no reason for breaking the cache for different datasets

SimonThordal (approved):

zavod/zavod/dedupe.py
pudo: Maybe it's actually safer to get rid of the cache entirely

SimonThordal: Yeah, if we know to only create the linker once.

Issue #430, opened 2023-10-09
Did Liechtenstein PDF out on us? 


pudo:
@brawer do I see this right that Liechtenstein removed the HTML version of the Sperren and replaced them with a PDF? We don't really have a solid PDF processing strategy in Zavod right now, so this leaves me at a loss a bit. 

https://www.llv.li/de/unternehmen/berichtspflichten-bewilligungen/grenzueberschreitende-dienstleistungen/entsendung-von-arbeitnehmern/sanktionsliste-gemaess-art.-7-abs.-4-entsg

pudo:
Replaced with a static extract from before the switch. Crawler is now inactive :(

brawer:
See [#1012 ](https://github.com/opensanctions/opensanctions/pull/1012) for an attempt at PDF processing.

PR #1012, opened 2024-07-08
Crawl PDF lists with Liechtenstein posting sanctions

brawer:
Although the PDF layout analysis is not perfect, its quality appears to be decent. For the debarments list, we can extract 100% of sanctions. However, for the infractions list, we fail extracting 7 sanctions because their names are so long that two table columns get merged together.

https://github.com/opensanctions/opensanctions/issues/430

### DIFF
--- a/datasets/li/posting_sanctions/crawler.py
+++ b/datasets/li/posting_sanctions/crawler.py
@@ -1,25 +1,28 @@
 import re
-import lxml
-import shutil
 from datetime import datetime
+from io import IOBase
 from typing import Optional
+
+from pdfminer.high_level import extract_pages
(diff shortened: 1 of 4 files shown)


pudo:
Hi @brawer - I'm so sorry we didn't merge this PR for such a long time. We wanted to come up with a common solution to parsing PDF tables, and have now settled on `pdfplumber`. I've pushed a version of this PR that uses that lib to https://github.com/opensanctions/opensanctions/pull/1423 ..

Would you be OK if I re-merged the two datasets into one in the PR?

brawer:
Sure, no problem! Many thanks.

pudo:
Close in favor of #1423.

Issue #32, opened 2019-03-19
CSV Sanctions Lists from GB down


guiserguez:
Hello,

I was trying to download the United Kingdom Consolidated List of Targets in the website but it is down.

Could you please upload it back?

pudo:
Should be working atm. Closing

PR #1014, opened 2024-07-08
Add us_ice crawler

atehe:
Closes [US Immigration and Customs Enforcement #32](https://github.com/opensanctions/crawler-planning/issues/32)

### DIFF
--- a/datasets/us/ice_wanted/crawler.py
+++ b/datasets/us/ice_wanted/crawler.py
@@ -1,0 +1,155 @@
+from zavod import Context
+from xml.etree import ElementTree
+from normality import collapse_spaces
+
+
+def get_element_text(doc: ElementTree, xpath_value: str, to_remove=[]) -> str:
+    """Extract text from each child nodes of an xpath and joins them together
+
+    Args:
+        doc (ElementTree): HTML Tree
+        xpath_value (str):  xpath to extract text from
+        to_remove (list, optional): string to remove in the extracted text.
+    """
+    element_tags = doc.xpath(xpath_value)
+
+    tag_list = []
+    for tag in element_tags:
+        tag_list.append(tag.text_content())
+
+    element_text = "".join(tag_list)
+
+    for string in to_remove:
+        element_text = element_text.replace(string, "")
+
+    return collapse_spaces(element_text.strip())
+
+
+def generate_mw_notes(**kwargs):
+    note_string = f"""Additional Information:
+- Wanted For: {kwargs.get('crime_summary')}
+- Wanted Heading: {kwargs.get('crime_details')}
+- Wanted Category: {kwargs.get("category")}
+- Wanted Status: {kwargs.get("status")}
+- Age: {kwargs.get("age")}
+- Height: {kwargs.get("height")}
+- Weight: {kwargs.get("weight")}
+- Skin tone: {kwargs.get("skin_tone")}
+- Eyes: {kwargs.get("eyes")}
+- Hair: {kwargs.get("hair")}
+- Last Known Location: {kwargs.get('last_known_location')}"""
+
+    return note_string
+
+
+def crawl_person(context: Context, url: str, crime_summary: str):
+    doc = context.fetch_html(url, cache_days=1)
+
+    names = get_element_text(
+        doc,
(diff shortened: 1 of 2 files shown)


datasets/us/ice_wanted/crawler.py
```diff
@@ -0,0 +1,153 @@
...
+        xpath_value (str):  xpath to extract text from
+        to_remove (list, optional): string to remove in the extracted text.
+    """
+    element_tags = doc.xpath(xpath_value)
+
+    tag_list = []
+    for tag in element_tags:
+        tag_list.append(tag.text_content())
+
+    element_text = "".join(tag_list)
```
jbothma, line 20: let's join on `" "` so we don't inadvertently get something like `end of something.New something` - or is that prevented in some other way?

datasets/us/ice_wanted/crawler.py
```diff
@@ -0,0 +1,153 @@
...
+    for string in to_remove:
+        element_text = element_text.replace(string, "")
+
+    return collapse_spaces(element_text.strip())
+
+
+def generate_mw_notes(**kwargs):
+    note_string = f"""Additional Information:
+- Wanted For: {kwargs.get('crime_summary')}
+- Wanted Heading: {kwargs.get('crime_details')}
```
jbothma, line 31: we don't really define or render this as markdown and we don't render newlines on the website. What could work nicely is if this was a list of strings without the dash instead. Then each attribute would be separated and easy to render in reasonable ways for everyone.

datasets/us/ice_wanted/crawler.py
```diff
@@ -0,0 +1,153 @@
...
+    doc = context.fetch_html(url, cache_days=1)
+
+    names = get_element_text(
+        doc,
+        '//div[contains(@class, "field--name-field-most-wanted-name")]//div[contains(@class, "field__item")]',
+    )
+    names_split = names.split(",")
+
+    name = names_split[0]
+    alias = names_split[1:]
```
jbothma, line 55: as far as I can tell, the name is actually `surname(s), forename(s)` rather tan `name, aliases`. Or do you have different examples?

There's an alias field which I think we're missing right now.

datasets/us/ice_wanted/us_ice_wanted.yml
```diff
@@ -0,0 +1,45 @@
+title: "U.S. Immigration and Customs Enforcement Most Wanted Fugitives"
+entry_point: crawler.py
+prefix: us-ice-want
```
jbothma, line 3:
```suggestion
prefix: us-icew
```

datasets/us/ice_wanted/crawler.py
```diff
@@ -0,0 +1,153 @@
...
+- Wanted Heading: {kwargs.get('crime_details')}
+- Wanted Category: {kwargs.get("category")}
+- Wanted Status: {kwargs.get("status")}
+- Age: {kwargs.get("age")}
+- Height: {kwargs.get("height")}
+- Weight: {kwargs.get("weight")}
+- Skin tone: {kwargs.get("skin_tone")}
+- Eyes: {kwargs.get("eyes")}
+- Hair: {kwargs.get("hair")}
+- Last Known Location: {kwargs.get('last_known_location')}"""
```
jbothma, line 40: i think you can stuff this into `Person:address`

datasets/us/ice_wanted/crawler.py
jbothma: and the descriptive attributes in the description property, while the category, status, crime etc can go in the notes property

jbothma, line 31:
```suggestion
- Crime details: {kwargs.get('crime_details')}
```

jbothma (commented):
very nice, thanks! Just some tweaks then I think it's ready

datasets/us/ice_wanted/crawler.py
atehe: You're right.
Some of them had two names before the comma, eg https://www.ice.gov/most-wanted/garcia-jerez-anibal,  so I though it was an alias

datasets/us/ice_wanted/crawler.py
atehe: Got it

PR #1017, opened 2024-07-10
Latvia's Magnitsky Law Sanctions List

nvmbrasserie:
The first working version of the Latvia's Magnitsky Law Sanctions List.


datasets/lv/magnitsky_list/crawler.py
```diff
@@ -0,0 +1,66 @@
...
+# Define constants for date parsing
+DATE_FORMATS = ["%Y-%m-%d", "%Y"]
+
+# Normalized country names
+NORMALIZED_COUNTRIES = {"Krievijas Federācija": "Russia"}
+
+
+def normalize_country(country: str) -> str:
+    """Normalize the country name to a standard format"""
+    return NORMALIZED_COUNTRIES.get(country, country)
```
pudo, line 15: Ha! Finally I get to show you another mechanism :) 

We have this mechanism called data patches that's used for various data types - countries are often the most prominent. The way it works is that you can define overrides/cleanups in the YAML of the crawler metadata that are then applied inside `entity.add()` if you're trying to set a value of a specific type. Have a look at: 

https://github.com/opensanctions/opensanctions/blob/main/datasets/_externals/wikidata.yml#L86-L654

This example is a bit bonkers, but shows the principle. Wanna try?

datasets/lv/magnitsky_list/crawler.py
```diff
@@ -0,0 +1,66 @@
...
+    entity.id = context.make_slug(original_name, birth_date)
+
+    # Add names
+    entity.add("name", original_name, lang="lav")
+    entity.add("name", name_in_brackets, lang="eng")
+    # entity.add("alias", name_in_brackets, lang="eng")
+
+    # Add birth date
+    if birth_date:
+        entity.add("birthDate", birth_date)
```
pudo, line 42: Doing `entity.add(prop, None)` is safe (just has no effect) - so you don't need these checks!

pudo (changes requested):

pudo:
Very nice work! Thank you.

PR #1021, opened 2024-07-10
feat: add a bit of timing on exporters

SimonThordal:


zavod/zavod/exporters/__init__.py
```diff
@@ -72,8 +75,13 @@ def export_data(context: Context, view: View) -> None:
         if idx > 0 and idx % 10000 == 0:
             log.info("Exported %s entities..." % idx, dataset=context.dataset.name)
         for exporter in exporters:
+            t1 = time()
             exporter.feed(entity)
+            t2 = time()
+            timings[type(exporter).__name__] += t2 - t1
 
+    for exporter, seconds in timings.items():
+        log.info("Exporter %s took %.2f seconds" % (exporter, seconds))
```
jbothma, line 84: we should check if the resolution of this is sufficient - the slices will potentially be quite small per entity

jbothma, line 84: structured data would be nice for use in a metric

PR #1029, opened 2024-07-12
Ivan/FinCEN-311-Special-Measures-#204

nvmbrasserie:
firt commit


datasets/us/fincen_special_measures/crawler.py
```diff
@@ -0,0 +1,147 @@
...
+            pdf_link = pdf_link_element.get("href")
+            if pdf_link.endswith(".pdf"):
+                if i == 0:
+                    pdf_link_finding = pdf_link
+                elif i == 1:
+                    pdf_link_nprm = pdf_link
+                elif i == 2:
+                    pdf_link_final_rule = pdf_link
+                elif i == 3:
+                    pdf_link_rescinded = pdf_link
```
jbothma, line 127: I think it's better to keep the table parsing simpler, and do cell.findall("a") in crawl_item instead. Then you don't have to worry about offsets (and sometimes there are more than one link in a cell), and you know exactly which column the link related to.

```python
def parse_table(table: HtmlElement) -> Generator[Dict[str, str], None, None]:
    headers = None
    for row in table.findall(".//tr"):
        if headers is None:
            headers = []
            for el in row.findall("./th"):
                # Workaround because lxml-stubs doesn't yet support HtmlElement
                # https://github.com/lxml/lxml-stubs/pull/71
                eltree = cast(HtmlElement, el)
                headers.append(slugify(eltree.text_content()))
            continue

        cells = row.findall("./td")  # <--- edited here
        assert len(headers) == len(cells), (headers, cells)
        yield {hdr: c for hdr, c in zip(headers, cells)}
```

and then `finding_date = row.get("finding").text_content()` and `pdf_link_finding = row.get("finding").find("./a/@href")` or something

The risk with assuming column order is that they might add/change columns and if you don't have an assertion on the column headings being in the order you assume, the code might continue emitting data but it becomes invalid.

datasets/us/fincen_special_measures/crawler.py
```diff
@@ -0,0 +1,147 @@
...
+
+    # Extract PDF links
+    pdf_link_finding = row.get("pdf_link_finding", None)
+    pdf_link_nprm = row.get("pdf_link_nprm", None)
+    pdf_link_final_rule = row.get("pdf_link_final_rule", None)
+    pdf_link_rescinded = row.get("pdf_link_rescinded", None)
+
+    # Convert relative URLs to full URLs
+    if pdf_link_finding and pdf_link_finding.startswith("/"):
+        pdf_link_finding = BASE_URL + pdf_link_finding
```
jbothma, line 45: an easier way to do this is `doc.make_links_absolute(...)`
https://github.com/opensanctions/opensanctions/blob/main/datasets/at/meine_abgeordneten/crawler.py#L185

datasets/us/fincen_special_measures/crawler.py
```diff
@@ -0,0 +1,147 @@
...
+    if pdf_link_nprm and pdf_link_nprm.startswith("/"):
+        pdf_link_nprm = BASE_URL + pdf_link_nprm
+    if pdf_link_final_rule and pdf_link_final_rule.startswith("/"):
+        pdf_link_final_rule = BASE_URL + pdf_link_final_rule
+    if pdf_link_rescinded and pdf_link_rescinded.startswith("/"):
+        pdf_link_rescinded = BASE_URL + pdf_link_rescinded
+
+    # Create and add details to the sanction
+    sanction = h.make_sanction(context, entity)
+    sanction.add("country", "us")
```
jbothma, line 55: sanction country is automatically added by the helper, so shouldn't be needed.

datasets/us/fincen_special_measures/crawler.py
```diff
@@ -0,0 +1,147 @@
...
+    if final_rule_date != "---":
+        sanction.add("startDate", convert_date(final_rule_date))
+
+    rescinded_date = row.get("rescinded", "")
+    if rescinded_date != "---":
+        sanction.add("endDate", convert_date(rescinded_date))
+
+    # Add description with the PDF link
+    if pdf_link_finding:
+        sanction.add("description", pdf_link_finding)  # Finding PDF link
```
jbothma, line 71: you could add these links to the sourceUrl property instead

datasets/us/fincen_special_measures/crawler.py
```diff
@@ -0,0 +1,147 @@
...
+    if pdf_link_nprm:
+        sanction.add("description", pdf_link_nprm)  # NPRM PDF link
+    if pdf_link_final_rule:
+        sanction.add("description", pdf_link_final_rule)  # Final Rule PDF link
+    if pdf_link_rescinded:
+        sanction.add("description", pdf_link_rescinded)  # Rescinded PDF link
+
+    # Emit the entity and the sanction
+    context.emit(entity, target=True)
+    context.emit(sanction)
```
jbothma, line 81: only set target=True and add the topic if there's no rescinded date

datasets/us/fincen_special_measures/crawler.py
```diff
@@ -0,0 +1,147 @@
...
+BASE_URL = "https://www.fincen.gov"
+
+
+def convert_date(date_str: str) -> str or None:
+    """Convert various date formats to 'YYYY-MM-DD'."""
+    # Regular expression to find dates in the format 'MM/DD/YYYY'
+    date_pattern = re.compile(r"(\d{1,2}/\d{1,2}/\d{4})")
+    match = date_pattern.search(date_str)
+    if match:
+        date_str = match.group(1)  # Extract the first matching date
```
jbothma, line 16: can you easily enough extract and return all matching dates? we support multiple values for the same property for just this scenario.

datasets/us/fincen_special_measures/us_fincen_special_measures.yml
```diff
@@ -0,0 +1,41 @@
...
+entry_point: crawler.py
+prefix: us-fincen
+publisher:
+  name: Financial Crimes Enforcement Network
+  acronym: FinCEN
+  official: true
+  description: |
+    The Financial Crimes Enforcement Network (FinCEN) is a bureau of the United States Department of the Treasury that collects and 
+    analyzes information about financial transactions in order to combat domestic and international money laundering, terrorist financing, 
+    and other financial crimes.
```
jbothma, line 25: could you wrap the yaml text at around 80-90 chars?

datasets/us/fincen_special_measures/us_fincen_special_measures.yml
```diff
@@ -0,0 +1,41 @@
+title: "FinCEN 311 and 9714 Special Measures"
```
jbothma, line 1: could you add data assertions for this? https://zavod.opensanctions.org/best_practices/merge_checklist/

datasets/us/fincen_special_measures/crawler.py
```diff
@@ -0,0 +1,147 @@
...
+def crawl_item(context: Context, row: Dict[str, str]):
+    # Create the entity based on the schema
+    name = row.pop("company")
+    schema = context.lookup_value("target_type", name)
+    if schema is None:
+        schema = "Company"
+    entity = context.make(schema)
+    entity.id = context.make_id(name)
+    entity.add("name", name)
+    entity.add("topics", "sanction")
```
jbothma, line 35: if "final ruling" -> sanction else reg.warn

jbothma (changes requested):
nice work! I think the code can be done a bit more simply and safely, but shout if I'm missing somethng

datasets/us/fincen_special_measures/crawler.py
```diff
@@ -0,0 +1,113 @@
...
+
+    rescinded_date = row.get("rescinded").text_content()
+    if rescinded_date != "---":
+        sanction.add("endDate", convert_date(rescinded_date))
+        context.emit(entity, target=True)
+    else:
+        context.emit(entity, target=False)
+
+    # Emit the entity and the sanction
+    # context.emit(entity)
```
jbothma, line 79:
```suggestion
```

datasets/us/fincen_special_measures/crawler.py
```diff
@@ -0,0 +1,113 @@
...
+        schema = "Company"
+    entity = context.make(schema)
+    entity.id = context.make_id(name)
+    entity.add("name", name)
+    # Adjust the topic based on the presence of "final rule"
+    final_rule = row.get("final_rule", "").strip().lower()
+    if final_rule and final_rule != "---" and final_rule != "":
+        entity.add("topics", "sanction")
+    else:
+        entity.add("topics", "reg.warn")
```
jbothma, line 38: if final rule and not rescinded

datasets/us/fincen_special_measures/crawler.py
```diff
@@ -0,0 +1,113 @@
...
+    listing_date = finding_date if finding_date else nprm_date
+    listing_date = convert_date(listing_date)
+    sanction.add("listingDate", listing_date)
+
+    final_rule_date = row.get("final-rule").text_content()
+    if final_rule_date != "---":
+        sanction.add("startDate", convert_date(final_rule_date))
+
+    rescinded_date = row.get("rescinded").text_content()
+    if rescinded_date != "---":
```
jbothma, line 72: it can also be blank

datasets/us/fincen_special_measures/crawler.py
jbothma: maybe re.findall would do the trick here?

datasets/us/fincen_special_measures/crawler.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def convert_date(date_str: str) -> List[str]:
+    """Convert various date formats to 'YYYY-MM-DD'."""
+    # Regular expression to find dates in the format 'MM/DD/YYYY'
+    date_pattern = re.compile(r"(\d{1,2}/\d{1,2}/\d{4})")
+    date_pattern.search(date_str)
+
+    formats = [
+        "%m/%d/%Y",  # 'MM/DD/YYYY' format
```
jbothma, line 16: this is the only one you need though, right?

datasets/us/fincen_special_measures/crawler.py
```diff
@@ -0,0 +1,113 @@
...
+    for row in table.findall(".//tr"):
+        if headers is None:
+            headers = []
+            for el in row.findall("./th"):
+                # Workaround because lxml-stubs doesn't yet support HtmlElement
+                # https://github.com/lxml/lxml-stubs/pull/71
+                eltree = cast(html.HtmlElement, el)
+                headers.append(slugify(eltree.text_content()))
+            assert headers[0] is None, headers
+            assert len(set(headers)) == len(headers), headers
```
jbothma, line 95:
```suggestion
            assert headers[0] is None, headers
            # no duplicate column headers
            assert len(set(headers)) == len(headers), headers
```

jbothma (commented):
Looking good - the most important ones still are to double-check we're only treating the ones with rescinded date as rescinded, and also try and get all the dates for a particular field. Also a few nitpicks.

datasets/us/fincen_special_measures/crawler.py
```diff
@@ -0,0 +1,114 @@
...
+
+def convert_date(date_str: str) -> List[str]:
+    """Convert various date formats to 'YYYY-MM-DD'."""
+    # Regular expression to find dates in the format 'MM/DD/YYYY'
+    date_pattern = re.compile(r"(\d{1,2}/\d{1,2}/\d{4})")
+    date_pattern.findall(date_str)
+
+    formats = ["%m/%d/%Y"]  # 'MM/DD/YYYY' format
+    date = h.parse_date(date_str, formats)
+    return date
```
jbothma, line 17: so a couple of things

- the result of the regex findall wasn't being used, so we're getting one value (just a year) when multiple dates are available
- it's nice to compile regexes once at the top of the module, rather than each time they're used. I missed this earlier

```suggestion
from typing import List


REGEX_DATE = re.compile(r"(\d{1,2}/\d{1,2}/\d{4})")


def convert_date(date_str: str) -> List[str]:
    """Convert various date formats to 'YYYY-MM-DD'."""
    dates = date_pattern.findall(date_str)
    parsed = []
    formats = ["%m/%d/%Y"]  # 'MM/DD/YYYY' format
    for str in dates:
        parsed.extend(h.parse_date(str, formats)
    return parsed
```

PR #1034, opened 2024-07-12
Retry transient errors according to zyte docs

jbothma:


pudo (approved):
Let's give it a shot. Do you have time maybe in the next week or two to look into a blob downloader using zyte, too? For the PH Police images & some other stuff :)

PR #1048, opened 2024-07-17
Stop expanding to sibling securities via issuer

jbothma:


pudo (approved):
I find it really hard to conceptualise the impact fully (beyond securities) but it is a very sane policy to take. Let's FAFO.

jbothma:
This should have the same effect as this line https://github.com/opensanctions/nomenklatura/blob/main/nomenklatura/entity.py#L389

PR #1060, opened 2024-07-19
Skip local enricher search results below the average for that search

jbothma:
We want to speed up the local enricher. We're currently spending most of the time in regression-v1, and also a significant portion of time fetching matches from the store.

For each entity, we fetch up to 50 results from the tantivy search index and score them. By only scoring the results who scored better than the mean of the tantivy scores for that search, we reduce the number by roughly half.

If the scores are a descending curve, we'll score a bit more, and ascending, we'll score a bit less. We haven't tested whether that really has an impact.

What we've seen is that this doesn't reduce the matches with positive decisions for the current enrichers using LocalEnricher, nor for ext_cz_business_register.

### DIFF
--- a/zavod/zavod/runner/local_enricher.py
+++ b/zavod/zavod/runner/local_enricher.py
@@ -89,7 +89,16 @@ class LocalEnricher(Enricher):
             if same_id_match is not None:
                 yield self.entity_from_statements(type(entity), same_id_match)
 
-        for match_id, index_score in self._index.match(store_type_entity):
+        search_results = self._index.match(store_type_entity)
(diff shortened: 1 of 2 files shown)


pudo:
If I understand this correctly, we're sort of optimizing for a world here in which a particular enrichment query produces a few good matches, and then a load of bad ones. While that scenario definitely exists, I wonder if it would be even more effective to consider the (what feels like 95%) case in which none of the results are any good, and even the top-scored one will already be a thing that tantivy returns only because we asked it to. Or do we have good enough filters on the queries to prevent junk results in the first place?

PR #1073, opened 2024-07-23
Romania Terrorism Related Sanctions List #216

nvmbrasserie:
fixes opensanctions/crawler-planning#216


datasets/ro/onpcsb_sanctions/ro_onpcsb_sanctions.yml
```diff
@@ -0,0 +1,41 @@
...
+  frequency: never
+  start: 2024-07-24
+summary: >
+  The dataset contains information about suspected individuals and organizations involved in terrorism, approved by the Government of Romania
+  and based on UN Security Council data.
+description: |
+  The dataset contains a list of individuals and organizations suspected of committing or funding terrorist activities. It was approved 
+  by the Government of Romania through Decision No. 1.272/2005 and is based on information from the UN Security Council. For individuals, 
+  it includes personal details such as names, dates and places of birth, nationalities, and passport numbers. For entities, the entries include 
+  names, alternate names (if any), addresses, and other relevant details. The purpose of this list is to help prevent and combat terrorism.
```
jbothma, line 16:
```suggestion
  names, alternate names (if any), addresses, and other relevant details. The purpose of this list is to help prevent and combat terrorism.
  
  According to the law referenced by this publication, the transfer of assets by these entities are
  prohibited and are to be blocked.
```

datasets/ro/onpcsb_sanctions/ro_onpcsb_sanctions.yml
```diff
@@ -0,0 +1,41 @@
+title: "Romania Government Decision No. 1.272/2005: List of Suspected Terrorists"
+url: https://www.onpcsb.ro/en/a/126/legislatie
+entry_point: crawler.py
+prefix: ro-onpcsb
+load_db_uri: ${OPENSANCTIONS_DATABASE_URI}
+coverage:
+  frequency: never
+  start: 2024-07-24
+summary: >
+  The dataset contains information about suspected individuals and organizations involved in terrorism, approved by the Government of Romania
```
jbothma, line 10: it's nice if the text in these files can be wrapped roughly around 90 chars wide. Just makes it easier to edit in smaller windows, e.g. split screen.

datasets/ro/onpcsb_sanctions/crawler.py
```diff
@@ -0,0 +1,88 @@
...
+    if not text:
+        return []
+    date = h.parse_date(text, DATE_FORMATS)
+    return date
+
+
+def crawl_row(context: Context, row: Dict[str, str]):
+    full_name = row.pop("name")
+    other_name = row.pop("other name")
+    birth_date_1 = parse_date(row.pop("date of birth"))
```
jbothma, line 21: you might not need the parse_date function if you 

```suggestion
    birth_date_1 = h.parse_date(row.pop("date of birth") or None, FORMATS)
```

datasets/ro/onpcsb_sanctions/crawler.py
```diff
@@ -0,0 +1,88 @@
...
+        entity.add("birthDate", birth_date_1)
+        if birth_date_2:
+            entity.add("birthDate", birth_date_2)
+        entity.add("birthPlace", birth_place)
+        # Handle multiple nationalities
+        nationalities = nationality.split("/")
+        for nation in nationalities:
+            entity.add("nationality", nation.strip())
+        entity.add("passportNumber", passport_number)
+        entity.add("address", address)
```
jbothma, line 60: I think you need to emit the address as well. Instead of entity.add, I think h.apply_address will add and emit for you in one call.

datasets/ro/onpcsb_sanctions/crawler.py
```diff
@@ -0,0 +1,88 @@
...
+        entity.add("name", full_name)
+        entity.add("alias", other_name)  # separated by comma
+        entity.add("birthDate", birth_date_1)
+        if birth_date_2:
+            entity.add("birthDate", birth_date_2)
+        entity.add("birthPlace", birth_place)
+        # Handle multiple nationalities
+        nationalities = nationality.split("/")
+        for nation in nationalities:
+            entity.add("nationality", nation.strip())
```
jbothma, line 58: A touch more concise would be

```suggestion
        entity.add("nationality", [n.strip() for n in nationality.split("/")])
```

datasets/ro/onpcsb_sanctions/crawler.py
```diff
@@ -0,0 +1,88 @@
...
+    # add a function to handle multiple entires
+    if not text:
+        return []
+    date = h.parse_date(text, DATE_FORMATS)
+    return date
+
+
+def crawl_row(context: Context, row: Dict[str, str]):
+    full_name = row.pop("name")
+    other_name = row.pop("other name")
```
jbothma, line 20: looks like you can

```suggestion
    other_name = h.multi_split(row.pop("other name"), [",", ";"])
```

this

datasets/ro/onpcsb_sanctions/crawler.py
```diff
@@ -0,0 +1,88 @@
...
+        entity.add("name", full_name)
+        entity.add("alias", other_name)  # separated by semicolon
+        entity.add("address", address)
+        entity.add("country", country)
+    else:
+        context.log.warning("Unhandled entity type", type=entity_type)
+
+    if entity:
+        entity.add("topics", "sanction")
+        sanction = h.make_sanction(context, entity)
```
jbothma, line 77: if we don't have any interesting data about the sanction like start date or reason, there's no point in making these entities. Then just put the sanctions program in the entity:program property. That seems to be the case with this dataset.

datasets/ro/onpcsb_sanctions/crawler.py
```diff
@@ -0,0 +1,88 @@
...
+        city=city,
+        country=country,
+    )
+    entity_type = row.pop("type")
+    raw_data = row.pop("parsed data")
+
+    entity = None
+    if entity_type == "Person":
+        entity = context.make("Person")
+        entity.id = context.make_id(raw_data)
```
jbothma, line 48: I'd suggest using name, date and place of birth here (all of them), and for company, something like name and a key part of the address like the post code.

The benefit of that, is if we get the same data in future in another form, e.g. they add some entities and publish a nice excel file or something, or even a new PDF, the ID hash is more likely to stay the same, meaning dedupe we do on these entities (with other datasets) continues to apply.

jbothma (commented):
Nice! Some small changes I think then we can ship

PR #1083, opened 2024-07-25
feat: emit delta stats from exporter

SimonThordal:


pudo:
Hey! Have you checked if this would be possible in Google logging by configuring a log sink based on structured log contents? I'm just a bit worried that this introduces a super surprising dependency between all the exporters and this metrics DB - thus making `zavod` a good bit less usable as open source software outside the context of our stack.

SimonThordal:
> Hey! Have you checked if this would be possible in Google logging by configuring a log sink based on structured log contents? I'm just a bit worried that this introduces a super surprising dependency between all the exporters and this metrics DB - thus making `zavod` a good bit less usable as open source software outside the context of our stack.

Is it that different than the cache, except we default to emitting no metrics instead where the cache uses sqlite? 

We can also do it through cloud logging though! Then we'll need to properly log the operations and upgrade our log bucket to use [log analytics](https://cloud.google.com/logging/docs/log-analytics#analytics). As far as I can tell there's no additional cost to us.

pudo:
Should we close this in favor of #1085?

PR #1085, opened 2024-07-26
Emit stats in log

SimonThordal:
Another possible implementation of logs. Emit them to log, count them as a metric or through log analytics. We're not looking at excluding the merged entities just yet, I'd rather just get this out first so we can see if we can generate useful metrics from it.

### DIFF
--- a/zavod/zavod/exporters/delta.py
+++ b/zavod/zavod/exporters/delta.py
@@ -1,5 +1,4 @@
 from typing import Any, Generator
-
 from zavod.entity import Entity
 from zavod.archive import DELTA_EXPORT_FILE
 from zavod.exporters.common import Exporter
@@ -16,6 +15,11 @@ class DeltaExporter(Exporter):
         super().setup()
         self.delta = HashDelta(self.dataset)
         self.delta.backfill()
+        self.counts = {
+            "ADD": 0,
+            "MOD": 0,
+            "DEL": 0,
+        }
 
     def feed(self, entity: Entity) -> None:
         self.delta.feed(entity)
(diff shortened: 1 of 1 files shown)


zavod/zavod/exporters/delta.py
```diff
@@ -33,6 +37,15 @@ def generate(self) -> Generator[Any, None, None]:
...
             for op in self.generate():
+                if op["op"] in self.counts:
+                    self.counts[op["op"]] += 1
                 write_json(op, fh)
         self.delta.close()
+        self.context.log.info(
+            "Delta export complete",
+            version=str(self.context.version),
+            dataset=self.dataset.name,
+            **self.counts,
```
pudo, line 48: Should we lowercase these?

zavod/zavod/exporters/delta.py
```diff
@@ -33,6 +37,15 @@ def generate(self) -> Generator[Any, None, None]:
     def finish(self) -> None:
         with open(self.path, "wb") as fh:
             for op in self.generate():
+                if op["op"] in self.counts:
```
pudo, line 40: Kind of wonder if we don't want the check - if there's an unknown op, let's crash.

zavod/zavod/exporters/delta.py
```diff
@@ -33,6 +37,15 @@ def generate(self) -> Generator[Any, None, None]:
...
+                    self.counts[op["op"]] += 1
                 write_json(op, fh)
         self.delta.close()
+        self.context.log.info(
+            "Delta export complete",
```
pudo, line 45: Should we add something to this log that makes it easy to filter for? metric="export_ops" ?

pudo (commented):
Ok this is a lot easier :)

Issue #105, opened 2021-08-25
Prototype an API server


pudo:
Things an API server could offer:

* An OpenSanctions-specific matching API
* Support for the OpenRefine reconciliation API (see [Aleph implementation](https://github.com/alephdata/aleph/blob/main/aleph/views/reconcile_api.py))
* Export on demand with filters (country, schema) 

Should track what's happening in alephdata/followthemoney-compare that we can adopt downstream.

Implementation thoughts?

* Does this want to be implemented in some funky fast programming language (go, rust)? 
* Does this keep the data in memory (or use heavy memoization) to be blazing fast?
* Do we ship this to users or run it on an OS-sponsored server? If the latter, is there going to be API accounting?

Requires #106 or similar.

pudo:
Fixed now, iteration 2 should have specific issues. Lots of work to be done on match quality.

PR #1089, opened 2024-07-26
Nepal MHA #105

nvmbrasserie:
fixes opensanctions/crawler-planning#105


datasets/np/mha_sanctions/crawler.py
```diff
@@ -0,0 +1,81 @@
...
+        entity.add("notes", remarks)
+        if sanction_status == "ACTIVE":
+            entity.add("topics", "sanction")
+            sanction = h.make_sanction(context, entity)
+            sanction.add("listingDate", listing_date)
+            sanction.add("unscId", unsc_id)
+            context.emit(entity, target=True)
+            context.emit(sanction)
+        elif sanction_status == "REMOVED":
+            context.emit(entity, target=False)
```
jbothma, line 52: it's nice to still have the sanction entity with the information about the sanction, but just not give them the sanction topic.

When customers ask how they should know why someone came up as a match in the dataset, we tell them to look primarily at the topic property. If there's no topic, the entity is usually there because they're either somehow related to some risky entity, but not officially a risk themselves, or they used to be sanctioned/debarraed, but aren't any more.

datasets/np/mha_sanctions/crawler.py
```diff
@@ -0,0 +1,81 @@
...
+            entity.add("address", a.strip())
+        if sanction_status == "ACTIVE":
+            entity.add("topics", "sanction")
+            sanction = h.make_sanction(context, entity)
+            sanction.add("listingDate", listing_date)
+            sanction.add("unscId", unsc_id)
+            context.emit(entity, target=True)
+            context.emit(sanction)
+        elif sanction_status == "REMOVED":
+            context.emit(entity, target=False)
```
jbothma, line 71: same as above.

datasets/np/mha_sanctions/crawler.py
jbothma: since we're still including the person in the dataset, it's nice to include the sanction information explaining why they're here, but also why they don't have the topic any more.

jbothma (commented):
very nice! Just a couple of small things then this goes live

datasets/np/mha_sanctions/crawler.py
```diff
@@ -0,0 +1,81 @@
...
+        for a in alias:  # Add aliases
+            entity.add("alias", a.strip())
+        for d in birth_date:
+            entity.add("birthDate", d.strip())
+        for p in birth_place:
+            entity.add("birthPlace", p.strip())
+        for n in nationality:
+            entity.add("nationality", n.strip())
+        for p in pass_no:
+            entity.add("passportNumber", p.strip())
```
jbothma, line 39: Sorry, only spotted this now.

What could be really nice is for the passport numbers where `p.tolower().split(" number: ")` gives you two values, treat those values as `county, number` and

- put the number only in the passportNumber field - that will improve matching
- h.make_identification with the country name and number - that's more for end users making sense of the match result

image

datasets/np/mha_sanctions/crawler.py
```diff
@@ -0,0 +1,110 @@
...
+            p_parts = p.lower().split(" number: ")
+            if len(p_parts) == 2:
+                country = p_parts[0].strip()
+                number = p_parts[1].strip()
+                entity.add("passportNumber", number)
+
+                # Create an identification object associated with the entity
+                passport = h.make_identification(
+                    context, entity=entity, country=country, number=number
+                )
```
jbothma, line 64:
```suggestion
                # Create an identification object associated with the entity
                passport = h.make_identification(
                    context, entity=entity, country=country, number=number, passport=True
                )
```

PR #1090, opened 2024-07-26
 Organizations designated as terrorist by Bahrain #159

nvmbrasserie:
fixes opensanctions/crawler-planning#159


jbothma (commented):
Thanks for this!

jbothma:
I'm gonna close for now - the source doesn't cover the individuals, so I think it's best to hold off until we have a better source.

PR #1094, opened 2024-07-29
Jul 29 fixes

nvmbrasserie:


datasets/ua/nsdc_sanctions/ua_nsdc_sanctions.yml
```diff
@@ -29,8 +29,61 @@ data:
   format: JSON
 
 lookups:
-  # type.name:
-  #   options:
+  type.name:
+    options:
+      - name: "Філія Державного унітарного підприємства \"Республіки Крим\" \"Кримські морські порти\" \"Держгідрографія\" (Филиал Государственного унитарного предприятия \"Республики Крым\" \"Крымские морские порты\" \"Госгидрография\", Філія Державного унітарного підприємства \"Республіки Крим\" \"Кримські морські порти\" \"Держгідрографія\")"
```
jbothma, line 34: @nvmbrasserie I think it's missing the part where it should match the original name - you need a `match` or `regex` or `contains` key

Issue #101, opened 2021-07-17
json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 2 column 1 (char 2)


francofabio:
I'm  trying load the OpenSanctions JSON-model but I'm receiving this error `json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 2 column 1 (char 2)`

I'm using the command `cat opensanctions.json | ftm store write -d opensanctions`

Would you guys help me to understand what's going on?

pudo:
Hmm, that sounds like some broken JSON. Can you tell me what `opensanctions.json` is and where you got it?

francofabio:
@pudo I got the `opensanctions.json` on opensanctions.org

pudo:
Ah, I assume you downloaded the index file (https://data.opensanctions.org/datasets/latest/index.json). Try using one of the collections:

* All data: https://data.opensanctions.org/datasets/latest/all/entities.ftm.json
* All sanctions: https://data.opensanctions.org/datasets/latest/sanctions/entities.ftm.json

Can't wait till I have finished the home page :)

francofabio:
Wow! Thank you so much @pudo 
I will try it

PR #1101, opened 2024-07-30
Withhold Release Orders and Findings List #101

nvmbrasserie:
fixes opensanctions/crawler-planning#101

### DIFF
--- a/datasets/us/cbp_withhold_release/crawler.py
+++ b/datasets/us/cbp_withhold_release/crawler.py
@@ -1,0 +1,105 @@
+from lxml import html
+import re
+from zavod import Context, helpers as h
+from normality import slugify
+from typing import Dict, Generator, cast
+from typing import List
+
+
+REGEX_DATE = re.compile(r"(\d{1,2}/\d{1,2}/\d{4})")
+
+
+def convert_date(date_str: str) -> List[str]:
+    """Convert various date formats to 'YYYY-MM-DD'."""
+    dates = REGEX_DATE.findall(date_str)
+    parsed = []
+    formats = ["%m/%d/%Y"]  # 'MM/DD/YYYY' format
+    for str in dates:
+        parsed.extend(h.parse_date(str, formats))
+    return parsed
+
+
+def crawl_item(context: Context, row: Dict[str, str]):
+    # Create the entity based on the schema
+    name = row.pop("name").text_content()
+    schema = context.lookup_value("target_type", name)
+    if schema is None:
+        schema = "Company"
+    entity = context.make(schema)
+    entity.id = context.make_id(name)
+    entity.add("name", name)
+    # Adjust the topic based on the presence of "final rule"
+    final_rule = row.get("final_rule", "").strip().lower()
+    rescinded_date = row.get("rescinded").text_content()
+    if (
+        final_rule
+        and final_rule != "---"
+        and (not rescinded_date or rescinded_date == "---")
+    ):
+        entity.add("topics", "sanction")
+    else:
+        entity.add("topics", "reg.warn")
+
+    # Create and add details to the sanction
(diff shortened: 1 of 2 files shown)


datasets/us/cbp_forced_labor/us_cbp_forced_labor.yaml
```diff
@@ -0,0 +1,354 @@
...
+          - name: 
+              - Central Romana Corporation Limited
+      - match: 'Mangalore Ganesh Beedie Works'
+        entities: 
+          - name: 
+              - Mangalore Ganesh Beedie Works
+      - match: 'Fuchu Prison; Union Kogyo Co., Ltd.'
+        entities: 
+          - name: 
+              - Fuchu Prison
```
jbothma, line 250: If there's generally a single value for name, you can just make the name the value - no point putting it in an array. This will save a lot of visual space reading the lookups

If there is a case where you want two name values, putting it in an array is fine. Should just work.

```suggestion
          - name: Fuchu Prison
```

datasets/us/cbp_forced_labor/crawler.py
```diff
@@ -0,0 +1,137 @@
...
+                        entity.id = context.make_id(
+                            match_entity.get("name"), internal_id
+                        )
+                        entity.add("name", match_entity.get("name"))
+                        if status in ["Active", "Partially Active"]:
+                            entity.add("topics", "sanction")
+                            sanction = h.make_sanction(context, entity)
+                            sanction.add("listingDate", listing_date)
+                            entity.add("notes", status_notes)
+                        context.emit(entity, target=True)
```
jbothma, line 43: also emit the sanction. Here and for companies

datasets/us/cbp_forced_labor/crawler.py
```diff
@@ -0,0 +1,137 @@
...
+        name_result = context.lookup("name", name)
+        listing_date = h.parse_date(row.pop("date"), formats=["%m/%d/%Y"])
+        merchandise = row.pop("merchandise")
+        status = row.pop("status")
+        status_notes = row.pop("status-notes")
+        status_notes_link = row.pop("status_notes_link", None)
+
+        # print(f"Processing entity: {name}, ID: {internal_id}")
+
+        if re.search(pattern, country, re.IGNORECASE):
```
jbothma, line 29: Does this need to be a regex? Or can it just be a string comparison like `if "fishing vessels" in country.lower()`?

If it really needs to be a regex, Id name the pattern so it reads more clearly, and compile it at module-level
```suggestion

REGEX_VESSEL = re.compile(r"\bFishing\s+Vessels\b")

def ...
    ...
        if REGEX_VESSEL.search(country, re.IGNORECASE):
```

jbothma (commented):
Excellent job! Just small improvement suggestions

datasets/us/cbp_forced_labor/crawler.py
nvmbrasserie: I assume there're some whitespaces, it only worked with the regex, so I'd keep it

datasets/us/cbp_forced_labor/crawler.py
```diff
@@ -0,0 +1,158 @@
...
+    context: Context, table: HtmlElement, main_header: str
+) -> Generator[Dict[str, Any], None, None]:
+    headers = []
+    header_found = False
+
+    for row in table.findall(".//tr"):
+        # Check if this row is the main header row
+        main_header_cell = row.find(".//th[@colspan='6'][@scope='col']")
+        if main_header_cell is not None:
+            continue  # Skip the main header row as it's already provided
```
jbothma, line 106: cool way to skip only what you know is the spanning header

PR #1106, opened 2024-07-31
The PEP stats exporter now runs as a separate cron job

pudo:


jbothma (approved):
The PEPs exporter is dead! Long live the PEPs exporter!

Issue #1095, opened 2024-07-29
Go through wanted lists with new FtM properties


pudo:
The latest FtM brings in some new properties: 

* height, weight, eyeColor, hairColor, identifyingMarks - none of which have defined units, just use as per list 
* `wanted` as a topic for entities - less definite than `crime`.

PR #1135, opened 2024-08-07
Go through wanted lists with new FtM properties #1095

nvmbrasserie:
updated FtM properties for each wanted list:
height, weight, eyeColor, hairColor, identifyingMarks were added where applicable.

`"wanted"` was added as one of the topics


nvmbrasserie:
pk_proscribed_persons is down and still to be updated. For now it's skipped with `ci_test: false`

PR #1136, opened 2024-08-08
Common HTML table and helpers

jbothma:
There are currently 17 `def parse_table`s.

This replaces each crawler's own parse_table with h.parse_table. By returning Elements for cell values instead of strings, crawlers that need to be fancy like pick out one or a flexible number of links from cells, can do so.

Testing:

for each modified crawler

1. crawl fully locally with prod ZAVOD_OPENSANCTIONS_API_KEY, ZAVOD_ARCHIVE_BUCKET and ZAVOD_ARCHIVE_BACKEND
2. `zavod validate`
3. check for zero or explainable changes in entities.deltas.json


zavod/zavod/helpers/html.py
```diff
@@ -3,7 +3,7 @@
 from lxml.html import HtmlElement
 
 
-def parse_table(table: HtmlElement) -> Generator[Dict[str, str], None, None]:
+def parse_table(table: HtmlElement) -> Generator[Dict[str, HtmlElement], None, None]:
```
jbothma, line 6: parse_html_table

PR #1144, opened 2024-08-09
Date parser refactor

pudo:


datasets/at/nbter_sanctions/at_nbter_sanctions.yml
```diff
@@ -32,6 +32,21 @@ data:
...
+    "03": März
+    "04": April
+    "05": Mai
+    "06": Juni
+    "07": Juli
+    "08": August
+    "09": September
+    "10": Oktober
+    "11": November
+    "12": Dezember
```
jbothma, line 49: I can see us adding a `lang` option instead of `months` in future, with standard dates in zavod.meta.dates - we already have three sets of german mappings (one with an austrian option).

```yaml
  lang: deu
```

jbothma (approved):
very cool

datasets/at/nbter_sanctions/at_nbter_sanctions.yml
pudo: This must exist somewhere .... I just want to avoid `dateparser` (https://dateparser.readthedocs.io/en/v0.3.0/index.html) because it's a slow mess. 

FWIW, they seem to have sourced this from CLRD: https://github.com/scrapinghub/dateparser/blob/master/dateparser_data/cldr_language_data/date_translation_data/de.json

PR #1145, opened 2024-08-09
eiti-soe-database

nvmbrasserie:
fixes opensanctions/crawler-planning#209

### DIFF
--- a/datasets/_global/eiti_soe_list/crawler.py
+++ b/datasets/_global/eiti_soe_list/crawler.py
@@ -6,13 +6,12 @@ def crawl_entity(context: Context, item: Dict[str, Any]) -> None:
     entity = context.make("LegalEntity")
 
     soe = item.pop("SOE")
-    year = item.pop("Year")
     country = item.pop("Country")
     fin_statement = item.pop("Audited Financial Statement or Equivalent")
     website = item.pop("Public Listing or Website")
 
     # Making a unique ID for the entity
-    entity.id = context.make_id(soe, year, country)
+    entity.id = context.make_id(soe, country)
 
     # Adding fields to the entity
     entity.add("name", soe)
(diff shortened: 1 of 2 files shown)


datasets/_global/eiti_soe_list/crawler.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def crawl(context: Context):
+    data = context.fetch_json(context.data_url)
+    if not data or not data.get("rows"):
+        context.log.error("No data found or 'rows' key is missing.")
+        return
+
+    # Processing each item in the 'rows' list
+    for item in data["rows"]:
```
jbothma, line 38: Pop raises an exception when the key is missing, logging automatically.

```suggestion
    if not data:
        context.log.error("No data found")
        return

    # Processing each item in the 'rows' list
    for item in data.pop("rows"):
```

datasets/_global/eiti_soe_list/crawler.py
```diff
@@ -0,0 +1,39 @@
...
+    entity = context.make("LegalEntity")
+
+    soe = item.pop("SOE")
+    year = item.pop("Year")
+    country = item.pop("Country")
+    fin_statement = item.pop("Audited Financial Statement or Equivalent")
+    website = item.pop("Public Listing or Website")
+
+    # Making a unique ID for the entity
+    entity.id = context.make_id(soe, year, country)
```
jbothma, line 15: I think let's drop year from key so that the same SOE over multiple years gets the same key. Or did you include it because there's some risk of different SOEs having the same name in the same country over time?

datasets/_global/eiti_soe_list/eiti_soe_list.yaml
```diff
@@ -0,0 +1,40 @@
...
+  frequency: daily
+  start: "2024-08-09"
+deploy:
+  schedule: "@weekly"
+load_db_uri: ${OPENSANCTIONS_DATABASE_URI}
+summary: >
+  List of all SOEs operating in EITI implementing countries.
+description: |
+  This dataset covers SOEs’ payments to government and other disclosures related to oil, 
+  gas and mining activities in EITI implementing countries.
```
jbothma, line 14: This dataset doesn't really cover the payments - it only really lists the entities.

You could expand the dataset description a bit with something like "this is made available to facilitate identification of channels for corruption."

jbothma (changes requested):
cool! Just a couple of small things I think

Issue #121, opened 2021-09-07
Re-do feature mappings in OFAC & CSL


pudo:
* Include CryptoWallets and Securities
* Maybe share mappings between OFAC and CSL?

pudo:
Done :) Feature definitions now live in `common.yml` :)

PR #1146, opened 2024-08-09
Add monaco_sanction crawler

atehe:
close [Monaco sanctions list #121](https://github.com/opensanctions/crawler-planning/issues/121)
Hey @jbothma , I'm getting a weird reaction when I add datapatch lookups for `type.country`.
When I add a lookup for rejected value, I rejects even more values of that patch I applied.
Example, I add a lookup that contains RUSSIE to patch to ru, but the other RUSIE values that were not rejected gets rejected after the patch.
I may be doing something wrongly but what do you think ?

### DIFF
--- a/datasets/mc/fund_freezes/crawler.py
+++ b/datasets/mc/fund_freezes/crawler.py
@@ -1,0 +1,84 @@
+from zavod import Context
+from zavod import helpers as h
+import re
+from typing import Dict
+
+
+def crawl_person(context: Context, record):
+    person_details = record.get("mesureDetails")
+    first_name = person_details.pop("prenom")
+    gender = person_details.pop("sexe")
+    dob = person_details.pop("dateNaissance")
+    if dob:
+        dob = h.parse_date(dob, ["%d/%m/%Y"])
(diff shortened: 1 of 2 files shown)


jbothma:
Hi!

I did

```diff
diff --git a/datasets/mc/fund_freezes/mc_fund_freezes.yml b/datasets/mc/fund_freezes/mc_fund_freezes.yml
index 69030290..2570144c 100644
--- a/datasets/mc/fund_freezes/mc_fund_freezes.yml
+++ b/datasets/mc/fund_freezes/mc_fund_freezes.yml
@@ -34,6 +34,9 @@ lookups:
       - contains:
           - RPDC
         value: kp
+      - contains:
+          - RUSSIE
+        value: ru
```

and it seems to be working fine. There are lots of other issues though so it's hard to see new ones. The new main branch (with new followthemoney) also warns that the names are too long which you probably didn't have yet.

If you're still having trouble, perhaps push a branch with the issue and show an example of the output?

atehe:
Ahh, got it...
Will try the new branch.
Thank you

PR #1177, opened 2024-08-23
Ru egrul name length

jbothma:
ext_ru_egrul has lots of too-long names

from @nvmbrasserie: we need to be careful with automatic trimming. Although lots of prefixes are generic, like "state" or "primary trade union", trimming too much can change the meaning too much

the plan:

- manually made list of prefixes that are safe enough to trim in ru_egrul
- use short names if they're long enough, putting the too-long name in the description
- the remainder get patched in ext_ru_egrul - would this work?! How is it even warning in ext_?

what hasnt worked:
- if there's an odd number of "s and the part before the first " is shorter than the rest - not always safe to trim, doesn't cover a large % of issues
- longest repeating phrase - actually usually doesn't match a prefix, so we'd be stripping something mid-phrase and making language meaningless or broken
- keep only the bit in parens or "s - often not sufficient information

With the prefix trimming we'll probably still have 250+ name edits for ext_ru_egrul.


datasets/ru/egrul/crawler.py
```diff
@@ -1,21 +1,136 @@
...
+    "ФЕДЕРАЛЬНОЕ ГОСУДАРСТВЕННОЕ БЮДЖЕТНОЕ УЧРЕЖДЕНИЕ",
+    "ФЕДЕРАЛЬНОЕ ГОСУДАРСТВЕННОЕ КАЗЕННОЕ ВОЕННОЕ ОБРАЗОВАТЕЛЬНОЕ УЧРЕЖДЕНИЕ ВЫСШЕГО ОБРАЗОВАНИЯ",
+    "ФЕДЕРАЛЬНОЕ ГОСУДАРСТВЕННОЕ КАЗЕННОЕ ОБРАЗОВАТЕЛЬНОЕ УЧРЕЖДЕНИЕ ДОПОЛНИТЕЛЬНОГО ПРОФЕССИОНАЛЬНОГО ОБРАЗОВАНИЯ (ПЕРЕПОДГОТОВКИ И ПОВЫШЕНИЯ КВАЛИФИКАЦИИ)",
+    "ФЕДЕРАЛЬНОЕ ГОСУДАРСТВЕННОЕ ОБРАЗОВАТЕЛЬНОЕ УЧРЕЖДЕНИЕ ВЫСШЕГО ПРОФЕССИОНАЛЬНОГО ОБРАЗОВАНИЯ",
+    "ФЕДЕРАЛЬНОЕ ГОСУДАРСТВЕННОЕ ОБРАЗОВАТЕЛЬНОЕ УЧРЕЖДЕНИЕ ДОПОЛНИТЕЛЬНОГО ПРОФЕССИОНАЛЬНОГО ОБРАЗОВАНИЯ СПЕЦИАЛИСТОВ",
+    "ФЕДЕРАЛЬНОЕ ГОСУДАРСТВЕННОЕ ОБРАЗОВАТЕЛЬНОЕ УЧРЕЖДЕНИЕ ДОПОЛНИТЕЛЬНОГО ПРОФЕССИОНАЛЬНОГО ОБРАЗОВАНИЯ СПЕЦИАЛИСТОВ",
+    "ФЕДЕРАЛЬНОЕ ГОСУДАРСТВЕННОЕ УНИТАРНОЕ ПРЕДПРИЯТИЕ",
+    "ФЕДЕРАЛЬНОЕ ГОСУДАРСТВЕННОЕ",  # Federal state
+    "ФЕДЕРАЛЬНОЕ ДОЧЕРНЕЕ ГОСУДАРСТВЕННОЕ УНИТАРНОЕ ПРЕДПРИЯТИЕ",
+    'ФЕДЕРАЛЬНОЕ КАЗЕННОЕ УЧРЕЖДЕНИЕ "ОТДЕЛ ОХРАНЫ ФЕДЕРАЛЬНОГО КАЗЕННОГО УЧРЕЖДЕНИЯ',
```
nvmbrasserie, line 114: this contains a part of the actual name, I'd remove `"ОТДЕЛ ОХРАНЫ ФЕДЕРАЛЬНОГО КАЗЕННОГО УЧРЕЖДЕНИЯ'`

PR #1185, opened 2024-08-26
adding New York medicare exclusion list crawler

bgmello:


datasets/us/medicare_exclusions/new_york/crawler.py
```diff
@@ -0,0 +1,83 @@
...
+    sanction = h.make_sanction(context, entity)
+    sanction.add(
+        "startDate",
+        h.parse_date(row.pop("exclusion-effective-date"), formats=["%m/%d/%Y"]),
+    )
+
+    context.emit(entity, target=True)
+    context.emit(sanction)
+
+    context.audit_data(row, ignore=["license-num", "provider-type"])
```
jbothma, line 72: let's put provider type in sector
let's put license num in idNumber

jbothma (changes requested):
Thanks! Same changes as Kansas please

PR #1192, opened 2024-08-27
Generate training pairs by clustering chronologically

jbothma:


contrib/make_training_pairs.py
```diff
@@ -1,93 +1,154 @@
...
+
+
+def generate_pairs(
+    resolver_path: Path,
+    dataset: Dataset,
+) -> Generator[Tuple[Entity, Entity, Judgement], None, None]:
+    chrono_resolver = ChronoResolver(resolver_path)
+    store = get_store(dataset, Linker({}))
+    store.sync()
+    view = store.view(dataset, external=False)
```
pudo, line 64: Oh this is kind of interesting: a good number of the negative/unsure judgements should be `external`-based, and then the external fragment isn't re-stored when the enricher runs again. So we lose these.

contrib/make_training_pairs.py
```diff
@@ -1,93 +1,154 @@
...
+        )
+
+    log.info("Generating disjoint subgraphs...")
+    for ix, subgraph_nodes in enumerate(nx.connected_components(g)):
+        for _, _, data in g.subgraph(subgraph_nodes).copy().edges(data=True):
+            yield {
+                "left": data["source"].to_dict(),
+                "right": data["target"].to_dict(),
+                "judgement": data["judgement"].value,
+                "group": ix,
```
pudo, line 135: uh, fascinating, what's happening here? I'm a fan of loading this into NetworkX :)

pudo:
This looks ready to merge, should we?

PR #1203, opened 2024-09-02
adding Michigan Medicare Exclusions crawler

bgmello:
There are multiple sanction dates in the data. I used only the first one because I'm not sure what are the others.

### DIFF
--- a/datasets/us/mi/med_exclusions/crawler.py
+++ b/datasets/us/mi/med_exclusions/crawler.py
@@ -1,12 +1,8 @@
 from typing import Dict
 from rigour.mime.types import XLSX
 from openpyxl import load_workbook
-import re
 
 from zavod import Context, helpers as h
-
-# Regular expression to match the comma before "Inc."
-INC_PATTERN = r",\s*Inc\."
 
 
 def crawl_item(row: Dict[str, str], context: Context):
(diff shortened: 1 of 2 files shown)


datasets/us/mi/med_exclusions/crawler.py
```diff
@@ -0,0 +1,88 @@
...
+from rigour.mime.types import XLSX
+from openpyxl import load_workbook
+
+from zavod import Context, helpers as h
+
+
+def crawl_item(row: Dict[str, str], context: Context):
+
+    if not row.get("entity_name"):
+        return
```
jbothma, line 11: it looks like sometimes there's a first name but not a legal entity

datasets/us/mi/med_exclusions/crawler.py
```diff
@@ -0,0 +1,88 @@
...
+
+    entity_name = row.pop("entity_name")
+    entity = context.make("LegalEntity")
```
jbothma, line 14: Looks safe to be more specific

```suggestion
    entity = context.make("Company")
```

jbothma (changes requested):
Thanks! Mainly missing some entities

bgmello:
There are some dates I couldn't parse, I think they are an error but I'm not sure what to do in that case, do I just leave the warning?
Screenshot 2024-09-12 at 15 17 35

jbothma:
good question.
The next step is to add a `type.date` lookup in the metadata. Any values you feel you can accurately parse yourself, you can make the `value` the date, e.g. `2023-06-27`. The rest, just make `value: null`

https://zavod.opensanctions.org/best_practices/datapatch_lookups/

bgmello:
I added the lookup but it's still showing the warnings for the non-null values.

PR #1206, opened 2024-09-02
adding Indiana medicare exclusions crawler

bgmello:


jbothma (commented):
sorry, forgot to publish comments

jbothma:
the file says this shouldn't be used as an exclusion list. Just confirming with them what that means and whether we can use it as planned.

image

jbothma:
800-457-4515 option 8 https://www.in.gov/medicaid/providers/files/quick-reference.pdf

jbothma:
let's add these without the debarment topic, and with target=False

for that to work, also add

```yaml
exports:
  - statistics.json
  - entities.delta.json
  - entities.ftm.json
  - names.txt
  - senzing.json
```

to the metadata so that we exclude targets.nested.json, which will be empty.

PR #1204, opened 2024-09-02
adding Massachusetts medicare exclusions crawler

bgmello:
I wasn't able to use the `context.fetch_html` and `context.fetch_resource` to get the data because it was returning with status code = 403.

PR #1212, opened 2024-09-03
adding New Hampshire Medicare Exclusions Crawler

bgmello:


bgmello:
Same as PR #1204, I wasn't able to use the `context` methods

jbothma (commented):
I'm missing Jennifer Lynn from the resulting data. Maybe I'm being silly but could you check if there's a bug?

bgmello:
I can see Jennifer Lynn in the `entities.ftm.json`:

Screenshot 2024-09-04 at 10 55 01

The id is `us-mednh-bb589665b25a0bacf2bb74e48cb4f137a6e1fb39`

jbothma:
Weird - after deleting the data file and letting it re-fetch, I find them.

Thanks for the fixes. Looks like some imports are snagging. Once it's green I can merge.

PR #1223, opened 2024-09-04
Interpol - Format weight as string

khanxmetu:
Hi,
Weight values could also be float therefore should be parsed as string to avoid decimal discrepancies. Same idea as commit 3b903bf0780e2c23de2aa2609f38fd2c4cd0117a

Example of affected entity:
https://www.opensanctions.org/statements/Q5546542
Source url: https://ws-public.interpol.int/notices/v1/red/2005-30989


jbothma:
Thank you!

khanxmetu:
I figured the discrepancy is coming due to this issue: https://github.com/pudo/normality/pull/21
If we get it fixed, we won't have to necessarily re-format floats in our crawlers :)

PR #1226, opened 2024-09-05
adding Mexico Tax Offender list crawler

bgmello:
closes [crawler-planning#232](https://github.com/opensanctions/crawler-planning/issues/232)

### DIFF
--- a/datasets/mx/tax_offender/crawler.py
+++ b/datasets/mx/tax_offender/crawler.py
@@ -45,11 +45,9 @@ def crawl_item(input_dict: dict, context: Context):
     sanction = h.make_sanction(context, entity)
     sanction.add("reason", input_dict.pop("reason"))
     if input_dict.get("start_date"):
-        sanction.add(
-            "startDate",
-            h.parse_date(input_dict.pop("start_date"), formats=["%d/%m/%Y", "%Y"]),
-        )
+        h.apply_date(sanction, "start_date", input_dict.pop("start_date"))
     sanction.add("authority", input_dict.pop("authority"))
+    sanction.add("description", "Value: " + input_dict.pop("value"))
 
     context.emit(entity, target=True)
     context.emit(sanction)
@@ -57,7 +55,6 @@ def crawl_item(input_dict: dict, context: Context):
     context.audit_data(
         input_dict,
         ignore=[
-            "value",
             "Fecha de publicación (Con monto de acuerdo a la Ley de Transparencia",
             "FECHA DE PUBLICACIÓN",
         ],
--- a/datasets/mx/tax_offender/crawler.py
+++ b/datasets/mx/tax_offender/crawler.py
@@ -40,7 +40,13 @@ def replace_key(d: dict, context: Context) -> dict:
 
 def crawl_item(input_dict: dict, context: Context):
 
-    schema = "Person" if input_dict.pop("person_type") == "F" else "Company"
+    schema = context.lookup_value("person_type", input_dict.get("person_type"))
+
+    if not schema:
+        context.log.info(input_dict.get("person_type"))
+        return
+    input_dict.pop("person_type")
+
 
     entity = context.make(schema)
     entity.id = context.make_id(input_dict.get("RFC"), input_dict.get("name"))
@@ -55,7 +61,8 @@ def crawl_item(input_dict: dict, context: Context):
     if input_dict.get("listing_date"):
         h.apply_date(sanction, "listingDate", input_dict.pop("listing_date"))
     sanction.add("authority", input_dict.pop("authority"))
-    sanction.add("description", "Value: " + input_dict.pop("value"))
+    if input_dict.get("value"):
+        sanction.add("description", "Value: " + input_dict.pop("value"))
     sanction.add("sourceUrl", input_dict.pop("source_url"))
 
     context.emit(entity, target=True)
--- a/datasets/mx/tax_offender/crawler.py
+++ b/datasets/mx/tax_offender/crawler.py
@@ -40,6 +40,7 @@ def crawl_item(input_dict: dict, context: Context):
     entity.id = context.make_id(input_dict.get("RFC"), input_dict.get("name"))
     entity.add("name", input_dict.pop("name"))
     entity.add("taxNumber", input_dict.pop("RFC"))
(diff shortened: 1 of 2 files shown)


datasets/mx/tax_offender/crawler.py
```diff
@@ -0,0 +1,124 @@
...
+        )
+    sanction.add("authority", input_dict.pop("authority"))
+
+    context.emit(entity, target=True)
+    context.emit(sanction)
+
+    context.audit_data(
+        input_dict,
+        ignore=[
+            "value",
```
jbothma, line 60: let's put this in sanction:description with a prefix `Value: `

datasets/mx/tax_offender/crawler.py
```diff
@@ -0,0 +1,124 @@
...
+    for possible_old_key in possible_old_keys:
+        if possible_old_key in d:
+            d[new_key] = d.pop(possible_old_key)
+            break
+    return d
+
+
+def crawl_item(input_dict: dict, context: Context):
+
+    schema = "Person" if input_dict.pop("person_type") == "F" else "Company"
```
jbothma, line 37: Some values I've seen:

Tipo de persona
Persona moral
Persona fisica

Let's use a datapatch lookup to map values to schemata. 

```
Persona moral: LegalEntity
Persona fisica: Person
```
etc

datasets/mx/tax_offender/crawler.py
```diff
@@ -0,0 +1,124 @@
...
+
+    entity = context.make(schema)
+    entity.id = context.make_id(input_dict.get("RFC"), input_dict.get("name"))
+    entity.add("name", input_dict.pop("name"))
+    entity.add("taxNumber", input_dict.pop("RFC"))
+    entity.add("topics", "debarment")
```
jbothma, line 43: I think this should be `reg.action` but let's wait for confirmation

```suggestion
    entity.add("topics", "reg.action")
```

datasets/mx/tax_offender/crawler.py
```diff
@@ -0,0 +1,124 @@
...
+
+    sanction = h.make_sanction(context, entity)
+    sanction.add("reason", input_dict.pop("reason"))
```
jbothma, line 46: let's add the CSV URL as sourceUrl to the sanction to make the reason and provenance super clear

datasets/mx/tax_offender/crawler.py
```diff
@@ -0,0 +1,124 @@
...
+    "http://omawww.sat.gob.mx/cifras_sat/Documents/Condonadosart21CFF.csv",
+    "http://omawww.sat.gob.mx/cifras_sat/Documents/CondonadosporDecreto.csv",
+    "http://omawww.sat.gob.mx/cifras_sat/Documents/Condonados_07_15.csv",
+    "http://omawww.sat.gob.mx/cifras_sat/Documents/Cancelados_07_15.csv",
+    "http://omawww.sat.gob.mx/cifras_sat/Documents/Retornoinversiones.csv",
+    "http://omawww.sat.gob.mx/cifras_sat/Documents/Exigibles.csv",
+    "http://omawww.sat.gob.mx/cifras_sat/Documents/Firmes.csv",
+    "http://omawww.sat.gob.mx/cifras_sat/Documents/No%20localizados.csv",
+    "http://omawww.sat.gob.mx/cifras_sat/Documents/Sentencias.csv",
+]
```
jbothma, line 19: can we take these from the HTML automatically? Something like all urls from the table with an allow-list and deny list of labels, so we exclude `Documento técnico y normativo` automatically, and get a warning when something that's neither in the allow list nor the deny list shows up? That way we can make 100% sure we don't miss new files, changed links, or accidentally add files that shouldn't be included automatically. a `context.log.warning("Unknown file found", label=label, url=url)`

jbothma (commented):
Nice! thank you.

datasets/mx/tax_offender/crawler.py
bgmello: I tried fetching the HTML using the `fetch_html` but it needs some kind of cookie, I can try to see if I can find a workaround if you want.

datasets/mx/tax_offender/mx_tax_offender.yaml
```diff
@@ -0,0 +1,43 @@
...
+  name: Servicio de Administración Tributaria
+  acronym: SAT
+  description: Tax Administration Service
+  url: https://www.sat.gob.mx/home
+  official: true
+  country: "mx"
+url: "http://omawww.sat.gob.mx/cifras_sat/Paginas/datos/vinculo.html?page=ListCompleta69.html"
+data:
+  url: http://omawww.sat.gob.mx/cifras_sat/Paginas/datos/vinculo.html?page=ListCompleta69.html
+  format: HTML
```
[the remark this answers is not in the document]
bgmello: When I try to export to check the number of entities I get the following error, do you know what it could be?

Screenshot 2024-09-12 at 11 27 51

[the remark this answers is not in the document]
jbothma: this happens if you generate the same ID for an entity with two different incompatible schemata, e.g. Company and Person.
I was wondering about using Company in any case. Some of the entities might just be legal entities and not companies per se.
But this might also indicate a bug, so I think before changing the Company schema, see if we're correctly generating the same ID for two different rows of data.

datasets/mx/tax_offender/crawler.py
jbothma: aha let's use http://omawww.sat.gob.mx/cifras_sat/Paginas/datos/ListCompleta69.html for the data URL

bgmello: It worked!

datasets/mx/tax_offender/mx_tax_offender.yaml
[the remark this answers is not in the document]
bgmello: It worked now, it was because it was choosing the wrong schema for a person that appears twice, but I fixed it in the commit https://github.com/opensanctions/opensanctions/pull/1226/commits/964080ea17b5a71fc71277c9d6ccffadebfe045d

PR #1228, opened 2024-09-05
adding description to medicare exclusions list publishers

bgmello:


datasets/us/ny/med_exclusions/us_ny_med_exclusions.yaml
```diff
@@ -13,6 +13,10 @@ description: |
   reimbursement for Medicaid services in any capacity.
 publisher:
   name: New York State Office of the Medicaid Inspector General
+  description: >
+   > The Office of the Medicaid Inspector General is an independent entity created within
+   > the New York State Department of Health to promote and protect the integrity of the Medicaid
+   > program in New York State.
```
jbothma, line 19: It turns out because of the > line break modifier, it can't have quote prefixes on each line - the yaml parser will merge all the lines into one.

```suggestion
  description: >
   > The Office of the Medicaid Inspector General is an independent entity created within
   the New York State Department of Health to promote and protect the integrity of the Medicaid
   program in New York State.
```

PR #1237, opened 2024-09-10
fixes-sep-10

nvmbrasserie:


datasets/jp/mof_sanctions/crawler.py
```diff
@@ -192,7 +192,9 @@ def emit_row(context: Context, sheet: str, section: str, row: Dict[str, List[str
     sanction.add("reason", row.pop("root_nomination", None))
     sanction.add("reason", row.pop("reason_res1483", None))
     sanction.add("authorityId", row.pop("notification_number", None))
-    sanction.add("unscId", h.multi_split(row.pop("designated_un", None), DATE_SPLITS))
+    sanction.add(
+        "startDate", h.multi_split(row.pop("designated_un", None), DATE_SPLITS)
+    )
 
     sanction.add("startDate", parse_date(row.pop("notification_date", [])))
     sanction.add("startDate", parse_date(row.pop("designated_date", [])))
```
jbothma, line 200: it looks like there isn't a column headed 指定日 so I think the solution is

1) remove 
```
      - match: 指定日
        value: designated_date
```

2) change designated_un to designated_date in the yaml

3) remove designated_date from the python.

PR #1241, opened 2024-09-11
sep-11-fix

nvmbrasserie:


datasets/tr/fcib/crawler.py
```diff
@@ -14,26 +14,23 @@
...
-        "https://ms.hmb.gov.tr/uploads/sites/2/2024/05/B-YABANCI-ULKE-TALEPLERINE-ISTINADEN-MALVARLIKLARI-DONDURULANLAR-6415-SAYILI-KANUN-6.-MADDE.xlsx",
-        "Asset freezes pursuant to Article 6 of Law No. 6415, targeting individuals"
-        " and entities based on requests made by foreign governments.",
-        "B - Foreign government requests",
+
+# Mapping of letter prefix to full label (program) and short label
+LABEL_MAPPING = {
+    "6MADDE_ING": (
+        "Asset freezes pursuant to Article 6 of Law No. 6415, targeting individuals and entities based on requests made by foreign governments.",
+        "Foreign government requests",
```
jbothma, line 22: I think their numbering (letters) is useful for quick reference

```suggestion
        "B - Foreign government requests",
```

datasets/tr/fcib/crawler.py
```diff
@@ -135,40 +132,66 @@ def crawl(context: Context):
...
+    # Process each found link
+    for url, program, short in found_bcd_links:
+        context.log.info(f"Processing URL: {url} - Program: {program}")
+        # Call the crawl_xlsx function to process the link
         crawl_xlsx(context, url, program, short)
+
     context.log.info("Finished processing the Excel files")
 
-    # UN Security Council stubs
+    # UN Security Council stubs (keep this part if necessary)
```
jbothma, line 192: Is this a question?

datasets/tr/fcib/crawler.py
```diff
@@ -135,40 +132,66 @@ def crawl(context: Context):
...
+        if len(doc_links) != 1:
+            context.log.warning(
+                f"Expected to find 1 .xlsx link in section {url}, but found {len(doc_links)}"
+            )
+
+        # Extract the link
+        doc_link = doc_links[0].get("href")
+
+        # Extract the prefix letter (e.g., B, C, D) from the filename or URL structure
+        prefix = url.split("/")[-1].split("-")[0].upper()
```
jbothma, line 164: This part isn't relevant to the HTML URL, is it?

```suggestion
        slug = url.split("/")[-1].upper()
```

Then a better name would also be slug, since it's the page slug rather than the section number or prefix

datasets/tr/fcib/crawler.py
jbothma: for the others too

PR #1246, opened 2024-09-12
h-apply-date-02

nvmbrasserie:


datasets/au/dfat_sanctions/crawler.py
```diff
@@ -49,7 +47,7 @@ def clean_date(date):
         part = part.strip().strip(",")
         if not len(part):
             continue
-        dates.update(h.parse_date(part, FORMATS))
+        dates.update(h.parse_date(part, context.dataset.dates.formats))
```
pudo, line 50: Creative! It doesn't allow us to see the original value, though, but not the end of the world :)

pudo (approved):

datasets/au/dfat_sanctions/crawler.py
nvmbrasserie: should I go ahead with this solution or is there another way we should be doing it?

PR #1249, opened 2024-09-12
Il sanction

atehe:
closes [Israel Weapons of Mass Destruction Banned Entities List #167](https://github.com/opensanctions/crawler-planning/issues/167)


datasets/il/wmd_banned/il_wmd_sanction.yml
```diff
@@ -0,0 +1,47 @@
...
+          - Iran
+        value: ir
+      - contains:
+          - 1969
+        value: null
+assertions:
+  min:
+    schema_entities:
+      Person: 100
+      Organization: 100
```
jbothma, line 47: let's add a max to this too - the max is helpful to discover with then min isn't sufficiently large any more.

datasets/il/wmd_banned/il_wmd_sanction.yml
```diff
@@ -0,0 +1,47 @@
...
+  name: Ministry of Finance
+  description: |
+    The Israeli Ministry of Finance is the primary economic ministry in Israel, overseeing the government's economic policies, fiscal targets, and budget management. 
+    It handles state revenues, tax collection, nonresident investments, and international economic relations.  [Read More](https://en.wikipedia.org/wiki/Ministry_of_Finance_(Israel))
+  country: il
+  url: https://www.gov.il/en/departments/ministry_of_finance/govil-landing-page
+  official: true
+url: https://www.gov.il/en/pages/declared_elements_list
+data:
+  url: https://www.gov.il/BlobFolder/generalpage/declared_elements_list/en/sanctions_headquarters_declared_elements_list-file2.xlsx
```
jbothma, line 27: it looks like this URL could change if they upload a new version. Could you look up the .xlsx URL on the page and make https://www.gov.il/en/pages/declared_elements_list the data URL so that we pick up changes automatically? If you search for xlsx in the codebase you should find examples - expecially in the us/*/med* path

datasets/il/wmd_banned/crawler.py
```diff
@@ -0,0 +1,244 @@
...
+        date_str = re.sub(heb_month, f"{eng_month} ", date_str, flags=re.IGNORECASE)
+        date_str = re.sub("ב", "", date_str, flags=re.IGNORECASE)
+
+    date_fmt = ["%d %B %Y", "%d %B. %Y"]
+    for fmt in date_fmt:
+        try:
+            date_obj = datetime.datetime.strptime(date_str, fmt)
+            return date_obj.date()
+        except ValueError:
+            pass
```
jbothma, line 49: can you replace most or all of this with this new approach? https://zavod.opensanctions.org/best_practices/dates_meta/

datasets/il/wmd_banned/crawler.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+def clean_name(name: str | List[str]):
+    """
+    Clean the given name (or list of names) by removing newlines, quotes, colons, and trailing dots
+    """
+    if isinstance(name, list):
+        return [re.sub(r"[\n':]|^\.*|\.*$", "", match).strip() for match in name]
+
+    return re.sub(r"[\n':]|^\.*|\.*$", "", name).strip()
```
jbothma, line 95: do we want to remove these, or split on them?
I suspect h.multi_split(names_str, "1:", "2:", ..., ";", "\n", "ידוע גם:"]) would give us a lovely list of names.

jbothma: Possibly that removes the need for format_numbered_listing?

datasets/il/wmd_banned/crawler.py
```diff
@@ -0,0 +1,244 @@
...
+            if 'מס"ד' in cells:  # cell that contains headers, skip it
+                continue
+
+            none_count = sum(1 for item in cells if not item)
+            if (
+                none_count > len(headers) // 2
+            ):  #  cells that do not contain data up to half of the headers count, skip it as it is probably a subheader
+                continue
+
+            yield cells
```
jbothma, line 142: this was a good idea, but could you see if you can use [h.parse_xlsx_table](https://zavod.opensanctions.org/helpers/#zavod.helpers.parse_xlsx_sheet) with the column translation lookup argument and skip repeated header rows in the item loop?

https://zavod.opensanctions.org/best_practices/datapatch_lookups/#translate-headers-using-datapatch-lookups

datasets/il/wmd_banned/crawler.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+def parse_sheet_row(context: Context, row: Tuple[str | int, ...]):
+    assert (
+        len(row) == 11
+    ), "Crawler was developed with excel sheet of 11 columns, please check that attributes still conforms to index used"
+
+    other_info = row.pop(10)
+    address_nationality = row.pop(9)
```
jbothma, line 151: index-based lookup is always risky. I totally see that crawling from hebrew column names is tricky. Our usual approach is to translate the headers so we can still iterate over a list of dicts. Key-based pops helps discover bugs.

jbothma (commented):
Thanks for this! I've shared some links to our preferred ways of getting around the tricky parts of this. Sorry for the double work.

datasets/il/wmd_banned/crawler.py
atehe: Sure, it seems the lookup uses lower case so we have to add the property or metadata as lowercase.
I'm using lowercase for the property in my case

datasets/il/wmd_banned/crawler.py
atehe: ahh ok, makes sense

jbothma (commented):
Hi

Nice work - I know right to left text is tricky to work with, at least I find it so in vscode.

I've made a bunch of changes - some addressing issues noted in this review that I'll resolve in a moment, and some more with the goals of

- ensuring we separate ID numbers from passport numbers, rather than throwing ID numbers into the passport prop
- not dropping detail from the passport field - I'm not sure if having the textual detail in the passport field is an issue - I'll check
- arranging things a bit more consistently with our other crawlers, e.g.
  - ordering in the crawler like determining schema, adding id, then the rest
  - function naming
  - only assigning values we pop if it's used in more than one place or for readability, but then doing so in context

jbothma:
My changes in https://github.com/opensanctions/opensanctions/pull/1455

PR #1260, opened 2024-09-16
adding Pennsylvania Medicheck list crawler

bgmello:


datasets/us/pa/med_exclusions/crawler.py
```diff
@@ -0,0 +1,100 @@
...
+
+    if last_name and business_name:
+        link = context.make("UnknownLink")
+        link.id = context.make_id(company.id, person.id)
+        link.add("object", company)
+        link.add("subject", person)
+        context.emit(link)
+
+    context.audit_data(
+        row, ignore=["CAO", "IND_CHGD", "DTE_CHANGE_LAST", "ProviderName"]
```
jbothma, line 90: let's add an address entity with h.make_addess(..., country="us") and h.apply_address.
if cao contains "Out if State", reaplace it with "" and use the rest as state. Otherwise use it as city, and pass PA as state.

datasets/us/pa/med_exclusions/crawler.py
```diff
@@ -0,0 +1,100 @@
...
+
+        if row.get("NAM_PROVR_ALT"):
+            person.add("alias", row.pop("NAM_PROVR_ALT"))
+
+        person.add("npiCode", npi)
+        person.add("title", title)
+        person.add("topics", "debarment")
+        if license_number:
+            person.add("idNumber", license_number)
+        person.add("country", "us")
```
jbothma, line 40: also company

jbothma (commented):
Good stuff! Thank you. Just two small things. I added an address like this to another one that had a city in the data. I think thst'd help users with disambiguation.

datasets/us/pa/med_exclusions/crawler.py
```diff
@@ -16,6 +16,14 @@ def crawl_item(row: Dict[str, str], context: Context):
     npi = row.pop("IDN_NPI")
     title = row.pop("NAM_TITLE_PROVR")
 
+    if row.get("CAO").startswith("Out of State"):
+        state = row.pop("CAO").replace("Out of State ", "")
+        city = ""
+    else:
+        city = row.pop("CAO")
+        state = ""
```
jbothma, line 24: when it's not out of state, the state is pennsylvania :) I'll add it.

PR #1265, opened 2024-09-19
h-apply-date-05

nvmbrasserie:


datasets/ar/parliament/crawler.py
```diff
@@ -51,10 +49,14 @@ def crawl_person(context: Context, element) -> dict:
     person_data["term"] = _extract_text(element, ".//td[4]/text()")
```
pudo, line 49: This crawler is a bit of a mess because it creates this `person_data` intermediate object instead of parsing straight into an entity under construction. Would you perhaps be up for getting it rid of `person_data`? It's half a re-write, but we end up with a more sane crawler.

pudo (commented):
A little bit of obsessing about `context.dataset.dates.formats`, which I'd use as a last resort.

datasets/ar/parliament/crawler.py
```diff
@@ -1,142 +1,125 @@
...
+            row["diputado"] = cells[1].text_content().strip()
+            row["distrito"] = cells[2].text_content().strip()
+            row["mandato"] = cells[3].text_content().strip()
+            row["inicia_mandato"] = cells[4].text_content().strip()
+            row["finaliza_mandato"] = cells[5].text_content().strip()
+            row["bloque"] = cells[6].text_content().strip()
+        else:
+            context.log.warning(
+                "Row does not have enough cells", row_html=html.tostring(element)
+            )
```
jbothma, line 122: I think we can go a step further and replace this with h.parse_html_table now that it's in prod.

datasets/_global/iadb_sanctions/crawler.py
```diff
@@ -74,7 +72,7 @@ def crawl(context: Context):
             prop = "nationality" if schema == "Person" else "jurisdiction"
             entity.add(prop, country)
 
-        end_dates = h.parse_date(row.pop("to", None), FORMATS)
+        end_dates = h.parse_date(row.pop("to", None), context.dataset.dates.formats)
```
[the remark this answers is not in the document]
nvmbrasserie: I think it's better to use `context.dataset.dates.formats` in this case , since it makes the establishment of ```
 if is_active:
            entity.add("topics", "debarment")
``` much more intuitive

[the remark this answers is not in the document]
Comment (jbothma) [reply datasets/_global/iadb_sanctions/crawler.py:75]: specifically, the date can be in the future, so we compare to today.

File: datasets/_global/iadb_sanctions/crawler.py [hunk 4]
```diff
@@ -74,16 +72,12 @@ def crawl(context: Context):
...
         is_active = False
         if end_dates:
             end_date = end_dates[0]
             today = datetime.now().isoformat()[:10]
-            if len(end_date) == 4:
-                today = today[:4]
-                context.log.warning("Only parsed year", name=title, date=end_date)
-            is_active = today < end_date
-        else:
+            is_active = today < end_date or end_date == "Ongoing"
```
jbothma, line 80: how can the parse result end_date == "Ongoing"? do you mean to check if the unparsed value is "Ongoing"?

datasets/_global/iadb_sanctions/crawler.py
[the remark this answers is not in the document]
jbothma: in this case we have the entity at hand. Do we prefer

```python
h.apply_date(sanction, raw_end_date)
is_active = sanction.get("endDate") > [todays_date]
```
or
```
end_date = h.parse_date(raw_end_date, dataset.dates.formats)
is_active = end_date > [todays_date]
sanction.add("endDate")
```
The former feels a bit like we're shoe-horning into the helper, but it's more concise and has the benefit of more consistent date assignment across the codebase

datasets/_global/iadb_sanctions/crawler.py
jbothma: it turns out, contrary to my expectation, that the raw value is returned if no date or year could be parsed.

datasets/ar/parliament/crawler.py
```diff
@@ -1,142 +1,91 @@
...
-        person.add("alias", person_data["alias"])
+        person.id = context.make_id(name)
+        person.add("name", name)
         person.add("country", "ar")
-        h.apply_name(
-            person,
-            first_name=person_data["first_name"].strip(),
-            last_name=person_data["last_name"].strip(),
-        )
-        person.add("sourceUrl", person_data["url"])
```
jbothma, line 106: looks like we've lost sourceUrl an applying first and last name (constructing a correctly-ordered full name). Could you add that back?
https://www.opensanctions.org/statements/Q107840687/?dataset=ar_parliament

jbothma: also alias, although I'm not sure if that's worth any effort because in the case of Ignacio it's the same as his name - so either don't drop it, or only drop it if we're sure it's always the constructed name.

datasets/ar/parliament/crawler.py
```diff
@@ -1,142 +1,91 @@
...
+        else:
+            context.log.warning("No link found for this person", name=name)
+            return
 
+        # Create position and categorize
         position = h.make_position(
             context,
             name="Member of National Congress of Argentina",
             country="ar",
+            summary=mandate,
```
jbothma, line 70: please remove

datasets/ar/parliament/crawler.py
jbothma: after a rewrite like this, I would check each field individually to see if I've changed any behaviour. By comparing the data and/or by going through each person.add in the old and the new code to see that all the ones we want to keep are still there.

jbothma (commented):
other than these, it looks good to go - shout when you've replaced parse_date with extract_date!

datasets/ar/parliament/crawler.py
nvmbrasserie: I don't think original source has aliases, that's why I dropped them

PR #1270, opened 2024-09-20
ru-egrul-yaml-mapping

nvmbrasserie:


nvmbrasserie:
sources: https://www.prlib.ru/spisok-sokrashcheniy-v-naimenovaniyah-uchrezhdeniy

jbothma:
next steps here:

We need to be able to find these strings in entity names and replace them with the canonical abbreviations.

One way to do that is to build up a regular expression per canonical abbreviation, where the regex matches e.g.
`(МУНИЦИПАЛЬНОЕ БЮДЖЕТНОЕ ОБЩЕОБРАЗОВАТЕЛЬНОЕ УЧРЕЖДЕНИЕ|МУНИЦИПАЛЬНОЕ ОБЩЕОБРАЗОВАТЕЛЬНОЕ БЮДЖЕТНОЕ УЧРЕЖДЕНИЕ)` and replaces it with `МБОУ`

We can build up a list of tuples [(canonical, regex)] and for each name, loop over those.

To build up the regex, we should take care to escape regex-characters in the candidate strings. See https://docs.python.org/3/library/re.html#re.escape

datasets/ru/egrul/crawler.py
```diff
@@ -446,6 +447,52 @@ def parse_address(context: Context, entity: Entity, el: Element) -> None:
...
+
+def substitute_abbreviations(
+    name: str, abbreviations: List[Tuple[str, re.Pattern, List[str]]]
+) -> Tuple[str, Optional[str]]:
+    """
+    Substitute abbreviations in the name using the compiled regex patterns.
+    :param name: The input name where abbreviations should be substituted.
+    :param abbreviations: A list of tuples with canonical abbreviations, regex patterns,
+                          and original phrases.
+    :return: The modified text with substitutions applied.
```
jbothma, line 480:
```suggestion
    :return: Tuple with the modified text with substitutions applied, and the original if shortened, otherwise None.
```

datasets/ru/egrul/crawler.py
```diff
@@ -446,6 +447,52 @@ def parse_address(context: Context, entity: Entity, el: Element) -> None:
...
+def compile_abbreviations(context) -> List[Tuple[str, re.Pattern, List[str]]]:
+    """
+    Load abbreviations and compile regex patterns from the YAML config.
+    Returns:
+    List[Tuple[str, re.Pattern, List[str]]]: A list of tuples containing canonical abbreviations
+    and their compiled regex patterns for substitution.
+    """
+    yaml = context.dataset.config.get("types")
+    abbreviations = []
+    for canonical, phrases in yaml.items():
```
jbothma, line 459: I think inside the yaml, we wanna be more specific - what kind of types is it?
Inside the crawler, we can't call it yaml, because it isn't yaml anymore. It's now a Python Dictionary. But instead of calling it `dictionary` or `data`, let's call it what it is - types. Since this function is only about one type of type, we can be brief and call it `type`. In contrast, there are all kinda of "types" in different yaml files - that's why we want to be more specific there.

```suggestion
    types = context.dataset.config.get("company_types")
    abbreviations = []
    for canonical, phrases in yaml.items():
```

datasets/ru/egrul/crawler.py
```diff
@@ -446,6 +447,52 @@ def parse_address(context: Context, entity: Entity, el: Element) -> None:
...
+            compiled_pattern = re.compile(phrase_pattern, re.IGNORECASE)
+            # Append the canonical form, compiled regex pattern, and sorted phrases to the list
+            abbreviations.append((canonical, compiled_pattern, phrase))
+
+    abbreviations.sort(key=lambda x: len(x[2]), reverse=True)
+    return abbreviations
+
+
+def substitute_abbreviations(
+    name: str, abbreviations: List[Tuple[str, re.Pattern, List[str]]]
```
jbothma, line 473: we could declare a [type alias](https://docs.python.org/3/library/typing.html#type-aliases) AbbreviationList and then use that in the type hints to the functions where we compile and use them

datasets/ru/egrul/crawler.py
```diff
@@ -446,6 +447,52 @@ def parse_address(context: Context, entity: Entity, el: Element) -> None:
...
+    """
+    yaml = context.dataset.config.get("types")
+    abbreviations = []
+    for canonical, phrases in yaml.items():
+        # we want to match the whole word and allow for ["] or ['] at the beginning
+        for phrase in phrases:
+            phrase_pattern = rf"^[ \"']?{re.escape(phrase)}"
+
+            compiled_pattern = re.compile(phrase_pattern, re.IGNORECASE)
+            # Append the canonical form, compiled regex pattern, and sorted phrases to the list
```
jbothma, line 465:
```suggestion
            # Append the canonical form, compiled regex pattern, and phrase to the list
```

datasets/ru/egrul/crawler.py
```diff
@@ -446,6 +447,52 @@ def parse_address(context: Context, entity: Entity, el: Element) -> None:
...
+            abbreviations.append((canonical, compiled_pattern, phrase))
+
+    abbreviations.sort(key=lambda x: len(x[2]), reverse=True)
```
jbothma, line 468:
```suggestion
    # Reverse-sort by length so that...
    abbreviations.sort(key=lambda x: len(x[2]), reverse=True)
```

datasets/ru/egrul/crawler.py
```diff
@@ -446,6 +447,52 @@ def parse_address(context: Context, entity: Entity, el: Element) -> None:
...
+    """
+
+    # Iterate over all abbreviation groups
+    for canonical, pattern, phrases in abbreviations:
+        match = pattern.search(name)
+        if match:
+            # Check if this is a better match than previous matches (longer phrase)
+            matched_phrase = match.group(0)
+            modified_name = name.replace(matched_phrase, canonical)
+            return modified_name, name
```
jbothma, line 490: if might be more efficient (only match once per regex) to do something like

```python
modified_name = regex.sub(canonical, name)
if modified_name != name:
    return modified_name, name
```

datasets/ru/egrul/crawler.py
```diff
@@ -446,6 +447,52 @@ def parse_address(context: Context, entity: Entity, el: Element) -> None:
...
+    """
+    Substitute abbreviations in the name using the compiled regex patterns.
+    :param name: The input name where abbreviations should be substituted.
+    :param abbreviations: A list of tuples with canonical abbreviations, regex patterns,
+                          and original phrases.
+    :return: The modified text with substitutions applied.
+    """
+
+    # Iterate over all abbreviation groups
+    for canonical, pattern, phrases in abbreviations:
```
jbothma, line 484: I'd call this `regex` rather than `pattern` since a "pattern" is what gets compiled

```suggestion
    for canonical, regex, phrases in abbreviations:
```

jbothma: total nitpick, but FYI

datasets/ru/egrul/crawler.py
```diff
@@ -461,20 +508,30 @@ def parse_company(context: Context, el: Element) -> None:
     name_full: Optional[str] = None
     name_short: Optional[str] = None
 
+    # Load abbreviations once using the context
+    abbreviations = compile_abbreviations(context)
```
jbothma, line 512: let's do this once in crawl() and pass it in as an argument.

datasets/ru/egrul/crawler.py
```diff
@@ -461,20 +508,30 @@ def parse_company(context: Context, el: Element) -> None:
...
-    entity.add("name", name_full)
+    entity.add("name", name_full_short)
     entity.add("name", name_short)
     entity.add("ogrnCode", ogrn)
     entity.add("innCode", inn)
     entity.add("kppCode", el.get("КПП"))
     entity.add("legalForm", el.get("ПолнНаимОПФ"))
     entity.add("incorporationDate", el.get("ДатаОГРН"))
+    if name_full:
+        entity.add("description", initial_name)
```
jbothma, line 534: another naming nitpick

datasets/ru/egrul/crawler.py
```diff
@@ -605,14 +676,8 @@ def parse_examples(context: Context):
...
-        "7709383684",
-        "7704667322",
-        "9710075695",
-        "7813654884",
-        "1122031001454",
-        "1025002029580",
-        "1131001011283",
-        "1088601000047",
+        "7714034350",
+        "7729348110",
```
jbothma, line 680: Let's keep the old ones around, but add the new ones - perhaps with a commend beside the new ones to indicate what's interesting about them

jbothma (commented):
Very nice!

Looks like it's time to productionise this.

- let's compile the regexes only once, in crawl(), and pass in as an argument - that's a relatively compute-intensive task so we don't want to do it for each entity in a massive dataset
- there might be a few more cases where we need to shorten names, e.g. public bodies

jbothma:
Looks good! Giving this a thorough run-through this weekend then we ship it unless it looks off 🚢

PR #1275, opened 2024-09-23
Iran Business Registry

nvmbrasserie:
opensanctions/crawler-planning#256

### DIFF
--- a/datasets/ir/business_registry/crawler.py
+++ b/datasets/ir/business_registry/crawler.py
@@ -48,12 +48,14 @@ def crawl(context: Context):
             entity.add("country", nationality)
             entity.add("sourceUrl", company_link)
             entity.add("description", stock_symbol)
-            if is_withdrawn:
+            if is_withdrawn is True:
                 entity.add("topics", "export.control")
+                target = False
             else:
                 entity.add("topics", "export.risk")
+                target = True
 
-            context.emit(entity, target=not is_withdrawn)
+            context.emit(entity, target=target)
             context.audit_data(str_row)
 
         page_number += 1
--- a/datasets/ir/business_registry/crawler.py
+++ b/datasets/ir/business_registry/crawler.py
@@ -1,6 +1,12 @@
 from urllib.parse import urljoin
+from lxml import etree
 
 from zavod import Context, helpers as h
+from zavod.shed.zyte_api import fetch_html
+
+
+def unblock_validator(doc: etree._Element) -> bool:
+    return doc.find(".//div[@class='view-content']//table") is not None
 
 
 def crawl(context: Context):
@@ -13,7 +19,7 @@ def crawl(context: Context):
         context.log.info(f"Fetching URL: {url}")
 
         # Fetch the HTML and get the table
-        doc = context.fetch_html(url)
+        doc = fetch_html(context, url, unblock_validator, cache_days=3)
         table = doc.find(".//div[@class='view-content']//table")
         if table is None:
             context.log.info("No more tables found.")
--- a/datasets/ir/business_registry/crawler.py
+++ b/datasets/ir/business_registry/crawler.py
@@ -1,0 +1,30 @@
+from lxml import html
+from rigour.mime.types import HTML
+from zavod import Context, helpers as h
+
+
+def crawl(context: Context):
+    path = context.fetch_resource("source.html", context.data_url)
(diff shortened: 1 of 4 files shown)


datasets/ir/business_registry/crawler.py
```diff
@@ -0,0 +1,66 @@
...
+            company_name = str_row.pop("company_sort_descending")
+            nationality = str_row.pop("nationality")
+            stock_symbol = str_row.pop("stock_symbol")
+
+            # Create and emit an entity
+            entity = context.make("Company")
+            entity.id = context.make_id(company_name, nationality)
+            entity.add("name", company_name)
+            entity.add("country", nationality)
+            entity.add("sourceUrl", company_link)
```
jbothma, line 52: do we wanna crawl the page too? Looks like we can get a company website there. not sure how useful that is.

would need to catch and ignore errors like this page https://www.unitedagainstnucleariran.com/company/540-sports-and-entertainment

datasets/ir/business_registry/crawler.py
```diff
@@ -0,0 +1,66 @@
...
+            entity.add("description", stock_symbol)
+            if is_withdrawn:
+                entity.add("topics", "export.control")
+            else:
+                entity.add("topics", "export.risk")
+
+            context.emit(entity)
```
jbothma, line 59: We usually do this - did you discuss not doing it with pudo?

```suggestion
            context.emit(entity, target=not is_withdrawn)
```

datasets/ir/business_registry/ir_business_registry.yaml
```diff
@@ -0,0 +1,63 @@
...
+publisher:
+  name: United Against Nuclear Ira
+  acronym: UANI
+  official: false
+  description: |
+    United Against Nuclear Iran (“UANI”) is a nonprofit and non-partisan policy organization formed 
+    to combat the threats posed by the Islamic Republic of Iran. UANI's diverse Advisory Board includes 
+    policy experts and distinguished former government officials.
+    [Source](https://www.unitedagainstnucleariran.com/about)
+  country: ir
```
jbothma, line 30:
```suggestion
  country: us
```
image

datasets/ir/business_registry/crawler.py
```diff
@@ -0,0 +1,63 @@
...
+            entity.add("name", company_name)
+            entity.add("country", nationality)
+            entity.add("sourceUrl", company_link)
+            entity.add("description", stock_symbol)
+            if is_withdrawn:
+                entity.add("topics", "export.control")
+            else:
+                entity.add("topics", "export.risk")
+
+            context.emit(entity, target=not is_withdrawn)
```
jbothma, line 56: are you sure is_withdrawn is working correctly?
now none of the entities are targets, so I think you need a bit more logic to set is_withdrawn properly. 

Remember all lxml elements are Falsy, so testing for non-presence has to be is None. In this case you don't want to check if the cell exists, you want to check the cell contents.

datasets/ir/business_registry/crawler.py
```diff
@@ -0,0 +1,63 @@
...
+            context.audit_data(str_row)
+
+        page_number += 1
+        pages_processed += 1
```
jbothma, line 60: do we need both of these or can we drop one and use it both for the URL and the assertion?

datasets/ir/business_registry/crawler.py
pudo: I think crawling the subpages for a description (notes) and website url could be good!

datasets/ir/business_registry/crawler.py
```diff
@@ -0,0 +1,156 @@
...
+from zavod.shed.zyte_api import fetch_html
+
+
+def unblock_validator(doc: etree._Element) -> bool:
+    return doc.find(".//div[@class='o-grid']") is not None
+
+
+def extract_text(doc, xpath_query):
+    result = doc.xpath(xpath_query)
+    return result[0].strip() if result else None
```
jbothma, line 14: whenever you get a list of things, be careful not to assume it will have exactly one element, or never more than one. In this case, affiliates can have more than one span. I like to just stick an `assert len(thing) == 0` in there when it would be a surprise if there are more or less elements. That sometimes helps me catch invalid assumptions.

PR #1278, opened 2024-09-24
adding Wyoming Medicare Exclusions crawler

bgmello:


bgmello:
There are some alias that they are putting in the business name but I couldn't find a way to address that automatically because it doesn't follow a pattern (sometimes they use a.k.a, aka, aka: etc). Should I do something manual in this case?

datasets/us/wy/med_exclusions/crawler.py
```diff
@@ -0,0 +1,75 @@
...
+        person.add("topics", "debarment")
+        person.add("country", "us")
+
+        sanction = h.make_sanction(context, person)
+        h.apply_date(sanction, "startDate", exclusion_date)
+
+        context.emit(person, target=True)
+        context.emit(sanction)
+
+    if business_name := row.pop("business_name"):
```
jbothma, line 41: it looks like if the "business name" matches `\ba\.?k\.?a[\. -]*` we can strip the matching txt, split on comma, and add as additional names of Persons

datasets/us/wy/med_exclusions/us_wy_med_exclusions.yaml
```diff
@@ -0,0 +1,40 @@
...
+  Medical providers deemed ineligible to participate in Wyoming State's Medicaid program.
+description: |
+  The Wyoming Department of Health maintains a list of providers that have been excluded from
+  the state Medicaid program. Providers listed are not allowed to receive reimbursement for
+  Medicaid services in any capacity.
+publisher:
+  name: Wyoming Department of Health
+  description: >
+    > Wyoming Medicaid is a joint federal and state government program that pays for medical
+    > care for some low-income and medically needy individuals and families.
```
jbothma, line 18: let's try and make the description more about Wyoming Department of Health https://health.wyo.gov/admin/about/

datasets/us/wy/med_exclusions/us_wy_med_exclusions.yaml
```diff
@@ -0,0 +1,40 @@
...
+  acronym: WDH
+  url: https://health.wyo.gov
+  official: true
+  country: "us"
+ci_test: false
+url: "https://health.wyo.gov/healthcarefin/medicaid/"
+data:
+  url: https://health.wyo.gov/wp-content/uploads/2021/07/2021_07_07_Wyoming-Medicaid-Exclusion-List-PDF_AP.pdf
```
jbothma, line 26: it looks like we can crawl up to date links from this page
image

datasets/us/wy/med_exclusions/crawler.py
```diff
@@ -0,0 +1,75 @@
...
+ a JSON list (`providers`) in which each object represents an medical provider.
+ Each object should have the following fields: `last_name`, `first_name`,
+ `business_name`, `provider_type`, `provider_number`, `city`, `state`, `exclusion_date`.
+ Return an empty string for unset fields.
+"""
+
+
+def crawl_item(row: Dict[str, str], context: Context):
+
+    address = h.make_address(context, city=row.pop("city"), state=row.pop("state"))
```
jbothma, line 20: I think if you pass country_code=US as an argument to this, the "full" address will be {city}, {state} which is a bit nicer

datasets/us/wy/med_exclusions/crawler.py
```diff
@@ -0,0 +1,75 @@
...
+
+        context.emit(link)
+
+    context.audit_data(row)
+
+
+def crawl(context: Context) -> None:
+    path = context.fetch_resource("source.pdf", context.data_url)
+    context.export_resource(path, PDF, title=context.SOURCE_TITLE)
+    for page_path in h.make_pdf_page_images(path)[1:]:
```
jbothma, line 71: why are we skipping the first page? There seems to be data there which we'd like to include.

jbothma (changes requested):
very nice, thank you

datasets/us/wy/med_exclusions/crawler.py
bgmello: Sorry, that was because I used the Alaska crawler as template and there we needed to skip the first page.

jbothma:
gpt is skipping this row
image

and merging the alias in the business name field here
```python
{'last_name': 'Berry', 'first_name': 'Charles Wayne aka - Wayne Berry', 'business_name': '', ...}
```
image

jbothma:
@bgmello did the alias in company name code work for you?
For me it doesn't work because GPT is joining the alias with the first name and outputting it in the name field so I end up with
image

jbothma:
I'm interested in whether we're seeing varying output from gpt

bgmello:
It's not working for me either. It looks like it only merged the columns in this row you showed

PR #1285, opened 2024-09-26
Develop a better toolkit for Wikidata crawling

pudo:
This tries to improve the way we handle countries, and later also positions.

### DIFF
--- a/datasets/_wikidata/categories/crawler.py
+++ b/datasets/_wikidata/categories/crawler.py
@@ -1,15 +1,17 @@
 import csv
+from functools import cache
 from io import StringIO
 from datetime import timedelta
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Set
 from urllib.parse import urlencode
 from nomenklatura.enrich.wikidata import WikidataEnricher
 from nomenklatura.enrich.wikidata.model import Item
 
 from zavod import Context, Entity
 from zavod import helpers as h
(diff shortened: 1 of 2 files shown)


zavod/zavod/shed/wikidata/position.py
```diff
@@ -0,0 +1,72 @@
...
+    else:
+        for label in item.labels:
+            label.apply(position, "name")
+
+    countries = item_countries(enricher, item)
+    for country in countries:
+        country.apply(position, "country")
+
+    if need_country and not position.has("country"):
+        return None
```
jbothma, line 61: which positions don't need a country? can't we just always fail here? Or would we use need_country=False later if we have logic to then add a country?

zavod/zavod/shed/wikidata/position.py
```diff
@@ -0,0 +1,72 @@
...
+
+    for sub_type, topics in SUB_TYPES.items():
+        if sub_type in types:
+            position.add("topics", topics)
+
+    is_pep = True if "role.pep" in position.get("topics") else None
```
jbothma, line 67: we don't currently put role.pep in Positions - normally only on the Person. 

Why do SUB_TYPES not have role.pep? Can't we assume any SUBTYPE match is a PEP position?

jbothma (commented):
This is much more grokable than I expected from wikidata traversal

PR #1286, opened 2024-09-26
adding West Virginia Medicare Exclusions crawler

bgmello:


datasets/us/wv/med_exclusions/crawler.py
```diff
@@ -0,0 +1,80 @@
...
+        "https://www.wvmmis.com/WV%20Medicaid%20Provider%20SanctionedExclusion/"
+        + info_txt[15 : info_txt.find(",") - 1]
+    )
+
+
+def crawl(context: Context) -> None:
+    path = context.fetch_resource("source.pdf", crawl_pdf_url(context))
+    context.export_resource(path, PDF, title=context.SOURCE_TITLE)
+
+    for page_path in h.make_pdf_page_images(path)[1:]:
```
jbothma, line 76:
```suggestion
    for page_path in h.make_pdf_page_images(path):
```

The way I discovered this is by checking if the first and last entry in the original dataset is in the output. This is a handy approach for ETL testing generally. Reading the code should also have caught it

datasets/us/wv/med_exclusions/crawler.py
```diff
@@ -0,0 +1,80 @@
...
+        data = run_image_prompt(context, prompt, page_path)
+        assert "providers" in data, data
+        for item in data.get("providers", []):
```
jbothma, line 79:
```suggestion
        for item in data.get("providers"):
```

datasets/us/wv/med_exclusions/crawler.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+        h.apply_name(
+            entity, first_name=row.pop("first_name"), last_name=row.pop("last_name")
+        )
+
+    if npi := row.pop("npi"):
+        entity.add("npiCode", npi)
+    entity.add("topics", "debarment")
+    entity.add("country", "us")
+    h.apply_address(context, entity, address)
```
jbothma, line 49: when we have addresses that include streets, let's also do h.copy_address

jbothma (commented):
I'll apply these changes and merge

PR #1294, opened 2024-09-27
h-apply-date-10

nvmbrasserie:


datasets/ru/acf_bribetakers/crawler.py
```diff
@@ -51,7 +42,9 @@ def parse_result(context: Context, row: Dict[str, Any]):
     for tl in transliterations.split("\n"):
         tl = h.remove_bracketed(tl)
         entity.add("alias", tl)
-    entity.add("birthDate", parse_date(dob), original_value=dob)
+    if dob and NO_YEAR.match(dob):
+        dob = None
+    entity.add("birthDate", h.extract_date(context.dataset, dob), original_value=dob)
```
nvmbrasserie, line 47: I decided to keep `entity.add` instead of `h.apply_date` not to lose the `original_value`

datasets/us/state_dept/crawler.py
```diff
@@ -82,8 +82,9 @@ def crawl_bio_page(context: Context, url: str):
...
     if end_date == "Present":
         end_date = None
-    start_date = h.parse_date(start_date, DATE_FORMAT)[0]
-    end_date = h.parse_date(end_date, DATE_FORMAT)[0] if end_date else None
+    start_date = h.extract_date(context.dataset, start_date)[0]
+
+    end_date = h.extract_date(context.dataset, start_date)[0] if end_date else None
 
     occupancy = h.make_occupancy(
         context,
```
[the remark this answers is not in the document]
pudo: @nvmbrasserie This can now be cleaned up - we can feed uncleaned values into `make_occupancy`.

pudo (changes requested):

datasets/ru/acf_bribetakers/crawler.py
```diff
@@ -51,7 +42,9 @@ def parse_result(context: Context, row: Dict[str, Any]):
     for tl in transliterations.split("\n"):
         tl = h.remove_bracketed(tl)
         entity.add("alias", tl)
-    entity.add("birthDate", parse_date(dob), original_value=dob)
+    if dob and NO_YEAR.match(dob):
```
pudo, line 45: this needs to be `not NO_YEAR.match`

nvmbrasserie: it matches values like `value=17.07.`, so I think it should be as it is

PR #1306, opened 2024-09-30
adding Maryland medicare exclusions crawler

bgmello:


jbothma:
Looks like there's an excel file that's more up to date than the PDF - any reason we can't use that?
image

datasets/us/md/med_exclusions/crawler.py
```diff
@@ -0,0 +1,76 @@
...
+    if row.get("address") or row.get("city_state_zip"):
+        address = h.make_address(
+            context,
+            (row.pop("address") or "") + " " + (row.pop("city_state_zip") or ""),
+        )
+        entity.add("address", address)
+
+    sanction = h.make_sanction(context, entity)
+    h.apply_date(sanction, "startDate", row.pop("termination_sanction_date"))
+    sanction.add("description", "Sanction Type: " + row.pop("sanction_type"))
```
jbothma, line 50: perhaps defaulting to the code if the label isn't mapped to support new/misspelt codes

datasets/us/md/med_exclusions/us_md_med_exclusions.yaml
```diff
@@ -0,0 +1,37 @@
...
+coverage:
+  frequency: daily
+  start: "2024-09-30"
+load_db_uri: ${OPENSANCTIONS_DATABASE_URI}
+summary: >
+  Medical providers deemed ineligible to participate in Maryland's Medicaid program.
+description: |
+  Maryland Medicaid suspends or excludes some providers from working with the Medicaid
+  program. Providers listed are not allowed to receive reimbursement for Medicaid
+  services in any capacity.
```
jbothma, line 13: let's add a paragraph 

> For sanction type LB on the Sanctioned Providers list, verify the provid​er’s license status through the [Maryland State Board](https://health.maryland.gov/Pages/boards.ASPX).

datasets/us/md/med_exclusions/us_md_med_exclusions.yaml
```diff
@@ -0,0 +1,37 @@
...
+dates:
+  formats: ["%m/%d/%Y"]
+
+assertions:
+  min:
+    schema_entities:
+      LegalEntity: 7500
+  max:
+    schema_entities:
+      LegalEntity: 9000
```
jbothma, line 37: This isn't currently validating. I spot 1567 entities in the dataset, although there are repetitions, and the stats will refer to the more specific schemata used - just using LegalEntity doesn't work out of the box here.

datasets/us/md/med_exclusions/crawler.py
```diff
@@ -0,0 +1,76 @@
...
+
+    if row.get("license_no"):
+        entity.add("description", "License No: " + row.pop("license_no"))
+
+    if row.get("address") or row.get("city_state_zip"):
+        address = h.make_address(
+            context,
+            (row.pop("address") or "") + " " + (row.pop("city_state_zip") or ""),
+        )
+        entity.add("address", address)
```
jbothma, line 46: watch out, `Address` entities must be added to the `addressEntity` property. Address `string`s are added to the `address` property.

So for address strings like this, merely adding to `address` prop is good.

We use `Address` entities when we have more granular address data like a separate zip code and/or state. But parsing addresses like this into their parts is out of scope at the moment.

PR #1307, opened 2024-10-01
fixes-1-oct

nvmbrasserie:


datasets/_wikidata/categories/wd_categories.yml
```diff
@@ -43,12 +43,18 @@ lookups:
           - member states of the United Nations
           - member of the United Nations Security Council
         value: zz
-      - match: Pakistan Occupied Kashmir
+      - match: 
+          - Pakistan Occupied Kashmir
+          - Kashmir
```
jbothma, line 48: let's leave this as a warning for the moment until a new ftm is out with a placeholder country code for this disputed territory, rather than defaulting to one of the claimants

datasets/_wikidata/categories/wd_categories.yml
```diff
@@ -43,12 +43,18 @@ lookups:
...
         value: pk
       - match:
           - Galicia
           - Ceuta
         value: es
+      - match: Romanov Empire
+        value: ru
+      - match: Persian Empire
+        value: ir
```
jbothma, line 57: I almost think it's better to let ancient countries be null in this dataset, rather than mapping them to current countries. It would have the effect of them still exiting in the dataset, but at least not being listed under Iran in search or https://www.opensanctions.org/countries/ir/

jbothma (changes requested):
Thanks for these.

PR #1308, opened 2024-10-01
h-apply-date-12

nvmbrasserie:


datasets/qa/nctc_sanctions/crawler.py
```diff
@@ -63,12 +60,13 @@ def crawl(context: Context):
...
         dob_format = item.pop("dobFormat")
-        dobs = parse_dates(dob_format)
+        dobs = parse_dates(context, dob_format)
         if entity.schema.is_a("Person"):
             entity.add("passportNumber", item.pop("passportNo"))
             entity.add("idNumber", item.pop("qid"))
             entity.add("nationality", item.pop("nationality"))
-            entity.add("birthDate", dobs, original_value=dob_format)
+            h.apply_dates(entity, "birthDate", dobs)
+            # entity.add("birthDate", dobs, original_value=dob_format)
```
jbothma, line 69:
```suggestion
```

datasets/qa/nctc_sanctions/crawler.py
```diff
@@ -1,37 +1,34 @@
...
     if text == "X_X_X_X":
         return []
-    for part in h.multi_split(text, [":", ";", "؛", " to "]):
+    for part in h.multi_split(
+        text,
+        [":", ";", "؛", " to ", "حوالي", "_", "والي"],
+    ):
         part = part.replace("___", " ")
         part = part.replace("_X_X_X", " ")
+        part = part.replace("X_X_", " ")
```
jbothma, line 25: if we split on _ we'll never see these replacements. Should we rethink the split or the replacements?

datasets/qa/nctc_sanctions/crawler.py
```diff
@@ -1,37 +1,31 @@
...
             continue
-        parsed = parse_formats(part, FORMATS)
-        if parsed.text is not None:
-            dates.add(parsed.text)
-        else:
-            dates.update(h.extract_years(part))
-    # if not len(dates):
-    #     print(text)
-    # return [text]
+        parsed_dates = h.extract_date(context.dataset, part)
```
pudo, line 26: Just as an idea for a pattern, this function could also return a list of unparsed dates that have been split and pre-cleaned, and then send those into `h.apply_date`.

PR #1312, opened 2024-10-01
h-apply-date-14

nvmbrasserie:
- il_mod_terrorists - MODs: more complete date parsing


datasets/hr/public_officials/crawler.py
```diff
@@ -72,8 +70,8 @@ def make_affiliation_entities(
         no_end_implies_current=True,
         categorisation=categorisation,
         propagate_country=True,
-        start_date=h.parse_date(start_date, DATE_FORMATS).pop(),
-        end_date=h.parse_date(end_date, DATE_FORMATS).pop(),
+        start_date=h.extract_date(context.dataset, start_date).pop(),
+        end_date=h.extract_date(context.dataset, end_date).pop(),
```
jbothma, line 74: might be solved if we move extract_date into make_occupancy

datasets/il/mod_terrorists/crawler.py
```diff
@@ -9,28 +9,27 @@
...
 
 
-def parse_date(date):
+def parse_date(context: Context, date):
     dates = []
-    for part in h.multi_split(date, ["OR", ";", " - "]):
-        dates.extend(h.parse_date(part, FORMATS))
+    for part in h.multi_split(date, ["OR", ";", " - ", "a) ", "b) ", "c) "]):
+        dates.extend(h.extract_date(context.dataset, part))
     return dates
```
jbothma, line 21: do we need this to extract? are we using the dates anywhere in a way they need to be parsed, or can we just split and apply in this crawler?

datasets/il/mod_terrorists/il_mod_terrorists.yml
```diff
@@ -115,6 +121,20 @@ lookups:
         value: null
       - match: 16/6/67
         value: "1967-06-16"
+      - match: 23 Apr. 1968
+        value: "1968-04-23"
+      - match: 25 Jun. 1964
+        value: "1964-06-25"
+      - match: 5 Dec. 1969
```
jbothma, line 128: extract_years was probably hiding these. How about instead adding a format string for them?

datasets/kz/afmrk_sanctions/crawler.py
```diff
@@ -6,7 +6,7 @@
 
 from urllib.parse import urljoin
 
-FORMATS = ["%d.%m.%Y"]
+# FORMATS = ["%d.%m.%Y"]
```
jbothma, line 9:
```suggestion
```

datasets/kz/afmrk_sanctions/crawler.py
```diff
@@ -32,8 +32,8 @@ def make_entity(context: Context, el, schema, entity_id, topics, program):
     sanction = h.make_sanction(context, entity)
     sanction.add("summary", el.findtext("./correction"))
     sanction.add("program", program)
-    listingDate = h.parse_date(el.findtext("./added_to_list"), FORMATS)
-    note_date = added_date_from_note(el.findtext("./correction"))
+    listingDate = h.extract_date(context.dataset, el.findtext("./added_to_list"))
+    note_date = added_date_from_note(context, el.findtext("./correction"))
     sanction.add("listingDate", listingDate or note_date)
```
jbothma, line 37: are we relying on the parsing to make listingDate falsy or is the unparsed value already falsy?

line 37 can already be apply, and depending on the above, we might also be able to skip the extract_date on the new line 35.

datasets/hr/public_officials/crawler.py
nvmbrasserie: what do you mean? it's already inside `make.occupancy`, isn't it?

jbothma: we've made this change now.
in this comment's selection, we're calling extract_date _in the call to_ make_occupany. After the changes now in this PR, we're calling extract_date (via apply_date) _within_ make_occupancy

PR #1316, opened 2024-10-02
h-apply-date-17

nvmbrasserie:


datasets/us/trade_csl/crawler.py
```diff
@@ -9,18 +9,17 @@
...
-def parse_date(text: Optional[str]) -> List[str]:
+def parse_date(context: Context, text: Optional[str]) -> List[str]:
     if text is None or not len(text):
         return []
     text = text.replace("circa", "")
     text = text.strip()
     dates: List[str] = []
     for part in h.multi_split(text, [" to "]):
-        dates.extend(h.parse_date(part, FORMATS))
+        dates.extend(h.extract_date(context.dataset, part))
```
jbothma, line 22: any reason why we can't skip the extract here and just make line 130 apply? If we do that, perhaps it's more appropriate to rename parse_date to split_dates or something, right

PR #1319, opened 2024-10-02
h-apply-date-18

nvmbrasserie:
- md_terror_sanctions: 5 persons getting re-keyed because of improved birth date splitting
- ua_sfms website is down, can't check it yet.

### DIFF
--- a/datasets/md/terror_sanctions/crawler.py
+++ b/datasets/md/terror_sanctions/crawler.py
@@ -24,10 +24,10 @@ SPLITS = [
 ]
 
 
-def parse_birth_dates(context: Context, string: str) -> List[str]:
-    strings = h.multi_split(string, SPLITS)
-    # flatten
-    return [date for s in strings for date in h.extract_date(context.dataset, s)]
+# def parse_birth_dates(context: Context, string: str) -> List[str]:
+#     strings = h.multi_split(string, SPLITS)
+#     # flatten
+#     return [date for s in strings for date in h.extract_date(context.dataset, s)]
 
 
 def clean_name(string: str):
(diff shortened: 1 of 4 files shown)


datasets/ua/sfms_blacklist/crawler.py
```diff
@@ -88,7 +98,7 @@ def parse_entry(context: Context, entry: ElementOrTree) -> None:
         entity.add_cast("Person", "birthPlace", pob.text)
 
     for dob in entry.findall("./date-of-birth-list"):
-        date = parse_date(dob.text)
+        date = parse_date(context, dob.text)
```
pudo, line 101: this can be apply_date if we split it into two calls:

```
if dob.text is not None:
   entity.add_schema('Person')
   h.apply_date(...)
```

pudo (changes requested):

datasets/md/terror_sanctions/crawler.py
```diff
@@ -6,15 +6,28 @@
...
+    "din pasaport fals",
+    "presupusă:",
+]
 
-FORMATS = ["%d.%m.%Y"]
-SPLITS = ["si", ";", "sau", "a)", "b)", "c)"]
 
-
-def parse_birth_dates(string: str) -> List[str]:
+def parse_birth_dates(context: Context, string: str) -> List[str]:
```
pudo, line 27: Similar as with -12, I wonder if this should return unparsed dates and feed them into `h.apply_date`.

datasets/ua/sfms_blacklist/ua_sfms_blacklist.yml
```diff
@@ -88,43 +90,52 @@ lookups:
         value: Iraq
   type.date:
```
pudo, line 91: Nice cleanup here! :)

pudo (approved):

PR #1320, opened 2024-10-02
h-apply-date-19

nvmbrasserie:


datasets/ro/onpcsb_sanctions/crawler.py
```diff
@@ -41,12 +36,11 @@ def crawl_row(context: Context, row: Dict[str, str]):
 
     if entity_type == "Person":
         entity = context.make("Person")
-        entity.id = context.make_id(full_name, birth_date_1, birth_place)
```
pudo, line 44: Wait this will re-key the entire dataset and make keys that are less unique. anyway we can just put the raw value of `data of birth` from the `row` in there?

datasets/ro/onpcsb_sanctions/crawler.py
```diff
@@ -44,9 +40,8 @@ def crawl_row(context: Context, row: Dict[str, str]):
         entity.id = context.make_id(full_name, birth_date_1, birth_place)
         entity.add("name", full_name)
         entity.add("alias", other_name)
-        entity.add("birthDate", birth_date_1)
-        if birth_date_2:
-            entity.add("birthDate", birth_date_2)
+        h.apply_date(entity, "birthDate", row.pop("date of birth", None))
```
jbothma, line 43: this is now always going to be None because we've already popped it.

datasets/ro/onpcsb_sanctions/crawler.py
jbothma: we generated the ID on the parsed form, so we'll need to parse that one still.

PR #1325, opened 2024-10-04
adding New Jersey medicare exclsions crawler

bgmello:


datasets/us/nj/med_exclusions/crawler.py
```diff
@@ -0,0 +1,76 @@
...
+    context.emit(entity, target=is_debarred)
+    context.emit(sanction)
+
+    context.audit_data(row)
+
+
+def crawl(context: Context) -> None:
+    path = context.fetch_resource("source.pdf", context.data_url)
+    context.export_resource(path, PDF, title=context.SOURCE_TITLE)
+    for page_path in h.make_pdf_page_images(path)[1:]:
```
jbothma, line 72: the data starts on the first page

datasets/us/nj/med_exclusions/crawler.py
```diff
@@ -0,0 +1,76 @@
...
+    h.apply_date(sanction, "startDate", row.pop("effective_date"))
+    sanction.add("provisions", row.pop("action"))
+
+    if row.get("expiration_date") and row.get("expiration_date").upper() not in [
+        "PERMANENT",
+        "DECEASED",
+    ]:
+        is_debarred = (
+            datetime.strptime(row.get("expiration_date"), "%m/%d/%Y") > datetime.today()
+        )
```
jbothma, line 54: let's rather parse with h.extract_date - a new function that's actually used by apply_date so it would produce the same result.

jbothma (commented):
The PDF suggests to email for an Excel copy. This site has a TSV download https://www.nj.gov/treasury/revenue/debarment/debarsearch-medical.shtml

I've emailed to ask if it's the same, or if they have an excel download option.

datasets/us/nj/med_exclusions/crawler.py
jbothma: or better, use the approach I used on nevada.

bgmello:
I'm getting an warning about an invalid date, even though I added it to the lookup:

Screenshot 2024-10-21 at 09 54 51

PR #1327, opened 2024-10-04
date parsing within h.make_occupancy enhancement

nvmbrasserie:
- dk_pep 9 modified - birth dates now parsed completely instead of extract_years fallback
- lt_magnitsky_amendments 293 modified - adding start date to sanctions
-

### DIFF
--- a/datasets/lt/magnitsky_amendments/crawler.py
+++ b/datasets/lt/magnitsky_amendments/crawler.py
@@ -34,7 +34,6 @@ def crawl(context: Context):
     check_data_limits(data)
 
     for entry in data.get("list", []):
-
         # Create a Person entity
         person = context.make("Person")
         person_id = context.make_id(entry.get("vardas"), entry.get("pavarde"))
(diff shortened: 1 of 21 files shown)


zavod/zavod/logic/pep.py
```diff
@@ -153,7 +153,7 @@ def occupancy_status(
     categorisation: Optional[PositionCategorisation] = None,
 ) -> Optional[OccupancyStatus]:
     if death_date is not None and death_date < backdate(current_time, AFTER_DEATH):
```
jbothma, line 155: since the original string is returned if a date doesn't parse, we probably want to

1. start warning if dates don't parse
2. and eventually parse and compare them as prefixdates instead of ISO strings.

Currently we can be comparing a date string to an unsuccessfully-parsed date

jbothma: doesn't have to be in this change

datasets/lt/magnitsky_amendments/crawler.py
```diff
@@ -62,8 +61,9 @@ def crawl(context: Context):
         # Create a Sanction entity
         sanction = h.make_sanction(context, person)
         sanction.add("reason", entry.get("priezastis"), lang="lit")
-        # sanction.add("startDate", h.parse_date("2018-01-01", DATE_FORMATS))
-        sanction.add("endDate", h.parse_date(entry.get("uzdraustaIki"), DATE_FORMATS))
+        for date in h.multi_split(entry.get("priezastis"), ["galiojanti nuo "]):
+            h.apply_date(sanction, "startDate", date)
```
jbothma, line 65: is this the start date of the law or the ban? it looks like new people have recently been added, but they still get the 2018 date which makes me think it's not really right thing thing for an individual's sanction

PR #1330, opened 2024-10-04
h apply date 21

nvmbrasserie:


pudo:
@nvmbrasserie This seems to have a different solution for `id_dttot` than what is now merged. Can you have a look and pick what's better?

PR #1335, opened 2024-10-07
The Cabinet of Ministers of Thailand

nvmbrasserie:
opensanctions/crawler-planning#257

We've decided to stick with the original names for now, as some of them are being translated very poorly. @jbothma If you could take another look at the prompt, that would be great. I've tried a number of things, but none of them work well enough. If it doesn't work though, we can just merge it without translation

### DIFF
--- a/datasets/th/cabinet/crawler.py
+++ b/datasets/th/cabinet/crawler.py
@@ -8,7 +8,7 @@ from zavod.shed.zyte_api import fetch_html
 ROLE_PATTERNS = re.compile(
     r"(?P<name>.+?)\s*(?P<role>รองนายกรัฐมนตรี.*|รัฐมนตรีว่าการ.*|รัฐมนตรีประจำ.*|รัฐมนตรีช่วยว่าการ.*|นายกรัฐมนตรี)"
 )
-# Pattern specifically for "นายกรัฐมนตรี"
+# Pattern specifically for "นายกรัฐมนตรี" (Prime Minister)
 PRIME_MINISTER_PATTERN = re.compile(r"(?P<role>นายกรัฐมนตรี)\s*(?P<name>.*)")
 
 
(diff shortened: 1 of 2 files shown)


datasets/th/cabinet/crawler.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+def crawl(context: Context):
+    doc = fetch_html(
+        context, context.data_url, unblock_validator=unblock_validator, cache_days=3
+    )
+    main_container = doc.find(".//div[@class='wonderplugintabs-panel-inner']")
+
+    prime_minister = main_container.find(
+        ".//div[@style='min-height: 480; max-height: 600; display: flex; align-items: center; flex-direction: column; justify-content: center; width: 100%;']"  # to be updated
```
pudo, line 62: this selector is super specific and likely to break. is there some way to select things here independent of their CSS properties?

datasets/th/cabinet/crawler.py
```diff
@@ -0,0 +1,127 @@
...
+
+# ```json
+# {
+#   "original_name": "นายอนุทิน ชาญวีรกูล",
+#   "transliterated_name": "Anutin Charnvirakul"
+# }
+# ```
+
+# Focus on providing both an accurate English translation and common romanized form where applicable,
+# maintaining standard conventions in English nomenclature.
```
pudo, line 32: let's kick this all out

jbothma (commented):
Ran out of time to finish today, will wrap tomorrow.

datasets/th/cabinet/crawler.py
```diff
@@ -0,0 +1,92 @@
...
+            occupancy = h.make_occupancy(context, person, position)
+            if occupancy:
+                context.emit(person, target=True)
+                context.emit(position)
+                context.emit(occupancy)
+
+        # Find the date of the last update
+    last_upd = doc.find(".//h5[@style='text-align: center;']")
+    date_last_upd = last_upd.text_content().strip()
+    assert date_last_upd == "ปรับปรุงข้อมูล ณ วันที่ 4 กันยายน 2567"  # September 4, 2024
```
jbothma, line 92: why? This will stop updates to the data from taking effect without us updating the crawler - is that intentional?

jbothma: this was to detect changes since it seems like they do a fair amount of ad-hoc markup (variety of heading levels)

I've made it check for matching a title in the name, e.g. "Ms." and skip and warn if it can't do so. That, plus the same for position prefixes, should be sufficient to automatically update only as far as we can trust.

PR #1339, opened 2024-10-07
adding Washington state medicare exclusions crawler

bgmello:


pudo (approved):
This looks great and simple. Only possible nitpick would be to title-case the dataset title: 

`US Washington State Provider Termination and Exclusion List`

PR #1344, opened 2024-10-08
adding Hawaii medicare exclusion list

bgmello:
It is emitting some warnings because of inconsistent data, but it looks like it's a problem with GPT.

### DIFF
--- a/datasets/us/hi/med_exclusions/crawler.py
+++ b/datasets/us/hi/med_exclusions/crawler.py
@@ -1,19 +1,34 @@
+import pdfplumber
+from pathlib import Path
 from typing import Dict
 from rigour.mime.types import PDF
 
 from zavod import Context, helpers as h
+from normality import collapse_spaces, slugify
 
-from zavod.shed.gpt import run_image_prompt
 
-prompt = """
-Extract structured data from the following page of a PDF document. Return 
-a JSON list (`providers`) in which each object represents an medical provider.
-The name of a provider might be split onto the next row within a record.
-Each object should have the following fields: `last_name`,
-`first_name`, `middle_initial`, `medicaid_provider_id`, `provider_type`,
-`exclusion_date`, `reinstatement_date`.
-Return an empty string for unset fields.
-"""
+def parse_pdf_table(
+    context: Context, path: Path, save_debug_images=False, headers=None
+):
+    pdf = pdfplumber.open(path.as_posix())
+    settings = {}
+    for page_num, page in enumerate(pdf.pages, 1):
+        # Find the bottom of the bottom-most rectangle on the page
+        bottom = max(page.height - rect["y0"] for rect in page.rects)
+        settings["explicit_horizontal_lines"] = [bottom]
+        if save_debug_images:
+            im = page.to_image()
+            im.draw_hline(bottom, stroke=(0, 0, 255), stroke_width=1)
+            im.draw_rects(page.find_table(settings).cells)
+            im.save(f"page-{page_num}.png")
+        assert bottom < (page.height - 5), (bottom, page.height)
+
+        for row in page.extract_table(settings)[1:]:
+            if headers is None:
+                headers = [slugify(collapse_spaces(cell), sep="_") for cell in row]
+                continue
+            assert len(headers) == len(row), (headers, row)
+            yield dict(zip(headers, row))
 
 
 def crawl_item(row: Dict[str, str], context: Context):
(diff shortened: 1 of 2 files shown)


datasets/us/hi/med_exclusions/us_hi_med_exclusions.yaml
```diff
@@ -0,0 +1,43 @@
...
+    > families, including healthcare, financial assistance, and social services,
+    > with a focus on improving economic security and fostering self-sufficiency.
+  acronym: DHS
+  url: https://humanservices.hawaii.gov
+  official: true
+  country: "us"
+url: "https://medquest.hawaii.gov/en/plans-providers/provider-exclusion-reinstatement-list.html"
+data:
+  url: https://medquest.hawaii.gov/content/dam/formsanddocuments/plans-and-providers/Med%20Prov%20Excl-Rein%20List-UPDATED-09.26.2024.pdf
+  format: PDF
```
jbothma, line 30: Let's automate finding the "Med Prov Excl-Rein List" PDF link on the page

jbothma:
hmmm

image

```javascript
{'last_name': 'Bergama', 'first_name': 'Raquel', 'middle_initial': 'H', 'medicaid_provider_id': 'NONE', 'provider_type': 'Owner', 'exclusion_date': '01/2003', 'reinstatement_date': 'Indefinite'}
```

jbothma:
and with 300dpi 

```javascript
{'last_name': 'BERGMAN', 'first_name': 'RAQUEL', 'middle_initial': '', 'medicaid_provider_id': 'NONE', 'provider_type': 'OWNER', 'exclusion_date': '01/2003', 'reinstatement_date': 'Indefinite'}
```

datasets/us/hi/med_exclusions/crawler.py
```diff
@@ -0,0 +1,64 @@
...
+
+    sanction = h.make_sanction(context, entity)
+    h.apply_date(sanction, "startDate", row.pop("exclusion_date"))
+    if row.get("reinstatement_date") != "Indefinite":
+        h.apply_date(sanction, "endDate", row.pop("reinstatement_date"))
+        is_debarred = False
+    else:
+        row.pop("reinstatement_date")
+        is_debarred = True
+        entity.add("topics", "debarment")
```
jbothma, line 49: we also need to check whether reinstatement date is in the past.
See how I did Nevada.

datasets/us/hi/med_exclusions/crawler.py
```diff
@@ -0,0 +1,64 @@
...
+            last_name=row.pop("last_name"),
+            middle_name=row.pop("middle_initial"),
+        )
+    else:
+        entity = context.make("Company")
+        entity.id = context.make_id(row.get("last_name"))
+        entity.add("name", row.pop("last_name"))
+
+    entity.add("sector", row.pop("provider_type"))
+    entity.add("description", "Provider ID: " + row.pop("medicaid_provider_id"))
```
jbothma, line 38: let's skip NONE

datasets/us/hi/med_exclusions/crawler.py
```diff
@@ -0,0 +1,64 @@
...
+def crawl_item(row: Dict[str, str], context: Context):
+
+    if first_name := row.pop("first_name"):
+        entity = context.make("Person")
+        entity.id = context.make_id(
+            row.get("last_name"), first_name, row.get("exclusion_date")
+        )
+        h.apply_name(
+            entity,
+            first_name=first_name,
```
jbothma, line 28: let's use the original first name value for the id, but split the first and middle names on aka, a.k.a etc when applying names. I think you can call apply_name as many times as there are name combinations

datasets/us/hi/med_exclusions/us_hi_med_exclusions.yaml
```diff
@@ -0,0 +1,43 @@
...
+  start: "2024-10-08"
+load_db_uri: ${OPENSANCTIONS_DATABASE_URI}
+summary: >
+  Medical providers excluded from Hawaii's Medicaid program.
+description: |
+  The State of Hawai'i Department of Human Services maintains a list of providers
+  who have been excluded from the state program.
+   > An excluded provider is an individual or entity that is not allowed to receive
+   > reimbursement for providing Medicare and Medicaid services in any capacity,
+   > even if they are not on this listing by the Med-QUEST Division.
```
jbothma, line 15: Please add 

> Any provider participating or applying to participate in the Medicaid program must search Hawaiʻi’s excluded provider list monthly and the List of Excluded Individuals and Entities (LEIE) on an annual basis to determine if any existing employee or contractor has been excluded from participation in the Medicaid program. In addition, any provider participating or applying to participate in the Medicaid program must search both lists prior to hiring staff to ensure that any potential employees or contractors have not been excluded from participating in the Medicaid program.

datasets/us/hi/med_exclusions/us_hi_med_exclusions.yaml
```diff
@@ -0,0 +1,43 @@
...
+prefix: us-medhi
+coverage:
+  frequency: daily
+  start: "2024-10-08"
+load_db_uri: ${OPENSANCTIONS_DATABASE_URI}
+summary: >
+  Medical providers excluded from Hawaii's Medicaid program.
+description: |
+  The State of Hawai'i Department of Human Services maintains a list of providers
+  who have been excluded from the state program.
```
jbothma, line 12: Please add

> Excluded individuals and entities are listed according to the type of provider they were at the time of exclusion; however, they are excluded from participating in providing services in the Medicaid program in all categories of service and in any capacity.

with a blank line in between the existing and the new paragraphs so that we get separation of the paragraphs and quotes on the web page.

datasets/us/hi/med_exclusions/us_hi_med_exclusions.yaml
```diff
@@ -0,0 +1,43 @@
...
+  name: Hawai'i Department of Human Services
+  description: >
+    > The Hawaii Department of Human Services is a state agency that provides
+    > essential programs and services to support the well-being of individuals and
+    > families, including healthcare, financial assistance, and social services,
+    > with a focus on improving economic security and fostering self-sufficiency.
+  acronym: DHS
+  url: https://humanservices.hawaii.gov
+  official: true
+  country: "us"
```
jbothma, line 26: where is this quote from? I can't find it anywhere.

jbothma (commented):
here's a commit changing it to use pdfplumber to extract the data from the PDF https://github.com/opensanctions/opensanctions/commit/f27ac4a9eaae2265ff1246c966b57739af3474d0
Let's use this approach instead of gpt to finish this.

datasets/us/hi/med_exclusions/us_hi_med_exclusions.yaml
jbothma: I asked because mentioning financial assistance is a very specific thing. If it's something we've synthesised, sure, it's not a quote, but I'm still unsure where it came from, because financial assistance isn't mentioned on their about page.

Anyway I'm making it more specific to the department.

datasets/us/hi/med_exclusions/crawler.py
```diff
@@ -0,0 +1,120 @@
...
+        path,
+        headers=[
+            "last_name",
+            "first_name",
+            "middle_initial",
+            "medicaid_provider_id",
+            "provider_type",
+            "exclusion_date",
+            "reinstatement_date",
+        ],
```
jbothma, line 118: remember the risk with hardcoding header order is that if they change it, e.g. by insertiing a new column in the middle, things might become nonsense.

The fix is just to reset headers=None for each page in this case. The other PDFs have had a header row only on the first page.

jbothma (commented):
Nice work on this! I've made some changes in https://github.com/opensanctions/opensanctions/pull/1439 which could be worth checking out

PR #1372, opened 2024-10-14
h apply date 22

nvmbrasserie:


datasets/gb/hmt/crawler.py
```diff
@@ -156,7 +152,7 @@ def parse_row(context: Context, row: Dict[str, Any]):
     pobs = split_items(row.pop("Individual_TownOfBirth", None))
     entity.add_cast("Person", "birthPlace", pobs)
 
-    dob = h.parse_date(row.pop("Individual_DateOfBirth", None), FORMATS)
```
pudo, line 159: I'd split this up and pop the dob, check if it's non-empty, if so apply "Person", and then do apply_date

pudo (changes requested):

PR #1389, opened 2024-10-15
ch seco other info

nvmbrasserie:


datasets/ch/seco_sanctions/crawler.py
```diff
@@ -6,6 +6,7 @@
 from followthemoney.types import registry
 from followthemoney.util import join_text
 from lxml.etree import _Element as Element
+from slugify import slugify
```
pudo, line 9: Can you use `from normality import slugify`? It's all over the rest of our codebase.

pudo (changes requested):

datasets/ch/seco_sanctions/crawler.py
```diff
@@ -32,19 +33,66 @@
...
-REGEX_INN = re.compile(r"Taxpayer [Ii]dentification [Nn]umber ?: (\d+)\.?")
-REGEX_REGNUM = re.compile(
-    r"(ОГРН/main )?([Ss]tate |Business )?[Rr]egistration number ?: (\d+)\.?"
-)
-REGEX_TAX = re.compile(r"Tax [Rr]egistration [Nn]umber ?: (\d+)\.?")
-REGEX_IMO = re.compile(r"IMO [Nn]umber ?: (\d+)\.?")
+
+
+OTHER_INFO_REGEXES = [
+    re.compile(
```
jbothma, line 39: here's a way to trim and simplify this https://github.com/opensanctions/opensanctions/pull/1412/files#diff-f272cb5f296048636b78fa34143f7a7c7d76efa85f316c2013a28538ee2a0470R19

jbothma:
This is paused while we consider alternatives. We probably don't want to carry on down this path.

We didn't discuss guidelines for how many items justify a regex, so we might have gone overboard there, just because it wasn't clear what the right "gut feel" was.

pudo:
Ok, it's time: I'm closing this in favor of `zavod-ui`. Thanks for doing all the work on it, @nvmbrasserie.

PR #1392, opened 2024-10-15
h-apply-date-28

nvmbrasserie:


datasets/us/fincen_special_measures/crawler.py
```diff
@@ -97,7 +95,7 @@ def parse_table(table: html.HtmlElement) -> Generator[Dict[str, str], None, None
 
 # Main crawl function to fetch and process data.
 def crawl(context: Context):
-    doc = context.fetch_html(context.data_url)
+    doc = context.fetch_html(context.data_url, cache_days=3)
```
pudo, line 98: Is this for testing? Perhaps we can take it back out?

PR #1407, opened 2024-10-17
singapore economic development board

nvmbrasserie:


datasets/sg/dev_board/crawler.py
```diff
@@ -0,0 +1,71 @@
...
+            profiles = section.findall(".//div[@class='profile-content-container']")
+            for profile in profiles:
+                # Extract the name, role, and link
+                name = profile.find(".//h5").text_content().strip()
+                name_match = TITLE_REGEX.match(name)
+                if name_match:
+                    name = name_match.group("name")
+                    title = name_match.group(1)
+                else:
+                    context.log.warn(f"Could not extract title from name: {name}")
```
jbothma, line 64: did you split the "emit" part to its own function because this was getting a bit long? Or some other reason?

we normally prefer a "crawl_person" which extracts and emits the data in one place.

In this case I think there's a bug where if one iteration of the loop fails to extract title, the previous title will be used in that iteration. That sort of bug is less likely when each person is handled in their own function call because the title variable won't be initialised if it doesn't match.

datasets/sg/dev_board/crawler.py
```diff
@@ -0,0 +1,71 @@
...
+from zavod import Context, helpers as h
+from zavod.logic.pep import categorise
+from zavod.shed.zyte_api import fetch_html
+
+
+LINKS = {
+    # Mapping URLs to their respective position names
+    "https://www.edb.gov.sg/en/about-edb/our-team/executive-management.html": "Member of the Executive Committee",
+    "https://www.edb.gov.sg/en/about-edb/our-team/board-members.html": "Board Member",
+    "https://www.edb.gov.sg/en/about-edb/our-team/international-advisory-council.html": "Member of the International Advisory Council",
```
jbothma, line 12: I think it's best to include the board name in the position name in this case. I see https://zavod.opensanctions.org/peps/#selecting-a-position-name is a bit vague with "include the organisational body where needed". But I think in this case it's helpful I don't think there are hundreds of boards (https://www.sgdi.gov.sg/statutory-boards) so the number of positions won't get out of hand.

```suggestion
    "https://www.edb.gov.sg/en/about-edb/our-team/executive-management.html": "Member of the Executive Committee of the Economic Development Board",
    "https://www.edb.gov.sg/en/about-edb/our-team/board-members.html": "Board Member of the Economic Development Board",
    "https://www.edb.gov.sg/en/about-edb/our-team/international-advisory-council.html": "Member of the International Advisory Council of the Economic Development Board",
```

datasets/sg/dev_board/crawler.py
jbothma: https://zavod.opensanctions.org/best_practices/patterns/#common-crawler-code-structure

jbothma:
we added the full gov directory instead

PR #1411, opened 2024-10-17
adding Arizona medicare suspensions crawler

bgmello:


bgmello:
I don't think it's possible to determine if the provider is a Person or Company using the data. There are some Persons names that have a `/`, but it's not all of them:
Screenshot 2024-10-17 at 17 51 22
Screenshot 2024-10-17 at 17 51 30

jbothma:
It looks like we need to work a bit more on the merge/split logic. We're currently incorrectly splitting these names, increasing the chance of false positive or false negative negative matches.

image

image

Let's see if gpt does the right thing with little effort, by just scraping and looking at the returned dicts. If not, I think putting more effort into the pdfplumber approach is better than struggling with gpt.

I don't see anything obvious from pdfplumber about handling merged/spanning rows, so I think the approach you're taking with a bit more work is best.

It looks like the Action Type column has one value when the rows need to be merged to get the full name, and it's repeated when it pertains to distinct entities (needs to be split).

jbothma:
here's the geolocation code I used https://github.com/opensanctions/opensanctions/commit/2c5a5ede66d6da89955b970455b9e07052531431

bgmello:
I added another condition to the if statement to check if there is multiple values in the Action Type column but it doesn't work every time because of this example:
Screenshot 2024-10-23 at 13 28 23

It generates like this, with the two companies names on both dictionaries:
Screenshot 2024-10-23 at 13 28 37

I'm not seeing any solution other than trying to solve this cases manually with a lookup.

jbothma:
Thanks for trying it out. 

I think you're right - it's best to be cautious. So we need some logic to make sure 

1. any rows that might be split/merged are not added automatically
2. we check for a lookup value for them and use that instead of what's extracted
3. we warn if no lookup is found

one Idea I have is something like this

1. loop through all the rows
  1.  mark any partial rows as needing manual extraction
  1. mark the row before any partial rows as needing manual extraction
1. run through again
  1. extract/emit what we can automatically
  2. take all contiguous rows needing manual extraction, combine their content somehow, e.g. slugify(json.stringify(rows, sort_keys=True))
  3. do a lookup on that string
  4. add an array of dictionaries with schema, name, etc with the manually extracted data

But I'm happy if you have another idea for doing this

jbothma:
let's wait until Wednesday to see if they respond - I've emailed to ask for an excel version

jbothma:
No response. Let's try this sort of thing https://github.com/opensanctions/opensanctions/pull/1411#issuecomment-2442133819

bgmello:
Done

PR #1423, opened 2024-10-21
Liechtenstein EntsG via pdfplumber

pudo:


datasets/li/posting_sanctions/crawler.py
```diff
@@ -77,88 +75,98 @@ def parse_target(
...
+    settings = {
+        "vertical_strategy": "lines",
+        "horizontal_strategy": "lines",
+        "text_tolerance": 1,
+    }
+    headers: Optional[List[str]] = None
+    rows: List[Dict[str, str]] = []
+    for page in pdf.pages:
+        cropped = page.crop((0, 50, page.width, page.height - 10))
+        for row in cropped.extract_table(settings):
```
jbothma, line 101: out of interest, was the cropping necessary, or just a bit of cleanup? extract_table returns the biggest table discovered by default, so I'd expect it to pick the table and nothing else, handling margin changes out of the box.

datasets/li/posting_sanctions/crawler.py
```diff
@@ -77,88 +75,98 @@ def parse_target(
...
         sanction.add("description", "Debarment")
         sanction.add("program", "EntsG Sanctions")
         reason = (
             "Repeated or severe infraction against "
-            f"Liechtenstein Posted Workers Act, {law}"
+            f"Liechtenstein Posted Workers Act, {violation}"
         )
         sanction.add("reason", reason)
         context.emit(sanction)
         context.emit(company, target=True)
```
jbothma, line 137: we probably want some `ended = max(end) < settings.RUN_TIME_ISO` on topics and target, right?

I'll get to that helper any day now

datasets/li/posting_sanctions/crawler.py
```diff
@@ -77,88 +75,98 @@ def parse_target(
...
+        )
+        sanction.add("date", date)
         sanction.add("description", "Administrative Penalty")
         sanction.add("program", "EntsG Sanctions")
         sanction.add(
-            "reason", f"Infraction against Liechtenstein Posted Workers Act, {law}"
+            "reason",
+            f"Infraction against Liechtenstein Posted Workers Act, {violation}",
         )
         company.add("topics", "debarment")
```
jbothma, line 165: should we add reg.warn and announce a migration to that topic for this file?

datasets/li/posting_sanctions/li_entsg.yml
```diff
@@ -32,12 +29,11 @@ publisher:
   country: li
   url: https://www.llv.li/de/landesverwaltung/amt-fuer-volkswirtschaft
   official: true
-
 url: https://www.llv.li/de/unternehmen/berichtspflichten-bewilligungen/grenzueberschreitende-dienstleistungen/entsendung-von-arbeitnehmern/sanktionsliste-gemaess-art.-7-abs.-4-entsg
```
jbothma, line 32: this seems to be 404. now https://www.llv.li/de/landesverwaltung/amt-fuer-volkswirtschaft/grenzueberschreitende-dienstleistungen/entsendung-von-arbeitnehmern/sanktionsliste-gemaess-art.-7-abs.-4-entsg

datasets/li/posting_sanctions/crawler.py
jbothma: it looks a bit like art 9 is violations but not yet sanctions, but my german is a tad rustier than yours

datasets/li/posting_sanctions/crawler.py
pudo: It actually fixed a fixed a parsing app. The first page of the PDF contains a really heavy line at the top as a visual element, and that was used by plumber as the first cell divider line unless I cut it off.

pudo:
Man that was shoddy work. Pushing a revised version.

PR #1426, opened 2024-10-22
Emit warnings if identifiers don't match the format of the prop.

pudo:


pudo:
This has escalated a bit into allowing for moving properties to a new field based on a type mapping. The idea is that this would enable cleanups where either we fix invalid type values, or we just move them to a more generic field for which no `format` is defined.

jbothma (commented):
you know what's coming next, right? `result.clean_fun` and `result.eval`.... :trollface:

jbothma:
Looks good to me!

PR #1440, opened 2024-10-23
adding South Carolina medicare exclusions crawler

bgmello:
I tried to find a way to identify if it's a company or a person but I don't think it's possible because there are companies with `,` in the name, for example:

Screenshot 2024-10-23 at 17 45 34

From what I see it is because of the suffixes `, Inc`, `, LLC` etc. Would it be ok to assume that if there is a `,` and if it's not those suffixes, then it is a person?

### DIFF
--- a/datasets/us/sc/med_exclusions/crawler.py
+++ b/datasets/us/sc/med_exclusions/crawler.py
@@ -1,0 +1,79 @@
+from typing import Dict
+from rigour.mime.types import XLSX
+from openpyxl import load_workbook
+
+from zavod import Context, helpers as h
+from zavod.shed.zyte_api import fetch_html, fetch_resource
+import re
+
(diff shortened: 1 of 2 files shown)


jbothma:
I think it's not really worth the risk - better to just make them all LegalEntity then.

While it's nice to encode them using specific schemata when we can, getting it wrong causes issues if we later merge the same entity from two sources and one of them is wrong.

So if you can be 100% sure from the data, be specific, otherwise LegalEntity.

One pattern when you have schema-specific props, is to make everything LegalEntity, then use h.add_cast(entity, birthDate) or whatever on each row, and if there's a birthDate value, we'll change the schema, otherwise it'll just carry on as it was.

jbothma (commented):
great stuff, thank you!

I've made the changes for the couple of comments, and small tweaks applicable in this case, in https://github.com/opensanctions/opensanctions/pull/1476

PR #1454, opened 2024-10-25
daily-fixes-oct-25

nvmbrasserie:


datasets/at/meine_abgeordneten/crawler.py
```diff
@@ -200,7 +200,7 @@ def crawl(context: Context):
     response = context.fetch_html(context.data_url)
 
     # XPath to the url for the pages of each politician
-    xpath_politician_page = '//*[contains(@class, "abgeordneter")]/*/a/@href'
+    xpath_politician_page = '//div[(@class="abgeordneter row")]/*/a/@href'
```
jbothma, line 203: why are we making this more strict? I think checking for row as well is fine, but we shouldn't make assumptions about what else is in the class attribute, or what the ordering of classes is. So rather 

```
//div[contains(@class, "abgeordneter")][contains(@class, "row")]/*/a/@href

File: datasets/us/sec_harmed_investors/us_sec_harmed_investors.yaml [hunk 2]
```diff
@@ -1023,3 +1023,8 @@ lookups:
       - match: moravec, matthew
         names:
           - Matthew Moravec
+      - match: sterling watters group lp; sterling watters capital advisors llc; and sterling watters capital management inc.
+        names:
+          - Sterling Watters Group LP
+          - Sterling Watters Capital Advisors LLC
+          - Sterling Watters Capital Management Inc.
```
jbothma, line 1030: I already did this one :) adjacent to the existing sterling watters entry. I try to colocate them

PR #1463, opened 2024-10-28
h parse date 10

nvmbrasserie:


datasets/ar/repet/crawler.py
```diff
@@ -27,11 +27,11 @@ def parse_date(date):
     if isinstance(date, list):
         dates = []
         for d in date:
-            dates.extend(parse_date(d))
+            dates.append(parse_date(d))
         return dates
     if isinstance(date, dict):
         date = date.get("VALUE")
-    return h.parse_date(date, ["%d/%m/%Y"])
```
jbothma, line 34: this changes the return type from `List[str]` to `List[str] | str` making it harder to reason about its logic. How about returning `[date]` and keeping it `dates.extend`? Then it's a little easier to reason about the correctness of this function.

datasets/ar/repet/crawler.py
```diff
@@ -86,11 +88,15 @@ def crawl_common(context: Context, data: Dict[str, str], part: str, schema: str)
...
     modified_at = parse_date(data.pop("LAST_DAY_UPDATED"))
-    entity.add("createdAt", submitted_on or listed_on or modified_at)
-    entity.add("modifiedAt", modified_at)
-
-    sanction.add("listingDate", submitted_on or listed_on)
-    sanction.add("startDate", listed_on)
+    h.apply_dates(entity, "createdAt", submitted_on)
+    h.apply_dates(entity, "createdAt", listed_on)
+    if modified_at != []:
+        h.apply_date(entity, "createdAt", modified_at[0])
```
pudo, line 94: maybe use `min(modified_at)` and then `max(modified_at)` for the next line?

nvmbrasserie: I'd suggest making both of them `modified_at`, because sometimes we have cases like: 

```
Person:createdAt,2012-02-13,
Person:createdAt,2001-02-23,
Person:modifiedAt,2012-02-13
```

PR #1464, opened 2024-10-28
h parse date 11

nvmbrasserie:


datasets/ae/local_terrorists/crawler.py
```diff
@@ -27,15 +35,22 @@ def parse_row(
...
-        if header in ("program", "listingDate", "endDate", "provisions"):
+        if header in ("program", "provisions"):
+            sanction.add(header, value, lang=lang)
+            continue
+        if header in ("listingDate", "endDate"):
+            for date in DATES:
+                match = date.match(value[0])
+                if match is not None:
+                    value = match.group(1)
             sanction.add(header, value, lang=lang)
```
jbothma, line 54: how about not parsing it above at `if type_ == date`, 

then here at `if header in ("listingDate...`, instead of trying to match, you just replace "مدرج بموجب قرار مجلس الوزراء رقم" and the other one with "" and calling `apply_date(sanction, ...)`, 

then you can handle `\(\d+\) لسنة (\d{4})` using `Dataset.dates.formats`

it's a bit simpler and avoids an unnecessary h.extract_date

PR #1466, opened 2024-10-29
h parse date 12

nvmbrasserie:


datasets/at/meine_abgeordneten/crawler.py
```diff
@@ -14,22 +14,21 @@
...
         assume_current = True
     elif inactive_dates_el is not None:
-        inactive_dates = inactive_dates_el.text_content()
+        inactive_dates = inactive_dates_el.text_content().replace("ab ", "")
         if " - " in inactive_dates:
             start_date, end_date = inactive_dates.split(" - ")
-            start_date = h.parse_date(start_date, context.dataset.dates.formats)
-            end_date = h.parse_date(end_date, context.dataset.dates.formats)
+            start_date = h.extract_date(context.dataset, start_date.strip())
+            end_date = h.extract_date(context.dataset, end_date.strip())
```
jbothma, line 31: if we leave the parsing to make_occupancy() and move the `? -> None` to lookups, can we get the same output? that'd be a nice big cleanup.

I think the real logic in this function is about setting assume_current true/false. But it doesn't look like parsing the dates according to our date formats is actually part of that logic, unless I missed something

PR #1467, opened 2024-10-29
h parse date 13

nvmbrasserie:


pudo (commented):
This one we could also just set `data.lang = ron` instead of the individual fields, right?

PR #1469, opened 2024-10-29
h parse date 15

nvmbrasserie:


datasets/ru/nsd_isin/crawler.py
```diff
@@ -11,19 +11,12 @@
...
-    )
-    date_info = h.parse_formats(text, context.dataset.dates.formats)
-
-    if date_info and date_info.dt:
-        return date_info.text
-
-    context.log.warning("Failed to parse date", raw_date=text)
-    return None
+    text = text.replace("\xa0", " ").replace("г.", "").strip().lower()
+    date_info = text
```
pudo, line 18: What does this do?

datasets/ru/nsd_isin/crawler.py
```diff
@@ -87,7 +80,7 @@ def crawl_item(context: Context, url: str):
         issuer.id = context.make_id(isin_code, issuer_name)
     issuer.add("country", "ru")
     for prop, prop_val in values["issuer"].items():
-        issuer.add(prop, prop_val)
+        issuer.add(prop, prop_val, lang="rus")
```
pudo, line 83: We have `data.lang = rus` in the dataset metadata, so in the crawler itself we don't really need to override, I think :)

datasets/ru/nsd_isin/crawler.py
```diff
@@ -55,7 +48,7 @@ def crawl_item(context: Context, url: str):
 
         if result.type == "date":
             try:
-                value = parse_date(value, context)
+                value = h.extract_date(context.dataset, parse_date(value))[0]
             except ValueError:
```
pudo, line 52: Does this ever raise a value error? Do we want to instead do this when we actually .add the properties, based on looking at `prop.type` ?

pudo (changes requested):
Good

datasets/ru/nsd_isin/crawler.py
```diff
@@ -73,7 +58,11 @@ def crawl_item(context: Context, url: str):
     security.add("sourceUrl", url)
     security.add("topics", "sanction")
     for prop, prop_val in values["security"].items():
-        security.add(prop, prop_val)
+        if prop in ["issueDate", "maturityDate", "createdAt"]:
```
pudo, line 61: Nice, that works and it short and sweet! Let's keep. What would have also worked (just because we talked at lunch):

```
prop_ = security.schema.get(prop)
if prop_ is not None and prop.type.name == 'date':
```

PR #1472, opened 2024-10-29
in mha banned update

nvmbrasserie:


datasets/in/mha_banned/crawler.py
```diff
@@ -16,37 +16,53 @@ def parse_names(field: str) -> List[str]:
...
+    expected_sources: Set[bool] = set()
 
     expected_sources.add(
         assert_link_hash(
             context,
             doc,
             "UNLAWFUL ASSOCIATIONS UNDER SECTION 3 OF UNLAWFUL ACTIVITIES (PREVENTION) ACT, 1967",
-            "582fd09cb946aa1bfc186da33d43642fa160cb3c",
+            "50cbbee5d7777188e5fd8d2e5b74ddcc6b537716",
+            xpath='.//div[@id="block-mhanew-content"]//div[3]//table',
```
jbothma, line 56: index-based xpath is a red flag. It's brittle, i.e. it could easily stop matching if they make uninteresting content changes, and I don't think it's contributing anything to the selection. It's not like you need to say "the table in the third div in some list of divs under block-mha-new-content, right? Is there more than one table in block-mhanew-content?

```suggestion
            xpath='.//div[@id="block-mhanew-content"]//table',
```

datasets/in/mha_banned/crawler.py
```diff
@@ -16,37 +16,53 @@ def parse_names(field: str) -> List[str]:
...
         )
     )
     expected_sources.add(
         assert_link_hash(
             context,
             doc,
             "TERRORIST ORGANISATIONS LISTED IN THE FIRST SCHEDULE OF THE UNLAWFUL ACTIVITIES (PREVENTION) ACT, 1967",
-            "46e270bfefd3174adcb0e8c80c8e952bca5a772f",
+            "66cc60e0bb6937ec88c620eadc3ec213a1ede29c",
+            xpath='.//div[@id="block-mhanew-content"]//div[2]//div[2]//table',
```
jbothma, line 65: ditto

```suggestion
            xpath='.//div[@id="block-mhanew-content"]//table',
```

datasets/in/mha_banned/crawler.py
```diff
@@ -65,11 +81,6 @@ def crawl(context: Context) -> None:
         for row in csv.DictReader(fh):
             entity = context.make(row.pop("Type", "LegalEntity"))
             source_url = row.pop("SourceURL")
-            if source_url not in expected_sources:
-                context.log.warn(
-                    "Source URL not in overview page. Perhaps it's out of date?",
-                    url=source_url,
-                )
```
jbothma, line 72: why are we removing this? This is our sanity check that all our sources are still linked from the landing page. We don't want to give this up.

datasets/in/mha_banned/crawler.py
```diff
@@ -16,37 +16,53 @@ def parse_names(field: str) -> List[str]:
...
+        if not success:
+            print(f"Hash assertion for {url} failed")
+
+    return url if success else None
 
 
 def crawl(context: Context) -> None:
     doc = context.fetch_html(context.dataset.url)
     doc.make_links_absolute(context.dataset.url)
-    expected_sources: Set[str] = set()
```
jbothma, line 34: we still want this to be a set of url strings

datasets/in/mha_banned/crawler.py
```diff
@@ -16,37 +16,53 @@ def parse_names(field: str) -> List[str]:
...
+    if xpath:
+        success = h.assert_html_url_hash(context, url, expected, path=xpath)
+        if not success:
+            print(f"Hash assertion for {url} with XPath {xpath} failed")
+    else:
+        success = h.assert_url_hash(context, url, expected)
+        if not success:
+            print(f"Hash assertion for {url} failed")
+
+    return url if success else None
```
jbothma, line 42: we don't want to leave this as print for production. assert_... would be logging a warning for us if the hash doesn't match. 

we also always want to add the link we find on the overview page to the sources set.

we're doing two important things in this function:

- looking up the link to the document based on the label (where the label is a part of the legislation we care about)
- checking if the hash of that link's content has changed
I think that's making it a bit difficult to understand.

Perhaps now or later, we could split this into two functions:
- get_link_by_label()
- assert_link_hash()
That would also make it clearer why we return the url despite the hash not matching.

Perhaps renaming "expected_sources" to "linked_sources" would make more sense?

PR #1493, opened 2024-11-04
fixes-nov-4

nvmbrasserie:


datasets/us/ia/med_exclusions/crawler.py
```diff
@@ -95,7 +95,7 @@ def crawl_item(row: Dict[str, str], context: Context):
 def crawl_excel_url(context: Context):
     doc = context.fetch_html(context.data_url)
     doc.make_links_absolute(context.data_url)
-    return doc.find(".//a[@title='Program Integrity - Sanctions List']").get("href")
+    return doc.find(".//a[@title='Iowa Medicaid Sanction List 10.24']").get("href")
```
jbothma, line 98: cardcoding this months date means fixing it again next month. Could you change it to contains sanction list? Element.find doesn't support contains, only Element.xpath does.

PR #1501, opened 2024-11-06
us plum book

nvmbrasserie:
opensanctions/crawler-planning#273

We've decided to keep both positions for now in order to dedupe based on the `person.add("position")`

### DIFF
--- a/datasets/us/plum_book/crawler.py
+++ b/datasets/us/plum_book/crawler.py
@@ -40,7 +40,6 @@ def crawl(context: Context):
             position = h.make_position(
                 context,
                 name=position_name,
-                description=appointment_type,
                 subnational_area=location,
                 country="us",
             )
@@ -51,6 +50,8 @@ def crawl(context: Context):
             occupancy = h.make_occupancy(
                 context, person, position, end_date=expiration_date
             )
+            if occupancy is not None:
+                occupancy.add("description", appointment_type)
             if not occupancy:
                 continue
             context.emit(person, target=True)
--- a/datasets/us/plum_book/crawler.py
+++ b/datasets/us/plum_book/crawler.py
@@ -6,7 +6,6 @@ from zavod.logic.pep import categorise
 IGNORE = [
     "OrganizationName",
     "PositionStatus",
-    "AppointmentTypeDescription",
     "PaymentPlanDescription",
     "LevelGradePay",
     "Tenure",
@@ -21,7 +20,7 @@ def crawl(context: Context):
             agency_name = row.pop("\ufeffAgencyName")
             # org_name = row.pop("OrganizationName")
             position_title = row.pop("PositionTitle", None)
-            # appointment_type = row.pop("AppointmentTypeDescription")
+            appointment_type = row.pop("AppointmentTypeDescription")
             expiration_date = row.pop("ExpirationDate")
             location = row.pop("Location")
             incumbent_first_name = row.pop("IncumbentFirstName")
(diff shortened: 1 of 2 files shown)


datasets/us/plum_book/crawler.py
```diff
@@ -0,0 +1,54 @@
...
+            person.id = context.make_slug(incumbent_first_name, incumbent_last_name)
+            h.apply_name(
+                person, first_name=incumbent_first_name, last_name=incumbent_last_name
+            )
+            person.add("position", f"{position_title}, {agency_name}")  # for dedupe
+
+            position = h.make_position(
+                context,
+                name=f"{position_title}, {agency_name}",
+                description=appointment_type,
```
jbothma, line 40: I think appointment type technically belongs on the occupancy, not the position. There are probably easier ways to determine this, but here are examples where one position name would come up with multiple appointment types. That means one Position entity would end up with all three of these in its description property.

```
grep '"Position"' data/datasets/us_plum_book/entities.ftm.json | jq '(.properties.description | length | tostring) + " " + .properties.name[0]' | sort -n
...
"3 SENIOR ADVISOR, DEPARTMENT OF THE INTERIOR"
"3 SENIOR ADVISOR, OFFICE OF PERSONNEL MANAGEMENT"
"3 SENIOR ADVISOR, OFFICE OF THE SECRETARY OF DEFENSE"
"4 SENIOR ADVISOR, DEPARTMENT OF HOUSING AND URBAN DEVELOPMENT"
```

datasets/us/plum_book/crawler.py
```diff
@@ -0,0 +1,54 @@
...
+
+            categorisation = categorise(context, position, is_pep=True)
+            if categorisation.is_pep:
+                occupancy = h.make_occupancy(
+                    context, person, position, end_date=expiration_date
+                )
+                if occupancy:
+                    context.emit(person, target=True)
+                    context.emit(position)
+                    context.emit(occupancy)
```
jbothma, line 53: it can be nice to invert the conditions, then the nesting doesn't get deeper and deeper.

datasets/us/plum_book/us_plum_book.yml
```diff
@@ -0,0 +1,64 @@
...
+entry_point: crawler.py
+prefix: plum
+coverage:
+  frequency: daily
+  start: 2024-11-06
+load_db_uri: ${OPENSANCTIONS_DATABASE_URI}
+summary: >
+  The Plum Book lists federal positions filled by political appointees, Schedule C 
+  employees, and other executives.
+description: |
```
jbothma, line 11: super nice

datasets/us/plum_book/crawler.py
```diff
@@ -0,0 +1,54 @@
...
+            appointment_type = row.pop("AppointmentTypeDescription")
+            expiration_date = row.pop("ExpirationDate")
+            location = row.pop("Location")
+            incumbent_first_name = row.pop("IncumbentFirstName")
+            incumbent_last_name = row.pop("IncumbentLastName")
+
+            if not incumbent_first_name or not incumbent_last_name:
+                continue
+            person = context.make("Person")
+            person.id = context.make_slug(incumbent_first_name, incumbent_last_name)
```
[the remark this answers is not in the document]
jbothma: It sucks, but I think we need to put more information into the ID. I think the best thing to add is position_name, since so many positions are going to be in washington, and there's just a really good chance of the same name for two individuals coming up.

```
grep '"Occupancy"' data/datasets/us_plum_book/entities.ftm.json | jq .properties.holder[0]|sort|uniq -c|sort -n
...
   2 "plum-thomas-sullivan"
   2 "plum-william-brown"
   2 "plum-william-harris"
   3 "plum-michael-horowitz"
```

jbothma (changes requested):
Super nice, thanks.

I think we need some easy tweaks then it's ready to go.

PR #1509, opened 2024-11-08
fixes-nov-8

nvmbrasserie:


datasets/gb/hmt/crawler.py
```diff
@@ -123,8 +152,23 @@ def parse_row(context: Context, row: Dict[str, Any]):
...
+                prop = context.lookup_value("properties", slugify(result["key"]))
+                if prop is not None:
+                    if prop == "kppCode":
+                        entity.add_cast("Company", prop, result["value"])
+                    else:
+                        entity.add_cast("LegalEntity", prop, result["value"])
+                else:
+                    context.log.warning(
+                        "Unknown property", value=slugify(result["key"])
+                    )
```
jbothma, line 169: I think after splitting, you can remap these props with this lookup on type.identifier https://opensanctions-corp.slack.com/archives/C07C5G9KZED/p1731064875859569 rather than adding this code

datasets/gb/hmt/gb_hmt_sanctions.yml
```diff
@@ -39,6 +39,23 @@ data:
...
+          - ogrn
+        value: ogrnCode
+      - match: russia-kpp
+        value: kppCode
+      - match: russia-okpo
+        value: okpoCode
+      - match: 
+          - tax-id-no
+          - tin-taxpayer-identification-number
+        value: taxNumber
```
jbothma, line 58: this might not be needed

datasets/gb/hmt/crawler.py
jbothma: here https://github.com/opensanctions/opensanctions/pull/1509/files#diff-573cc64c90b8256f0913c009aa2ce0d475bdf166f59e20e29a505f61f96748e4R59

jbothma (commented):
nice stuff - I think it could be a tad simpler though - see comments

PR #1516, opened 2024-11-09
adding Missouri medicare exclusions crawler

bgmello:
I wasn't able to fetch the excel file even though I was using zavod, I think it might be a problem with the website.


jbothma:
they might have updated in the meantime. Try this? https://github.com/opensanctions/opensanctions/commit/ce1ecfcbe0d09df5662b221c9e2fc7a728e9e4a0

bgmello:
There is a column called letter date but I'm not sure what it means.

I also think there isn't a way to check if the provider is a `Company` or `Person`, maybe using the `,` that appears in the persons names, but it might be risky.

PR #1517, opened 2024-11-09
adding Alabama medicare exclusions crawler

bgmello:


bgmello:
I wasn't able to read the data from the xls file because it is poorly formatted:
Screenshot 2024-11-09 at 18 21 01

jbothma:
I've emailed them to ask if they could re-release the data since the file is corrupt. The PDF also seems tricky to use.

PR #1529, opened 2024-11-12
h-parse-date-22

nvmbrasserie:


datasets/gb/proscribed_orgs/crawler.py
```diff
@@ -27,8 +27,7 @@ def parse_comment(context: Context, text: str) -> List[str]:
     m = ADDITION_RE.search(text)
     if m is None:
         return []
-    return h.extract_date(context.dataset, m.group(2))
-    # return h.parse_date(m.group(2), DATE_FORMATS)
+    return [m.group(2)]
```
jbothma, line 30: this can now simply be str | None, so that we don't have to do add_date[0]

datasets/fr/assemblee/crawler.py
```diff
@@ -117,7 +117,7 @@ def crawl_acteur(context, data: Dict[str, Any]):
     context.audit_data(birth, ["depNais"])
     dod = ec.pop("dateDeces")
     if is_not_nil(dod):
-        person.set("deathDate", h.parse_date(dod, ["%Y-%m-%d"]))
+        h.apply_date(person, "deathDate", dod)
```
jbothma, line 120: we can get rid of is_not_nil using a couple of type.date lookups, right? at least these calls but I think more

nvmbrasserie: it gets into the values and double-checks `{'@xmlns:xsi': 'http://www.w3.org/2001/XMLSchema-instance', '@xsi:nil': 'true'}` whether `xsi:nil` is true. I don't think we can solve it with the lookups

jbothma: ah yeah - dict is unhashable, it says.

PR #1532, opened 2024-11-12
h parse date 23

nvmbrasserie:


datasets/im/disqualified_directors/crawler.py
```diff
@@ -5,82 +5,58 @@
...
-    if item["Date of Birth"] != "Not known":
-        birth_date = h.parse_date(item.pop("Date of Birth"), formats=["%d %B %Y"])
-    else:
-        birth_date = None
-        item.pop("Date of Birth")
+    birth_date = item.pop("Date of Birth")
     address = item.pop("Address (at date of disqualification)")
 
     person = context.make("Person")
-    person.id = context.make_id(name, birth_date, address)
```
jbothma, line 66: why remove the address? they don't always have a birth date, so it's nice to have more disambiguating info in the ID. I don't see the harm in including it?

datasets/im/disqualified_directors/crawler.py
```diff
@@ -5,82 +5,58 @@
...
     data_dict = {}
-    current_key = None
 
     # Iterate through each p element
     for p in ele:
+        text_content = p.text_content().strip()
 
-        strong_tag = p.find("strong")
+        # Split on the first colon to get the key and value
+        if ":" in text_content:
```
jbothma, line 14: nice solution!

datasets/im/disqualified_directors/crawler.py
nvmbrasserie: I think I removed it temporarily and forgot to add it back

PR #1540, opened 2024-11-13
h-parse-date-28

nvmbrasserie:


pudo:
This is so nice, getting rid of all these inline format strings :)

PR #1555, opened 2024-11-15
War and Sanctions

nvmbrasserie:


datasets/ua/war_sanctions/crawler.py
```diff
@@ -140,12 +145,24 @@ def crawl_vessel(context: Context, details_container, link):
...
+    for name in h.multi_split(data.pop("Former ship names"), [" / "]):
+        vessel.add("previousName", name)
+    vessel.add(
+        "description", f"Commercial ship manager (IMO / Country / Date): {com_manager}"
+    )
+    vessel.add(
+        "description",
+        f"Ship Safety Management Manager (IMO / Country / Date): {safety_manager}",
+    )
+    vessel.add("topics", "sanction")
```
pudo, line 165: is this data that wants to be a relationship? Otherwise let's skip it and avoid semi-structured messes in `description` :)

jbothma (commented):
Just a start - I haven't gotten to the detail yet.

Let's only emit the `poi` topic. We could add to the description that the website is advocating for broader sanctions being imposed on the entities listed there.

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,397 @@
...
+    sanction.add("reason", data.pop("Reasons"))
+    sanction.add("sourceUrl", link)
+
+    context.emit(company, target=True)
+    context.emit(sanction)
+    context.audit_data(data)
+
+
+def extract_next_page_url(doc, base_url, next_xpath):
+    doc.make_links_absolute(base_url)
```
jbothma, line 356: let's do this right after we fetch the page. We want absolute links for the rest of the process too, and that'll be consistent with other crawlers.

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,397 @@
...
+            next_xpath = "//ul[@class='pagination']//li[@class='next']/a"
+
+            # get the next page URL, if exists
+            next_url = extract_next_page_url(doc, base_url, next_xpath)
+
+            if next_url:
+                current_url = next_url
+                visited_pages += 1
+            else:
+                break
```
jbothma, line 397: Nice, you're already checking if current_url is set, so you can trim this down

```suggestion
            next_url = extract_next_page_url(doc, base_url, next_xpath)
            current_url = next_url
            visited_pages += 1
```

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,397 @@
...
+    vessel.add("imoNumber", imo_num)
+    vessel.add("type", type)
+    vessel.add("description", description)
+    vessel.add("description", data.pop("Vessel information"))
+    vessel.add("callSign", data.pop("Call sign"))
+    vessel.add("flag", data.pop("Flag (Current)"))
+    vessel.add("mmsi", data.pop("MMSI"))
+    vessel.add("buildDate", data.pop("Build year"))
+    for web_resource in data.pop("Web resources"):
+        vessel.add("sourceUrl", web_resource)
```
jbothma, line 138: @pudo am I right in thinking since these aren't our sources, we only want the detail URL we used like https://war-sanctions.gur.gov.ua/en/transport/ships/455 in sourceUrl?

Do we want to place these URLs anywhere else? Perhaps a additionalUrl prop? Or just stick to the identifying information and let people find those links at the source?

image

Style-wise, @nvmbrasserie let's not add data to a dict just to take it from the dict a few rows later, when we could just use `web_resources` in the first place. Entity.add also accepts lists of values so no need for the loop.

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,397 @@
...
+    return data
+
+
+def crawl_index_page(context: Context, index_page, data_type):
+    index_page = context.fetch_html(index_page, cache_days=3)
+    main_grid = index_page.find('.//div[@id="main-grid"]')
+    if main_grid is not None:
+        for a in main_grid.findall(".//a"):
+            href = [a.get("href")]
+            for link in href:
```
jbothma, line 64: for link in href is unnecessary because we know there's exactly one element in href.

But this could be simplified to

```suggestion
        for link in main_grid.findall(".//a/@href"):
```

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,397 @@
...
+                    elif data_type == "company":
+                        details_container = detail_page.find(
+                            ".//div[@class='col-12 col-lg-9']"
+                        )
+                        crawl_company(context, details_container, link)
+                    if data_type == "vessel":
+                        details_container = detail_page.find(
+                            ".//div[@id='js_visibility']"
+                        )
+                        crawl_vessel(context, details_container, link)
```
jbothma, line 81: I wouldn't fetch and extract data from detail pages inside crawl_index_page. Let's move the fetch and container extraction to the respective detail functions, then 

- crawl_index_page will be focused on index pages
- crawl_ for detail pages will be focused on detail pages

let's also try and find more semantic selectors.

e.g. for vessel, js_visibility is about showing/hiding stuff using javascript. They might add more instances of it unrelated to the content we know we can handle. "col-12 col-lg-9" are about column width and responsiveness. I think just selecting //main will do the right thing if you then look for //section//tools-* below that later, right?

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,397 @@
...
+            com_rep = context.make("Representation")
+            com_rep.id = context.make_id(
+                vessel.id, "commercially managed by", com_manager.id
+            )
+            com_rep.add("client", vessel.id)
+            com_rep.add("agent", com_manager.id)
+            com_rep.add("role", "Commercial ship manager")
+            h.apply_date(com_rep, "startDate", com_date)
+
+            context.emit(com_rep)
```
jbothma: this looks ripe for factoring out to a dedicated function `crawl_ship_manager(context, vessel, com_manager)`

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,397 @@
...
+            safety_rep = context.make("Representation")
+            safety_rep.id = context.make_id(
+                vessel.id, "safety managed by", safety_manager.id
+            )
+            safety_rep.add("client", vessel.id)
+            safety_rep.add("agent", safety_manager.id)
+            safety_rep.add("role", "Ship Safety Management Manager")
+            h.apply_date(safety_rep, "startDate", safety_date)
+
+            context.emit(safety_rep)
```
jbothma: this looks ripe for factoring out to a dedicated function crawl_ship_safety_manager(context, vessel, com_manager)

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,397 @@
...
+            context.emit(owner, target=True)
+
+            ownership = context.make("Ownership")
+            ownership.id = context.make_id(vessel.id, "owned by", owner.id)
+            ownership.add("asset", vessel.id)
+            ownership.add("owner", owner.id)
+            ownership.add("ownershipType", "Owner")
+            h.apply_date(ownership, "startDate", owner_date)
+
+            context.emit(ownership)
```
jbothma: this looks ripe for factoring out to a dedicated function crawl_ship_owner(context, vessel, owner_info)

now that you have three functions that look almost exactly the same, maybe they can be refactored into one

```python
crawl_ship_relation(context, vessel, data: string, rel_schema, rel_role, from_prop, to_prop)
```

with three calls like

```python
crawl_ship_relation(
    context,
    vessel,
    data.pop("Shipowner (IMO / Country / Date)")
    "Owner",
    "asset",
    "owner",
)
```

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,397 @@
...
+            safety_rep.add("agent", safety_manager.id)
+            safety_rep.add("role", "Ship Safety Management Manager")
+            h.apply_date(safety_rep, "startDate", safety_date)
+
+            context.emit(safety_rep)
+
+    owner_info = data.pop("Shipowner (IMO / Country / Date)")
+    if owner_info != "":
+        owner_parts = owner_info.split(" / ")
+        if len(owner_parts) == 3:
```
jbothma, line 232: Once this is factored out to its own function, I would invert this:

```suggestion
        if len(owner_parts) != 3:
            context.log.warning(...)
            return
```

jbothma (commented):
Some more feedback

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,362 @@
...
+    return extracted
+
+
+def extract_label_value_pair(label_elem, value_elem, data):
+    label = label_elem.text_content().strip().replace("\n", " ")
+    value = value_elem.text_content().strip().replace("\n", " ")
+    value = " ".join(value.split())
+    # we use it for splitting some strings at a later stage
+    value = " | ".join(
+        [text.strip() for text in value_elem.itertext() if text.strip()]
```
jbothma, line 51: instead of _this join to later split_ thing, it'd be more standard to make the type if `value` be `str | List[str]` if simply `List[str]` isn't sufficient. `Entity.add` does the right thing with `List[str]` - no looping required so no logic needed for the different types

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,362 @@
...
+    ).strip()
+    data[label] = value
+
+    return data
```
jbothma, line 55: instead of mutating data in this function, it'd be more typical for a function like this to just clean and return the key and value as a tuple. Then you can add it to `data` where you call it.

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,362 @@
...
+        ".//section[@class='sections d-flex flex-wrap align-items-stretch justify-content-center mb-5 pl-5 pr-5 medium']"
+    )
+    h.assert_dom_hash(node, "dbb9a924d940b3a69a132a102e04dcf0f9fbfc5e")
+
+    for link_info in LINKS:
+        base_url = link_info["url"]
+        data_type = link_info["type"]
+        current_url = base_url
+        visited_pages = 0
+        while current_url and visited_pages < 100:  # emergency exit check
```
jbothma, line 342: remember to drop this check and rely on `if visited_pages >= 100` at the bottom

```suggestion
        while current_url:
```

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,362 @@
...
+            doc = context.fetch_html(current_url)
+            doc.make_links_absolute(base_url)
+            if doc is None:
+                context.log.warn(f"Failed to fetch {current_url}")
+                break
+            context.log.info(f"Processing {current_url}")
+            crawl_index_page(context, current_url, data_type)
+
+            # xpath to find the next page link
+            next_xpath = "//ul[@class='pagination']//li[@class='next']/a"
```
jbothma, line 352: can we hardcode this in the function dedicated to this task?

datasets/ua/war_sanctions/crawler.py
jbothma: we decided to include them all in sourceUrl

datasets/ua/war_sanctions/crawler.py
jbothma: we're still updating the dict in the function, but now we're also updating the dict where the function is called. I think you can drop the dict argument to this function now, right?

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,348 @@
...
+
+    return None
+
+
+def crawl(context: Context):
+    main_page = context.fetch_html(context.data_url, cache_days=3)
+    node = main_page.find(
+        ".//section[@class='sections d-flex flex-wrap align-items-stretch justify-content-center mb-5 pl-5 pr-5 medium']"
+    )
+    h.assert_dom_hash(node, "dbb9a924d940b3a69a132a102e04dcf0f9fbfc5e")
```
jbothma, line 324: maybe we can also `assert_html_url_hash("https://war-sanctions.gur.gov.ua/en/transport/ships?page=1&per-page=12", xpath_to_tabs)` this to detect when new content shows up?

image

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,348 @@
...
+
+    for link_info in LINKS:
+        base_url = link_info["url"]
+        data_type = link_info["type"]
+        current_url = base_url
+        visited_pages = 0
+        while current_url:
+            doc = context.fetch_html(current_url, cache_days=3)
```
jbothma, line 332: my experience is that it's best not to cache index pages because additions/removals can result in skipped entities

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,348 @@
...
+        ),
+        (
+            "justification_rows",
+            ".//div[contains(@class, 'tools-frame')]/div[contains(@class, 'mb-3')]",
+        ),
+        (
+            "additional_info_rows",
+            ".//div[contains(@class, 'tools-frame')]//div[@class='mb-3' or contains(@class, 'js_visibility')]",
+        ),
+    ]
```
jbothma, line 87: I don't really see a better way to do this. These are potentially brittle classes that could be used anywhere else too, but this is scoped to tools-frame and the .pop()s below should alert us to changes that make these lookups invalid.

👍

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,348 @@
...
+    vessel.add("sourceUrl", web_resources)
+    vessel.add("keywords", data.pop("Category"))
+    for pr_name in h.multi_split(data.pop("Former ship names"), [" / "]):
+        vessel.add("previousName", pr_name)
+    for past_flag in h.multi_split(data.pop("Flags (former)"), [" /"]):
+        vessel.add("pastFlags", past_flag)
+    vessel.add("topics", "poi")
+
+    sanction = h.make_sanction(context, vessel)
+    sanction.add("country", data.pop("Sanctions", None))
```
jbothma, line 131: this would have the effect of saying the sanction authority `Defence Intelligence of Ukraine` has sanctioned the vessel in all the countries where other countries' sanctions authorities have sanctioned the vessel. So let's not add the other sanctioning countries here.

Let's add Program to indicate that this is part of the `Vessels, Airports and Ports involved in the transportation of weapons, stolen Ukrainian products and in the circumvention of sanctions` list.
I got that from the heading on the homepage/list page

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,348 @@
+from zavod import Context, helpers as h
+
+
+LINKS = [
+    {  # child kidnappers
+        "url": "https://war-sanctions.gur.gov.ua/en/kidnappers/persons?page=1&per-page=12",
+        "type": "person",
+    },
```
jbothma, line 8: Let's add something like `List of persons involved in the deportation of Ukranian children` (sourced from the homepage) as `Sanction:program`

Same for the other lists.

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,348 @@
...
+        entity.id = context.make_id(entity_name, entity_country)
+        entity.add("name", entity_name)
+        entity.add("registrationNumber", entity_imo)
+        entity.add("country", entity_country)
+        entity.add("topics", "poi")
+        context.emit(entity, target=True)
+
+        # Create the relation representation
+        relation = context.make(rel_schema)
+        relation.id = context.make_id(vessel.id, f"{rel_role} by", entity.id)
```
jbothma, line 210: Could also just be this

```suggestion
        relation.id = context.make_id(vessel.id, rel_role, entity.id)
```

jbothma (commented):
Great stuff, we're just about there.

I think most of what's left is just being a bit careful about some conditions we have. Since we're parsing html with little semantics embedded in the markup, we need to be very careful to make sure the assumptions hold, and when they don't hold any more, we'll find out.

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,348 @@
...
+        rows = details_container.xpath(xpath_expr)
+        for row in rows:
+            divs = row.findall("div")
+            if len(divs) == 2:
+                label_elem, value_elem = divs
+                if "yellow" in value_elem.get("class"):
+                    label, value = extract_label_value_pair(
+                        label_elem, value_elem, data
+                    )
+                    data[label] = value
```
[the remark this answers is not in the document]
jbothma: I've added comments to highlight how we ensure we're capturing at least some of what we expect to, and nothing more than we expect to.

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,381 @@
...
+    linked_entity_name = data.pop(
+        "The person in connection with whom sanctions have been applied"
+    )
+    if linked_entity_name != "":
+        linked_entity = context.make("LegalEntity")
+        linked_entity.id = context.make_id(linked_entity_name)
+        for name in h.multi_split(linked_entity_name, [" | "]):
+            linked_entity.add("name", name)
+        linked_entity.add("topics", "poi")
+        context.emit(linked_entity, target=True)
```
jbothma, line 153: are we missing a relation entity for this? an UnknownLink perhaps? `role` could be `"The person in connection with whom sanctions have been applied"`

datasets/ua/war_sanctions/crawler.py
```diff
@@ -0,0 +1,381 @@
...
+    return None
+
+
+def crawl(context: Context):
+    main_page = context.fetch_html(context.data_url, cache_days=1)
+    node = main_page.find(
+        ".//section[@class='sections d-flex flex-wrap align-items-stretch justify-content-center mb-5 pl-5 pr-5 medium']"
+    )
+    api_section = main_page.find(
+        ".//div[@class='d-flex flex-column flex-sm-row menu-alt']"
```
jbothma, line 351: in the browser, these class values are not ordered. If you were writing CSS, which most or all of these are intended for, you could select this element with something like `div.sections.justify-content-center` or equally `div.justify-content-center.sections`

They can add more classes in between, rename some of these, change mb-5 to mb-4 to adjust margin ever so slightly, and now we have to come in and adjust this. Not such a big deal, but I think as a general rule, it's good to start thinking about classes in a similar way to how a web developer does. so instead of hard-matching the exact ordering of 5+ classes, let's see if we can make the selector as selective, but robust to change, as possible.

datasets/ua/war_sanctions/crawler.py
nvmbrasserie: Since these are usually entities Ian the data I said "The entity to which..."

PR #1573, opened 2024-11-22
ru-egrul-internal-bucket and address parsing upd

nvmbrasserie:


datasets/ru/egrul/crawler.py
```diff
@@ -697,17 +698,17 @@ def crawl_index(context: Context, url: str) -> Set[str]:
...
-        if not link_url.startswith(url):
-            continue
-        if link_url.endswith(".zip"):
-            archives.add(link_url)
-            continue
-        archives.update(crawl_index(context, link_url))
-    return archives
+    with open(url, "r", encoding="utf-8") as f:
+        doc = html.fromstring(f.read())
+        doc.make_links_absolute(context.data_url)
```
jbothma, line 703: you're making the links absolute, relative to http://internal-data.opensanctions.org/ru_egrul/egrul.itsoft.ru/EGRUL_406/01.01.2022_FULL/

but you actually want to be making `fetch_internal_data()` calls using relative **paths** similar to the call for `ru_egrul/egrul.itsoft.ru/EGRUL_406/01.01.2022_FULL/index.html` e.g. `fetch_internal_data("ru_egrul/egrul.itsoft.ru/EGRUL_406/01.01.2022_FULL/EGRUL_FULL_2022-01-01_1.zip")`.

datasets/ru/egrul/crawler.py
```diff
@@ -688,25 +693,34 @@ def parse_examples(context: Context):
...
     Args:
         context: The processing context.
-        url: The URL to crawl.
+
     Returns:
-        A set of ZIP archive URLs.
+        A set of relative paths to the ZIP archives listed in the index.
     """
+    local_index_path = context.get_resource_path("index.html")
+    fetch_internal_data(
```
pudo, line 710: The list function for the internal bucket is now in `main`. We can use that to avoid having to parse the index HTML and just parse all the `.zip` in the folder :)

PR #1587, opened 2024-11-25
daily-fixes-nov-25

nvmbrasserie:
gender `Curis` is not there anymore

### DIFF
--- a/datasets/_externals/wikidata.yml
+++ b/datasets/_externals/wikidata.yml
@@ -654,7 +654,6 @@ lookups:
           - Kingdom of Numidia
           - Spanish protectorate in Morocco
           - State of Judaea
-          - Uyghur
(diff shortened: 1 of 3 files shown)


datasets/_externals/wikidata.yml
```diff
@@ -653,6 +653,8 @@ lookups:
           - Electorate of Mainz
           - Kingdom of Numidia
           - Spanish protectorate in Morocco
+          - State of Judaea
+          - Uyghur
```
jbothma, line 657: Removing Uyghur since in this case they referred to Uyghur language for citizenship. I removed the invalid statement in wikidata instead.

We might want to map Uyghur to China geographically, instead of nothing. But we can unpack that if it comes up again.
```suggestion
```

PR #1603, opened 2024-11-28
Adding Delaware medicare exclusions crawler

bgmello:


nvmbrasserie (changes requested):

datasets/us/de/med_exclusions/crawler.py
```diff
@@ -0,0 +1,89 @@
...
+    if row.get("npi") != "N/A":
+        npis = h.multi_split(row.pop("npi"), [";", "&", " and ", ","])
+        entity.add("npiCode", npis)
+    else:
+        row.pop("npi")
+
+    if row.get("license") != "N/A":
+        entity.add("description", "License Number: " + row.pop("license"))
+    else:
+        row.pop("license")
```
jbothma, line 54: We've been discussing internally that we don't want to build up `description` to be a bag of key-value pairs - that's made a mess of some of the upstream data and we don't want to repeat that. So let's put `license` into `idNumber` for now - that prop is fine for identifiers that don't fit into a more specific type like `registrationNumber` or `npiCode`

nvmbrasserie (changes requested):

datasets/us/de/med_exclusions/crawler.py
```diff
@@ -41,37 +41,51 @@ def crawl_item(row: Dict[str, str], context: Context):
...
-end_date = max(sanction.get('endDate'), None)
-if end_date is not None: 
-      target = datetime.strptime(reinstatement_date, "%Y-%m-%d") >= settings.RUN_TIME_ISO
-else:
-         target = True
+    reinstated_date = row.pop("reinstated_date")
+    h.apply_date(sanction, "endDate", reinstated_date)
+
+    end_date = sanction.get("endDate")
+    if end_date and reinstated_date not in [
```
bgmello, line 71: There are some reinstated dates that are like Suspended, Rescinded, Annulled. In this cases, I understand that the entity is not a target anymore. Correct? What I did was add that condition to the if statement.

nvmbrasserie: Since you've already set all these values to be `null`  in the lookups:  
```
        "Revoked",
        "Annulled",
        "Suspended",
        "Rescinded",
        "Preclusion"
```
this check `if end_date and reinstated_date not in [...]` won't make any difference. We can check it by printing both of them: 

```
    h.apply_date(sanction, "endDate", reinstated_date)
    print(reinstated_date)

    end_date = sanction.get("endDate")
    print(end_date)
```
we'll get: 
```
Indefinite
[]
Indefinite
[]
```
`end_date` is a processed `reinstated_date`. So, looking them up in the `yml` is enough

datasets/us/de/med_exclusions/crawler.py
```diff
@@ -0,0 +1,98 @@
...
+        "Revoked",
+        "Annulled",
+        "Suspended",
+        "Rescinded",
+        "Preclusion",
+    ]:
+        target = datetime.strptime(end_date[0], "%Y-%m-%d") >= datetime.today()
```
nvmbrasserie, line 78: It's better to compare with `settings.RUN_TIME_ISO` not with `datetime.today()`. They'll not always be the same

```suggestion
        target = datetime.strptime(end_date[0], "%Y-%m-%d") >= settings.RUN_TIME_ISO
```

datasets/us/de/med_exclusions/us_de_med_exclusions.yaml
```diff
@@ -0,0 +1,86 @@
...
+      - match: Revoked
+        value: null
+      - match: Annulled
+        value: null
+      - match: Suspended
+        value: null
+      - match: Rescinded
+        value: null
+      - match: Preclusion
+        value: null
```
nvmbrasserie, line 77: when you have several matches that should end up being nulls, you can also say: 
```
- match: 
       - Active
       - Probation
       - Revoked
       - Annulled
       - Rescinded
    value: null
```

nvmbrasserie (changes requested):

datasets/us/de/med_exclusions/crawler.py
bgmello: I'm getting an error saying that it can't find `settings`, is it some package?

nvmbrasserie: `from zavod import settings`

nvmbrasserie (approved):

datasets/us/de/med_exclusions/crawler.py
```diff
@@ -0,0 +1,91 @@
...
+        target = end_date[0] >= settings.RUN_TIME_ISO
+    else:
+        # It's not a target if the sanction was suspended
+        target = reinstated_date not in [
+            # "Revoked", usually meaning that the license was revoked
+            # "Annulled", usually meaning that the license was annulled
+            "Suspended",
+            "Rescinded",
+            "Preclusion",
+            "Probation",
```
jbothma, line 74: Suspended seems to be used in the comments to mean they are suspended from providing services. Preclusion seems to mean they are precluded from providing services. Probation seems to be the period during which they may not participate.

Rescinded is used in the comments to mean their sanction was rescinded - they are free to provide services.

So I think the only one to consider a target is `rescinded`

datasets/us/de/med_exclusions/crawler.py
jbothma: Suspended seems to be used in the comments to mean they are suspended from providing services. Preclusion seems to mean they are precluded from providing services. Probation seems to be the period during which they may not participate.

Rescinded is used in the comments to mean their sanction was rescinded - they are free to provide services.

So I think the only one to consider a target is rescinded

datasets/us/de/med_exclusions/crawler.py
```diff
@@ -0,0 +1,91 @@
...
+    entity.add("sector", row.pop("taxonomy"))
+    entity.add("idNumber", row.pop("dea"))
+    if npi != "N/A":
+        entity.add(
+            "npiCode",
+            h.multi_split(
+                npi,
+                [";", ",", "&"],
+            ),
+        )
```
jbothma, line 53: I think we can format this so it's not quite so extremely split into rows. The trailing comma stops black from reconsidering something to be on the same line. So I'd remove trailing commas and formatting again and then deciding if more effort is needed.

datasets/us/de/med_exclusions/us_de_med_exclusions.yaml
```diff
@@ -0,0 +1,80 @@
...
+  start: "2024-11-27"
+load_db_uri: ${OPENSANCTIONS_DATABASE_URI}
+ci_test: false
+summary: >
+  List of providers excluded from participating in the Delaware Medicaid program.
+description: |
+  This dataset provides details on health care providers who have been sanctioned
+  by Medicaid or the Office of Inspector General (OIG). It includes records of
+  both individual providers and entities, outlining their sanctions, descriptions
+  of violations, and associated regulatory actions.
```
jbothma, line 15: I get the idea that Medicaid isn't an organisation in its own right - it's legislation defining services that can be reimbursed, and each state has its own authorities who administer it.

```suggestion
  This dataset provides details on health care providers who have been excluded from
  participation in the Delaware Medicaid program. It lists provider names, identifiers
  such as NPI and License numbers, as well as the authority, duration, and nature of
  the sanction.
```

jbothma (commented):
Thanks guys - good discussion - we're learning constantly about doing this in better ways so I appreciate the exchanges.

I'll wrap up these last bits

jbothma:
https://github.com/opensanctions/opensanctions/pull/1661

PR #1609, opened 2024-11-29
fixes-nov-29

nvmbrasserie:


datasets/fr/tresor/crawler.py
```diff
@@ -8,6 +8,11 @@
 from zavod import Context, Entity
 from zavod import helpers as h
 
+HEADERS = {
+    "Accept": "application/json",
+    "User-Agent": "custom-agent",
```
jbothma, line 13: It's nice to include opensanctions.org in the user agent if we can make it work with that text in there. That way if we hurt them they can get in touch with us easily.

datasets/nz/russia_sanctions/nz_russia_sanctions.yml
```diff
@@ -42,6 +42,7 @@ lookups:
       - match:
           - unknown
           - 08/15/1977
+          - 45625
```
jbothma, line 45: fun fact: this is an excel date

```
 h.convert_excel_date(45625)
'2024-11-29T00:00:00'
```

datasets/fr/tresor/crawler.py
nvmbrasserie: It doesn't work unfortunately in this case

datasets/nz/russia_sanctions/nz_russia_sanctions.yml
nvmbrasserie: I suspected it, but couldn't find a way to verify it :)

PR #1627, opened 2024-12-04
fixes-dec-4

nvmbrasserie:


datasets/dk/pep/dk_pep.yaml
```diff
@@ -66,6 +66,7 @@ lookups:
           - Stillingsbetegnelse
           - roh
           - Stilling
+          - Sonne-Schmidt
```
jbothma, line 69: are you sure this a column heading and not a surname?

datasets/us/ia/med_exclusions/us_ia_med_exclusions.yaml
```diff
@@ -37,3 +37,9 @@ assertions:
     schema_entities:
       Organization: 400
       Person: 1000
+
+lookups:
+  sanction_end_date:  
+    options:  
+      - match: "2 Years"
+        value: 2026-08-06
```
jbothma, line 45: perhaps this needs to be `max(entity.get(start_date)) + timedelta(years=2)` and not 2 years after the start of the row that triggered this.

datasets/dk/pep/dk_pep.yaml
nvmbrasserie: probably a mistake on their side. it was there a couple of hours ago but it seems that they've already fixed it
Screenshot 2024-12-04 at 15 58 47

PR #1631, opened 2024-12-04
Add program and Improve end_date Handling for Sanction Status

nvmbrasserie:
Enhancements to h.make_sanction:
Add `program` as an optional argument to `h.make_sanction`

Then lookup based on the `program` to get `programId`  from `sanction.program` in the YAML configuration.

```
If `program` wasn't looked up: 
     context.log.warn("")

```

Sanction Status Handling:
Refines the logic for determining the active status of a sanction based on its `end_date`.
If `end_date` exists and is earlier than the current date, the sanction is marked **inactive**; otherwise, it remains **active**.

EU regimes: 

- https://data.europa.eu/apps/eusanctionstracker/
- https://sanctionsmap.eu/#/main?search=%7B%22value%22:%22%22,%22searchType%22:%7B%7D%7D
- https://data.europa.eu/apps/eusanctionstracker/individuals/

### DIFF
--- a/zavod/zavod/helpers/sanctions.py
+++ b/zavod/zavod/helpers/sanctions.py
@@ -48,20 +48,20 @@ def make_sanction(
     sanction.add("sourceUrl", dataset.url)
     if program is not None:
         sanction.set("program", program)
-    program_id = context.lookup_value("sanction.program", program_key)
-    if program_id is not None:
-        program_url = f"https://www.opensanctions.org/programs/{program_id}"
-        sanction.add("programUrl", program_url)
-        sanction.add("programId", program_id)
-    else:
-        context.log.warn(f"Program key '{program_key}' not found.", program=program)
+    if program_key is not None:
+        program_id = context.lookup_value("sanction.program", program_key)
+        if program_id is not None:
+            program_url = f"https://www.opensanctions.org/programs/{program_id}"
+            sanction.add("programUrl", program_url)
+            sanction.add("programId", program_id)
+        else:
+            context.log.warn(f"Program key {program_key!r} not found.", program=program)
     if start_date is not None:
         h.apply_date(sanction, "startDate", start_date)
     if end_date is not None:
         h.apply_date(sanction, "endDate", end_date)
         iso_end_date = max(sanction.get("endDate"))
-        end_date_obj = datetime.strptime(iso_end_date, "%Y-%m-%d")
-        is_active = end_date_obj >= settings.RUN_TIME
+        is_active = iso_end_date >= settings.RUN_TIME_ISO
         sanction.add("status", "active" if is_active else "inactive")
 
     return sanction
--- a/zavod/zavod/helpers/sanctions.py
+++ b/zavod/zavod/helpers/sanctions.py
@@ -1,6 +1,11 @@
 from typing import Optional
+from datetime import datetime
+
 from zavod.context import Context
 from zavod.entity import Entity
+from zavod import helpers as h
+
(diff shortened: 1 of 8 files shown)


zavod/zavod/helpers/sanctions.py
```diff
@@ -30,4 +43,19 @@ def make_sanction(
         sanction.add("country", dataset.publisher.country)
     sanction.add("authority", dataset.publisher.name)
     sanction.add("sourceUrl", dataset.url)
+
+    if program is not None:
+        sanction.set("program", program)
```
pudo, line 48: I'd strongly prefer if we used two separate fields handed into `h.make_sanction`: `program` to be set for the property and `program.key` to be used for the lookup. These /can/ be the same, but forcing them to be will be super complicated and messy. For one thing, it'll auto-activate lookups on all datasets where we use `program`.

zavod/zavod/helpers/sanctions.py
```diff
@@ -30,4 +43,19 @@ def make_sanction(
...
+        program_id = context.lookup_value("sanction", program)
```
pudo, line 49: Can we make it `sanction.program`? `sanction` on its own is not descriptive of what this does.

zavod/zavod/helpers/sanctions.py
```diff
@@ -30,4 +43,19 @@ def make_sanction(
...
+        if program_id is None:
+            context.log.warn(f"Program key '{program}' not found.")
+    if start_date is not None:
+        h.apply_date(sanction, "startDate", start_date)
+    if end_date is not None:
+        h.apply_date(sanction, "endDate", end_date)
+        iso_end_date = h.extract_date(context.dataset, end_date)[0]
+        end_date_obj = datetime.strptime(iso_end_date[:10], "%Y-%m-%d")
+        is_active = end_date_obj >= datetime.today()
```
pudo, line 58: Can we compare with `settings.RUN_TIME` not with `datetime.today()`? They'll not always be the same, and there's lots of timezone trouble in this code. 

FWIW it should be OK to compare ISO timestamps instead of parsing the date. And maybe instead of using `h.extract_date` here we can use `max(sanction.get('endDate'))` ?

pudo (changes requested):

pudo (commented):
This looks great now. Would you be up for adding it to one or two more crawlers before we merge? 

One idea: We can auto derive `programUrl` from `programId` via ` f"https://www.opensanctions.org/programs/{programId}" ` : )

zavod/zavod/helpers/sanctions.py
```diff
@@ -30,4 +46,22 @@ def make_sanction(
...
     sanction.add("sourceUrl", dataset.url)
+
+    if program is not None:
+        sanction.set("program", program)
+    if program_key is not None:
+        program_id = context.lookup_value("sanction.program", program_key)
+        program_url = f"https://www.opensanctions.org/programs/{program_id}"
+        sanction.add("programUrl", program_url)
+        if program_id is None:
+            context.log.warn(f"Program key '{program_key}' not found.")
```
pudo, line 57: This would make urls a la `https://www.opensanctions.org/programs/None` I think. How about:

```
program_id = context.lookup_value("sanction.program", program_key)
if program_id is not None:
    program_url = f"https://www.opensanctions.org/programs/{program_id}"
    sanction.add("programUrl", program_url)
    sanction.add("programId", program_id)
else:
     context.log.warn(f"Program key '{program_key}' not found.", program=program)

PR #1644, opened 2024-12-06
post-validation-fixes

nvmbrasserie:


datasets/ch/seco_sanctions/ch_seco_sanctions.yml
```diff
@@ -104,3 +103,8 @@ lookups:
         value: male
       - match: Female
         value: female
+  type.identifier: 
+    options: 
+      - match: 3232824038
+        prop: taxNumber
+        value: 3232824038
```
jbothma, line 110: I don't think you need to repeat the value if you just want to map the exact value to a different prop. https://github.com/opensanctions/opensanctions/blob/main/datasets/us/ofac/us_ofac_sdn.yml#L430

PR #1645, opened 2024-12-06
ru_nsd_isin post validation

nvmbrasserie:


datasets/ru/nsd_isin/crawler.py
```diff
@@ -67,7 +67,8 @@ def crawl_item(context: Context, url: str):
 
     issuer = context.make("LegalEntity")
     inn_code = first(values["issuer"].get("innCode", []))
```
pudo, line 69: what if actually checked if it's valid here, e.g. `is_inn_code = INN.is_valid(inn_code)` (after importing `from rigour.ids import INN`) and then used that instead of the length games?

PR #1654, opened 2024-12-09
validation-fixes-09-12

nvmbrasserie:


jbothma:
Let's keep the values that aren't blatant nonsense like `1` or `12345` in the most appropriate generic identifier prop.

jbothma:
Thank you!

PR #1671, opened 2024-12-12
Helena Kennedy Centre Companies Operating in the Uyghur Region

nvmbrasserie:
fixes opensanctions/crawler-planning#284


pudo:
Nice work splitting the addresses! I'm wondering if naming this `cn_` is a bit misleading since the prefix normally refers to the publisher country. Maybe `shu_`?

datasets/_global/shu_uyghur_companies/shu_uyghur_companies.yml
```diff
@@ -0,0 +1,81 @@
...
+  Its purpose is to provide a better understanding of the economic landscape in the 
+  Uyghur Region amid ongoing human rights concerns. This list represents publicly 
+  available information and is not an original research product of the authors or 
+  of Sheffield Hallam University.
+
+  **Note:** The list is drawn from a human resources website, accessed via the internet in 
+  December 2021. This list is not exhaustive, nor does it provide data on companies that were 
+  registered, de-registered, or otherwise changed status after that time.
+  English translation is machine-generated and may not be accurate.
+  [Source: Helena Kennedy Centre for International Justice](https://www.shu.ac.uk/helena-kennedy-centre-international-justice/research-and-projects/all-projects/useful-resources)
```
jbothma, line 26: I wonder if we should assert a hash and indicate the static nature of this website in the metadata. Then the hash assertion will tell is if they do another set of research and update it letting us know to update the metadata.

datasets/_global/shu_uyghur_companies/crawler.py
```diff
@@ -0,0 +1,42 @@
...
+
+def crawl(context: Context):
+    doc = context.fetch_html(context.data_url, cache_days=1)
+    url = doc.xpath(
+        './/a[contains(text(), "Companies Operating in the Uyghur Region")]/@href'
+    )
+    path = context.fetch_resource("source.xlsx", url[0])
+    workbook: openpyxl.Workbook = openpyxl.load_workbook(path, read_only=True)
+    for row in h.parse_xlsx_sheet(
+        context, sheet=workbook["1. Companies Operating in XUAR"], skiprows=1
```
jbothma, line 14: I think it's a good idea to get into a habit of asserting that new worksheets aren't introduced that we would have wanted to scrape. One way of doing that is asserting that set(workbook.sheets) = {'1. Companies Operating in XUAR', '2. Export Relevant Categories'}

datasets/_global/shu_uyghur_companies/crawler.py
```diff
@@ -0,0 +1,42 @@
...
+
+        entity = context.make("Company")
+        entity.id = context.make_id(name, addr, sector)
+        entity.add("name", name, lang="zhu")
+        entity.add("name", name_en, lang="eng")
+        entity.add("address", addr, lang="zhu")
+        entity.add("address", addr_en, lang="eng")
+        entity.add("sector", sector, lang="zhu")
+        entity.add("sector", sector_en, lang="eng")
+        entity.add("topics", "export.control")
```
jbothma, line 33: can some of these fit on one neat row if we pop and add on the same line? It would make the code a bit more concise and that's what we usually do, except for readability (when the lines break awkwardly) and when a value is used more than once

datasets/_global/shu_uyghur_companies/crawler.py
```diff
@@ -0,0 +1,42 @@
...
+
+        address_ent = h.make_address(context, full=addr, city=city, lang="zhu")
+        address_en_ent = h.make_address(context, full=addr_en, city=city_en, lang="eng")
+        h.copy_address(entity, address_ent)
+        h.copy_address(entity, address_en_ent)
```
jbothma, line 38: unless there's a specific reason not to, let's use h.apply_address which will 1) make the entitiy, 2) emit the entity, 3) set it in addressEntity, 4) set the full address in the address prop, all in one

datasets/_global/shu_uyghur_companies/crawler.py
jbothma, line 33: decision now is to do both files, but only tag the files from the second file, and distinguish the entities with the entity program prop with something like `Companies operating in Uyghur region` and
 `Companies Named in Media and Academic Reports as engaging in Labour Transfers or other XUAR Government Programs`

datasets/_global/shu_uyghur_companies/shu_uyghur_companies.yml
```diff
@@ -0,0 +1,81 @@
+title: Helena Kennedy Centre Companies Operating in the Uyghur Region
+entry_point: crawler.py
+prefix: shu-xj
+coverage:
+  frequency: daily
+  start: 2024-12-12
+load_db_uri: ${OPENSANCTIONS_DATABASE_URI}
+ci_test: false
+summary: >-
+  Over 50,000 companies operating across various sectors in the Uyghur Region.
```
jbothma, line 10:
```suggestion
  Companies operating in the Uyghur Region including some identified as engaging in labour transfer or other XUAR Government Programs
```

jbothma (commented):
Nice and clean so far!

datasets/_global/shu_uyghur_companies/shu_uyghur_companies.yml
```diff
@@ -0,0 +1,104 @@
+title: Helena Kennedy Centre Companies Operating in the Uyghur Region
```
pudo, line 1: Maybe just: `Companies Operating in the Uyghur Region` ?

datasets/_global/shu_uyghur_companies/shu_uyghur_companies.yml
```diff
@@ -0,0 +1,104 @@
...
+
+  **Note:** This data is static and will be updated when new files become available.
+  This list is not exhaustive, nor does it provide data on companies that were 
+  registered, de-registered, or otherwise changed status after that time.
+  English translation is machine-generated and may not be accurate.
+  [Source: Helena Kennedy Centre for International Justice](https://www.shu.ac.uk/helena-kennedy-centre-international-justice/research-and-projects/all-projects/useful-resources)
+url: https://www.shu.ac.uk/helena-kennedy-centre-international-justice/research-and-projects/all-projects/useful-resources
+publisher:
+  name: Helena Kennedy Centre for International Justice at Sheffield Hallam University
+  acronym: HKC
```
pudo, line 37: SHU

datasets/_global/shu_uyghur_companies/crawler.py
```diff
@@ -0,0 +1,156 @@
...
+    context.audit_data(row)
+
+
+def crawl(context: Context):
+    doc = context.fetch_html(context.data_url, cache_days=1)
+    companies_url = doc.xpath(
+        './/a[contains(text(), "Companies Operating in the Uyghur Region")]/@href'
+    )
+    h.assert_url_hash(
+        context, companies_url[0], "e71aab15641be03abeb8b6fce1564aad8714a124"
```
pudo, line 146: What's the rationale for putting a hash on these - are we sure the parser will break?

pudo (changes requested):

datasets/_global/shu_uyghur_companies/crawler.py
nvmbrasserie: https://github.com/opensanctions/opensanctions/pull/1671#discussion_r1883977497 JD suggested to add it so that we are notified when there's another set of research and update the metadata accordingly

PR #1682, opened 2024-12-13
playwright for browser scraping

jbothma:


datasets/us/il/med_exclusions/crawler.py
```diff
@@ -0,0 +1,126 @@
...
+import zipfile
+
+from rigour.mime.types import XLSX
+
+from zavod.shed.playwright import click_and_download
+from zavod import Context, helpers as h
+from zavod.archive import dataset_data_path
+
+BRIGHT_USERNAME = os.environ.get("BRIGHTDATA_BROWSER_USERNAME")
+BRIGHT_PASSWORD = os.environ.get("BRIGHTDATA_BROWSER_PASSWORD")
```
jbothma, line 16: this is split in two because geolocation is controlled via suffixes like -country-cl added to the username

PR #1694, opened 2024-12-18
dbas-removal

nvmbrasserie:


datasets/us/ar/med_exclusions/crawler.py
```diff
@@ -38,49 +36,32 @@ def crawl_item(row: Dict[str, str], context: Context):
...
-            link.id = context.make_id(company.id, dba_person.id)
-            link.add("object", company)
-            link.add("subject", dba_person)
-            link.add("role", "d/b/a")
-            context.emit(dba_person)
-            context.emit(link)
-
-        sanction = h.make_sanction(context, company)
+        if "d/b/a" in facility_name:
+            dba = context.lookup_value("names", facility_name)
```
pudo, line 48: So `context.lookup_value` will identify a single value from the lookup. To get multiple values, use: 
```
result = context.lookup("names", facility_name)
if result is not None:
   result.values
```

PR #1704, opened 2024-12-20
War-and-sanctions-expansion

nvmbrasserie:


datasets/ua/war_sanctions/crawler.py
```diff
@@ -63,6 +83,57 @@ def crawl_index_page(context: Context, index_page, data_type, program):
...
+            apply_dob_pob(captain, dob_pob)
+
+        sanction = h.make_sanction(context, captain)
+        sanction.add("program", program)
+
+        context.emit(captain, target=True)
+        context.emit(sanction)
+
+        vessel = context.make("Vessel")
+        vessel.id = context.make_id(vessel_name, vessel_category)
```
pudo, line 123: Do we have the URL for the vessel page that we could key on?

datasets/ua/war_sanctions/crawler.py
```diff
@@ -34,6 +39,19 @@ def extract_label_value_pair(label_elem, value_elem, data):
     return label, value
 
 
+def apply_dob_pob(entity, dob_pob):
+    if dob_pob:
```
pudo, line 43: Just an aesthetic point: if you do `if dob_pob` then the whole rest of the function is one indented. If, instead, you do `if not dob_pob` you can just put a `return` and keep the function more legible.

pudo (commented):
Looks really good, let's merge next week

PR #1710, opened 2024-12-23
Slovakia public officials database

nvmbrasserie:
fixes opensanctions/crawler-planning#244


jbothma:
I think there's a position we need to trim up a bit

image

image

datasets/sk/public_officials/crawler.py
```diff
@@ -0,0 +1,99 @@
...
+    person.add("position", position)
+    person.add("topics", "role.pep")
+    person.add("sourceUrl", href)
+
+    position = h.make_position(
+        context,
+        position,
+        lang="slk",
+        country="SK",
+    )
```
jbothma, line 58: could be nice to translate the position label as we do for georgian positions

jbothma:
Which option is this using? I think that has implications for our assumption of whether occupancy says the positions are current (default when there's no end date). That, and the (default latest) year on the profile

image

image

datasets/sk/public_officials/crawler.py
```diff
@@ -0,0 +1,99 @@
...
+
+    crawl_person(context, data, href, name_raw)
+
+
+def crawl_person(context: Context, data, href, name_raw):
+    # name = data.pop("titul, meno, priezvisko") # name with title
+    # year = data.pop("oznámenie za rok")
+    position = data.pop("vykonávaná verejná funkcia")
+    person = context.make("Person")
+    person.id = context.make_id(name_raw)
```
jbothma, line 46: I think we need more disambiguating information in the ID than just name, for such a broad pool of people.

Perhaps Internal number is consistent for a given person each year? e.g. ZP235369
Just confirm.

The alternative is urlparse-ing the URL to get userid e.g. Peter.Demcak, but I think using internal ID is fine since they're publishing it front and centre.

datasets/sk/public_officials/crawler.py
```diff
@@ -0,0 +1,99 @@
...
+    categorisation = categorise(context, position, is_pep=True)
+    if not categorisation.is_pep:
+        return
+
+    occupancy = h.make_occupancy(
+        context,
+        person,
+        position,
+        categorisation=categorisation,
+    )
```
jbothma, line 68: can we assume all the positions are current? see https://github.com/opensanctions/opensanctions/pull/1710#issuecomment-2566356934

jbothma:
This should be multiple positions and occupancies, not all in one position 
image

datasets/sk/public_officials/crawler.py
nvmbrasserie: yes, internal number is consistent for a given person each year, just added it to the id

datasets/sk/public_officials/crawler.py
```diff
@@ -0,0 +1,120 @@
...
+        label = None  # Reset the label for the next row
+    crawl_person(context, data, href, name_raw)
+
+
+def crawl_person(context: Context, data, href, name_raw):
+    # name = data.pop("titul, meno, priezvisko") # name with title
+    # year = data.pop("oznámenie za rok")
+    position_slk = data.pop("vykonávaná verejná funkcia")
+    int_id = data.pop("Interné číslo")
+    print(int_id)
```
jbothma, line 61:
```suggestion
```

datasets/sk/public_officials/sk_public_officials.yml
```diff
@@ -0,0 +1,50 @@
...
+      Person: 3000
+lookups:
+  type.name:
+    options:
+      - contains: "verejný funkcionár, ktorý nie je uvedený v písmenách a) až zo)"
+        value: null
+  type.string: 
+    options: 
+      - contains: "verejný funkcionár, ktorý nie je uvedený v písmenách a) až zo)"
+        value: null
```
jbothma, line 50: this will make the Position:name prop empty for the positions that contain this. This won't just strip out the matching bit. So I think we'll have to do this in Python

datasets/sk/public_officials/crawler.py
jbothma: with

```diff
--- a/datasets/sk/public_officials/crawler.py
+++ b/datasets/sk/public_officials/crawler.py
@@ -23,8 +23,11 @@ def parse_details(context: Context, link_el):
         return
+    year = doc.find(".//select[@id='_sectionLayoutContainer_ctl01_OznameniaList']/option[@selected='selected']").text
+    if year != "2023":
+        print(year, href)
```

I could see that https://www.nrsr.sk/web/Default.aspx?sid=vnf/oznamenie&UserId=Tomas.Hellebrandt and https://www.nrsr.sk/web/Default.aspx?sid=vnf/oznamenie&UserId=Jan.Imrich only have notices for 2022. But we can't assume that their positions ended then because 

a. tomas only entered the position in 2023 (maybe the notices are for non-calendar years like 2022-2023 or something) https://en.wikipedia.org/wiki/Tom%C3%A1%C5%A1_Hellebrandt
b. we don't know if they simply haven't filed a notice, or what.

jbothma: so I think let's override status in make_occupancy to OccupancyStatus.UNKNOWN and add the year of the notice to Occupancy:date - then users can at least see as at that date, it was considered to be the case...according to the data at least.

jbothma: so just fr completeness add `no_end_date_implies_current=False` to make_occupancy as well

datasets/sk/public_officials/sk_public_officials.yml
```diff
@@ -0,0 +1,53 @@
...
+load_db_uri: ${OPENSANCTIONS_DATABASE_URI}
+summary: >
+  List of public officials in Slovakia, including their names, positions,
+  and other relevant information.
+description: |
+  The dataset contains information about public officials in Slovakia, including
+  their names, positions, and other relevant information. The data is collected
+  from the official website of the National Council of the Slovak Republic.
+publisher:
+  name: National Council of the Slovak Republic
```
jbothma, line 16:
```suggestion
  name: Národná rada Slovenskej republiky
  name_en: National Council of the Slovak Republic
```

PR #1713, opened 2024-12-24
Suggestion: Fix extraction of 'LegAddition' tag to handle nested child elements

touchyongan:
### Issue Description

While running the dataset `gb/proscribed_orgs/crawler.py`, I noticed that the current implementation only extracts the top-level text from the LegAddition tag in the HTML structure. However, some LegAddition tags contain nested child elements that are not being extracted, resulting in incomplete data.

Example snippet of the HTML content:

```
<li>
    <p class="LegListTextStandard LegLevel3">
        <span class="LegAddition">
            <span class="LegChangeDelimiter">[</span>
            <a class="LegCommentaryLink" href="#commentary-key">
                F4
            </a>
            <span class="LegSubstitution">
                Harakat al Muqawama al-Islamiyya (Hamas)
            </span>
            <span class="LegChangeDelimiter">]</span>
        </span>
    </p>
</li>
```

The current implementation misses nested elements like LegSubstitution within LegAddition.

### Suggested Fix

To address this, I added a new recursive method that processes the nested child elements inside the LegAddition tag. 

Here’s the implementation:

```
def extract_leg_list_text_standard_info(context: Context,
                                        page: HtmlElement,
                                        entry: HtmlElement,
                                        text: List[str],
                                        change_stack: List[str]):
    for el in entry:
        classnames = el.get("class").split()
        if "LegChangeDelimiter" in classnames:
            bracket = el.text.strip()
            if bracket == "[":
                link = el.getnext()
                comment_id = link.get("href")[1:]
                commentary = page.find(f".//div[@id='{comment_id}']")
                comment_text = commentary.text_content()
                context.log.debug(f"Start subgroup: {comment_text}")
                change_stack.append(comment_text)
            elif bracket == "]":
                crawl_group(context, "".join(text), change_stack)
                context.log.debug("End subgroup")
                change_stack.pop()
                text.clear()
        elif "LegAddition" in classnames:
            context.log.debug(f"Addition: {el.text}")
            if el.text is not None:
                text.append(el.text)
            elif len(el):
                extract_leg_list_text_standard_info(context, page, el, text, change_stack)
        elif "LegSubstitution" in classnames:
            context.log.debug(f"Substitution: {el.text}")
            if el.text is not None:
                text.append(el.text)
```

### DIFF
--- a/datasets/gb/proscribed_orgs/crawler.py
+++ b/datasets/gb/proscribed_orgs/crawler.py
@@ -19,7 +19,7 @@ ALIAS_RE = re.compile(r"\s+\(([^)]+)\)")
 JUNK = " ,.;\t\r\n"
 
 
-def parse_comment(context: Context, text: str) -> Optional[str | None]:
+def parse_comment(text: str) -> Optional[str | None]:
     """
     Parse stack of  commentary from the act looking for the most
     recent date when groups were added.
@@ -39,7 +39,7 @@ def crawl_group(context: Context, text: str, change_stack: List[str]):
         return
     add_date = None
     for item in change_stack:
-        add_date = parse_comment(context, item)
+        add_date = parse_comment(item)
         if add_date:
             break
     context.log.info(f"Adding group: {text} (starting {add_date})")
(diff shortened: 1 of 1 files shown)


jbothma:
Thank you for identifying the problem and for the thorough and clear pull request! This was a pleasure to receive.

I've simplified the code right down to make it much less likely that we miss out on names in future. That means dropping the edit history / start dates, but I think the priority of extracting all names is much higher. https://github.com/opensanctions/opensanctions/pull/1720

Thanks again for raising that this was missing data.

PR #1717, opened 2024-12-27
cn_peoples_congress_wikipedia expansion

nvmbrasserie:


datasets/cn/peoples_congress_wikipedia/crawler.py
```diff
@@ -119,6 +161,60 @@ def parse_table(
...
+                label = strip_note(eltree.text_content())
+                headers.append(context.lookup_value("headers", label))
+            continue
+
+        cells = row.findall("./td")
+        row_data = {}
+
+        cell_index = 0  # To track which header corresponds to which cell
+        for idx, cell in enumerate(cells):
+            while cell_index in rowspan_cells and rowspan_cells[cell_index][1] > 0:
```
jbothma, line 186: I'm not sure I understand this - why two loops? Can you explain how it works in words, perhaps with excel style cell references for the example?

    for each cell in the row
        for each what?

datasets/cn/peoples_congress_wikipedia/crawler.py
```diff
@@ -139,3 +235,19 @@ def crawl(context: Context):
     for id, count in ids.items():
         if count > 1 and id not in IGNORE_DUPES:
             context.log.info(f"ID {id} emitted {count} times")
+    for h3 in doc.findall(".//h3"):
+        delegation_name = None
+        h3_text = h3.text_content().strip()
+        if any(heading in h3_text for heading in CHANGES_IN_REPRESENTATION):
+            print(f"Match found: {h3_text}")
```
jbothma, line 242: let's try not to merge prints in prod code

datasets/cn/peoples_congress_wikipedia/crawler.py
```diff
@@ -119,6 +161,60 @@ def parse_table(
         yield row
 
 
+def parse_table_with_rowspan(
+    context: Context,
+    table: HtmlElement,
+) -> Generator[Dict[str, HtmlElement], None, None]:
+    headers = None
+    rowspan_cells = {}  # Dictionary to keep track of cells with rowspan
```
jbothma, line 169: I think it'd be good to document a data structure underpinning something like this - it looks like keys are column indices and values are tuples where the first element is what? and the second element is the span value?

datasets/cn/peoples_congress_wikipedia/crawler.py
jbothma: Perhaps the general mechanism could be documented in the function-level docstring?

datasets/cn/peoples_congress_wikipedia/crawler.py
```diff
@@ -71,7 +104,8 @@ def crawl_item(
         entity,
         position,
         False,
-        start_date="2023",
+        start_date=date_of_by_election or "2023",
+        end_date=date_of_resignation or date_of_dismissal or date_of_death,
         categorisation=categorisation,
         status=OccupancyStatus.UNKNOWN,
```
jbothma, line 110: this overrides the status. So to get ENDED, only set status to non-None if you don't have an end date.

datasets/cn/peoples_congress_wikipedia/crawler.py
```diff
@@ -9,7 +9,17 @@
 from zavod.shed.trans import apply_translit_full_name
 
 
-REGEX_DELEGATION_HEADING = re.compile(r"(\w+)（\d+名）$")
+REGEX_DELEGATION_HEADING = re.compile(r"(\w+)（\d+名）$")  #
+REGEX_BRACKETS = re.compile(r"\[.*?\]")
+CHANGES_IN_REPRESENTATION = [
+    "补选",  # by-election
+    # # Not supporting this for now because of the data structure
```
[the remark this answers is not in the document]
nvmbrasserie: these are just reassigned people, we're still getting them from the original crawl, I'll remove it from this list

datasets/cn/peoples_congress_wikipedia/crawler.py
```diff
@@ -27,14 +35,36 @@ def clean_text(text: str) -> str:
...
+
+
 def crawl_item(
     context: Context,
     input_dict: dict,
     delegation: str,
 ):
     name = clean_text(input_dict.pop("name").text_content())
+    if "[" in name:
+        name = name.split("[", 1)[0]
```
jbothma, line 57: Alternatively we could do

```
reference = name_el.find(".//sup")
if reference:
    reference.getparent().remove(reference)
name = name_el.text_content()
```

datasets/cn/peoples_congress_wikipedia/crawler.py
```diff
@@ -139,3 +178,13 @@ def crawl(context: Context):
...
+    for h3 in doc.findall(".//h3"):
+        delegation_name = None
+        h3_text = h3.text_content().strip()
+        if any(heading in h3_text for heading in CHANGES_IN_REPRESENTATION):
+            table = h3.getparent().getnext()
+            if table.tag != "table":
+                table = table.getnext()
+            assert table.tag == "table"
+            for row in parse_table(context, table):
+                crawl_item(context, row, delegation_name)
```
jbothma, line 190: we can merge this with the block above by adding `any(heading in h3_text for heading in CHANGES_IN_REPRESENTATION)` to the `if not delegation_match:` then we only have the horrible code block once, right?

datasets/cn/peoples_congress_wikipedia/crawler.py
```diff
@@ -124,17 +172,25 @@ def crawl(context: Context):
...
-        delegation_name = delegation_match.group(1)
-        table = h3.getparent().getnext().getnext()
+
         if table.tag != "table":
             table = table.getnext()
         assert table.tag == "table"
         for row in parse_table(context, table):
             id = crawl_item(context, row, delegation_name)
-            ids[id] += 1
+            ids[id] = ids.get(id, 0) + 1
```
jbothma, line 192: ids is a `defaultdict` initialising values using `int()` i.e. `0` so we don't have to initialise it here

```suggestion
            ids[id] += 1
```

jbothma (approved):

PR #1727, opened 2025-01-02
Croatia State Registry of Public Officials extension

nvmbrasserie:


jbothma (commented):
I've done some work to simplify this a bit. I think we overcomplicated it at the start so it's been tricky to see opportunities. Here are some notes driving the changes

datasets/hr/public_officials/crawler.py
```diff
@@ -7,36 +7,65 @@
...
+    "Secondary Position Start Date",
+    "Secondary Position End Date",
+]
+
 
 def make_position_name(data: dict) -> Optional[str]:
-    title = data.pop("dužnost")
-    legal_entity_name = data.pop("Pravna osoba u kojoj obnaša dužnost")
+    title = data.pop("Position")
+    legal_entity_name = data.pop("Legal Entity")
```
jbothma, line 68: while we're changing all the keys, let's snake_case like we normally do

PR #1734, opened 2025-01-06
Montenegro Public Officials’ Income and Asset Declarations

nvmbrasserie:
fixes opensanctions/crawler-planning#286

depends on https://github.com/alephdata/followthemoney/pull/1627


datasets/me/acp_peps/crawler.py
```diff
@@ -0,0 +1,87 @@
...
+        latest_date = max(dates)
+        return latest_date.strftime("%Y-%m-%d")
+    return None
+
+
+def crawl_person(context: Context, person):
+    name = person.pop("imeIPrezime")
+    position = person.pop("nazivFunkcije")
+    dates = person.pop("izvjestajImovine")
+    latest_date = get_latest_date(dates)
```
jbothma, line 29: do you want to see if within a couple of hours you can:

- take the latest report id e.g. `stariIzvjestaj: 84990`
- request the CSV dump of the report using the ID https://portal.antikorupcija.me:9343/acamPublic/izvestajDetailsCSV.json?izvestajId=84990
- open the public functions file `csv_funkcije_funkcionera_....csv` (https://github.com/opensanctions/opensanctions/blob/main/datasets/_global/icij_offshoreleaks/crawler.py#L45)
- try and generate a position and occupancy using full start/end dates for each entry in that?

datasets/me/acp_peps/crawler.py
```diff
@@ -0,0 +1,87 @@
...
+            no_end_implies_current=False,
+            categorisation=categorisation,
+            status=OccupancyStatus.UNKNOWN,
+        )
+        occupancy.add("date", latest_date)
+
+        context.emit(position)
+        context.emit(occupancy)
+
+    context.emit(entity, target=True)
```
jbothma, line 65: only emit the person if there was an occupancy for them, and let make_occupancy add the topic. Even though the current implementation will always make an occupancy, it's good to still follow the pattern  because if we change the implementation later to use start/end dates (which means make_occupancy decides whether they're still a PEP), the logic for deciding whether to emit the person is still valid.

datasets/me/acp_peps/me_acp_peps.yml
```diff
@@ -0,0 +1,29 @@
+title: Declarable Montenegrin Public Officials
```
jbothma, line 1: We try to start with the country name in its basic form, then datasets group nicely in an alphabetic list

```suggestion
title: Montenegro Declarable Public Officials
```

datasets/me/acp_peps/crawler.py
jbothma: it looks like this wont work for everyone so maybe just do a test of what proportion we can get these data IDs for before doing a lot of work on the zip and csv 
image

jbothma: the family members data is also juicy - they get the RCA topic

datasets/me/acp_peps/crawler.py
```diff
@@ -0,0 +1,162 @@
...
+        with ZipFile(zip_path, "r") as zip:
+            # Find the file name that matches the pattern
+            file_name = next(
+                (
+                    name
+                    for name in zip.namelist()
+                    if name.startswith("csv_funkcije_funkcionera_")
+                ),
+                None,
+            )
```
jbothma, line 68: This is on the edge of being too clever. But It works and makes sense once I've processed what it does, and might make for tidy code once the expression is broken up a bit, so I think it's ok.

The nesting is a bit awkward to read so I'd suggest assigning the filtered list first and calling next on the next line or something?

datasets/me/acp_peps/crawler.py
```diff
@@ -0,0 +1,162 @@
...
+                latest_report_id = entry.get("stariIzvjestaj")
+                break
+        else:
+            # If all entries have -1, set the value to None
+            latest_report_id = None
+
+    return latest_date, latest_report_id
+
+
+def fetch_and_process_csv(context: Context, latest_report_id: int):
```
jbothma, line 52: I would just call this fetch_csv_rows - process sounds like we're doing a bit more with it than extracting it from the csv

datasets/me/acp_peps/crawler.py
```diff
@@ -0,0 +1,162 @@
...
+def extract_latest_filing(
+    dates: List[Dict[str, object]]
+) -> Tuple[Optional[str], Optional[int]]:
+    if not dates:
+        return None, None
+
+    # Sort the dates in descending order
+    dates_sorted = sorted(
+        dates,
+        key=lambda item: datetime.strptime(item["datum"], "%Y-%m-%d"),
```
jbothma, line 29: if they're already in this format, they're already sortable and there's no need to parse them, right? That could simplify this code a touch.

datasets/me/acp_peps/crawler.py
```diff
@@ -0,0 +1,162 @@
...
+        reverse=True,
+    )
+
+    # Assume the first entry is the latest
+    latest_entry = dates_sorted[0]
+    latest_date: Optional[str] = latest_entry.get("datum")
+    latest_report_id: Optional[int] = latest_entry.get("stariIzvjestaj")
```
jbothma, line 36: I think this can be handled by the loop, cant it?

```
for entry in dates_sorted:
    if report_id := entry.get("starIzcjestaj") != "-1":
        return entry.get(datum), report_id
return None, None # or just return the entry?

File: datasets/me/acp_peps/crawler.py [hunk 8]
```diff
@@ -0,0 +1,162 @@
...
+
+
+def crawl_person(context: Context, person):
+    name = person.pop("imeIPrezime")
+    # position = person.pop("nazivFunkcije")
+    dates = person.pop("izvjestajImovine")
+    latest_date, latest_report_id = extract_latest_filing(dates)
+    report_details = fetch_and_process_csv(context, latest_report_id)
+    if not report_details:
+        return
```
jbothma, line 90: does this mean we're skipping people who don't have reports? It'd be nice to keep the old logic because I think there were recent entries that didn't have the full details - only the function in the index data like you originally had it.

datasets/me/acp_peps/crawler.py
```diff
@@ -0,0 +1,162 @@
...
+
+    for row in report_details:
+        person_name = row.pop("FUNKCIONER_IME")
+        person_surname = row.pop("FUNKCIONER_PREZIME")
+        function = row.pop("FUNKCIJA")
+        # organization = row.pop("ORGANIZACIJA")
+        start_date = row.pop("DATUM_POCETKA_OBAVLJANJA", None)
+        end_date = row.pop("DATUM_PRESTANKA_OBAVLJNJA")
```
jbothma, line 98: so this just leaves the last row's function assigned to function, right? What I usually do is make an empty set, then in a loop over positions, add to the set if it passes make_position, categorise and make_occupancy https://github.com/opensanctions/opensanctions/blob/me-income-and-asset-reports/datasets/hr/public_officials/crawler.py#L148

and then emit the person (and the position entities) if there were any position entities in the set.

jbothma (commented):
looking good! There are some simplifications we can make, I think, and it'd be good to also capture the people who don't have any detailed reports (or am I mistaken in thinking there are some?)
But nice work so far!

datasets/me/acp_peps/crawler.py
```diff
@@ -0,0 +1,212 @@
...
+    )
+    apply_translit_full_name(
+        context, position, "cnr", position_name, TRANSLIT_OUTPUT, POSITION_PROMPT
+    )
+
+    categorisation = categorise(context, position, is_pep=True)
+    # Check categorisation topics match and fetch relatives
+    if "gov.national" in categorisation.topics:
+        relatives = fetch_csv(context, report_id, "csv_clanovi_porodice_")
+        crawl_relative(context, entity, relatives)
```
jbothma, line 139: we only want to emit the relatives if the PEP is still a PEP, so we need `if "gov.national" in categorisation.topics and entities`.

Maybe make_affiliation_entities can make a set of categorisation topics and return a tuple `(entities, topics)` then we call crawl_relative from crawl_person instead? Because just moving this down in this function will make us crawl the relative file for each function.

datasets/me/acp_peps/crawler.py
```diff
@@ -0,0 +1,212 @@
...
+    if occupancy:
+        # Switch to declarationDate, once it's introduced in FtM
+        occupancy.add("date", filing_date)
+        entities.extend([position, occupancy])
+    return entities
+
+
+def crawl_person(context: Context, person):
+    full_name = person.pop("imeIPrezime")
+    # position = person.pop("nazivFunkcije")
```
jbothma, line 161: reminder to use this as a fallback if there aren't downloadable declarations. But lets get the declaration crawling dialed first.

jbothma (commented):
Really good progress

jbothma:
```
crawl_person
    try and get latest filing id

    get family rows
    filter row where relation is Official (Funkcioner)
    filter rows where relation is not official
    if there's a family row for the official,
        make the entity from that make_id(name, city)
    else 
        make the entity from what's in the index json (make_id(name, function label from index)

    get function rows
    if there's a functions file
        make position entities from that, aggregate entities and topics
    else
        make position entity from the index json function label

    if there if there's a national position topic
        crawl relatives (make and emit)

    if there are current PEP positions
        emit person and pep entities
```

jbothma (commented):
I think these are the last bits of cleanup

datasets/me/acp_peps/crawler.py
```diff
@@ -0,0 +1,215 @@
...
+                )
+                return []
+
+            with zip.open(file_name) as zfh:
+                fh = io.TextIOWrapper(zfh, encoding="utf-8-sig")
+                reader = DictReader(fh, delimiter=",", quotechar='"')
+                return list(reader)
+    except BadZipFile:
+        context.log.warning(f"Failed to open {zip_path} as a ZIP file. Skipping.")
+        return []
```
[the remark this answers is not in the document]
jbothma: so when this fails, it seems to be a JSON response looking something liek {status: error}
I've changed to info level, and we warn if _no_ zip files were valid or contained rows we care about

PR #1737, opened 2025-01-07
Sg-gov-dir-missing-peps-categorization

nvmbrasserie:


jbothma:
on main as of 5 minutes ago, I got this delta:
```
added=96 dataset=sg_gov_dir deleted=0 metric=delta_counts modified=1
```

jbothma:
Let's add this:

```diff
diff --git a/datasets/sg/gov_dir/crawler.py b/datasets/sg/gov_dir/crawler.py
index 91baf053..c0170501 100644
--- a/datasets/sg/gov_dir/crawler.py
+++ b/datasets/sg/gov_dir/crawler.py
@@ -131,6 +131,8 @@ def crawl_person(context: Context, official, link, public_body, agency, section_
             context.emit(person, target=True)
             context.emit(position)
             context.emit(occupancy)
+    elif categorisation.is_pep is None:
+        context.log.warning("Uncategorised position", position=position_name, url=link)
```

basically if the allow and deny lists couldn't handle it, and the position database couldn't handle it, we'll warn about it.

Because with this dataset it seems more important than usual to keep up to date, and also quite risky to misclassify. Many listed positions are board-adjacent but not members of the board with that level of influence.

jbothma:
Let's clear up https://www.opensanctions.org/admin/positions/?offset=0&country=sg&is_pep=null before merging to avoid massive issue spam

datasets/sg/gov_dir/crawler.py
```diff
@@ -131,6 +138,10 @@ def crawl_person(context: Context, official, link, public_body, agency, section_
             context.emit(person, target=True)
             context.emit(position)
             context.emit(occupancy)
+    elif:
```
jbothma, line 141:
```suggestion
    elif categorisation.is_pep is None:
```

jbothma:
[uncategorised.txt](https://github.com/user-attachments/files/18336395/uncategorised.txt)

PR #1738, opened 2025-01-08
fixes 08 01

nvmbrasserie:


datasets/th/designated_person/crawler.py
```diff
@@ -32,6 +32,9 @@ def crawl_item(url: str, context: Context):
     info_dict = parse_table(response.find(".//table"))
     en_name = info_dict.pop("Individual/Entity Name (English)", "").strip()
     th_name = info_dict.pop("Individual/Entity Name (Thailand)").strip()
+    # Falsy entry at the very end of the list
+    if th_name == "1." and en_name is None:
```
jbothma: I think the none is breaking it

datasets/th/designated_person/crawler.py
```diff
@@ -77,7 +80,7 @@ def crawl_item(url: str, context: Context):
 
 
 def crawl(context: Context):
-    response = context.fetch_html(context.data_url)
+    response = context.fetch_html(context.data_url, cache_days=2)
```
jbothma: we don't want to wait 2 days to add a newly-sanctioned person, so we don't normally cache these for longer than a day

datasets/us/mi/med_exclusions/crawler.py
```diff
@@ -91,9 +91,9 @@ def crawl_item(row: Dict[str, str], context: Context):
 def crawl_excel_url(context: Context):
     doc = context.fetch_html(context.data_url)
     doc.make_links_absolute(context.data_url)
-    return doc.xpath("//*[text()='List of Sanctioned Providers (XLSX)']/../..")[0].get(
-        "href"
-    )
```
jbothma, line 96: I think we can write this more readably and robustly with with `/ancestor::a` and and a variable

PR #1741, opened 2025-01-09
Fixes-09-02

nvmbrasserie:


datasets/ae/local_terrorists/crawler.py
```diff
@@ -91,14 +90,14 @@ def parse_excel(context: Context, path: Path):
...
-            path = context.fetch_resource("source.xls", a_url)
+    doc.make_links_absolute(context.data_url)
+    for file_div in doc.findall(".//div[@class='download-file']"):
+        # Check the 'file-details' section for the correct file description
+        file_details = file_div.find(".//div[@class='file-details']/p")
+        if file_details is not None and file_details.text == "Download Excel File":
+            href_value = file_div.find(".//a[@href]").get("href")
+            if href_value is None:
+                context.log.error("Could not download Local Terror excel file!")
+            path = context.fetch_resource("source.xls", href_value)
```
jbothma, line 101: it'd be nice if this ensured we're downloading from the Local Terrorist section and not the UN section. Currently this should work because only the local section has an excel file and it's first, but we could e.g. select the h5 with text Local... and then find the link within the next element.

It looks like it'll crash if href_value is None, but but if the xpath stops matching anything, we'll just silently exit. Maybe keep the old logging logic and could you add data assertions?

jbothma: I'll do this quick

PR #1743, opened 2025-01-09
Russian Personal Sanctions Targeting US Citizens

nvmbrasserie:
fixes opensanctions/crawler-planning#246

The `[RU-MFA]` program is in a draft stage on the website, it should be published after this PR has been merged

### DIFF
--- a/datasets/ru/mfa_sanctions/crawler.py
+++ b/datasets/ru/mfa_sanctions/crawler.py
@@ -1,3 +1,4 @@
+import re
 from typing import Dict, Generator, List, Optional
 from lxml.html import HtmlElement
 
@@ -21,7 +22,6 @@ def parse_html_table(
         if any(cell.get("colspan") == "4" for cell in cells):
             continue
 
-        assert len(headers) == len(cells), (headers, cells)
         yield {hdr: c for hdr, c in zip(headers, cells)}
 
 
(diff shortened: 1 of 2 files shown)


datasets/ru/mfa_sanctions/crawler.py
```diff
@@ -0,0 +1,52 @@
...
+        yield {hdr: c for hdr, c in zip(headers, cells)}
+
+
+def crawl(context: Context):
+    doc = context.fetch_html(context.data_url, cache_days=1)
+    table = doc.xpath(".//table")
+    assert len(table) == 1
+    table = table[0]
+    for row in parse_html_table(table, headers=["index", "name", "-", "position"]):
+        row = h.cells_to_str(row)
```
jbothma, line 31: I'd like to maintain some confidence that the headings we assume are consistent with their table structure. I'm not really sure how, but one way might be something like

```
assert re.match("^\d\.$", cells[0]), cells[0]
assert re.match("^.+\(.+\)$, cells[1], cells[1]
...
```

but if you see a better way, by all means suggest it. I just don't want them to change the order or add a column and suddenly we're treating descriptions as the names or something weird like that.

jbothma: maybe I'm overdoing it? we're already asserting there are exactly as many columns as headers. I don't see them swapping the order of these columns in a way where role is column 2. So if we add assertions, we won't export if - becomes column 2.

So, let's add data assertions to the metadata and assert that name is longer than 3 letters and we're done

datasets/ru/mfa_sanctions/ru_mfa_sanctions.yml
```diff
@@ -0,0 +1,34 @@
+title: Russian Personal Sanctions Targeting US Citizens
```
jbothma, line 1:
```suggestion
title: Russia Personal Sanctions Targeting US Citizens
```

jbothma (commented):
Good stuff

datasets/ru/mfa_sanctions/crawler.py
```diff
@@ -0,0 +1,53 @@
...
+        person.add("name", name_en.rstrip(")"), lang="eng")
+        person.add("position", position, lang="rus")
+        person.add("citizenship", "us")
+        person.add("topics", "sanction.counter")
+        sanction = h.make_sanction(context, person)
+        sanction.add(
+            "program",
+            "Граждане США, находящиеся под персональными санкциями, включая запрет на въезд в Российскую Федерацию",
+            lang="rus",
+        )
```
jbothma, line 50: no need for a sanction object if there isn't more detail. This could be Person:program

nvmbrasserie: I think we could still use it to link the to the website using `program_key`, so that we're consistent with the way we treat programs

datasets/ru/mfa_sanctions/crawler.py
nvmbrasserie: maybe we could also assert the very first entry by hardcoding it into the assertions?

PR #1759, opened 2025-01-13
Reduce inclusion of stats in dataset metadata

pudo:
This implements: https://www.opensanctions.org/changelog/2/

### DIFF
--- a/zavod/zavod/exporters/metadata.py
+++ b/zavod/zavod/exporters/metadata.py
@@ -39,7 +39,14 @@ def get_base_dataset_metadata(
     if statistics_path.is_file():
         with open(statistics_path, "r") as fh:
             stats: Dict[str, Any] = json.load(fh)
-            meta.update(stats)
+            meta["entity_count"] = stats.get("entity_count", 0)
+            targets = stats.get("targets", {})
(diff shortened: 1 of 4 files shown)


jbothma (approved):

jbothma:
I think exporting the catalog might still depend on Dataset schemata https://github.com/opensanctions/opensanctions/blob/main/zavod/zavod/tools/export_catalog.py#L22

jbothma (commented):
Just digging a little deeper and I think generating the catalog might still depend on schemata https://github.com/opensanctions/opensanctions/blob/main/zavod/zavod/tools/export_catalog.py#L22

jbothma:
It doesn't break, but it outputs an empty `schemata` field https://github.com/opensanctions/opensanctions/blob/main/zavod/zavod/tools/export_catalog.py#L32

jbothma:
> Wooo nice catch. That's used to render /reference on the website, so not good to break. Maybe we need to fetch stats for all the datasets? Or we make the website run a schema facet on the API instead?

As far as I can tell, /reference is using default's statistics.schemata so that should be fine

PR #1761, opened 2025-01-13
Documentation-update

nvmbrasserie:


zavod/docs/best_practices/caching.md
```diff
@@ -0,0 +1,23 @@
+# Caching Considerations
+
+## General Caching Guidelines
+- We usually don’t cache sanctions index pages so that we can discover new entities as quickly as possible - we sometimes crawl these more than once a day.
```
pudo, line 4: -usually

zavod/docs/best_practices/caching.md
```diff
@@ -0,0 +1,23 @@
...
+
+- If we cache an index page, we usually don’t cache it for more than 1 day so that retries on a day are likely to succeed, but tomorrow we’ll get a fresh crawl.
+
+- We might cache detail pages (e.g. a profile of a person) longer, especially for bigger sites with thousands of pages, and if absolute freshness isn’t critical. The tradeoff is discovering new things vs completing crawls.
```
pudo, line 8: For slow-moving sources like PEP databases, ...

zavod/docs/best_practices/caching.md
```diff
@@ -0,0 +1,23 @@
...
+- We’d like to finish a crawl of a big site with many HTTP errors or hundreds of thousands of pages. Caching helps reduce the number of pages requested each crawl.
+
+!!! info "Please note"
+    It’s critical to discover newly sanctioned persons, and important to discover new 
+    high-level PEPs very quickly - hence selectively or never caching index/listing pages.
+
+## Issue with Pagination and Cached Pages
+When new content is added to a website, it can cause issues with how the pages are cached. For example, if new items push existing ones to a different page, the cached version of the page will not show the updated content immediately. This happens because the scraper continues to use the old version of the page until it's refreshed, causing some data to disappear and reappear later.
+
+## Solution
```
pudo, line 21: Doesn't need to be a heading, maybe just `Solution: ....`?

zavod/docs/best_practices/http_operations.md
```diff
@@ -0,0 +1,19 @@
...
+
+```python
+response = context.fetch_text(context.data_url, method="POST", headers=headers, data=body, cache_days=cache_days)
+```
+
+## Request Headers
+If you have to override the user-agent, include identifying information that allows the host to recognize us as the client. This helps them notify us if our requests are negatively impacting their server. Additionally, it's preferable to use the metadata HTTP configuration to override the user-agent when necessary.
+
+```python
+HEADERS = { "Accept": "application/json", "User-Agent": "opensanctions.org" }
```
pudo, line 14: didn't we have a way of setting the UA in the dataset yaml?

zavod/docs/best_practices/patterns.md
```diff
@@ -33,6 +33,111 @@ entity.add("name", row.pop("full_name"))
 entity.add("birthPlace", row.pop("place_of_birth"))
 ```
 
+The method `entity.add` works seamlessly with both a single string and a list of strings. This means that you don’t have to loop through a list to add multiple items individually. Instead, you can simply add the entire list at once.
```
pudo, line 36: In the long run, we probably want to make the typing of `entity.add` more strict and take only one argument ... perhaps we could already say that it's not worth forcing values into a list, and nicer to put them in individually if they come that way :)

zavod/docs/tutorial.md
```diff
@@ -19,6 +19,37 @@ companies into a database.
...
+For example:
+```
+projects/
+├── opensanctions/       # Main repository for active development
+├── opensanctions2/      # Secondary clone for experimental work
+├── nomenklatura/
+├── fingerprints/
+├── stdnum/
+└── … 
+```
```
pudo, line 34: I think this is a bit too deep for the tutorial. Maybe in best practices somewhere?

pudo (commented):
Thanks, this is all super helpful!

zavod/docs/best_practices/http_operations.md
jbothma: https://zavod.opensanctions.org/metadata/#http-options

zavod/docs/best_practices/patterns.md
jbothma: @pudo do you mean we should prefer

```python
for name in h.multi_split(name, SPLITS):
    entity.add(name)
```

or just to note inn these docs that lists are accepted but don't do `entity.add([name])`?

or something else?

zavod/docs/best_practices/patterns.md
```diff
@@ -33,6 +33,111 @@ entity.add("name", row.pop("full_name"))
...
+    address_ent = h.make_address(context, full=addr, city=city, lang="zhu")
+    # Copy address details to the entity
+    h.copy_address(entity, address_ent)
+    ```
+
+- Use `settings.RUN_TIME` to ensure consistent datetime comparisons.
+
+    ```python
+    # Determine if the end date is active by comparing with the configured runtime
+    is_active = end_date_obj >= settings.RUN_TIME
```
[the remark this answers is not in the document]
jbothma: we now have a helper for this which uses settings.RUN_TIME because it doesn't want to be affected by things like coverage.

Perhaps we can skip this entry since the cases where we're doing time comparison are so rare, and which field to use is quite dependent on the use case?

zavod/docs/tutorial.md
jbothma: or this could be in onboarding docs in notion, since it's quite specific to OpenSanctions devs rather than zavod users in general

zavod/docs/best_practices/http_operations.md
```diff
@@ -0,0 +1,19 @@
...
+```
+
+
+## Handling 403 Errors with Zyte
+If adding a browser-like `User-Agent` header (and additional headers) does not resolve a `403 Forbidden` error, switch to using Zyte for scraping. Zyte is specifically designed to bypass anti-scraping measures that commonly block typical scraping requests by simulating real user behavior.
```
jbothma, line 19:
```suggestion
## Handling bot blocking

Many sites employ bot blocking strategies. We believe this is primarily to mitigate Denial of Service attacks and manage server load, rather than protecting the content from extraction, since the purpose of the sites we scrape is dissemination of their block lists. As long as we are sensitive to our impact on their service and identifiable in their requests, we believe it is ok to work around their bot blocking strategies.

Blocking might result in error statuses like 403; redirects to error pages; or 200 status responses but with different content from what you've seen in the browser.

### Header-based restrictions

If a request using zavod fails but your browser succeeds, try making a request with more typical headers. Set a more browser-like user-agent header.

```yaml
http:
  user_agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/42.0.2311.90 Safari/537.36 (zavod; opensanctions.org)
```

If that doesn't work, try more of the common headers sent by browsers:

```
HEADERS = {
    "origin": "https://www.interpol.int",
    "referer": "https://www.interpol.int/",
    "sec-fetch-mode": "navigate",
    "sec-fetch-site": "none",
    "sec-fetch-user": "?1",
    "upgrade-insecure-requests": "1",
    "user-agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/126.0.0.0 Safari/537.36 (zavod; opensanctions.org)",
}
```

### Network blocking

If it fails in production but not locally, they might be blocking our production network IP range. It's common to block hosting provider networks for websites intended for humans only.

Use zyte with with `httpResponseBody` approach (default in `zavo.shed.zyte_api.fetch_*` functions except `fetch_html` whose `html_source` defaults to `browser_html`. `httpResponseBody is faster and cheaper than `browserHtml`.

It's also common to block requests from a country other than the publisher. If it works using a VPN exit point in that country, also try zyte using the `geolocation` argument.

### JavaScript challenges

If it works in the browser but you see different content when fetching using `zavod` or `curl`, there might be a javascript challenge that checks whether a full browser is rendering the page. This usually sets a cookie so the browser doesn't have to complete the challenge on each request. These challenges can also be intermittent.

For HTML, try requesting using `zyte_api.fetch_text` with `html_source="browserHtml"` (the default). This will render the page in a browser, execute any javascript, then turn the DOM back into HTML and return that.

If tricks like waiting for specific content or clicking on something to render the data is needed, look at the `actions` argument.
```

File: zavod/docs/best_practices/patterns.md [hunk 9]
```diff
@@ -41,6 +146,43 @@ If a variable number of fields can extracted automatically (e.g. from a list or
 * `dict_obj.pop()` individual fields when adding them to entities.
 * Log warnings if there are unhandled fields remaining in the `dict` so that we notice and improve the crawler. The context method [`context.audit_data()`][zavod.context.Context.audit_data] can be used to warn about extra fields in a `dict`. It takes the `ignore` argument to explicitly list fields that are unused.
 
+## Logs that matter
+
+Logs are essential for monitoring progress and debugging, but they are not always reliable for detecting omissions in large datasets. Use them to track the crawler’s status and provide alerts for issues that don’t stop the process.
```
jbothma, line 151:
```suggestion
## Logging and crawler feedback

It is good design to be told about issues, instead of having to go look to discover them.

Logs are essential for monitoring progress and debugging, but info-level logs don't help us notice if something is missing because we only see them when we choose to go and look at a crawler's logs. Use the appropriate log level for the purpose for cases that don’t stop the process.
```

zavod/docs/best_practices/patterns.md
```diff
@@ -41,6 +146,43 @@ If a variable number of fields can extracted automatically (e.g. from a list or
...
+    context.log.debug(f"Unique ID {person.id}")
+    ```
+
+* Info Logs: Monitor the crawler’s progress, especially on large sites.
+
+    ```python
+    context.log.info(f"Processed {page_number} pages")
+    ```
+
+* Warning Logs: Indicate potential issues that don't stop the crawl but may require attention.
```
jbothma, line 165:
```suggestion
* Warning Logs: Indicate potential issues that don't stop the crawl but may require attention. These are surfaced to the dev team on the [Issues](https://www.opensanctions.org/issues/) page and checked daily.
```

zavod/docs/best_practices/xpath_and_html.md
```diff
@@ -0,0 +1,30 @@
+# XPath and HTML
+
+## Selector Design
+Use semantic selectors that are resilient to HTML changes:
```
jbothma, line 4:
```suggestion
Try to make the selectors specific enough to reliably select the correct content (to avoid e.g. the wrong thing being stuffed in a person's `name` property) while making them as semantic, concise, and general as possible to be maintainable and resillient against irrelevant changes in the HTML.
```

zavod/docs/best_practices/xpath_and_html.md
```diff
@@ -0,0 +1,30 @@
...
+
+```python
+# Good: 
+table = doc.xpath('.//div[@id="block-content"]//table')
+
+# Avoid:
+table = doc.xpath('.//div[@id="block-content"]//div[3]//table')
+```
+
```
jbothma, line 13:
```suggestion
Prefer `div[contains(@class, 'abc')] over `div[@class='abc'] because classes can contain multiple values, but beware of conflicting class names like `abc-footer`
```

zavod/docs/best_practices/xpath_and_html.md
```diff
@@ -0,0 +1,30 @@
...
+## Assert the XPath Result
+Ensure that the XPath query returns exactly one element, especially when selecting from multiple possible matches (e.g., tables on a page). If more than one element is selected and we access `selection[0]`, we’re assuming the first one in the DOM order is the one we want, which may not be reliable with vague selectors. Always assert the correct number of elements to avoid errors.
+
+```python
+table = doc.xpath(".//table")
+assert len(table) == 1, table
+```
```
jbothma, line 20: ## Fail loudly if the selection is different from what was expected

Beware of selections being different from what you intend. Common issues are

- missing a bunch of entities because we're looping over a selection which turned out to be empty
- emitting garbage because content was added to the page which matched unintentionally
- cryptic errors because the wrong table is selected. It's ideal for an error to occur as close as possible to the offending code

One way to guard against these issues is assertions on the number of matches:

```python
items = doc.xpath(".//h2[@text()='The Section']/following-sibling::ul/li)
assert len(items) > 0, items
```
or

```python
table = doc.xpath(".//table")
assert len(table) == 1, table
```

zavod/docs/best_practices/xpath_and_html.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Assert DOM Hash
```
jbothma, line 22: I think the best headings are those that tell you which problem they solve

```suggestion
## Get notified of changes to content
```

zavod/docs/best_practices/xpath_and_html.md
jbothma: @pudo what do you think of a helper that asserts based on arguments?

```python
def xpath(doc, path, exactly: Optional[int] = None, at_least: Optional[int] = None) -> List[ElementTree]:
```

jbothma: and how horrid would it be if it did `matches[0]` for you if you passed `exactly=1`?

pudo: Looks nice! I really want the `element_to_string()` thing as well, I think we use a lot of approaches to this (e.g. `.xpath('string()')`, `.text_content()` etc. ... :)

zavod/docs/tutorial.md
nvmbrasserie: just moved it to the internal notion page

zavod/docs/tutorial.md
```diff
@@ -19,37 +19,6 @@ companies into a database.
...
-fixes opensanctions/crawler-planning#194
-```
-
-### Branch Naming Conventions
-Follow a consistent branch naming strategy to avoid issues with special characters in shell environments. Use **slug-like** names that are descriptive, hyphen-separated, and free of special characters such as `#`, `(`, or `)`:
-
-```bash
-git checkout planning-206-office-of-antiboycott-compliance
-```
-
```
nvmbrasserie: moved this bits here: https://github.com/opensanctions/crawler-planning/blob/main/README.md

PR #1764, opened 2025-01-14
[pl_finanse_sanctions] Use provided XLSX list instead of our own copy

leonhandreke:
I tackled this because the issues paged raised the warning that the page hash had changed, indicating an update to the page. I thought that they had addded an XLSX, but it seems it was something else. The warning has since disappeared, so it seems this was a temporary change to the site.

Anwway, parsing from the XLSX is nicer anyway

### DIFF
--- a/datasets/pl/finanse_sanctions/crawler.py
+++ b/datasets/pl/finanse_sanctions/crawler.py
@@ -16,16 +16,16 @@
 # Justification for entry on the list
 # Date of listing
 
-from normality import collapse_spaces
-from rigour.mime.types import CSV
 from typing import Dict
-import csv
 
+from normality import collapse_spaces
+from openpyxl import load_workbook
+
+from rigour.mime.types import XLSX
 from zavod import Context
 from zavod import helpers as h
 from zavod.shed.un_sc import get_legal_entities, get_persons, Regime, load_un_sc
 
-PDF_URL = "https://www.gov.pl/attachment/2fc03b3b-a5f6-4d08-80d1-728cdb71d2d6"
 POLAND_PROGRAM = "art. 118 ustawy z dnia 1 marca 2018 r. o przeciwdziałaniu praniu pieniędzy i finansowaniu terroryzmu"
 UN_SC_PREFIXES = [Regime.TALIBAN, Regime.DAESH_AL_QAIDA]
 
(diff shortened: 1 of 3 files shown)


datasets/pl/finanse_sanctions/pl_finanse_sanctions.yml
```diff
@@ -33,13 +33,26 @@ publisher:
   official: true
 url: https://www.gov.pl/web/finanse/lista-osob-i-podmiotow-wobec-ktorych-stosuje-sie-szczegolne-srodki-ograniczajace-na-podstawie-art-118-ustawy-z-dnia-1-marca-2018-r-o-przeciwdzialaniu-praniu-pieniedzy-i-finansowaniu-terroryzmu
 data:
-  url: https://docs.google.com/spreadsheets/d/e/2PACX-1vR1QhgLSbLVzombbJu20IB1eJ-ORMazTPj7TCTHDJ-y_bo4dEoISAZgE3DL9c7wszmXEfY13L6P78GR/pub?gid=0&single=true&output=csv
-  format: CSV
+  url: https://www.gov.pl/attachment/56238b34-8a26-4431-a05a-e1d039f0defa
```
jbothma, line 36: if we get the data url from https://www.gov.pl/web/finanse/lista-osob-i-podmiotow-wobec-ktorych-stosuje-sie-szczegolne-srodki-ograniczajace-na-podstawie-art-118-ustawy-z-dnia-1-marca-2018-r-o-przeciwdzialaniu-praniu-pieniedzy-i-finansowaniu-terroryzmu, just make that the data url to avoid the data url going stale.

jbothma:
particularly with sanctions datasets, but in general, when entities get re-keyed due to changes in a crawler (as is happening in this case), give Friedrich a heads-up before and when deploying so that he can dedupe when the new entities with new keys are out.

leonhandreke:
Good suggestions, I implemented them all.

datasets/pl/finanse_sanctions/crawler.py
```diff
@@ -16,88 +16,107 @@
...
-    path = context.fetch_resource("source.csv", context.data_url)
-    context.export_resource(path, CSV, title=context.SOURCE_TITLE)
-    with open(path, "r") as fh:
-        for row in csv.DictReader(fh):
-            crawl_row(context, row)
+    expected_page_hash = "ad7c8f0eecb445b4f99ef2f5e3fdc8a911660ff6"
+    if not h.assert_dom_hash(article, expected_page_hash, text_only=True):
+        context.log.warning(
+            "Page hash has changed. Verify that the lists referenced on the page are the 118 ust. 1 pkt 2, the UNSC Taliban list (resolution 1988 (2011)), and UNSC Al Qaida list (resolution 2253 (2015)). If any lists have been added or removed, update the crawler, otherwise just update the hash"
+        )
```
jbothma, line 111: hah I just meant as a comment, but I guess a super specific warning doesn't hurt.

datasets/pl/finanse_sanctions/crawler.py
```diff
@@ -16,88 +16,107 @@
...
-                context.log.warning(
-                    "Unknown materials download URL. Check if we want to scrape it.",
-                    url=url,
-                )
+    xlsx_link_element = doc.xpath(
+        ".//a[@class='file-download' and contains(text(), 'Lista osób i podmiotów, wobec których stosuje się szczególne środki ograniczające na podstawie art. 118 ust. 1 pkt 2 (wersja xlsx)')]"
+    )
+    if len(xlsx_link_element) != 1:
+        context.log.error("Could not find XLSX link element")
+        return
```
jbothma, line 102: A lot of the time it's best for a crawler to exit and don't publish when they're missing something important. E.g. in this case we don't want to publish a new dataset without the entities in this file, just because they changed the link text. It's better for it to exit loudly so we can come and fix it.

The way this is currently implemented, it should behave that way. If we don't find the file we exit before we export the UN stubs, so the dataset will be empty and won't publish. But it's worth bearing this requirement in mind.

There are exceptions - really big datasets where we sometimes struggle to do complete crawls sometimes allow for some failure, but we try to limit that.

datasets/pl/finanse_sanctions/crawler.py
```diff
@@ -16,88 +16,107 @@
...
+    entity.add("topics", "sanction")
     sanction = h.make_sanction(context, entity)
-    h.apply_date(sanction, "listingDate", row.pop("Data umieszczenia na liście"))
+    h.apply_date(sanction, "listingDate", row.pop("data_umieszczenia_na_liscie"))
     sanction.add(
-        "reason", collapse_spaces(row.pop("Uzasadnienie wpisu na listę")), lang="pol"
+        "reason",
+        collapse_spaces(row.pop("uzasadnienie_wpisu_na_liste_url")),
+        lang="pol",
     )
```
jbothma, line 78: ah sorry, the reason text is still nice in the reason prop. The link can go in sanction:sourceUrl

make_sanction automatically adds the publisher URL, and it's quite common to additionally add reference material linked in the data.

datasets/pl/finanse_sanctions/crawler.py
```diff
@@ -16,88 +16,107 @@
...
 
     context.emit(entity, target=True)
     context.emit(sanction)
 
-    context.audit_data(row)
+    context.audit_data(
+        row,
+        ignore=[
+            "lp",  # ID, we don't trust this to be stable
+            "uzasadnienie_wpisu_na_liste",  # Reason link text, we only care about the url
```
jbothma, line 88: I think we still want this

datasets/pl/finanse_sanctions/crawler.py
```diff
@@ -16,88 +16,107 @@
...
-            url = material.get("href")
-            if url in KNOWN_HASHES:
-                h.assert_url_hash(context, url, KNOWN_HASHES[url])
-            else:
-                context.log.warning(
-                    "Unknown materials download URL. Check if we want to scrape it.",
-                    url=url,
-                )
+    xlsx_link_element = doc.xpath(
+        ".//a[@class='file-download' and contains(text(), 'Lista osób i podmiotów, wobec których stosuje się szczególne środki ograniczające na podstawie art. 118 ust. 1 pkt 2 (wersja xlsx)')]"
```
jbothma, line 98: Maybe this is sufficient?

```suggestion
        ".//a[@class='file-download' and contains(text(), '118 ust. 1 pkt 2 (wersja xlsx)')]"
```

jbothma (commented):
Sorry for another round. But it's good to touch on reference URLs and exiting on errors.

BTW the thinking behind expanding the scope to the hash check changes was because need for hash checks changed with this changed from manual to automatic extraction.

datasets/pl/finanse_sanctions/crawler.py
leonhandreke: Thanks for the explanation! I'm not quite sure if you were happy with what I did, I just changed it to raise a `RuntimeError`, maybe that's even better?

jbothma: what you did was fine because it would do the right thing. Raising is definitely my personal preference because it's robust to future changes in the crawler, but I'm not sure if we have a general preference.

PR #1765, opened 2025-01-14
Infer entity.target from topics

pudo:
Implements: https://www.opensanctions.org/changelog/13/ 

This requires quite a bit of test patching but mostly due to a change in fixtures and overly specific assertions.


jbothma (approved):
nice!

PR #1766, opened 2025-01-14
More issues fixes 2025-01-14

leonhandreke:


datasets/us/cia_world_factbook/us_cia_world_factbook.yml
```diff
@@ -764,6 +764,30 @@ lookups:
...
+            start_date: 16 November 1994
+      # This is also Emomali RAHMON, split from the previous statement because there is a semicolon in the text
+      - match: head of state and Supreme Assembly Chairman since 20 November 1992)
+        holders:
+          - role: Head of State
+            name: Emomali RAHMON
+            start_date: 20 November 1992
+          - role: Supreme Assembly Chairman
+            name: Emomali RAHMON
+            start_date: 20 November 1992
```
jbothma, line 790: Looks like there are only two supreme assemblies so it's probably not likely this could be a false positive for another country, but just wanted to flag that in general we try to only output the facts contained in the matching text.

Really just a note though - you can keep this. The comments are very helpful.

jbothma (approved):

PR #1767, opened 2025-01-14
targets removals 

nvmbrasserie:


nvmbrasserie:
unsustainable way for CI tests to be completed

PR #1769, opened 2025-01-15
[gb_hmt_sanctions] Modernize and fix splitting logic

leonhandreke:
This fixes a data bug where we were dropping values. If you look hard enough at `split_items`, you'll see that the last item in a list never gets added. Same with values that are not a list. This means that we're currently dropping quite a few values for title, birthPlace, position, passportNumber and idNumber.

Here is a sample of old and new `targets.nested.json` for your diffing pleasure (unfortunately I haven't figured out how to make the ordering stable, so the diff is a bit dirty, but you'll get the idea):
[new-10k.json](https://github.com/user-attachments/files/18421948/new-10k.json)
[old-10k.json](https://github.com/user-attachments/files/18421949/old-10k.json)

Here is what the diff looks like for some entities:
2025-01-15T10:45:35,298376627+01:00

One question remains for me: `split_items` also performed `collapse_spaces` on its output, which is now omitted. Should we still do that, is that done somewhere else?

### DIFF
--- a/datasets/gb/hmt/crawler.py
+++ b/datasets/gb/hmt/crawler.py
@@ -1,6 +1,6 @@
 from typing import Optional, Dict, Any, List
 from banal import first
-from normality import stringify, collapse_spaces
+
 from rigour.mime.types import XML
 from followthemoney.util import join_text
 import re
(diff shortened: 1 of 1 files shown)


jbothma (approved):
Looks great.

I didn't see anything in the changed output needing collapse_spaces so I don't think adding it would be a significant improvement.

by the way, in addition to Delta export complete
```
 [gb_hmt_sanctions] added=0 dataset=gb_hmt_sanctions deleted=0 metric=delta_counts modified=3170
```

in the export output, there are a couple of delta scripts in contrib

A fun friday afternoon might be something that sorts array values before outputting diffs of changed entities.

PR #1771, opened 2025-01-15
bg-omnio-fixes

nvmbrasserie:


pudo (changes requested):
Can we just remove the sentence about last updated? It's now saying Jan 2022....

PR #1773, opened 2025-01-16
uavs manufacturers added

nvmbrasserie:


datasets/ua/war_sanctions/crawler.py
```diff
@@ -389,7 +394,9 @@ def crawl_legal_entity(context: Context, link, program):
             if "yellow" in label_elem.get("class"):
                 label, value = extract_label_value_pair(label_elem, value_elem, data)
                 data[label] = value
-    name = data.pop("Name")
+    name = data.pop("Name", None)
+    if name is None:
+        name = data.pop("Full name of legal entity")
```
jbothma, line 399: this is fine, but a common pattern is `name = data.pop("Full name of legal entity", data.pop("Name, None))` if that fits on a line

jbothma (approved):
very nice

PR #1778, opened 2025-01-16
bg asset declarations

nvmbrasserie:


datasets/bg/jud_declarations/bg_jud_declarations.yml
```diff
@@ -0,0 +1,69 @@
+title: Bulgaria Judicial Declarations
+entry_point: crawler.py
+prefix: jud-declarations
```
jbothma, line 3: let's make this country-scoped so `bg-`, and then something really short like `jud`

datasets/bg/jud_declarations/bg_jud_declarations.yml
```diff
@@ -0,0 +1,69 @@
+title: Bulgaria Judicial Declarations
```
jbothma, line 1:
```suggestion
title: Bulgaria Judiciary Asset Declarations
```

datasets/bg/jud_declarations/bg_jud_declarations.yml
```diff
@@ -0,0 +1,69 @@
...
+coverage:
+  frequency: weekly
+  start: 2025-01-17
+load_db_uri: ${OPENSANCTIONS_DATABASE_URI}
+summary: >
+  List of judges, prosecutors, and investigators in Bulgaria who are required to submit asset 
+  declarations.
+description: |
+  This dataset provides a list of judges, prosecutors, and investigators in Bulgaria. Published 
+  asset declarations under Article 175a, are linked to each individual's profile.
```
jbothma, line 13: Maybe we can flesh out a bit like this? But can you identify the act?

```suggestion
description: |
  This dataset provides a list of judges, prosecutors, and investigators in Bulgaria with links
  to their asset declarations.
  
  Article 175a of act ??? requires senior members of the judiciary to submit asset
  declarations annually. This provides us with a list of politically-exposed persons
  within the judiciary of Bulgaria.
```

datasets/bg/jud_declarations/crawler.py
```diff
@@ -0,0 +1,90 @@
...
+def parse_html_table(
+    table: HtmlElement,
+    skiprows: int = 0,
+    headers: Optional[List[str]] = None,
+) -> Generator[Dict[str, HtmlElement], None, None]:
+    for rownum, row in enumerate(table.findall(".//tr")):
+        if rownum < skiprows:
+            continue
+        cells = row.findall("./td")
+        yield {hdr: c for hdr, c in zip(headers, cells)}
```
jbothma, line 17: any reason h.parse_html_table doesn't work for this?

datasets/bg/jud_declarations/crawler.py
```diff
@@ -0,0 +1,90 @@
...
+        if date is None:
+            context.log.warning(f"Invalid doc_id_date: {doc_id_date}")
+        return
+    # Link is in the same cell as the name
+    name_link_elem = HtmlElement(row["name"]).find(".//a")
+    declaration_url = name_link_elem.get("href")
+
+    person = context.make("Person")
+    # We want the same person for 2 different years to have the same ID
+    person.id = context.make_id(name)
```
jbothma, line 41: as discussed in slack, let's put the location like `Sofia` in the id too

datasets/bg/jud_declarations/crawler.py
```diff
@@ -0,0 +1,90 @@
...
+    person.add("name", name, lang="bul")
+    person.add("topics", "role.pep")
+    person.add("sourceUrl", declaration_url)
+
+    position = h.make_position(
+        context,
+        name="Judiciary Official",
```
jbothma, line 48: as discussed in slack, let's put the position label in here. And if the quality of transliteration by gpt is good enough, let's add that too.

datasets/bg/jud_declarations/crawler.py
```diff
@@ -0,0 +1,90 @@
...
+        context.emit(occupancy)
+        context.emit(person)
+
+
+def crawl(context: Context):
+    doc = context.fetch_html(context.data_url, cache_days=1)
+    alphabet_links = doc.xpath(".//div[@itemprop='articleBody']/p//a[@href]")
+    # Bulgarian alphabet has 30 letters, but the name can start only with 29 of them
+    # We want to cover the last 2 years at any time
+    for a in alphabet_links[:58]:
```
jbothma, line 79: maybe we can also add an assertion that we got at least 58 links.

```python
[][:58] == []
```

datasets/bg/jud_declarations/crawler.py
```diff
@@ -0,0 +1,90 @@
...
+) -> Generator[Dict[str, HtmlElement], None, None]:
+    for rownum, row in enumerate(table.findall(".//tr")):
+        if rownum < skiprows:
+            continue
+        cells = row.findall("./td")
+        yield {hdr: c for hdr, c in zip(headers, cells)}
+
+
+def crawl_row(context: Context, row: Dict[str, HtmlElement], doc: HtmlElement):
+    doc.make_links_absolute(context.data_url)
```
jbothma, line 21: I'd do this in the loop, then it only happens once per page, and you don't have to pass `doc` to `crawl_row`

datasets/bg/jud_declarations/crawler.py
```diff
@@ -0,0 +1,90 @@
...
+        doc.make_links_absolute(context.data_url)
+        link = a.get("href")
+        doc = context.fetch_html(link, cache_days=1)
+        table = doc.xpath(".//table")
+        if len(table) == 0:
+            context.log.info("No tables found")
+            continue
+        assert len(table) == 1
+        table = table[0]
+        for row in parse_html_table(table, headers=["name", "doc_id_date"]):
```
jbothma, line 89: if we hardcode headers, it's also a good idea to assert that the columns are still arranged as they were when we wrote the crawler. The usual `parse ... row.pop("header")` pattern effectively does that because pop fails if the header is missing.

At a glance it looks like we can build the crawler to follow that pattern, otherwise we need to add some column-checking logic.

jbothma (changes requested):
nice work on a very limited and hand-edited (therefore error-prone) site.

datasets/bg/jud_declarations/crawler.py
nvmbrasserie: also some tables do not have any headers at all: http://62.176.124.194/index.php?option=com_content&view=article&layout=edit&id=204&Itemid=103

datasets/bg/jud_declarations/crawler.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def crawl(context: Context):
+    doc = context.fetch_html(context.data_url, cache_days=1)
+    alphabet_links = doc.xpath(".//div[@itemprop='articleBody']/p//a[@href]")
+    assert len(alphabet_links) >= 58, "Expected at least 58 links in `alphabet_links`."
+    # Bulgarian alphabet has 30 letters, but the name can start only with 29 of them
+    # We want to cover the last 2 years at any time
+    for a in alphabet_links[:58]:
+        doc.make_links_absolute(context.data_url)
```
leonhandreke, line 160: (nit) make_links_absolute can happen once I think

datasets/bg/jud_declarations/crawler.py
```diff
@@ -0,0 +1,171 @@
...
+        link = a.get("href")
+        doc = context.fetch_html(link, cache_days=1)
+        table = doc.xpath(".//table")
+        if len(table) == 0:
+            context.log.info("No tables found")
```
leonhandreke, line 165:
```suggestion
            context.log.info("No tables found", url=link)
```

datasets/bg/jud_declarations/crawler.py
```diff
@@ -0,0 +1,171 @@
...
+            continue
+        assert len(table) == 1
+        table = table[0]
+        for row in parse_html_table(table, headers=["name", "doc_id_date"]):
+            doc.make_links_absolute(context.data_url)
```
leonhandreke, line 170: (nit) Can happen once, not once per row I think.

datasets/bg/jud_declarations/crawler.py
```diff
@@ -0,0 +1,171 @@
...
+
+    occupancy = h.make_occupancy(
+        context,
+        person,
+        position,
+        no_end_implies_current=False,
+        categorisation=categorisation,
+        status=OccupancyStatus.UNKNOWN,
+    )
+    # Switch to the declarationDate once it's introduced
```
leonhandreke, line 144: Do we have a ticket for this TODO somewhere? If not, just close this, also fine.

datasets/bg/jud_declarations/crawler.py
leonhandreke: Not immediately obvious to me either but maybe this is actually resolved?

nvmbrasserie: yup, we discussed it below: https://github.com/opensanctions/opensanctions/pull/1778#discussion_r1952554790

datasets/bg/jud_declarations/crawler.py
nvmbrasserie: I haven't found a way to use it only once in the entire `crawl` function

datasets/bg/jud_declarations/crawler.py
nvmbrasserie: https://github.com/alephdata/followthemoney/issues/1631

datasets/bg/jud_declarations/crawler.py
```diff
@@ -0,0 +1,171 @@
...
+        context.log.warning(f"Skipping {pdf_path} due to error: {e}")
+        return {}
+
+    context.log.info(f"Extracted Data: {extracted_data}")
+    return extracted_data
+
+
+def parse_html_table(
+    table: HtmlElement,
+    skiprows: int = 0,
```
leonhandreke, line 50: (nit) we could get rid of skiprows.

datasets/bg/jud_declarations/crawler.py
```diff
@@ -0,0 +1,171 @@
...
+    headers: List[str] = [],
+) -> Generator[Dict[str, HtmlElement], None, None]:
+
+    first_row = table.find(".//tr[1]")
+    actual_headers = [td.text_content().strip() for td in first_row.findall("./td")]
```
leonhandreke, line 55: (nit) do we need to do text processing here?

`len(headers) == len(first_row.findall("./td"))` seems like it should be enough and expresses what we're actually checking here.

datasets/bg/jud_declarations/crawler.py
leonhandreke: I think the asserts on l74,75 look good to me (we could add a comment mentioning that they're quite important in this case in case somebody goes in and relaxes them in the future.

leonhandreke (approved):
Nice! Two more small thoughts, but I've approved this already, so feel free to merge whenever.

(Maybe this could be my first crawler release...?)

datasets/bg/jud_declarations/crawler.py
leonhandreke: What I meant was once per doc. We should only have to call it once after `fetch_html`, not once per letter of the alphabet, right? Same below.

PR #1791, opened 2025-01-21
ge_company_registry

nvmbrasserie:
fixes opensanctions/crawler-planning#238


jbothma:
double check the directory name spelling  :)

datasets/ge/company_regisrty/ge_company_registry.yml
```diff
@@ -0,0 +1,254 @@
+title: Georgian Company Registry
+entry_point: crawler.py
+prefix: ge-company-registry
+coverage:
+  frequency: weekly
+  start: 2025-01-20
+load_db_uri: ${OPENSANCTIONS_DATABASE_URI}
```
jbothma, line 7: we don't normally load the full company registers into the statements table - have you discussed whether we want to for this one?

datasets/ge/company_registry/crawler.py
```diff
@@ -0,0 +1,172 @@
...
+            entity,
+            "partner_id",
+            "partner_citizenship",
+            "partner_start_date",
+            "owner",
+        )
+        context.audit_data(
+            row,
+            ignore=[
+                "director_id",
```
nvmbrasserie, line 96: I'm not sure how we want to handle these IDs. They do not look like tax numbers, and some of them actually overlap with the ID numbers of unrelated entities. For example, there's a partner with `partner_id' that is the same as the `ID number' of https://www.opensanctions.org/entities/NK-SMThzFGG2WvqKfJtiyk8HV/

datasets/ge/company_registry/crawler.py
```diff
@@ -0,0 +1,172 @@
...
+    )
+
+    if relationship_type == "owner":
+        relationship.add("percentage", row.pop("partner_share"))
+
+    context.emit(relationship)
+
+
+def crawl(context: Context) -> None:
+    local_path = context.get_resource_path("companyinfo_v3.csv")
```
leonhandreke, line 143: For my understanding (this is a question for @jbothma maybe): What happens to the resources? Do they get archived somewhere? Because here we're overwriting this file continously.

If this path is not special in any way, then maybe we should have a context processor of the shape of:

```python
with InternalDataBlobAsLocalFile(blob, delete_after_use=True) as local_filepath:
   ...
```

or something like that. For elegance and all. (Should not be part of this PR of course)

datasets/ge/company_registry/crawler.py
```diff
@@ -0,0 +1,172 @@
...
+    entity.add("name", name)
+    entity.add("classification", legal_form)
+    if reg_date != "NULL":
+        h.apply_date(entity, "incorporationDate", reg_date)
+    for add in h.multi_split(row.pop("address"), ["; ", " / "]):
+        entity.add("address", add)
+    entity.add("status", row.pop("status"))
+    emails = email.replace(" ", "").strip()
+    for email in h.multi_split(emails, [";", ","]):
+        email_clean = registry.email.clean(email)
```
leonhandreke, line 55: Hm, doesn't `entity.add` do this automatically? (This is not a rhethorical question, I actually don't know but have a feeling that it could)

datasets/ge/company_registry/crawler.py
```diff
@@ -0,0 +1,172 @@
...
+    "კოოპერატივი",  # Cooperative
+    "უცხოური არასამეწარმეო იურიდიული პირის ფილიალი",  # Foreign Non-Commercial Legal Entity Branch
+    "უცხოური საწარმოს ფილიალი",  # Foreign Commercial Entity Branch
+    "სააქციო საზოგადოება",  # Joint Stock Company
+    "საჯარო სამართლის იურიდიული პირი",  # Public Legal Entity
+    "სოლიდარული პასუხისმგებლობის საზოგადოება",  # Solidarity Limited Liability Company
+    "კომანდიტური საზოგადოება",  # Limited Partnership
+    "შეზღუდული პასუხისმგებლობის საზოგადოება",  # General Partnership
+}
+ORGS = {
```
leonhandreke, line 19: (nit) rename to NON_COMPANY_ORGS since a company is also an organization?

leonhandreke (commented):
My first code review, bear with me :)

datasets/ge/company_registry/crawler.py
```diff
@@ -0,0 +1,172 @@
...
+            row,
+            ignore=[
+                "director_id",
+                "director_citizenship",
+                "director_start_date",
+            ],
+        )
+
+
+def emit_rel(
```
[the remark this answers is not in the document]
jbothma: we have a pattern of `crawl_...` functions which chain a bit and fan out with the structure of the data, e.g. crawl loops over records and calls crawl_company, which emits the company and loops over the directors calling crawl_directorship and emitting the director and relationship for each.

I also like a functional approach, although let's discuss that and see what it might look like with some of our crawlers.

For not it's perhaps best to be consistent with the rest of our crawlers, but factor in your suggestion of handling 

```
if name == "NULL":
        return None
```

inside `crawl_relationship` and popping the provided keys in there, whether name was `NULL` or not. That'd clean up the auditing at the call sites.

jbothma (commented):
Nice work @nvmbrasserie. Awesome comments as always @leonhandreke.

datasets/ge/company_registry/ge_company_registry.yml
```diff
@@ -0,0 +1,1032 @@
+title: Georgian Company Registry
+entry_point: crawler.py
+prefix: ge-company-registry
+coverage:
+  frequency: weekly
+  start: 2025-01-20
+load_db_uri: ${OPENSANCTIONS_DATABASE_URI}
+ci_test: false
```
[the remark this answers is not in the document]
nvmbrasserie: I was thinking initially to use gpt translation for companies names, but then forgot to remove when we decided not to do it

datasets/ge/company_registry/crawler.py
nvmbrasserie: It's a recent best practice: 'The method entity.add works seamlessly with both a single string and a list of strings. In the long run, however, we want to make the typing of entity.add more strict to accept only one argument at a time.' from the latest documentation update https://zavod.opensanctions.org/best_practices/patterns/

datasets/ge/company_registry/ge_company_registry.yml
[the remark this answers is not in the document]
nvmbrasserie: Looks like we actually need it for skipping the google cloud credentials

datasets/ge/company_registry/crawler.py
jbothma: isn't the question about calling `clean()`?

nvmbrasserie: sorry, I misinterpreted it. `entity.add` does `.strip() `indeed, but here we want to remove the spaces inside some broken emails (e.g. `personname @gmail.com`) so that after applying `registry.email.clean ` we still cover some initially broken ones

Issue #1796, opened 2025-01-24
BIS now sanctions addresses


pudo:
See: https://www.bis.gov/press-release/department-commerce-announces-additional-export-restrictions-counter-russian - need to make these into `Address`.

PR #1798, opened 2025-01-24
address fixes

nvmbrasserie:


pudo:
fixes #1796

datasets/us/trade_csl/crawler.py
```diff
@@ -76,6 +76,9 @@ def parse_result(context: Context, result: Dict[str, Any]):
     if schema is None:
         context.log.error("Unknown result type", type=type_)
         return
+    name = result.pop("name", None)
+    if name.startswith("Address 0"):
```
pudo, line 80: you won't believe it: I think we want a regex here! `re.match("Address \d+")`

datasets/us/trade_csl/crawler.py
```diff
@@ -92,7 +95,6 @@ def parse_result(context: Context, result: Dict[str, Any]):
         entity.id = context.make_slug(entity_number, prefix="ofac")
         is_ofac = True
 
-    name = result.pop("name", None)
```
pudo, line 95: Can we move the address to the `full` field for these?

pudo (changes requested):

datasets/us/trade_csl/crawler.py
```diff
@@ -163,7 +165,10 @@ def parse_result(context: Context, result: Dict[str, Any]):
 
     for obj in parse_addresses(context, result.pop("addresses", [])):
         # h.apply_address(context, entity, obj)
-        h.copy_address(entity, obj)
+        if not entity.schema.is_a("Address"):
+            h.copy_address(entity, obj)
+        else:
+            entity.add("full", obj.get("full"))
```
jbothma, line 171: It's easier to reason about "if is an address ... else ..." than "if not is an address ... else ..."

```suggestion
        if entity.schema.is_a("Address"):
            entity.add("full", obj.get("full"))
        else:
            h.copy_address(entity, obj)
```

PR #1805, opened 2025-01-27
adding Turkish Capital Markets Board banned list

bgmello:
Close opensanctions/crawler-planning#289


datasets/tr/capital_markets_banned_list/crawler.py
```diff
@@ -0,0 +1,54 @@
...
+    # Create sanction
+    sanction = h.make_sanction(context, entity)
+    sanction.add("authorityId", row.pop("kurulKararNo"))
+    sanction.add("status", row.pop("yargilamaAsamasi"))
+
+    # Add information about the related stock
+    if stock_name := row.pop("pay"):
+        sanction.add("description", f"Related stock: {stock_name}")
+    if stock_code := row.pop("payKodu"):
+        sanction.add("description", f"Stock code: {stock_code}")
```
leonhandreke, line 34: We try to avoid putting information into free text fields like this, as it will lead to our customers eventually parsing these fields and then we'll have gone full circle...

Do you know how exactly these stocks are related? Are they prohibited from being traded (in this case, they could be a Sanction)? Or are they the reason the entity was banned from trading? Let's get more info here so we can better map this to our data model (or expand it ;) )

datasets/tr/capital_markets_banned_list/crawler.py
```diff
@@ -0,0 +1,54 @@
...
+
+    # Add decision details
+    if decision_date := row.pop("kurulKararTarihi"):
+        h.apply_date(sanction, "startDate", decision_date)
+
+    # Add the reason/explanation
+    if explanation := row.pop("aciklama"):
+        sanction.add("reason", explanation)
+
+    entity.add("topics", "sanction")
```
leonhandreke, line 44: This should be `reg.action`. `sanction` is for super-serious stuff like terrorism. I discussed with @jbothma and we agreed to update the docs at https://www.opensanctions.org/docs/topics/ cause at the moment, it's not super clear.

leonhandreke (changes requested):
Thanks for this contribution! A few comments inline. 

(I guess a code review is very transactional, but since we haven't said hi: Hey, I'm new here! Looking forward to work with you :) )

datasets/tr/capital_markets_banned_list/crawler.py
jbothma: I know I previously suggested putting additional information into the description with a key prefix like this. Some issues with this approach have been pointed out to me. Basically, if something should be in our data model, let's rather suggest it as an addition and improve the data model. If we can't argue that it belongs there, it might not be interesting enough to include in the data.

In this case, we might want to model these as Company or Security entities, but it's worth understanding the relation here so we can describe that appropriately.

datasets/tr/capital_markets_banned_list/tr_capital_markets_banned_list.yaml
```diff
@@ -0,0 +1,39 @@
+title: Turkish Capital Markets Board Banned List
```
jbothma, line 1: We usually write the country name proper noun (Türkiye) rather than the adjective (Turkish) 

```suggestion
title: Türkiye Capital Markets Board Banned List
```

datasets/tr/capital_markets_banned_list/crawler.py
bgmello: From what I understood this is the stock related to the misconduct of the individual that has been sanctioned but I couldn't found a way to add this to any FtM model.

jbothma: ok - here are a couple of examples showing that the stock/company listed beside the person is indeed the shares being traded, and not the employer or co-accused of the listed persons

https://www.borsaningundemi.com/haber/spk-sonmez-filament-hissesinde-8-kisiyi-yakaladi-1558830
https://www.hurriyet.com.tr/ekonomi/spkdan-14-kisi-ve-2-sirkete-80-milyon-lira-para-cezasi-42327173

So yeah I don't think it makes sense to add this as a Security or Company or Asset. Something in the sanction description makes sense. How about `f"Due to activity relating to {ticker} ({stock_name})"`?

jbothma:
Sorry for catching this late. I don't see any reason for not including the other categories other than that the original ticket linked to one category 
image

Unless I'm missing something, could you add support for the other categories too? Then let's put the category label into `Sanction:provisions` (in turkish) and add a h.assert_dom_hash() on the menu so we get a warning if a new item is added.

datasets/tr/capital_markets_banned_list/crawler.py
```diff
@@ -0,0 +1,57 @@
...
+    entity.add("topics", "reg.action")
+
+    context.emit(entity, target=True)
+    context.emit(sanction)
+
+    # id = internal id
+    # kurulKararTarihiStr = decision date as string (we already have decision date)
+    # davaBilgisi = not used because it's always "Yok" (No)
+
+    if row.get("davaBilgisi") and row.get("davaBilgisi") != "Yok":
```
jbothma, line 50:
```suggestion
    # id = internal id
    # kurulKararTarihiStr (decision date) as string (we already have decision date)
    # davaBilgisi (Case Information) not used because it's always "Yok" (No)
    if row.get("davaBilgisi") and row.get("davaBilgisi") != "Yok":
```

jbothma (commented):
Thanks for the improvements!

bgmello:
Should I add them to this crawler? Or is better to create a new one?

jbothma:
Actually I thought the link we started with was the less serious one. Let's skip this addition and stick to what we have.

Also sorry for breaking the build with my committed suggestion :) I'll wrap this up and merge.

Issue #879, opened 2024-05-24
Make assertions on the properties of entities in the dataset


jbothma:
**Is your feature request related to a problem? Please describe.**

when only a portion of entities in a dataset have a particular value, it's easy to write a crawler in a way where you don't realise a value has been dropped, e.g. because of a typo, or because the source has changed.

e.g. 

- a typo in the xpath to get a detail URL fails to look up the href properly, so we keep adding None to the sourceUrl property

**Describe the solution you'd like**

a way to say that an expected proportion of entities have some value, or even a specified number of values, for a given property

e.g.

```yaml
assertions:
  min:
    schema_entities:
      Person: 100
    schema_properties:
      Person:
        name: 100  # this should perhaps be redundant because it's a required field
        sourceUrl: 100  # all entities are expected to have a 
        nationality: 80  # 80% of entities are expected to have a value for nationality
      Sanction:
        endDate: 1  # at least 1% of sanctions are expected to be inactive
```

It could also be useful to specify which values are expected (e.g. which topics), or how many (e.g. 2 URLs per sanction), but this may be out of scope of a first iteration.

**Describe alternatives you've considered**

None

**Additional context**

None

pudo:
FWIW this is a really cool proposal and I'd love to see it implemented!

PR #1811, opened 2025-01-28
Allow validating the number of property manifestation for a type

leonhandreke:
Some docs in the commit messages, the most important discussion item is probably:

> This carries a slight ambiguity with regards to inheritance: Can an
> assertion on LegalEntity.country be satisfied by Person and Company
> entities? At the moment, the implementation does not allow this.

Semantics aside, I also welcome comments regarding naming. schema, properties, values... such overload.

This makes our `statistics.json` quite a bit larger --- do we want that? I guess we could re-work the code to allow the validations to happen on the `Statistics` object, not on its `dict`-ified version. But then we'd lose the ability to do comparisons with the past in the validations, which is something I'd really like to tackle because I think it's actually more useful than specifying thresholds everywhere.

### DIFF
--- a/zavod/zavod/exporters/statistics.py
+++ b/zavod/zavod/exporters/statistics.py
@@ -42,7 +42,7 @@ def get_country_facets(countries: Dict[str, int]) -> List[Any]:
 SchemaProperty = namedtuple("SchemaProperty", ["schema", "property"])
 
 
-def get_property_values_facets(properties: Dict[SchemaProperty, int]) -> List[Any]:
+def get_entities_with_prop_facets(properties: Dict[SchemaProperty, int]) -> List[Any]:
     facets: List[Any] = []
     for prop, count in sorted(properties.items()):
         facet = {
@@ -64,7 +64,7 @@ class Statistics(object):
         self.thing_count = 0
         self.thing_countries: Dict[str, int] = defaultdict(int)
         self.thing_schemata: Dict[str, int] = defaultdict(int)
-        self.property_values_count: Dict[SchemaProperty, int] = defaultdict(int)
+        self.entities_with_prop_count: Dict[SchemaProperty, int] = defaultdict(int)
 
         self.target_count = 0
         self.target_countries: Dict[str, int] = defaultdict(int)
@@ -78,7 +78,7 @@ class Statistics(object):
         for prop_name, values in entity.properties.items():
             # We add 1 instead of len(values) here because we want to count the number of entities that have this
             # value set, not the number of values.
-            self.property_values_count[
+            self.entities_with_prop_count[
                 SchemaProperty(entity.schema.name, prop_name)
             ] += 1
 
@@ -116,8 +116,8 @@ class Statistics(object):
                 "total": self.thing_count,
                 "countries": get_country_facets(self.thing_countries),
                 "schemata": get_schema_facets(self.thing_schemata),
-                "property_values": get_property_values_facets(
-                    self.property_values_count
+                "entities_with_prop": get_entities_with_prop_facets(
+                    self.entities_with_prop_count
                 ),
             },
         }
--- a/zavod/zavod/exporters/statistics.py
+++ b/zavod/zavod/exporters/statistics.py
@@ -1,10 +1,10 @@
-from collections import defaultdict
+from collections import defaultdict, namedtuple
 from typing import Dict, List, Any, Optional, Set
+
 from followthemoney import model
 from followthemoney.types import registry
-
-from zavod.entity import Entity
 from zavod.archive import STATISTICS_FILE
+from zavod.entity import Entity
 from zavod.exporters.common import Exporter
 from zavod.util import write_json
 
@@ -37,6 +37,23 @@ def get_country_facets(countries: Dict[str, int]) -> List[Any]:
     return facets
 
 
+# We don't use followthemoney.Property because we want to track manifestations of property values on specific
+# schemas, even though the property might be defined somewhere higher in the type hierarchy.
+SchemaProperty = namedtuple("SchemaProperty", ["schema", "property"])
+
+
+def get_property_values_facets(properties: Dict[SchemaProperty, int]) -> List[Any]:
+    facets: List[Any] = []
+    for prop, count in sorted(properties.items()):
+        facet = {
+            "schema": prop.schema,
+            "property": prop.property,
+            "count": count,
+        }
+        facets.append(facet)
+    return facets
+
+
 class Statistics(object):
     def __init__(self) -> None:
         self.entity_count = 0
@@ -47,6 +64,7 @@ class Statistics(object):
         self.thing_count = 0
         self.thing_countries: Dict[str, int] = defaultdict(int)
         self.thing_schemata: Dict[str, int] = defaultdict(int)
+        self.property_values_count: Dict[SchemaProperty, int] = defaultdict(int)
 
         self.target_count = 0
         self.target_countries: Dict[str, int] = defaultdict(int)
@@ -57,6 +75,10 @@ class Statistics(object):
         self.schemata.add(entity.schema.name)
         for prop in entity.iterprops():
             self.qnames.add(prop.qname)
+        for prop_name, values in entity.properties.items():
+            self.property_values_count[
+                SchemaProperty(entity.schema.name, prop_name)
+            ] += len(values)
 
         if entity.schema.is_a("Thing"):
             self.thing_count += 1
@@ -92,6 +114,9 @@ class Statistics(object):
                 "total": self.thing_count,
                 "countries": get_country_facets(self.thing_countries),
                 "schemata": get_schema_facets(self.thing_schemata),
+                "property_values": get_property_values_facets(
+                    self.property_values_count
+                ),
             },
         }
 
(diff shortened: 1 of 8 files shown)


jbothma:
> This carries a slight ambiguity with regards to inheritance: Can an
> assertion on LegalEntity.country be satisfied by Person and Company
> entities? At the moment, the implementation does not allow this.

I think this specificity is useful. It's also consistent with that behaviour of the schema assertions. We could add the option to relax that on a case by case basis in future with an assertion like

```
min:
  inherited_property_values:
    LegalEntity:
      alias: 10
```

or you know, something. There exists a migration path.

jbothma:
> This makes our statistics.json quite a bit larger --- do we want that?

I'm really excited about this: (note citizenship and country)

```
"things": {
	    "total": 698,
	    [...]
	    "property_values": [
	      {
		"schema": "Person",
		"property": "citizenship",
		"count": 207
	      },
	      {
		"schema": "Person",
		"property": "country",
		"count": 199
	      },
```

It's describing the shape of the data and allows users to very quickly (depending on how well they read JSON) see how often `citizenship` is used, versus the more general property `country`. See https://www.notion.so/opensanctions/It-could-be-nice-to-learn-about-the-shape-of-the-data-075bd5c05c43444894d433223137b71f

I also like moving in a direction where historical comparison and even assertions is easy.

But let's get @pudo's input since this is a medium prominent part of a public interface.

jbothma:
cf https://github.com/opensanctions/opensanctions/issues/879

> It could also be useful to specify which values are expected (e.g. which topics), or how many (e.g. 2 URLs per sanction), but this may be out of scope of a first iteration.

What do you think of proposing the syntax for this, and maybe even the historical comparison assertions? Even if we don't implement them in this changeset, I think it might help us think through the syntax we're implementing in this round. e.g. we're adding `property_values` but that might possibly be a good key to use for when we're actually making assertions on the values. One option for the functionality implemented here is `property_instances`? or something.

jbothma:
> In the metadata YAML, an assertion specification like this means that at
> least 10 Person entities have to have a birthPlace set:

as opposed to 5 Person entities each having two distinct values for birthPlace, right? 

So 10 distinct Persons have one or more `birthPlace`.

I think the latter interpretation is the more useful one, I'm just being pedantic in case someone finds the alternative interpretation important.

zavod/zavod/validators/assertions.py
```diff
@@ -68,6 +68,9 @@ def feed(self, entity: Entity) -> None:
         self.stats.observe(entity)
 
     def finish(self) -> None:
+        if len(self.context.dataset.assertions) == 0:
+            self.context.log.warn("Dataset has no assertions.")
+
```
jbothma, line 84: could you break this out to its own PR since it's immediately useful and uncontroversial

zavod/zavod/meta/assertion.py
```diff
@@ -70,6 +71,20 @@ def parse_metrics(
                 yield from parse_filters(
                     Metric.ENTITY_COUNT, comparison, "country", value
                 )
+            case "property_values" if isinstance(value, dict) and value != {}:
+                for schema_name, props in value.items():
+                    for prop_name, threshold_value in props.items():
+                        yield Assertion(
+                            Metric.ENTITY_COUNT,
```
jbothma, line 78: I think this might warrant a different Metric so that the error messages make sense, and also that I'm not sure entity_count makes sense here. But it's been a bit of a job to refresh my memory of how this all fits together. Apologies if this is over-engineered.

https://github.com/opensanctions/opensanctions/blob/main/zavod/zavod/tests/test_validate.py#L79

zavod/zavod/meta/assertion.py
```diff
@@ -70,6 +71,20 @@ def parse_metrics(
...
+                            comparison,
+                            threshold_value,
+                            "property_values",
+                            # We don't just put a Property.qname in the shape of "Schema:name" here because we want to
+                            # assert not on qnames, which are the canonical name of a property (e.g. Thing.country), but
+                            # on property values in entities of a specific type, like Company.country.
+                            # TODO(Leon Handreke): Maybe raise if the property does not exist on the Schema?
```
jbothma, line 85: yeah warning or error - I'm not sure I have a strong opinion, but some kind of validation/feedback would be amazing

zavod/zavod/meta/assertion.py
jbothma: would also be nice to have a test case demonstrating this working, e.g. as per that link

jbothma (changes requested):
Nice! Thanks. It generally looks good and mergeable other than small matters.

Let's confirm with Friedrich before changing statistics.json - the initial implementation was based on the class used to calculate statistics, and I deferred splitting out the class until we know more about the future of assertions. I see the value of publishing these numbers, but it's worth a broader chat and not super urgent.

leonhandreke:
> > In the metadata YAML, an assertion specification like this means that at
> > least 10 Person entities have to have a birthPlace set:
> 
> (a) as opposed to 5 Person entities each having two distinct values for birthPlace, right?
> 
> (b) So 10 distinct Persons have one or more `birthPlace`.
> 
> I think the latter interpretation (b) is the more useful one (edit: adding - it's stricter and talks about how many entities have such a prop value - how good the coverage of that prop is). I'm just being pedantic in case someone finds the alternative interpretation important.
> 
> (edit: it looks like the implementation is actually (a) - let's talk about which is better, and then tighten up the docs. Speaking of docs, it'd be nice to add those on the metadata doc page once we've confirmed what we want here.)

Thanks for catching that, I consciously made a decision to implement (a) and then wrote (b) in the commit message.

I think (a) makes more sense when we're talking about "counting property values", which is what the name implies. But I'm also open to (b). WDYT?

leonhandreke:
> Nice! Thanks. It generally looks good and mergeable other than small matters.
> 
> Let's confirm with Friedrich before changing statistics.json - the initial implementation was based on the class used to calculate statistics, and I deferred splitting out the class until we know more about the future of assertions. I see the value of publishing these numbers, but it's worth a broader chat and not super urgent.

Cool, let's keep this around until we get input from him.

jbothma:
> Thanks for catching that, I consciously made a decision to implement (a) and then wrote (b) in the commit message.
> 
> I think (a) makes more sense when we're talking about "counting property values", which is what the name implies. But I'm also open to (b). WDYT?

I think (b) is the more useful - it's stricter and talks about how many entities have such a prop value - how complete the coverage of that prop is.

I don't feel super strongly about this, but it'd be cool to know if you had specific thoughts for preferring (a)?

jbothma:
Sorry for catching this late. I don't see any reason for not including the other categories other than that the original ticket linked to one category 
image

Unless I'm missing something, could you add support for the other categories too? Then let's put the category label into `Sanction:provisions` (in turkish) and add a h.assert_dom_hash() on the menu so we get a warning if a new item is added.

jbothma:
Nice, thanks for this!

Let's confirm with @pudo about expanding statistics.json or if he has any other concerns, otherwise it's an awesome addition.

leonhandreke:
> > Thanks for catching that, I consciously made a decision to implement (a) and then wrote (b) in the commit message.
> > I think (a) makes more sense when we're talking about "counting property values", which is what the name implies. But I'm also open to (b). WDYT?
> 
> I think (b) is the more useful - it's stricter and talks about how many entities have such a prop value - how complete the coverage of that prop is.
> 
> I don't feel super strongly about this, but it'd be cool to know if you had specific thoughts for preferring (a)?

Dunno honestly, felt more consistent at the time. with the name I had chosen, but that's not a good reason. You named a good reason (more useful, stricter), so I did (b) now (see newest commit)

pudo:
Lovely feature to have, thanks so much for implementing this, @leonhandreke!

I appreciate you waiting to change `statistics.json`, and would like to give the go-ahead on this change. I'm not sure, though, if it should be `property_values`, or just `properties`. The key `properties` is currently used to store a list of all `qnames` used in the dataset -- Which takes me to the `qname` vs. schema/property thing - bit torn on that. I appreciate and admire that the schema/property double is a more precise spec. A mapping of qnames with counts would be a lot more compact and simple. Maybe we can do one round of examples that we expect and see if the distinction is relevant:

* `Sanction.endDate` - bingo, that's really `Interval.endDate` which is a lot more ambiguous.
* `Thing.name` - yeah makes no sense. 

Ok then let's have the specific combo.

pudo:
Sorry this was very stream of consciousness. Long story short: no suggested changes.

zavod/zavod/exporters/statistics.py
```diff
@@ -57,6 +75,12 @@ def observe(self, entity: Entity) -> None:
         self.schemata.add(entity.schema.name)
         for prop in entity.iterprops():
             self.qnames.add(prop.qname)
+        for prop_name, values in entity.properties.items():
+            # We add 1 instead of len(values) here because we want to count the number of entities that have this
+            # value set, not the number of values.
+            self.property_values_count[
+                SchemaProperty(entity.schema.name, prop_name)
+            ] += 1
```
jbothma, line 83: Now that we've switched to the (b) semantics, can I argue for `entities_with_prop` instead of `property_values`? Because we're asserting the number of entities which have some value for this prop. 

Happy to be rejected. It's pedantry but it's nice for things to be self-explanatory.

jbothma (approved):
Thanks!

PR #1812, opened 2025-01-28
Remove h.parse_date and h.check_no_year

leonhandreke:
See explanations in the commit messages. Both were unused, basically.

### DIFF
--- a/datasets/ua/nabc_sanctions/crawler.py
+++ b/datasets/ua/nabc_sanctions/crawler.py
@@ -1,4 +1,5 @@
 import json
+
 from banal import ensure_list
 from typing import Any, Dict, List, Generator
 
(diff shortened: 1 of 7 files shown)


jbothma:
https://github.com/opensanctions/operations/issues/877

datasets/ua/nabc_sanctions/crawler.py
```diff
@@ -57,13 +58,9 @@ def clean_row(row: Dict[str, Any]) -> Dict[str, Any]:
     return data
 
 
-def parse_date(date):
-    if date is not None:
-        date = date.replace(" ", "")
-    dates = set()
-    for part in h.multi_split(date, [",", "\n", ";"]):
-        dates.update(h.parse_date(part, ["%d.%m.%Y", "dd.%m.%Y"]))
```
jbothma, line 65: I think we also need to add these format strings to the metadata `dates.formats` or something

jbothma (commented):
nice cleanup! Sad that we can't test the affected crawlers, but hopefully quick to just add the last bits to finish the migration

Issue #1797, opened 2025-01-24
Require at least one assertion per crawler


pudo:
Have crawlers emit a warning if they do not define at least one assertion on the count of entities emitted.

PR #1818, opened 2025-01-30
Emit a warning if a dataset does not specify any assertions

leonhandreke:
Fixes https://github.com/opensanctions/opensanctions/issues/1797


jbothma (approved):
Thanks!

Let's merge this Wednesday next week to give @nvmbrasserie a warm welcome

jbothma:
if you're bored, we can probably throw an assertion into one of the existing assertions tests to find out when we break this

leonhandreke:
Added a test!

PR #1827, opened 2025-02-03
[us_cuba_sanctions] Update hashes and document current state of affairs

leonhandreke:
A new entity was added in a recent press release. The list to which this press release refers to doesn't seem to be available for now, so we'll continue using our copy.

### DIFF
--- a/datasets/us/cuba_sanctions/crawler.py
+++ b/datasets/us/cuba_sanctions/crawler.py
@@ -6,9 +6,14 @@ from zavod import helpers as h
 from zavod.shed.zyte_api import fetch_html
 
 ORIGINAL_ACCOMMODATIONS_URL = (
-    "https://www.state.gov/cuba-sanctions/cuba-prohibited-accommodations-list/"
+    "https://www.state.gov/cuba-prohibited-accommodations-list-initial-publication/"
 )
 ACCOMMODATIONS_URL = "https://docs.google.com/spreadsheets/d/e/2PACX-1vQMquWjNWZ09dm9_mu9NKrxR33c6pe4hpiGFeheFT4tDZXwpelLudcYdCdME820aKJJo8TfMKbtoXTh/pub?gid=1890354374&single=true&output=csv"
+# The Restricted Entities list is currently nowhere to be found, the last known public snapshot is at
+# https://web.archive.org/web/20241209000143/https://www.state.gov/cuba-restricted-list/list-of-restricted-entities-and-subentities-associated-with-cuba-effective-january-8-2021/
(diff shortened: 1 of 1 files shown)


datasets/us/cuba_sanctions/crawler.py
```diff
@@ -5,10 +5,13 @@
 from zavod import helpers as h
 from zavod.shed.zyte_api import fetch_html
 
-ORIGINAL_ACCOMMODATIONS_URL = (
-    "https://www.state.gov/cuba-sanctions/cuba-prohibited-accommodations-list/"
-)
+ORIGINAL_ACCOMMODATIONS_URL = "https://www.state.gov/cuba-sanctions/cuba-prohibited-accommodations-list/cuba-prohibited-accommodations-list-initial-publication/"
```
jbothma, line 8: Why are we changing this to a broken link? Was this working yesterday? Perhaps the most sensible link for this is https://www.state.gov/cuba-prohibited-accommodations-list-initial-publication/ so that the hash check tells us if they modify the content and we need to update our spreadsheet.
image

datasets/us/cuba_sanctions/crawler.py
```diff
@@ -51,11 +54,14 @@ def crawl_restricted_entities(context: Context):
...
+    # The restricted list URL page currently points to an interim page that refers to ORIGINAL_ACCOMMODATIONS_URL.
+    # So the page doesn't really contain that list. We just want to alert when it changes
     restricted_list_url = landing_doc.xpath(".//a[text()='Cuba Restricted List']/@href")
     assert len(restricted_list_url) == 1, restricted_list_url
-    doc = fetch_html(context, restricted_list_url[0], CONTENT_XPATH, actions=ACTIONS)
-    node = doc.find(CONTENT_XPATH)
-    if not h.assert_dom_hash(node, "52894f39893ab2432b411441f3bc21c27c3bea7d"):
+    restricted_list_unblock_validator = ".//div[@id='content']"
+    doc = fetch_html(context, restricted_list_url[0], restricted_list_unblock_validator)
+    node = doc.find(restricted_list_unblock_validator)
```
jbothma, line 63: just a naming nit, in this context it's just the xpath to the content section of the page. It just so happens that a sensible "unblock validator" is usually the first xpath you're going to use once the page is fetched.
```suggestion
    content_xpath = ".//div[@id='content']"
    doc = fetch_html(context, restricted_list_url[0], content_xpath)
    node = doc.find(content_xpath)
```

jbothma (commented):
Thanks for adding the docs!

Issue #334, opened 2023-07-27
Find a new way of storing non-sanction adverse info


pudo:
We have a few data sources where we use the `Sanction` schema to describe non-sanction adverse information, such as a procurement debarment, a criminal record or a regulatory penalty. We should have a new schema type - of which Sanction is maybe a sub-type - for those sorts of facts.

brrttwrks:
I've been facing this a bit and agree, that some form of 'fine' 'rebuke' 'penalty'  or more broadly 'assessment', but in most cases (due to our work) negative assessment/judgement/resolution ... I feel we are in thesaurus territory.

brawer:
What is the next step to make this happen?

pudo:
I guess we just need a name, and figure out if we want `Sanction` to be a sub-type of it. Perhaps it also helps to clarify what's really broken about this. In my mind, it's the fact that while we use the term "this company got sanctioned for defrauding the World Bank" it's a different meaning of "sanction" than the US gov coming down on them. The broadest thing is `Punishment`, I guess. 

We'll also have to think about a migration story here, allowing our data re-users to adopt the new schema before we remove the old.

pudo:
Consolidating this at https://github.com/opensanctions/followthemoney/issues/7

PR #1834, opened 2025-02-05
[pl_wanted] Add Polish wanted list

leonhandreke:
Fixes https://github.com/opensanctions/crawler-planning/issues/294

For open questions, see TODOs in the code.

### DIFF
--- a/datasets/pl/wanted/crawler.py
+++ b/datasets/pl/wanted/crawler.py
@@ -30,7 +30,7 @@ def crawl_person(context: Context, url: str):
                 info[key] = text
 
     person = context.make("Person")
-    person.id = context.make_slug(info.get("full_name"), info.get("birth_date"))
+    person.id = context.make_id(info.get("full_name"), info.get("birth_date"))
     person.add("sourceUrl", url)
     person.add("topics", "crime")
     person.add("topics", "wanted")
@@ -65,7 +65,8 @@ def crawl_person(context: Context, url: str):
     )
     if not crimes:
         context.log.warn("No crimes found for person", entity_id=person.id, url=url)
-    person.add("notes", crimes)
+    for crime in crimes:
+        person.add("notes", f"Wanted for: {crime}")
 
     context.audit_data(info)
 
@@ -74,7 +75,7 @@ def crawl_person(context: Context, url: str):
 
 def crawl_index(context, url) -> str | None:
     context.log.info(f"Crawling index page {url}")
-    doc = context.fetch_html(url, cache_days=1)
+    doc = context.fetch_html(url)
     # makes it easier to extract dedicated details page
     doc.make_links_absolute(context.dataset.data.url)
     cells = doc.xpath("//li[.//a[contains(@href, '/pos/form/r')]]/a/@href")
--- a/datasets/pl/wanted/crawler.py
+++ b/datasets/pl/wanted/crawler.py
@@ -1,12 +1,24 @@
 import re
 from typing import List
 
-import countrynames
 
 from zavod import Context
 from zavod import helpers as h
 
 NULL_CITIZENSHIPS = ["NIEUSTALONE", "BEZPAŃSTWOWIEC"]
+CITIZENSHIPS_WITH_SPACES = [
+    "WIELKA BRYTANIA",
+    "SRI LANKA",
+    "ARABIA SAUDYJSKA",
+    "NOWA ZELANDIA",
+    "BOŚNIA I HERCEGOWINA",
+    "ZJEDNOCZONE EMIRATY ARABSKIE",
+    "KOREAŃSKA REPUBLIKA LUDOWO-DEMOKRATYCZNA",
+    "WYBRZEŻE KOŚCI SŁONIOWEJ",
+    "DEMOKRATYCZNA REPUBLIKA KONGA",
+    "SIERRA LEONE",
+    "KOREA POŁUDNIOWA",
+]
 
 
 def clean_and_split_citizenship(citizenship: str) -> str | List[str]:
@@ -20,62 +32,22 @@ def clean_and_split_citizenship(citizenship: str) -> str | List[str]:
         A string or a list of string containing the cleaned citizenship(s).
 
     """
+    res = []
     # Sometimes, we have another spelling of the country in brackets, like "USA (STANY ZJEDNOCZONE AMERYKI)"
-    alt_country_match = re.search(r"\((?P<alt_country>.+)\)", citizenship)
-    if alt_country_match:
-        cleaned_citizenship = re.sub(r" \(.+\)", "", citizenship)
-    else:
-        cleaned_citizenship = citizenship
+    cleaned_citizenship = re.sub(r" \(.+\)", "", citizenship)
+    # Split citizenships with spaces out before naively splitting on space
+    for c in CITIZENSHIPS_WITH_SPACES:
+        if c in cleaned_citizenship:
+            cleaned_citizenship = cleaned_citizenship.replace(c, "")
+            res.append(c)
 
-    if countrynames.to_code(cleaned_citizenship, fuzzy=True):
-        return cleaned_citizenship
+    res += cleaned_citizenship.split(" ")
 
-    if " " in cleaned_citizenship:
-        split_citizenships = cleaned_citizenship.split(" ")
-        # Somtimes a person is "NIEUSTALONE POLSKA" aka unknown polish
-        split_citizenships = [
-            c for c in split_citizenships if c not in NULL_CITIZENSHIPS
-        ]
-
-        # Multiple citizenships are space-delimited (yay), see if space-splitting them works
-        if all([countrynames.to_code(x) for x in split_citizenships]):
-            return split_citizenships
-        else:
-            # Sometimes, the citizenship looks like this "WIELKA BRYTANIA IZRAEL"
-            # Covering this special case here (just one split) is easy and cheap
-            for i in range(1, len(split_citizenships) + 1):
-                citizenship_1 = " ".join(split_citizenships[:i])
-                citizenship_2 = " ".join(split_citizenships[i:])
-                if countrynames.to_code(
-                    citizenship_1, fuzzy=True
-                ) and countrynames.to_code(citizenship_2, fuzzy=True):
-                    return [citizenship_1, citizenship_2]
-
-    # Sometimes, there is a full spelling of the country in parens
-    if alt_country_match:
-        alt_country = alt_country_match.groupdict().get("alt_country")
-        if countrynames.to_code(alt_country, fuzzy=True):
-            return alt_country
-
-    # If nothing worked, we just return the original value
-    return citizenship
+    return res
 
 
 def crawl_person(context: Context, url: str):
     context.log.debug(f"Crawling person page {url}")
-
-    person = context.make("Person")
-    # TODO(Leon Handreke): Do we use the r12323456 in the URL that looks like a record ID?
-
-    # There is an ID prefixed by r in the URL, which seems to refer to the record. The name component in the URL
-    # can be changed at will.
-    # URLs look like this: https://poszukiwani.policja.gov.pl/pos/form/r922391307881,OCHNIO-KRZYSZTOF.html
-    url_match = re.search(r"/pos/form/(?P<rid>r\d+),.+.html", url)
-    assert (
-        url_match and "rid" in url_match.groupdict()
-    ), "Could not extract r-ID from malformed person URL"
-    rid = url_match.groupdict()["rid"]
-    person.id = context.make_slug(rid)
 
     doc = context.fetch_html(url, cache_days=7)
     # Extract details using XPath based on the provided HTML structure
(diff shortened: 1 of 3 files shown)


datasets/pl/wanted/crawler.py
```diff
@@ -0,0 +1,160 @@
...
+    # On the last page, the next button will not have an <a>, so this will not match
+    next_button_href = doc.xpath(
+        "//li/a/span[contains(text(), 'następna')]/parent::a/@href"
+    )
+    return next_button_href[0] if next_button_href else None
+
+
+def crawl(context):
+    next_url = context.dataset.data.url
+    # Use this construction instead of recursion because Python sets a recursion limit
```
jbothma, line 158: if python had tail recursion 🥹

datasets/pl/wanted/crawler.py
```diff
@@ -0,0 +1,160 @@
...
+
+    crimes = doc.xpath(
+        "//p[contains(text(), 'Podstawy poszukiwań:')]/following-sibling::ul//a/text()"
+    )
+    if not crimes:
+        context.log.warn("No crimes found for person", entity_id=person.id, url=url)
+    # TODO(Leon Handreke): Every wanted list puts this in notes -- could there be a better place?
+    person.add("notes", crimes)
+
+    context.emit(person, target=True)
```
jbothma, line 137: no need for `target` any more. We [recently](https://www.opensanctions.org/changelog/13/) switched to setting `Entity.target` based on https://github.com/opensanctions/opensanctions/blob/main/zavod/zavod/settings.py#L33

Soon we should do a cleanup.

```suggestion
    context.emit(person)
```

datasets/pl/wanted/crawler.py
```diff
@@ -0,0 +1,160 @@
...
+
+    person.add("citizenship", cleaned_citizenship, original_value=citizenship)
+
+    crimes = doc.xpath(
+        "//p[contains(text(), 'Podstawy poszukiwań:')]/following-sibling::ul//a/text()"
+    )
+    if not crimes:
+        context.log.warn("No crimes found for person", entity_id=person.id, url=url)
+    # TODO(Leon Handreke): Every wanted list puts this in notes -- could there be a better place?
+    person.add("notes", crimes)
```
jbothma, line 135: potentially when we have a more general entity than Sanction for describing the risk posed by the entity (why they're in our database), this could be taken into account https://github.com/opensanctions/opensanctions/issues/334

datasets/pl/wanted/crawler.py
```diff
@@ -0,0 +1,160 @@
...
+        if q:
+            text = q[0].strip()
+            if text != "-":
+                info[key] = text
+
+    person.add("sourceUrl", url)
+    person.add("topics", "crime")
+    person.add("topics", "wanted")
+    # TODO(Leon Handreke): What does country mean in that context? The za_wanted crawler also does this.
+    person.add("country", "pl")
```
jbothma, line 108: we use country when there's some other association with a country, e.g. personally conducting business there, or in this case being wanted for a crime suggests some connection with the country. I think it can be useful for disambiguation of names.

datasets/pl/wanted/crawler.py
```diff
@@ -0,0 +1,160 @@
...
+
+    h.apply_name(person, full=info["full_name"], middle_name=info.get("middle_name"))
+    h.apply_date(person, "birthDate", info.get("date_of_birth"))
+    person.add("birthPlace", info.get("place_of_birth"))
```
jbothma, line 112: Not super important, but it might be a thing to use the same keys as our props. Not sure that gives us an automatic mapping with readable code, but it's at least consistent.

datasets/pl/wanted/crawler.py
jbothma, line 137: it's nice to context.audit dicts like info when we're done to help catch a code edit that forgets one of the fields

datasets/pl/wanted/crawler.py
```diff
@@ -0,0 +1,160 @@
...
+
+    # There is an ID prefixed by r in the URL, which seems to refer to the record. The name component in the URL
+    # can be changed at will.
+    # URLs look like this: https://poszukiwani.policja.gov.pl/pos/form/r922391307881,OCHNIO-KRZYSZTOF.html
+    url_match = re.search(r"/pos/form/(?P<rid>r\d+),.+.html", url)
+    assert (
+        url_match and "rid" in url_match.groupdict()
+    ), "Could not extract r-ID from malformed person URL"
+    rid = url_match.groupdict()["rid"]
+    person.id = context.make_slug(rid)
```
jbothma, line 78: it's tempting, but it could be an artifact of their software system rather than some operational consistent identifier. If it's the former, I worry a bit that it's not stable enough, and I'd feel more comfortable with context.make_id(full_name, birth_date) and maybe chucking place of birth in there too, especially if birth date isn't super consistently available.

datasets/pl/wanted/crawler.py
```diff
@@ -0,0 +1,160 @@
...
+        if all([countrynames.to_code(x) for x in split_citizenships]):
+            return split_citizenships
+        else:
+            # Sometimes, the citizenship looks like this "WIELKA BRYTANIA IZRAEL"
+            # Covering this special case here (just one split) is easy and cheap
+            for i in range(1, len(split_citizenships) + 1):
+                citizenship_1 = " ".join(split_citizenships[:i])
+                citizenship_2 = " ".join(split_citizenships[i:])
+                if countrynames.to_code(
+                    citizenship_1, fuzzy=True
```
jbothma, line 50: We prefer to approach this more like this:

```diff
index aada0edc..cbf6b7b8 100644
--- a/datasets/pl/wanted/crawler.py
+++ b/datasets/pl/wanted/crawler.py
@@ -120,11 +120,10 @@ def crawl_person(context: Context, url: str):
     person.add("eyeColor", info.get("eye_color"))

     citizenship = info.get("citizenship")
-    cleaned_citizenship = (
-        clean_and_split_citizenship(citizenship) if citizenship else None
-    )
-
-    person.add("citizenship", cleaned_citizenship, original_value=citizenship)
+    if citizenship and "POLSKA" in citizenship:
+        person.add("citizenship", "PL")
+        citizenship = citizenship.replace("POLSKA", "").strip()
+    person.add("citizenship", h.multi_split(citizenship, ["(", ")"]))

     crimes = doc.xpath(
         "//p[contains(text(), 'Podstawy poszukiwań:')]/following-sibling::ul//a/text()"
```

I get a small number more warnings, but it avoids splitting `SRI LANKA` on space, and skips a whole load of python code

jbothma: to be fair, I didn't run all thousands of pages, but about 100 pages in it looks sane

datasets/pl/wanted/pl_wanted.yml
```diff
@@ -0,0 +1,76 @@
...
+    options:
+      - match:
+          - NIEUSTALONE  # unknown
+          - BEZPAŃSTWOWIEC  # stateless
+        value: null
+      - match: "JUGOSŁAWIA (SERBIA I CZARNOGÓRA)"  # TODO: What the hell should we map this to?
+        value: yucs
+      - match: "JUGOSŁAWIA"  # TODO: What the hell should we map this to?
+        value: yucs
+      - match: "NIEMCY JUGOSŁAWIA (SERBIA I CZARNOGÓRA)"  # TODO: What the hell should we map this to?
```
jbothma, line 60: we got you, fam https://www.opensanctions.org/reference/#:~:text=Serbia%20and%20Montenegro

datasets/pl/wanted/pl_wanted.yml
```diff
@@ -0,0 +1,76 @@
...
+  official: true
+  url: https://www.policja.pl
+
+data:
+  url: https://poszukiwani.policja.gov.pl/pos/form/5,Poszukiwani.html
+  format: HTML
+  lang: pol
+
+dates:
+  formats: ["%Y-%m-%d"]
```
jbothma, line 33: _shouldn't_ be needed - we have that as a default

datasets/pl/wanted/crawler.py
```diff
@@ -0,0 +1,160 @@
...
+
+    h.apply_name(person, full=info["full_name"], middle_name=info.get("middle_name"))
+    h.apply_date(person, "birthDate", info.get("date_of_birth"))
+    person.add("birthPlace", info.get("place_of_birth"))
+    person.add("gender", info.get("gender"))
+    person.add("alias", info.get("alias"))
+    person.add("fatherName", info.get("father_name"))
+    person.add("motherName", info.get("mother_name"))
+    # TODO(Leon Handreke): Add mother maiden name or is that too detailed?
+    person.add("motherName", info.get("mother_maiden_name"))
```
jbothma, line 118: I think it's fair to add since it's available

datasets/pl/wanted/pl_wanted.yml
```diff
@@ -0,0 +1,76 @@
...
+      - match: "NIEMCY JUGOSŁAWIA (SERBIA I CZARNOGÓRA)"  # TODO: What the hell should we map this to?
+        values:
+          - de
+          - yucs
+      - match: "DEMOKRATYCZNA REPUBLIKA KONGA (ZAIR) POLSKA KONGO"
+        values:
+          - pl
+          - cd
+      # TODO: Move the below to opensanctions/countrynames
+      - match: KOSOWO
```
jbothma, line 69: maybe a github issue is more visible than a todo in code we wont look at for a year

jbothma (commented):
awesome, thanks!

datasets/pl/wanted/crawler.py
leonhandreke: First thing I searched for, apparently Guido has opinions on this that I did not care to investigate further...

datasets/pl/wanted/crawler.py
leonhandreke: This is the sort of thing that IntelliJ should be good at :)

datasets/pl/wanted/pl_wanted.yml
leonhandreke: Let's merge https://github.com/opensanctions/countrynames/pull/46 and be done with it?

datasets/pl/wanted/pl_wanted.yml
leonhandreke: With the new logic, it'll just use JUGOSŁAWIA and match that to yucs.

Serbia and Montenegro doesn't exist anymore either, so not sure that's more useful. WDYT?

datasets/pl/wanted/crawler.py
leonhandreke: That didn't really work, there are just far too many double citizenships that are not pl. I now specialcased all those with spaces. How do you feel about that?

datasets/pl/wanted/crawler.py
leonhandreke: Done.

datasets/pl/wanted/pl_wanted.yml
jbothma: we just try and represent the data as it is at the source, so we do normally map to these territories

datasets/pl/wanted/crawler.py
jbothma: fair enough. It's already much quicker to sanity check than the original.

The failure mode of this isn't ideal - if there's a country with a space where one of the parts is a valid country, we'll assign that country (which might be wrong) until we have time to fix it, assuming we get a warning for the second part.

How many is too many? The options I see are

- run the crawler through once, dump the countries from issues.json, unique and paste them into the yaml
- check all the country names at https://www.opensanctions.org/reference/#type.country for country parts that might be real countries and make sure we've added their polish form. Hope new ones don't occur.
- keep a rather complex country splitting function which also has cases like serbia and montenegro which doesn't accurately reflect the data source - hopefully it'd warn on "and"

leonhandreke: Yeah, I think the current way isn't too bad. Should I put the list in dataset.data in the YAML instead? And then call it a day?

datasets/pl/wanted/pl_wanted.yml
leonhandreke: I guess I'm not quite sure what the intention of the source is here.

Anyway "JUGOSŁAWIA (SERBIA I CZARNOGÓRA)" now maps to yucs (because we throw away anything in parens). Are we okay with that or should I modify to deal with this and map to both yucs and csxx?

jbothma: I guess they mean the person was from the part of yugoslavia which became "serbia and montenegro"? But I haven't confirmed that the bit in parents is always "another form" of the same thing.

do whatever's easier between verifying that the bit in parentheses always means "another form of the bit before the parens", or just playing it safe and always including both.

We try not to throw information away, and maybe it's ok here, but my gut says it's not that hard to include both.

datasets/pl/wanted/crawler.py
```diff
@@ -0,0 +1,95 @@
...
+    person.add("fatherName", info.pop("father_name", None))
+    person.add("motherName", info.pop("mother_name", None))
+    person.add("motherName", info.pop("mother_maiden_name", None))
+    person.add("height", info.pop("height", None))
+    person.add("eyeColor", info.pop("eye_color", None))
+
+    citizenship = info.pop("citizenship", None)
+    citizenship_original_value = citizenship
+    if citizenship and "POLSKA" in citizenship:
+        person.add("citizenship", "pl", original_value=citizenship_original_value)
```
jbothma, line 55: nice one capturing the original

jbothma (approved):
Thanks!

PR #1839, opened 2025-02-05
Add Sentry integration

leonhandreke:
This forks https://github.com/kiwicom/structlog-sentry (it's a single file thing). I split the import into a separate commit. I guess that's a bit ugly, but it seems small and simple enough to fork and I wanted to make some changes (see second commit in this series).

Something I haven't quite figured out is why the redactor only seems to operate on the stderr log and not on the issues files -- I guess this is intended behavior, but why? Maybe document that decision in the code?


jbothma:
> Something I haven't quite figured out is why the redactor only seems to operate on the stderr log and not on the issues files -- I guess this is intended behavior, but why? Maybe document that decision in the code?

Definitely unintended 🤦

jbothma:
some hidden datasets have many thousands of warnings

jbothma:
part of our approach to complete some difficult crawls (like interpol, where the design has been to build up the cache and retry the entire cron job a few times until int completes) will show up as errors even though the crawl eventually completed.

Maybe this is a good thing - we have better retry tooling now and maybe we can just adjust that to complete the job within one crawl.

We can also use "archive until rate increases" to hide these if we need to.

jbothma:
it'd be nice to be able to see issues in the ru_egrul data only for the entities that occur in ext_ru_egrul

e.g. by passing bad values through from ru_egrul, or by passing a list of ext_ru_Egrul entities to warn when running ru_egrul

zavod/zavod/sentry.py
```diff
@@ -0,0 +1,268 @@
...
+
+        event["message"] = str(event_dict.get("event"))
+        event["level"] = event_dict.get("level")  # type: ignore
+        if self._fingerprint is not None:
+            event["fingerprint"] = self._fingerprint
+        if "logger" in event_dict:
+            event["logger"] = event_dict["logger"]
+
+        if self._event_dict_as_extra:
+            print(event_dict)
```
jbothma, line 176: stray print

zavod/zavod/settings.py
```diff
@@ -116,3 +116,7 @@ def env_str(name: str, default: str) -> str:
 ZYTE_API_KEY = env.get("OPENSANCTIONS_ZYTE_API_KEY", None)
 OPENAI_API_KEY = env.get("OPENSANCTIONS_OPENAI_API_KEY", None)
 AZURE_OPENAI_ENDPOINT = env.get("OPENSANCTIONS_AZURE_OPENAI_ENDPOINT", None)
+
+ENABLE_SENTRY = env.get("ZAVOD_ENABLE_SENTRY", False)
```
jbothma, line 120:
```suggestion
ENABLE_SENTRY = as_bool(env.get("ZAVOD_ENABLE_SENTRY", False))
```

So that "false" isn't True

jbothma (commented):
Looks good!

Noted some things we need to look at before we merge as checklist.

Also wonder if we can get the dataset name listed in all cases? Not necessarily in that slot, but somewhere in the issue list. That helps with prioritisation, e.g. we want to make sure we tackle OFAC, GB, EU sanctions ASAP in the mornings
image

jbothma:
We probably also want to configure this so that we don't get emails, by default?

leonhandreke:
> Also wonder if we can get the dataset name listed in all cases? Not necessarily in that slot, but somewhere in the issue list. That helps with prioritisation, e.g. we want to make sure we tackle OFAC, GB, EU sanctions ASAP in the mornings

It just shows the `logger` field there. I could not find a way to change that. I changed it to just push the dataset field from the log message into that field, and the logger into a new property called `python_logger`. Not beautiful, but it should work.

> We probably also want to configure this so that we don't get emails, by default?

Are you still getting them? I thought I disabled them yesterday, but maybe something is still off.

jbothma:
>>  Also wonder if we can get the dataset name listed in all cases? Not necessarily in that slot, but somewhere in the issue list. That helps with prioritisation, e.g. we want to make sure we tackle OFAC, GB, EU sanctions ASAP in the mornings

> It just shows the logger field there. I could not find a way to change that. I changed it to just push the dataset field from the log message into that field, and the logger into a new property called python_logger. Not beautiful, but it should work.

Ah makes sense. Another option is prefixing the message with `[dataset_name] ` then issues are automatically distinct per dataset

jbothma:
https://sentry.io/settings/account/notifications/alerts/

image

Yeah I got some when I tried it. Maybe you disabled in your personal settings?

Looks like we can disable project-wide here so I think it'll be fine.

leonhandreke:
See https://github.com/opensanctions/opensanctions/pull/1899 for another pragmatic go at making ru_egrul not flood Sentry.

zavod/zavod/sentry.py
```diff
@@ -0,0 +1,272 @@
...
+
+        if self.tag_keys == "__all__":
+            event["tags"] = self._original_event_dict.copy()
+        if isinstance(self.tag_keys, list):
+            event["tags"] = {
+                key: event_dict[key] for key in self.tag_keys if key in event_dict
+            }
+        # Because we push the dataset into the Event.logger field (see above), we push the logger field into the tags.
+        if "logger" in event_dict:
+            event["tags"]["python_logger"] = event_dict["logger"]
```
jbothma, line 193: sorry for the late one - this feels like a bit of a hack. We might also miss out on built-in sentry functionality if we hack the logger name.

Is there a less hacky way to just make sure dataset name is in the message like `[dataset_name] Something happened` _and_ tags?

leonhandreke: You're completely right, fixed as suggested.

zavod/zavod/sentry.py
```diff
@@ -188,9 +192,8 @@ def _get_event_and_hint(
             event["tags"] = {
                 key: event_dict[key] for key in self.tag_keys if key in event_dict
             }
-        # Because we push the dataset into the Event.logger field (see above), we push the logger field into the tags.
         if "logger" in event_dict:
-            event["tags"]["python_logger"] = event_dict["logger"]
+            event["logger"] = event_dict["logger"]
```
jbothma, line 196: watch out we're still setting this on line 176 too

jbothma (approved):
can't wait!!

zavod/zavod/sentry.py
leonhandreke: Urgh, sorry. Good catch!

PR #1841, opened 2025-02-06
[nl_most_wanted] Update crawler to new webpage layout

leonhandreke:
Last export has 165 statements, this one has 168. Not quite sure what happened, and I don't have a diffing tool yet to figure out what happened.

All entities are being re-keyed because we're keying on source URL. Maybe keying on name is actually more stable and now is the time to do it?


jbothma:
if we're re-keying anyway then yes, I think name, birthDate, and birthPlace or citizenship depending on which one is more commonly available

PR #1842, opened 2025-02-06
Fixes-06-02

nvmbrasserie:


datasets/_externals/wikidata.yml
```diff
@@ -978,6 +979,8 @@ lookups:
           - South African Republic
           - Ciskei
         value: ZA
+      - match: Villavicencio
```
leonhandreke, line 982: I think I fixed this in the data https://www.wikidata.org/wiki/Q131917174

leonhandreke:
I renmoved the nl_most_wanted commits. LGTM from my side.

datasets/_externals/wikidata.yml
jbothma: it's probably going to take a few days to disappear from cache - well spotted

PR #1850, opened 2025-02-07
assertions extraction

nvmbrasserie:


jbothma:
Let's call it something like suggest_assertions.py - assert is very vague.

Is this rounding, or do you think it can round so that e..g. 123 becomes 100 and 123456 becomes 123000 or something like that? Or are you editing it manually to be a bit more sane?

jbothma:
Feel free to merge once you've renamed. Don't worry about rounding for now

PR #1860, opened 2025-02-10
[ru_egrul] More country mappings and a bit of refactor

leonhandreke:
This is in preparation making this crawler silent so we can also have it record to Sentry and don't have to special-case it (see https://github.com/opensanctions/operations/issues/874)

Do we get warnings if some countries are missing from ext_egrul? We do, right?


jbothma (approved):
thought I sent this earlier. 
Very nice.

PR #1868, opened 2025-02-11
Assertions-10

nvmbrasserie:


datasets/lv/fiu_sanctions/lv_fiu_sanctions.yml
```diff
@@ -35,3 +35,15 @@ publisher:
...
+  min:
+    schema_entities:
+      Person: 1
+      Address: 1
+      Organization: 1
+  max:
+    schema_entities:
+      Person: 15
+      Address: 15
+      Organization: 15
```
jbothma, line 49: here I can also imagine them issuing a bunch of sanctions - I wonder if a rule of thumb of a maximum of at least 50 per type is a good idea? Except for cases where we can reason that a lower max makes sense? What do you think?

PR #1871, opened 2025-02-11
[ru_egrul] Defensively use Entity.add to avoid emitting warnings

leonhandreke:


jbothma:
> I feel it's better to just shut up that logger instead of littering these checks all over the code.

Yeah 9 of these `if`'s isn't exactly making the code more readable. It's useful to see it in the flesh for this I think

so what are our options?

- shut up the logger - what does this look like? 
  - allow dataset metadata to say only error level is an issue, and info and warning still goes to stdout?
  - during dev someone can uncomment issue_level: INFO and get issues.json to munge structured logs like jq .issues[].message issues.json |sort|uniq -c|sort -n
  - I can't really think of anything we miss this way. Currently we don't see any warnings in zavod run within or outside the crawler because it's all hidden. Now we'll actually get alerted for errors in these big crawlers
- 9+ duplicate validations and if's making the code uglier and slower
- an argument like Entity.add(drop_invalid=True) - you suggested tacking this on in there is adding complexity to a super important and already complicated method

jbothma:
@leonhandreke I think if I understood the plan roughly correctly it's totally worth seeing what an implementation of that looks like

leonhandreke:
Maybe just a `logger.getLogger("zavod.runtime.cleaning").setLevel("error")` in the code would do the trick? Also easy to remove for development.

leonhandreke:
Like this: https://github.com/opensanctions/opensanctions/pull/1874

jbothma:
I like https://github.com/opensanctions/opensanctions/pull/1874

PR #1874, opened 2025-02-11
[ru_egrul] Silence warnings from cleaning

leonhandreke:
We don't want them being logged and warning us in Sentry, we only want that for ext_ru_egrul

### DIFF
--- a/datasets/ru/egrul/crawler.py
+++ b/datasets/ru/egrul/crawler.py
@@ -1,3 +1,4 @@
+import logging
 import re
 from typing import Dict, Optional, Set, IO, List, Any, Tuple
 from collections import defaultdict
@@ -6,7 +7,7 @@ from zipfile import ZipFile
 from lxml import etree
 from lxml.etree import _Element as Element, tostring
 
-
+import zavod.logs
 from zavod import Context, Entity
 from zavod import helpers as h
 from zavod.shed.internal_data import fetch_internal_data, list_internal_data
@@ -287,7 +288,7 @@ def parse_founder(context: Context, company: Entity, el: Element) -> None:
         return
 
     if owner.id is None:
-        context.log.warning(
+        context.log.info(
             "No ID for owner: %s" % company.id, el=tag_text(el), owner=owner.to_dict()
         )
         return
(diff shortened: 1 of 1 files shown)


jbothma (approved):
I can't find an argument for a more "official" approach. Looks good to me!

leonhandreke:
I thought about this a bit harder over night and I'm not sure silencing all logs is so smart anymore. Right now, logging enables us to jq issues.json, and for ru_egrul, this is valuable. I'll try to think of a better way...

datasets/ru/egrul/crawler.py
```diff
@@ -715,6 +716,9 @@ def crawl_archive(context: Context, blob_name: str) -> None:
 
 
 def crawl(context: Context) -> None:
+    # Shut up warnings being logged, we only care about them once they get pulled
+    # into our default dataset by ext_ru_egrul
+    zavod.logs.get_logger("zavod.runtime.cleaning").setLevel(logging.ERROR)
```
jbothma, line 721: I've been thinking, if we change the logging we might not think to update these log disables, and end up spamming ourselves (and possibly exhausting sentry quotas until we pay more or something) - missing critical feedback from crawlers.

I wonder if there's a way to make it more robust to such changes

pudo: This feels smelly to me. What I'd prefer is that we exempt specific classes of FtM validation warnings - the easiest is by pre-cleaning values like:

```
from followthemoney.types import registry

def crawl():
  ...
  address = node.text
  address_clean = registry.address.clean_text(address)
  if address_clean is not None:
    entity.add("address", address_clean, original_value=address)
```

Hope that makes sense?

jbothma: That implies 9 of those in this crawler, I believe. Just checking you're ok with that, or if that sways us to try harder to make "expempt specific classes" cleaner, or try other solutions?

https://github.com/opensanctions/opensanctions/pull/1871

pudo: I'm just really worried about messing with the logging system on a crawler-by-crawler basis, perhaps we can take that option off the menu. We could still think about adding a `mute_type_warnings` array to the dataset YAML, or make an argument to `.add(mute=True`?

PR #1876, opened 2025-02-12
Assertions-14

nvmbrasserie:


datasets/ps/local_freezing/ps_local_freezing.yaml
```diff
@@ -31,6 +31,13 @@ url: "https://www.pma.ps/en/Legislation//AMLCFT//LocalFreezingList"
...
 dates: 
   formats: ["%d/%m/%Y"]
+
+assertions:
+  min:
+    schema_entities:
+      Person: 4
+  max:
+    schema_entities:
+      Person: 10
```
jbothma, line 43: remember the min max is 50 - is there something about this list to argue for smaller?

nvmbrasserie: yup, it corresponds to a specific decree from 2015. I wouldn't expect it to grow

PR #1877, opened 2025-02-12
Assertions-15

nvmbrasserie:


datasets/rs/apml_domestic/rs_apml_domestic.yml
```diff
@@ -31,9 +31,17 @@ url: https://www.apml.gov.rs/english/targeted-financial-sanctions
...
 dates: 
   formats: ["%d.%m.%Y"]
+
+assertions:
+  min:
+    schema_entities:
+      Person: 5
+  max:
+    schema_entities:
+      Person: 15
```
jbothma, line 43: remember the minimum maximum rule of thumb is 50 - is this tighter because of something specific to this source?

nvmbrasserie: yeah, that's a pdf from 2018 that contain 7 people. I think in this case 15 is a fair maximum

PR #1881, opened 2025-02-13
[sg_gov_dir] Crawl subpages

jbothma:


jbothma:
I'd say this is ready to ship, but let's see if there's more position feedback we can work into default decisions tomorrow morning before shipping. Maybe we'll save ourselves a few hundred decisions.

PR #1888, opened 2025-02-13
Iraq Anti-Money Laundering and Terrorist Financing Office - Local and International Lists

nvmbrasserie:
fixes opensanctions/crawler-planning#295

### DIFF
--- a/datasets/iq/aml_list/crawler.py
+++ b/datasets/iq/aml_list/crawler.py
@@ -4,6 +4,7 @@ from rigour.mime.types import XLSX
 
 from zavod import Context, helpers as h
 
+SOURCE_URLS = ["https://aml.iq/?page_id=2169", "https://aml.iq/?page_id=2171"]
 PERSON_LISTS = [
     "2017",
     "2018",
@@ -15,6 +16,8 @@ PERSON_LISTS = [
     "2024",
     "2025",
 ]
+INT_LISTS = ["كيانات", "افراد"]
+IGNORE = ["الافراد", "القوائم المحلية "]
 YEAR_PATTERN = re.compile(r"\b(" + "|".join(PERSON_LISTS) + r")\b")
 COMPANY_NAME_REASON = r"\s*للتوسط ببيع وشراء العملات الاجنبية$"
 
@@ -45,10 +48,11 @@ def crawl_row(row: dict, context: Context):
         context.emit(entity)
         sanction = h.make_sanction(context, entity)
     else:
-        name = row.pop("asm_alshkhs")  # اسم الشخص
+        name = row.pop("asm_alshkhs", row.pop("asma_alashkhas", None))  # اسم الشخص
         person = context.make("Person")
         person.id = context.make_id(id, name)
         person.add("topics", "debarment")
+        person.add("nationality", row.pop("aljnsyt", None), lang="ara")  # الجنسية
         h.apply_date(person, "birthDate", row.pop("altwld"))  # التولد
         h.apply_name(
             person,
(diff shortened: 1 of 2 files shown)


datasets/iq/aml_list/crawler.py
```diff
@@ -0,0 +1,135 @@
...
+        "القائمة-الدولية",  # International list
+    )
+    process_xlsx(
+        context,
+        SOURCE_URLS[1],
+        "local.xlsx",
+        LOC_PERSON_LISTS + LOC_ENT_LIST,
+        "القوائم-المحلية",  # Local lists
+        LOC_IGNORE_LIST,
+    )
```
jbothma, line 135: can we just hardcode the source URLs here? It's so much easier to know what we're talking about. So much harder to break by swapping them in the list and forgetting to check if order matters.

datasets/iq/aml_list/iq_aml_list.yaml
```diff
@@ -0,0 +1,50 @@
...
+  max:
+    schema_entities:
+      Person: 10000
+      LegalEntity: 80
+
+lookups:
+  type.date: 
+    options: 
+      - match:
+          - 'لا يوجد'
```
jbothma, line 48: it's nice to have the translation in comments for quick reference

datasets/iq/aml_list/crawler.py
```diff
@@ -0,0 +1,135 @@
...
+def extract_listing_date(decision_number: str) -> Optional[str]:
+    """Extracts the listing year from the decision number."""
+    if decision_number is None:
+        return None
+    match = YEAR_PATTERN.search(decision_number)
+    return match.group(0) if match else None
+
+
+def crawl_row(row: dict, context: Context):
+    entity_name = row.pop("asm_alkyan", None)  # اسم الكيان
```
jbothma, line 54: could you pass a header_lookup arg to parse_xlsx_sheet to translate these to English?

datasets/iq/aml_list/crawler.py
```diff
@@ -0,0 +1,135 @@
...
+    "2020",
+    "2021",
+    "2022",
+    "2023",
+    "2024",
+    "2025",
+]
+LOC_ENT_LIST = ["الكيانات"]
+LOC_IGNORE_LIST = ["الافراد", "القوائم المحلية "]
+INT_LISTS = ["كيانات", "افراد"]
```
jbothma, line 23: it'd be nice to have translations of these in comments for quick reference

jbothma: and ENTITY_NAME_REASON. If I failed to do this in the past, I now apologise sincerely :)

datasets/iq/aml_list/crawler.py
```diff
@@ -0,0 +1,135 @@
...
+    "2017",
+    "2018",
+    "2019",
+    "2020",
+    "2021",
+    "2022",
+    "2023",
+    "2024",
+    "2025",
+]
```
jbothma, line 20: what happens next year?

Can we write the things that need this so it isn't time-dependent?

For the listing year, maybe we can match \b\d{4}\b? If that's too broad, maybe then do a if 2015 < value < 2050 or whatever?

For the sheets, maybe we can process all sheets except the ignore list? workbook.sheet_names or something.

datasets/iq/aml_list/crawler.py
```diff
@@ -0,0 +1,135 @@
...
+    decision_number = row.pop("rqm_alqrar")  # رقم القرار
+
+    reason = extract_reason(entity_name)
+    entity_name = clean_entity_name(entity_name)
+    listing_date = extract_listing_date(decision_number)
+
+    if entity_name:
+        entity = context.make("LegalEntity")
+        entity.id = context.make_id(id, entity_name, decision_number)
+        entity.add("topics", "debarment")
```
jbothma, line 65: this looks a lot like full-blown sanction to me. It's in the context of the UN sanctions without any suggestion it's less severe. It's fund freezes. Have you seen anything to suggest it's more of a debarment, as in exclusion from a market place like public procurement or a stock market?

jbothma (commented):
Thanks!

datasets/iq/aml_list/crawler.py
nvmbrasserie: nope, from the resolutions that I checked, it's always fund freezes

jbothma: I'm also changing Person to `sanction` - not sure if it was left as `debarment` intentionally

Issue #1896, opened 2025-02-17
Helper function to check if a value would pass validation


pudo:
We are currently using this pattern a lot to avoid verbose errors on very messy fields:

```
if registry.country.clean(crawled_value):
  entity.add("country", crawled_value)
```

Unfortunately, `type.clean` doesn't do all the checks that `entity.add` does. So we want to have a helper that actually produces the same outcome, e.g.:

```
if h.check_clean(entity, "country", crawled_value):
  entity.add("country", crawled_value)
```

Other proposals were `entity.add("country", crawled_value, mute_warnings=True)` and `entity.check_value(prop, value)`, as well as a context manager:

```
with entity.mute():
  entity.add(prop, value)
```

In any case, let's tackle this after having a basic sentry integration going (#1839) so we don't get los in yak shaving.

leonhandreke:
`type.clean` also does not consider datapatch overrides, date formats (although `Entity.add` doesn't either, you'd have to use `apply_date` for that) and identifier formats (e.g. INN).

leonhandreke:
This would also be nice to invalidate our cache of invalid data in the wikidata enricher.

goose-life:
refactoring `cleaning` will likely solve this

PR #1899, opened 2025-02-17
Allow noisy crawlers to dial down log level for Sentry

leonhandreke:


jbothma (approved):
Cool, yeah this works, and https://github.com/opensanctions/opensanctions/issues/1896 is in To Do.

PR #1901, opened 2025-02-17
[docs] Add entities_with_prop asserts to documentation

leonhandreke:
And restructure the docs a bit to make the sentences a bit more straightforward and make it clear what the countries magic actually does.


nvmbrasserie (approved):

jbothma:
So much better

PR #1903, opened 2025-02-17
Make max assertions only warn, without aborting a crawler run

jbothma:
before and after

image


jbothma:
Yeah the docs help a lot, thanks.

I think making it configurable would add more complexity than we want at the moment. I don't think we've seen a use case where it's important to be able to make max abort, or sufficiently important to be able to make min warn-only.

Since we want different defaults for min and max, to be able to override that, we'd need another level in the syntax, or complicate things in another way. Can you see this merged until we have a specific use case for configuring it?

I'll document this behaviour in the user docs though.

leonhandreke:
Yeah, fair, it was more me thinking aloud. I guess we could also think about just dropping the assertion if we don't care enough to fail on it, instead of introducing this behavior? 

But also I don't have a strong opinion here. If you think it's useful like this, LGTM :)

leonhandreke (approved):

PR #1915, opened 2025-02-18
[sk_rpvs] Fix some GLEIF codes

leonhandreke:
I noticed this while going through the KYB datasets with many issues. Currently, this is causing 3346 issues in sk_rpvs. I'm unsure why they introduced this different schema name. The schema type on these records is still XI-LEI, and they're valid LEI codes pointing to the right place in the GLEIF database. There is also no other LEI code on them, so this is not a second LEI code associated with the entity. Not sure what the suffix means :(


pudo:
Note to self: we should consider using the original source at some point for this crawler: https://www.justice.gov.sk/sluzby/register-partnerov-verejneho-sektora/open-data/ 

The nomenclature in the data is weird but the LEIs look legit. Thanks for the PR!

PR #1917, opened 2025-02-18
Make log messages for cleaning more unique

leonhandreke:
Prepare for reporting to Sentry.

### DIFF
--- a/zavod/zavod/runtime/cleaning.py
+++ b/zavod/zavod/runtime/cleaning.py
@@ -33,7 +33,7 @@ def clean_identifier(prop: Property, value: str) -> Optional[str]:
         normalized = format_.normalize(value)
     if normalized is None:
         log.warning(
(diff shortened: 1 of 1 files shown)


zavod/zavod/runtime/cleaning.py
```diff
@@ -68,7 +68,7 @@ def value_clean(
         if clean is not None:
             if len(clean) > prop_.max_length:
                 log.warning(
-                    "Property value exceeds type length",
+                    f"Property value for {prop_.name} exceeds type length on {entity.id}",
```
pudo, line 71: I think we gotta put the value here, too :/

pudo (approved):

pudo:
Can you also do the identifier warning on line 36 in that file?

pudo:
Nice!

PR #1924, opened 2025-02-19
[cli] Add dataset context to CLI logs, even after Context teardown

leonhandreke:
It doesn't seem awful, but again, I'm using `global`, so not sure if I maybe should feel awful about this...

### DIFF
--- a/zavod/zavod/cli.py
+++ b/zavod/zavod/cli.py
@@ -1,9 +1,9 @@
+import structlog
 import sys
 import click
 import logging
 from pathlib import Path
 from typing import Optional, List
-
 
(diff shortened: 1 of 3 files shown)


pudo:
But but but ... if we want that outcome, why not just remove `clear_contextvars()` from `context.close`? :)

leonhandreke:
I briefly talked about this with @pudo at the office. My takeaway from the conversation was that he was fine with this change, but had larger more general concerns that `zavod.Context` was growing in complexity. I'll merge this for now, let's talk again about the bigger picture in the future. If my recollection of the conversation was not right, please reach out to me and I'll revert this.

PR #1925, opened 2025-02-19
log.error for assertions that should abort

leonhandreke:
Also, get rid of "At least one assertion failed", it's just another item in Sentry.

Also consider this a test case for [d0e603a0](https://github.com/opensanctions/opensanctions/commit/d0e603a0c7b2525ad4cb8c71424fd25511018fba) :)


jbothma (approved):
nice change

PR #1940, opened 2025-02-20
sg_gov_dir expansion

nvmbrasserie:


datasets/sg/gov_dir/crawler.py
```diff
@@ -159,7 +160,14 @@ def crawl_person(
...
+    if email_elem is not None:
+        email = email_elem.text_content().strip()
+    else:
+        # For spokepersons, the email is in a different format
+        email = official.xpath(
+            ".//div[@class='name']//span[@class='fas fa-envelope']/following-sibling::text()"
+        )[0]
 
     pep_status = is_pep(context, rank)
     position_name = make_position_name(
```
leonhandreke, line 173: I saw that `make_position_name` omits the public_body if agency is given. But "[Public Communications Directorate](https://www.sgdi.gov.sg/spokespersons/mindef/others/public-communications-directorate)" only makes sense in the context of "Ministry of Defence". Is this doing the right thing?

datasets/sg/gov_dir/crawler.py
```diff
@@ -159,7 +160,14 @@ def crawl_person(
 
     rank = official.find(".//div[@class='rank']").text_content().strip()
     full_name = official.find(".//div[@class='name']").text_content().strip()
```
leonhandreke, line 162: I saw on the pages that there are some entries like "Feedback Hotline" or "Duty Media Relations Officer (24-Hour Hotline)". If I understand correctly, we don't yet filter these out. Maybe we could have a deny-list that we feed with a regex on the name? They all seem to have all-caps last names.

nvmbrasserie: Could you provide an example? We have some logic for these cases: 
```
regex:
          - "^pa to"
          - "^personal assistant"
          - "^secretary to"
          - "^please contact"
          - "^administrative professional"
          - "^manager"
```
But I haven't seen any of them in the processed data

datasets/sg/gov_dir/crawler.py
nvmbrasserie: I was concerned about this too at first, but most agencies have quite self-explanatory names (e.g. `Health Sciences Authority` or `Immigration & Checkpoints Authority)`, so I think it's sufficient. The logic in `make_position_name` follows this pattern because in every case (except for spokespersons) we had also a `section_name` or a `hierarchy` available, that served as a link between the two. In case of Public Communications Directorate, there's only one with the name like that. So I think we're still good. I don't want to rebuild the way `make_position_name` works, because there is already quite a lot of reliance on it, and categorisation also depends on it

datasets/sg/gov_dir/crawler.py
leonhandreke: But maybe it's not ingesting these for some other reason? You'll know better.

datasets/sg/gov_dir/crawler.py
leonhandreke: Okay, sounds good.

leonhandreke (approved):

datasets/sg/gov_dir/crawler.py
nvmbrasserie: yeah, currently it's split into allow and deny lists. Everything in-between goes to the manual categorisation, including `feedback hotline`. I'll add it to the deny list, thank you for spotting it!

Issue #1943, opened 2025-02-21
Handle more transient issues using urllib retries


jbothma:
Context: We currently retry failed crawler jobs twice. We have learnt a lot about urllib retries which can cover a lot of transient issues.

one important difference between sentry and issues.json is the latter "clears" an warning error if the job completes after a retry, e.g. a 500 error from the dutch police website.

This error was not in issues because a retry of the job resulted in a successful run.

![Image](https://github.com/user-attachments/assets/6ef5451e-d7e9-491e-a1df-45b62f040456)

This isn't always transient. This job didn't complete and the 502 was also in issues.json:

![Image](https://github.com/user-attachments/assets/d353fd7a-709e-4026-85f0-ac268de3c60c)

---

![Image](https://github.com/user-attachments/assets/413397fd-6d02-4bda-817e-3558e72456a4)

## Problem

- if a job retry succeeds, we still have a sentry log to clear, and we have to check that the dataset was actually built successfully after the sentry entry.
- cron job retries are a sledgehammer. For big sites, we're very reliant on careful caching to make "continuing" a crawl (not requesting every page already requested) painful for us and them.

## Proposed solution

- Make the default exponential backoff factor configurable via environment variable - we want this to enable a "low-and-slow max backoff of like 5 minutes" in production, but in development we want more like a few seconds by default otherwise it's just silently taking for ever in dev
- add 500 and 502 to retry potentially transient server and gateway errors
- add 520 to zyte API requests to retry [potentially transient errors](https://docs.zyte.com/zyte-api/usage/errors.html#ban-responses)

leonhandreke:
We already retry on Zyte 520, see https://github.com/opensanctions/opensanctions/blob/e6131a82a48c531c0270eb8f5f6ec2a01a8c5936/zavod/zavod/shed/zyte_api.py#L46

So I think this is done. @jbothma, can we close this for now and see if that helped?

PR #1944, opened 2025-02-21
Retry on HTTP 500 & 502 and make backoff env-configurable

leonhandreke:
See https://github.com/opensanctions/opensanctions/issues/1943

A good combination for prod might be:
```
HTTP_RETRY_BACKOFF_FACTOR=15
HTTP_RETRY_BACKOFF_MAX=300
HTTP_RETRY_TOTAL=5
```

leading to retries after 0, 30, 60, 120, 240, 500 seconds (if I understand https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry correctly)


zavod/zavod/settings.py
```diff
@@ -91,6 +91,9 @@ def env_str(name: str, default: str) -> str:
 HTTP_TIMEOUT = 1200
 HTTP_USER_AGENT = "Mozilla/5.0 (zavod)"
 HTTP_USER_AGENT = env_str("ZAVOD_HTTP_USER_AGENT", HTTP_USER_AGENT)
+HTTP_RETRY_TOTAL = int(env.get("ZAVOD_HTTP_RETRY_TOTAL", 3))
+HTTP_RETRY_BACKOFF_FACTOR = float(env.get("ZAVOD_HTTP_RETRY_BACKOFF_FACTOR", 1.0))
+HTTP_RETRY_BACKOFF_MAX = int(env.get("ZAVOD_HTTP_RETRY_TOTAL", 120))
```
jbothma, line 96:
```suggestion
HTTP_RETRY_BACKOFF_MAX = int(env.get("ZAVOD_HTTP_RETRY_BACKOFF_MAX", 120))
```

zavod/zavod/meta/http.py
```diff
@@ -1,16 +1,21 @@
...
-        self.total_retries: int = data.get("total_retries", 3)
-        self.backoff_factor: float = data.get("backoff_factor", 1)
-        statuses = ensure_list(
-            data.get("retry_statuses", list(Retry.RETRY_AFTER_STATUS_CODES))
+        self.total_retries: int = data.get("total_retries", settings.HTTP_RETRY_TOTAL)
+        self.backoff_factor: float = data.get(
+            "backoff_factor", settings.HTTP_RETRY_BACKOFF_FACTOR
         )
+        statuses = ensure_list(data.get("retry_statuses", DEFAULT_RETRY_STATUS_CODES))
+        self.backoff_max: int = settings.HTTP_RETRY_BACKOFF_MAX
```
jbothma, line 17: I'm a little nervous of being limited by this. Do you feel there's a risk of accidentally backing off too much?

If we do want it, we should probably document that there's a max (https://zavod.opensanctions.org/metadata/#http-options)

For context though, I think this limit isn't limiting yet (heh) - this is the longest that comes to mind https://github.com/opensanctions/nomenklatura/pull/186/files#diff-b93af8d376050f2a2530fac104d06349b73203c4d6b4a5c5ebd9ee244da8e925R86

zavod/zavod/meta/http.py
```diff
@@ -1,16 +1,21 @@
 from typing import Any, Dict, List
+
 from urllib3.util import Retry
 from banal import ensure_list
 from zavod import settings
 
+DEFAULT_RETRY_STATUS_CODES = [500, 502] + list(Retry.RETRY_AFTER_STATUS_CODES)
```
jbothma, line 7: let's add the additional default codes to the docs

jbothma (commented):
Thanks!

zavod/zavod/settings.py
leonhandreke: Urgh thanks for the catch.

zavod/zavod/meta/http.py
leonhandreke: This is the default, see [the docs](https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry) and [the code](https://github.com/urllib3/urllib3/blob/04662c9ae08c9f63fa254772d7618db65123a35e/src/urllib3/util/retry.py#L213). I now added a comment.

I considered introducing a None to keep this at the default, but we don't really have this pattern anywhere, so I went with this.

jbothma: interesting! Then it's nice that it's configurable while we're here :)

jbothma (approved):
Nice, thank you!

PR #1949, opened 2025-02-21
More Sentry-compatible logging when crawl experiences HTTP errors

leonhandreke:
Using log.exception transmits the exception to Sentry. Only put the type
name (instead of the full object repr, which includes the memory
address) in the log message to get a nicer message in the issues
overview


jbothma (approved):
image

PR #1966, opened 2025-02-25
Fixes-25-02

nvmbrasserie:


datasets/us/de/med_exclusions/crawler.py
```diff
@@ -48,7 +48,7 @@ def crawl_item(row: Dict[str, str], context: Context):
     entity.add("npiCode", h.multi_split(npi, [";", ",", "&"]))
 
     # Deal with sometimes intentional line breaks, and sometimes unwanted wrapping:
-    license_numbers = row.pop("license")
+    license_numbers = row.pop("license", row.pop("license_rn707737_pa", None))
```
jbothma, line 51: 😭

jbothma: maybe punt them an email asking that they try and make sure the whole value is always visible when they export the PDF, to make us feel a little more ok about this fix. I can't think of a better one.

image

jbothma (approved):

datasets/us/de/med_exclusions/crawler.py
nvmbrasserie: sure, will do

PR #1968, opened 2025-02-25
Final-fixes-25-02

nvmbrasserie:


datasets/lt/magnitsky_amendments/crawler.py
```diff
@@ -11,56 +11,54 @@ def map_gender(gender_code):
...
-
-def crawl(context: Context):
-    path = context.fetch_resource("source.json", context.data_url)
-    context.export_resource(
-        path, mime_type="application/json", title=context.SOURCE_TITLE
-    )
+def crawl_page(context: Context, entry):
+    # Create a Person entity
+    person = context.make("Person")
+    person_id = context.make_id(entry.get("vardas"), entry.get("pavarde"))
```
leonhandreke, line 17: Does it make sense to include date of birth in the ID? I've seen that we do that in other places, in addition to the name, but you'll have a better overview.

And small nit, I think the comment is a bit superflous, but a comment or named variable to explain "vardas" and "pavarde" wouldn't hurt :)

leonhandreke (approved):

PR #1972, opened 2025-02-26
ky judicial updated website

nvmbrasserie:


datasets/ky/judicial/crawler.py
```diff
@@ -73,10 +76,17 @@ def crawl_judges(context: Context, url, default_position, min, max):
...
+    doc = context.fetch_html(context.data_url, cache_days=1)
+    profile_links = [
+        link
+        for link in doc.xpath('//ul[@id="menu-judicial-officers"]//a/@href')
+        if link != "#"
+    ]
+    assert len(profile_links) >= 6, profile_links
+    for url in profile_links:
+        doc = context.fetch_html(url, cache_days=1)
+        profile_links = doc.xpath(
```
leonhandreke, line 88: (nit) let's use a different variable name, since you're currently iterating over it. It works, since you're not actually changing profile_links (then it would explode I think) but just assigning a new value to it, but it's an extra thing to think about. Maybe `person_urls` is better?

leonhandreke (approved):

PR #1978, opened 2025-02-28
Organizations designated as terrorist by Bahrain

nvmbrasserie:
fixes opensanctions/crawler-planning#159


nvmbrasserie:
OCR provided partial success, but there's no reliable way to verify the accuracy of the extracted text.

I have emailed the MFA of Bahrein requesting a republished version of the document with selectable text. We'll proceed as soon as we hear back.

PR #1979, opened 2025-02-28
[fr_tresor_gels_avoir] Implement new splitting logic

leonhandreke:
We use the previous lookups where available, but now we also split into segments and try to lookup and then parse those.

### DIFF
--- a/datasets/fr/tresor/crawler.py
+++ b/datasets/fr/tresor/crawler.py
@@ -29,8 +29,8 @@ SEPARATORS = "/:;-. "
 TEXT_KEYS = {
     "Type d'entité": "legalForm",
     "Date d'enregistrement": "incorporationDate",
-     # address, which is a free text field we don't auto-parse because it's too easy for the value
-     # to silently include a key we don't know about yet.
+    # address, which is a free text field we don't auto-parse because it's too easy for the value
+    # to silently include a key we don't know about yet.
     "Lieu d'enregistrement": None,
     "Principal établissement": "jurisdiction",
     "Principaux établissements": "jurisdiction",
--- a/datasets/fr/tresor/crawler.py
+++ b/datasets/fr/tresor/crawler.py
@@ -29,7 +29,9 @@ SEPARATORS = "/:;-. "
 TEXT_KEYS = {
     "Type d'entité": "legalForm",
     "Date d'enregistrement": "incorporationDate",
-    "Lieu d'enregistrement": None,  # address, which is a free text field we don't auto-parse
+     # address, which is a free text field we don't auto-parse because it's too easy for the value
+     # to silently include a key we don't know about yet.
+    "Lieu d'enregistrement": None,
     "Principal établissement": "jurisdiction",
     "Principaux établissements": "jurisdiction",
     "Établissement principal": "jurisdiction",
--- a/datasets/fr/tresor/crawler.py
+++ b/datasets/fr/tresor/crawler.py
@@ -23,7 +23,8 @@ SCHEMATA = {
     "Navire": "Vessel",
 }
 
-REGEX_IDENTIFIER = re.compile(r"^[\w\.]+$")
+# Only word-character and dots
+IDENTIFIER_SINGLE_VALUE_REGEX = re.compile(r"^[\w\.]+$")
 SEPARATORS = "/:;-. "
 # Order matters, earlier entries will be matched first
 # We don't match free-text fields (currently address) because they could swallow up extra data of unmapped keys.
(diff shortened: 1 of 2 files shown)


datasets/fr/tresor/crawler.py
```diff
@@ -101,6 +126,70 @@ def parse_identification(
...
+    if followthemoney.model.get("LegalEntity") in entity.schema.schemata:
+        registration_number = entity.first("registrationNumber")
+        if registration_number and rigour.ids.OGRN.is_valid(registration_number):
+            entity.add("ogrnCode", registration_number)
+            entity.remove("registrationNumber", registration_number)
+
+        tax_number = entity.first("taxNumber")
+        if tax_number and rigour.ids.INN.is_valid(tax_number):
+            entity.add("innCode", tax_number)
+            entity.remove("taxNumber", tax_number)
```
jbothma, line 191: Does this maybe fit better with the `if propname == "kppCode":` sort of stuff or does that make the code more awkward in a way I don't quickly see? So that the special cases are together and perhaps more importantly, we don't do the awkward add + remove + add

datasets/fr/tresor/crawler.py
```diff
@@ -47,36 +71,37 @@ def clean_key(key: str) -> Optional[str]:
...
 
+    # We try to find any of the keys in TEXT_KEYS and split on them. The basic idea is to not split on separators,
+    # but just find a key and keep reading until we see the next key.
+    # NOTE: ideally, we would never split within a previously matched key. For example, "Numero OGRN: 1234" will
+    # be split to ["Numero", "OGRN: 1234"], because the key "OGRN" is a suffix of "Numero OGRN". Doesn't happen
+    # that often, so we just solve it with a lookup instead of making the code here more complex.
+    # [0] is important so that if no key is found at the start of the string, it is still passed through
+    indices = [0] + sorted([full.lower().find(key.lower()) for key in TEXT_KEYS.keys()])
+    indices = [i for i in indices if i != -1]
+    segments = [full[i:j] for i, j in zip(indices, indices[1:] + [None])]
```
jbothma, line 83: something this doesn't cover is when we don't already have a lookup, and we don't already know a key that comes after a known key. And from time to time, new variations of keys are introduced.

e.g. let's imagine we don't have the lookups for the registration number `10320661315`, and we don't know about the key `"Numéro d'enregistrement`, then we end up silently with the value `adddress": "Iran: Numéro d'enregistrement: 10320661315 ou 414950 ou 47779"`

not a perfect example - there are other items with that key warning about it, but it's entirely possible and means we end up with crappy data which either doesn't match as well as it should, and doesn't look very impressive when we aim for quality data.

I wonder if you have any ideas for somehow guaranteeing we parse or warn for all keys? My ideas are:

- check for separators in values, e.g. `:` (present in this example, but not present in all)
- check for tokens of keys, e.g. numero, inn, etc in values (would probabably catch most, but not guaranteed)

I'm not sure where to draw the line, but I think we need to have more safety here, and it'd be nice to have no risk at all.

datasets/fr/tresor/crawler.py
leonhandreke: The advantage here is that we also catch those registrationNumbers that we mapped manually --- do we want that even?

jbothma:
aha - fair point. This is totally personal opinion so your call, but my
feeling is the remove/add is too icky compared to the value add of a more
specific prop.

datasets/fr/tresor/crawler.py
leonhandreke: I went with what you suggested.

datasets/fr/tresor/crawler.py
leonhandreke: Done, addresses are no longer auto-parsed. I checked `registrationNumber`. It does not allow spaces, so that should be fine, right?

jbothma: Sorry I never replied to the question.

I think without the stricter validation on values, it's still making me worry a bit.

E.g. if I remove some of the current manual extraction (representing new values we could conceivably get), I get the following:

```
Key: Numéros d'enregistrement
Value: 'Immatriculation (numéro OGRN): 1247700291200'
```

```
Key: Numéro d'enregistrement
Value: 'ИНН 7717653542 КПП 771701001'
===
Key: Date d'enregistrement
Value: '22.7.1992 (Ukraine), 13.8.2014 (Fédération de Russie)'
```

```
===
Key: NIF:
Value: '503210805145): 3001041615 ('
===
Key: Numéro d'identification fiscale
Value: ')'
===
Key: Numéro d'enregistrement
Value: 'SRN/OGRN 1021602510533'
```

maybe something like "reject if it's not a single \w+ or a valid country" would get us a long way?

Am I right in thinking the `Failed to parse identification segment` bit can only be reached if the new key is the first segment? because without the validation we're discussing, new keys can still be eaten as values after the first segment.

datasets/fr/tresor/crawler.py
```diff
@@ -101,6 +131,71 @@ def parse_identification(
...
+                    h.apply_date(entity, propname, value)
+
+                # Special handling for Russian tax/registration numbers, which is most companies on the list
+                elif propname == "registrationNumber":
+                    if rigour.ids.OGRN.is_valid(value):
+                        entity.add("ogrnCode", value)
+                    else:
+                        entity.add("registrationNumber", value)
+                elif propname == "taxNumber":
+                    if rigour.ids.INN.is_valid(value):
```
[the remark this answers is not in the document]
leonhandreke: Yeah, you're right. Maybe this should be a generic mechanism as you described above? Anyway, removed it for now.

Issue #1819, opened 2025-01-30
We've deployed wikidata YYYY-01-01 cleaner but we still have such dates. It's running to completion - double check where the dates are coming from.


jbothma:

pudo:
I'm pretty sure this was because of `wd_peps`, which was using its own logic to parse DOBs.

PR #1987, opened 2025-03-04
Shorten DOB if it ends in -01-01

pudo:
Fixes: https://github.com/opensanctions/opensanctions/issues/1819

### DIFF
--- a/datasets/_wikidata/peps/crawler.py
+++ b/datasets/_wikidata/peps/crawler.py
@@ -64,7 +64,9 @@ def crawl_holder(
         # Avoid constructing a proxy which emits a warning
         # before we discard it.
         return
-    entity.add("birthDate", birth_date)
+    if birth_date is not None and birth_date[:10].endswith("-01-01"):
+        birth_date = birth_date[:4]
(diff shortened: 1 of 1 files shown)


datasets/_wikidata/peps/crawler.py
```diff
@@ -64,7 +64,9 @@ def crawl_holder(
         # Avoid constructing a proxy which emits a warning
         # before we discard it.
         return
-    entity.add("birthDate", birth_date)
+    if birth_date is not None and birth_date[:10].endswith("-01-01"):
+        birth_date = birth_date[:4]
+    entity.add("birthDate", birth_date, original_value=holder.get("person_birth"))
```
jbothma, line 69: so, there is `?timenode wikibase:timePrecision     ?timeprecision.` but we're querying 7 dates in the query where this comes from, so for what we'd gain I'd say this is a fair simplification

pudo: The other problem is humans: a ton of people in Wikidata actually have `-01-01` entered as their date, although it is wrong... so the precision thing is also just used wrong.

PR #1992, opened 2025-03-06
[shed/zyte] Unify text/json/html fetchers and don't cache broken JSON

leonhandreke:
This is the long way to fix the il_knesset_members crawler, where we've cached broken JSON. Reduces some duplicate code in `fetch_text` and `fetch_html`. The outside interface should be exactly the same.


jbothma (approved):
such an improvement ❤️

PR #2008, opened 2025-03-10
Japan METI Russian List

nvmbrasserie:


pudo:
Some comments:

* `name_raw` seems to contain a numeric prefix that should be split out 
* `name_en` needs to be normalized so it's names, perhaps split with a semicolon
* addresses seem to have a common prefix that just means `Location: `. Don't need that.

pudo (commented):
We can merge this, but to me it represents a little bit of an anti-pattern: we have manually processed the PDF data into a Google Doc, where we can now edit it. Instead of just doing that, there's 400 lines of data patches here, patching data that we have write access to. I think just adding `_clean` columns with the corrected values could have done the same thing and be much more readable and maintainable...

nvmbrasserie:
I decided to do the cleanup in the lookups rather than in the Google Sheet, so I wouldn't accidentally miss anything when formatting. It felt more reliable that way.  Next time I'll use the Google Sheet formatting for similar cases

PR #2015, opened 2025-03-12
Fix and test log redaction of env vars in issues.json (and Sentry)

leonhandreke:
The issue was that in the processor chain, the two emitting processors `log_isue` and `sentry_processor` were inserted before the redactor. The consequence was that the redaction would work for the command line, but unredacted log messages would be sent to `issues.json` and Sentry.

As far as I can tell, this bug has been present since the initial introduction of this feature in
https://github.com/opensanctions/opensanctions/commit/8bf3ed700459d0abf867f27366c3ee17df3d48fd on 2024-05-10.

This fixes https://github.com/opensanctions/operations/issues/1086.


pudo (approved):
Thanks, that seems to fix the regression

PR #2024, opened 2025-03-13
gem energy tracker update

nvmbrasserie:


datasets/_global/gem_energy_ownership/crawler.py
```diff
@@ -88,27 +73,37 @@ def crawl_company(context: Context, row: Dict[str, str], skipped: Set[str]):
...
+    entity.add("alias", row.pop("full_name"))
     entity.add("name", original_name)
     entity.add("alias", row.pop("name_other"))
     entity.add("weakAlias", row.pop("abbreviation"))
-    if lei_code != "unknown":
+    if lei_code != "not found":
         entity.add("leiCode", lei_code)
     if entity_type != "unknown entity":
         entity.add("description", entity_type)
+    entity.add("classification", row.pop("legal_entity_type"))
```
jbothma, line 90: I think we normally put this into legalForm

jbothma (commented):
nice work! how are you checking the changes?

This diff was helpful:

```bash
colordiff <(csvcut --no-header-row -c1-4 statements.pack) <(csvcut --no-header-row -c1-4 data/datasets/gem_energy_ownership/statements.pack | sort)|less -R
```

this is a good candidate for assertions on number of entities by country - are you checking that at least the rough proportions by country are consistent with the previous version?

datasets/_global/gem_energy_ownership/crawler.py
```diff
@@ -96,16 +90,19 @@ def crawl_company(context: Context, row: Dict[str, str], skipped: Set[str]):
...
     if schema != "Person":
-        entity.add("permId", perm_id)
-        # find a way to remap invalid ones
-        entity.add("ogrnCode", ru_id)
+        entity.add_cast("Company", "permId", perm_id)
+        # Check if ru_id is a valid OGRN
+        if ru_id and rigour.ids.OGRN.is_valid(ru_id):
+            entity.add("ogrnCode", ru_id)
+        else:  # Remap invalid ones
+            entity.add("registrationNumber", ru_id)
```
jbothma: sorry for changing my mind at the last minute. This was about 20 items, right? If it's more like 100, maybe this is good, but for 20ish I think we should use the lookups mechanism instead of adding another special case to the crawler. We can kinda do it in bulk, right? like

```yaml
    - match:
        - id1
        - id2
        - ...
      prop: registrationNumber
```

datasets/_global/gem_energy_ownership/crawler.py
```diff
@@ -96,16 +90,19 @@ def crawl_company(context: Context, row: Dict[str, str], skipped: Set[str]):
...
         entity.add("registrationNumber", br_id)
         entity.add("registrationNumber", uk_id)
         entity.add("registrationNumber", in_id)
         entity.add("registrationNumber", us_eia_id)
         entity.add("registrationNumber", sp_cap)
         if schema != "PublicBody":
-            entity.add("cikCode", us_sec_id)
+            entity.add_cast("Company", "cikCode", us_sec_id)
         else:  # PublicBody
             entity.add("registrationNumber", us_sec_id)
```
jbothma: not required, but as a tiny tidy, if the first clause is `if entity.schema.is_a("PublicBody")`, it's more obvious that the second clause is not a public body, so the comment isn't needed. No double negative in your head :)
the is_a is also more future-proof

datasets/_global/gem_energy_ownership/crawler.py
```diff
@@ -14,10 +18,7 @@
     "gem_parents",
     "gem_parents_ids",
 ]
-
-# Unique entity types
-# {"person", "unknown entity", "state", "legal entity", "arrangement", "state body"}
-
+ALIAS_SPLITS = ["[former],", "[former]", "[former name]", "(former)"]
```
jbothma, line 21: since all of these are former, it might be nice to put them into `previousName` instead.

datasets/_global/gem_energy_ownership/crawler.py
```diff
@@ -74,10 +77,32 @@ def crawl_company(context: Context, row: Dict[str, str], skipped: Set[str]):
     entity = context.make(schema)
     entity.id = context.make_slug(id)
 
+    names_to_lookup = [full_name, original_name]
+    associates = set()
+    for name in names_to_lookup:
+        result = context.lookup("associates", name)
+        if result and result.associates:
+            associates.update(result.associates)
```
jbothma, line 85: is there anything in the data we can use to determine when we should expect an associates lookup to exist? e.g. `"(" in name and "," in name` or something, and then warn? Not sure how many false positives we'll get, and I don't think it's worthwhile for one or two in 1000. But try it - maybe we spot more instances where we need to map out related entities as opposed to aliases.

datasets/_global/gem_energy_ownership/gem_energy_ownership.yml
```diff
@@ -49,17 +49,31 @@ assertions:
...
-          - 'Tiemenguan City Finance Bureau'
-          - 'Xinjiang Production and Construction Corps Second Division State-owned Assets Supervision and Administration Commission'
-          - 'Tiemenguan City State-owned Assets Supervision and Administration Commission'
-          - 'Xinjiang Production and Construction Corps Second Division Financial Work Office'
-          - 'Tiemenguan City Financial Work Office'
-      - match: '中电国际新能源（上海）控股有限公司 [former], China Power International New Energy (Shanghai) Holdings [former], China Electric Power International New Energy Holdings, China Electric Power International New Energy Holdings Limited, China Electric Power International New Energy Holdings LTD, China Power International New Energy CO LTD, China Power International New Energy Holding Ltd., China Power International New Energy Holding LTD, China Power International New Energy Holdings CO LTD, CLP International New Energy Holdings CO LTD'
+      - match: 新疆生产建设兵团第二师财政局（铁门关市财政局、新疆生产建设兵团第二师国有资产监督管理委员会、铁门关市国有资产监督管理委员会、新疆生产建设兵团第二师金融工作办公室、铁门关市金融工作办公室）
+        values:
+          - 新疆生产建设兵团第二师财政局
```
jbothma, line 75: I think it's a bit awkward if the mechanism is that you manually override a name with associates to just the original. Can the code handling the associate lookup overwrite the name used for the main entity?

jbothma (commented):
Thanks - good progress. I know it's getting tricky so shout if you want to chat through anything.

datasets/_global/gem_energy_ownership/gem_energy_ownership.yml
```diff
@@ -44,21 +44,116 @@ assertions:
   min:
     schema_entities:
       Company: 12000
+      LegalEntity: 20
+      PublicBody: 900
+      Person: 1
+    country_entities:
+      cn: 6400
```
nvmbrasserie, line 51: I added assertions for countries. They expanded the coverage quite a bit, but it only grew bigger for each country in comparison with the previous version

datasets/_global/gem_energy_ownership/crawler.py
jbothma: my bad - I didn't realise this was originally connected to alias and the field name was name_other which suggests that most of the time, it'll actually be aliases but sometimes there are former name splits.

This kinda highlights an issue with declaring and extracting lots of values at the top, and using them later. I've changed the ID extracts to be mostly in line with where we use them. having the original key and our prop in close proximity makes it easier to see how we're mapping it without scrolling up and down a page worth of code.

datasets/_global/gem_energy_ownership/crawler.py
```diff
@@ -5,110 +5,154 @@
...
+        if name and re.search(NAME_PATTERN, name):
+            full_name, original_name, associates = get_associates(
+                context, name, associates, full_name, original_name
+            )
+
+    if associates:
+        for associate in associates:
+            other = context.make("LegalEntity")
+            other.id = context.make_slug("named", associate)
+            other.add("name", associate)
```
jbothma, line 101: could be nice adding country here too

datasets/_global/gem_energy_ownership/crawler.py
```diff
@@ -5,110 +5,154 @@
...
+    for name in [full_name, original_name]:
+        if name and re.search(NAME_PATTERN, name):
+            full_name, original_name, associates = get_associates(
+                context, name, associates, full_name, original_name
+            )
+
+    if associates:
+        for associate in associates:
+            other = context.make("LegalEntity")
+            other.id = context.make_slug("named", associate)
```
jbothma, line 100: let's use make_id instead of make_slug for names, personal identification numbers, etc - things considered possibly personal

datasets/_global/gem_energy_ownership/crawler.py
```diff
@@ -5,110 +5,154 @@
...
 
     entity = context.make(schema)
     entity.id = context.make_slug(id)
 
+    associates: Set[str] = set()
+    for name in [full_name, original_name]:
+        if name and re.search(NAME_PATTERN, name):
+            full_name, original_name, associates = get_associates(
+                context, name, associates, full_name, original_name
+            )
```
jbothma: a couple of things make this difficult to reason about:

1. we usually name things original_ when we want to pass it in as the original uncleaned string to the original_value argument of Entity.add(). Perhaps we can name this local_name as they do in the data - then its meaning is clearer
2. it's a tad confusing that we're both looping over full_name and original_name, and passing both in. What we're trying to accomplish here is. Perhaps if we don't try and choose a name, but rather just aggregate what we find, it gets simpler? something like

```suggestion
    names = Set[str]
    for name in [full_name, original_name]:
        if name and re.search(NAME_PATTERN, name):
            name_, associates_ = get_associates(
                context, name
            )
            names.add(name_)
            associates.update(associates_)
        else:
            names.add(name)
```

I think then get_associates also becomes simpler. Just something like

```
def get_associates(context, name):
    res = lookup(..., name)
    if res is None...
        return name, []
    else:
        return res.entity_name, res.associates
```

would we lose out on any important behaviour if we do this?

jbothma (commented):
Thanks for this! Shout if the name stuff is unclear.

It could also be nice to actually put the actual original name in the original_value arg for name. maybe that requires inlining the name get_associates function and it's not critical - skip if it's daunting.

PR #2038, opened 2025-03-18
Deep-ish copy before log redaction to avoid modifying logged data

leonhandreke:
This is affecting data quality, so here is a quickfix before we probably move log redaction into the issues writer.

### DIFF
--- a/zavod/zavod/logs.py
+++ b/zavod/zavod/logs.py
@@ -53,7 +53,9 @@ class RedactingProcessor:
         self.repl_regexes = {re.compile(p): r for p, r in repl_pattrns.items()}
 
     def __call__(self, logger: Any, method_name: str, event_dict: Event) -> Event:
-        return self.redact_dict(event_dict)
+        # stringify performs a deep-ish copy to avoid redacting the original data
+        data: Dict[str, Any] = stringify(dict(event_dict))
+        return self.redact_dict(data)
 
     def redact_dict(self, dict_: Event) -> Event:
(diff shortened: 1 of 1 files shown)


zavod/zavod/logs.py
```diff
@@ -53,7 +53,9 @@ def __init__(self, repl_pattrns: Dict[str, str | Callable[[str], str]]) -> None:
         self.repl_regexes = {re.compile(p): r for p, r in repl_pattrns.items()}
 
     def __call__(self, logger: Any, method_name: str, event_dict: Event) -> Event:
-        return self.redact_dict(event_dict)
+        # stringify performs a deep-ish copy to avoid redacting the original data
+        data: Dict[str, Any] = stringify(dict(event_dict))
+        return self.redact_dict(data)
```
jbothma, line 58: I don't think stringify copies, and it doesn't guarantee stringification. So it's not really consistent with the comment, nor contributing in a practical way to redaction. It's only special cases we've seen crash in json serialization. Maybe it should have been better called "prepare_for_json" or something.

Perhaps best to just use the thing designed for the thing here.

```suggestion
        data: Dict[str, Any] = copy.deepcopy(event_dict)
        return self.redact_dict(data)
```

leonhandreke: So, it does create copies, but only of lists, which seems to be just enough to prevent the bug from occuring. Anyway, let's move this to the issues writer instead as discussed.

PR #2040, opened 2025-03-19
Pin third-party actions to a SHA as recommended by GitHub

leonhandreke:
- **Pin tj-actions/changed-files action to specific SHA**
- **Pin peter-evans/repository-dispatch action to specific SHA**

See https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsuses

> Using the commit SHA of a released action version is the safest for stability and security.

Also https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -137,7 +137,8 @@ jobs:
     needs: [docker, python]
     steps:
       - name: Dispatch add-on build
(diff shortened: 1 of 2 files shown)


jbothma (approved):
thanks!

PR #2051, opened 2025-03-21
Fixes-21-03

nvmbrasserie:


datasets/ir/sanctions/crawler.py
```diff
@@ -64,7 +64,10 @@ def crawl(context: Context):
         qs = urlencode(params)
         url = f"{context.data_url}?{qs}"
         pagenums_xpath = '//*[contains(@class, "pageinationnum")]'
-        doc = fetch_html(context, url, pagenums_xpath, cache_days=1, retries=6)
+        # Geolocation bypasses Cloudflare checks
```
jbothma, line 67: 🙌

datasets/ee/international_sanctions/ee_international_sanctions.yaml
```diff
@@ -42,3 +42,5 @@ lookups:
     options:
       - match: ka ШАБУНЯ Виктория Валерьевна
         value: ШАБУНЯ Виктория Валерьевна
+      - match: also БОДАК Ала Мiкалаеўна
+        value: БОДАК Ала Мiкалаеўна
```
jbothma, line 46: I would suggest simplifying the regex to just match the prefix, and then split the name with a list of splits like `" (also", "(", ")"`

jbothma (approved):
all looks fine, but I think the estona name matching can be simplified

datasets/ee/international_sanctions/ee_international_sanctions.yaml
nvmbrasserie: it's only one case in the whole list that contains `also`, and it works very well for splitting everything else. I'd rather keep it as it is, but can also take a look and adjust

PR #2057, opened 2025-03-24
Bulgaria-asset-declarations-fixes

nvmbrasserie:


datasets/bg/jud_declarations/crawler.py
```diff
@@ -89,8 +123,25 @@ def crawl_row(context: Context, row: Dict[str, HtmlElement]):
...
+        normalize(name_dec),
+        max_length=MAX_NAME_LENGTH,
+        max_edits=12,
+        max_percent=0.4,
+    )
+    # Full last names should be considered the same if they have at least 70% similarity.
+    # Example: "Дебора Миленова Вълкова" and "Дебора Миленова Вълкова-Терзиева"
+    # should be treated as a match despite the additional surname.
+    if similarity < 0.7:
+        if (name, name_dec) in ALLOW_LIST:
```
leonhandreke, line 138: Super cool, I'm glad this turns out working quite well!

I guess the way we often do it is to make this a lookup to keep "data" out of the code, is there a reason it's not in this case? Anyway, this is cool too I think.

leonhandreke (approved):

datasets/bg/jud_declarations/crawler.py
nvmbrasserie: yeah, I agree. But I think keeping it in the set pairs is much more straightforward approach in this case

leonhandreke: Sounds good!

PR #2061, opened 2025-03-24
Temporarily disable pip isolation and pin setuptools

jbothma:
This should get us building again, but let's see if we can pin (haha) down the culprit and fix before merging this)


jbothma:
removing stringcase seems to have done the job

PR #2063, opened 2025-03-25
Fix up some more entity types

jbothma:


pudo:
outch, that looks like a mess :/

PR #2093, opened 2025-03-31
Drop pdfminer warning that doesn't seem to be an issue for our use case.

jbothma:
I can't see why this is such a big problem - the output looks fine

### DIFF
--- a/zavod/zavod/helpers/pdf.py
+++ b/zavod/zavod/helpers/pdf.py
@@ -1,12 +1,24 @@
-from normality import collapse_spaces, slugify
+import logging
+import subprocess
 from pathlib import Path
-from pdfplumber.page import Page
 from tempfile import mkdtemp
 from typing import Any, Callable, Dict, Generator, List, Optional, Tuple
+
 import pdfplumber
-import subprocess
+from normality import collapse_spaces, slugify
+from pdfplumber.page import Page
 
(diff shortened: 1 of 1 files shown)


pudo:
I'm so sorry to do this, but Claude: 

This warning message "CropBox missing from /Page, defaulting to MediaBox" is coming from pdfminer and isn't typically a severe issue.

In PDF documents, there are different "boxes" that define page boundaries:

- **MediaBox**: Defines the physical size of the page (required in all PDF files)
- **CropBox**: Defines the visible region of the page (optional)

When the CropBox is missing, pdfminer defaults to using the MediaBox dimensions instead. This is expected behavior and the library is simply informing you of this fallback.

This warning generally means:

1. The PDF was created without explicitly defining a CropBox
2. The parser will use the MediaBox dimensions to determine the page boundaries

This shouldn't affect functionality in most cases, as the MediaBox is a reasonable fallback. You'll still be able to extract text and other content from the PDF normally.

Only consider this a potential issue if you're doing precise layout analysis where the exact page boundaries matter.

jbothma:
Lol, good Claude. That's what I got from google too - I was more asking for your eyes in case excluding logs in this way or in this place is a bad idea. But it looks like things are setting up filters where they use the vocal thing, so.

PR #2094, opened 2025-03-31
Build out custom maritime export

pudo:


zavod/zavod/exporters/maritime.py
```diff
@@ -0,0 +1,104 @@
...
+        for name in entity.get_type_values(registry.name, matchable=True):
+            if name == entity.caption:
+                continue
+            name_ = collapse_spaces(name)
+            if name_ is not None:
+                names.add(name_)
+        return names
+
+    def feed(self, entity: Entity) -> None:
+        if "imoNumber" not in entity.schema.properties:
```
jbothma, line 66: this plus row type is fun

jbothma:
I'm not sure of the use case specifically so maybe the limited fields makes sense.

Previous names might be good. One of the deltas I see most commonly on the war_sanctions data is a vessel name changing.

There seems to be a range of companies linked to vessels - commercial manager, safety management, owner, P&I Club, "c/o" company I believe means care-of?

Could it be interesting to nest them in there? Maybe it's not so useful when the data is oriented by vessel. It's probably quite interesting when looking at the number of risky vessels associated with these companies though, which might be a different report.

Issue #2072, opened 2025-03-26
[de_abgeordnetenwatch] Runner failed: list indices must be integers or slices, not str


jbothma:
I've emailed them:

> We noticed a big drop in entities, and on investigation, it seems the API is not returning all the data available on the website (or which used to be in the API).
>
>We iterate over candidacies-mandates and identify positions and mandates relevant to our needs that way.  A couple of examples:
>
> https://www.abgeordnetenwatch.de/profile/carsten-schneider - the website shows Abgeordneter Bundestag entries, but https://www.abgeordnetenwatch.de/api/v2/candidacies-mandates?politician=79014 has zero results.
> There's another entry https://www.abgeordnetenwatch.de/profile/carsten-schneider-0 and https://www.abgeordnetenwatch.de/api/v2/candidacies-mandates?politician=182957 does have an entry for him.
>
> Is this just a transition period with new data being loaded? Is there anything we should do keep the old results and also get the new data?

pudo:
Maybe https://www.abgeordnetenwatch.de/api/v2/politicians/79014 ?

jbothma:
What's weird is that his mandates aren't showing up as they used to. So we could change the implementation to find the politicians via /politicians, but we still need the mandates.

jbothma:
The issue above still occurs, it's just that we're not hitting it because we're hitting this error at the moment.

jbothma:
We can construct URLs instead of the detail URLs using the IDs and their query syntax, e.g. `https://www.abgeordnetenwatch.de/api/v2/parliament-periods?id[eq]=133` for "parliamentary period 133"

pudo:
I'm closing this because it ran successfully yesterday and this issue seems orphaned.

PR #2095, opened 2025-03-31
Fixes-31-march

nvmbrasserie:


datasets/de/abgeordnetenwatch/crawler.py
```diff
@@ -32,9 +33,11 @@ def crawl_mandate(context: Context, mandate: Dict[str, Any]):
     politician = mandate.pop("politician")
     period = mandate.pop("parliament_period")
     period_detail = context.fetch_json(period["api_url"], cache_days=1).pop("data")
-    position = make_position(context, period_detail)
-    if not position:
-        return
```
nvmbrasserie, line 37: I assume that `period_detail` used to be a single dictionary, and that made `make_position` happy. However, now `period_detail` returns a list of dictionaries, which caused it to fail

jbothma: https://github.com/opensanctions/opensanctions/issues/2072#issuecomment-2767022037

PR #2096, opened 2025-03-31
ua-war-sanctions-expansion

nvmbrasserie:


datasets/ua/war_sanctions/ua_war_sanctions.yaml
```diff
@@ -175,13 +178,27 @@ lookups:
...
-          - 1959-08-21
-          - 2024-01-08
+        value: 1959-08-21
       - match: 08.011.2024)
         value: 2024-01-08
       - match: 07/03/2023)
         value: 2023-03-07
+      - match: 29.06.1961- 17.08.2018
+        values: 1961-06-29
+      - match: 06.05.1960 - 21.11.2021
```
jbothma, line 188: are there cases where `date - date` for `Date and place of birth` isn't a situation where we want to split on `-` and take the first? Any idea what they're doing with the multiple dates here?

nvmbrasserie: Let me check. I think it's date of birth and date of death. Is there a way we can automatically remap it to different props using one lookup?

nvmbrasserie: it's safe to split them. There are 4 scenarios we want to cover: 

- list with 2 values, none of which is a date range `['22.05.1957', 'Beregovoye, Crimea, Ukrainian SSR, USSR']`
- list with 2 values, one of which is a date range `['01.09.1960 - 30.03.2015 - 30.03.2015', 'Poltava, Poltava region, Ukrainian SSR, USSR']`
- single date `24.12.1945`
- single date range `15.07.1985 -  2016`

Current update covers them all

jbothma:
Let's maybe have a call and look together how to implement this. Is it safe to assume the second date is a date of decease? It feels surprising that they'd list someone on the site who died in 2016, but I'm not sure I can confirm one way or the other easily

jbothma (approved):
Thanks for this! 

It'd be nice if we could document the content of the section list we're monitoring so it's easier to identify whether changes are relevant. 

With or without that, ship it!

PR #2109, opened 2025-04-04
[us_ms_med_exclusions] more cases for dates handling

leonhandreke:
Pulled out from https://github.com/opensanctions/opensanctions/pull/2106 cause I'm not quite sure what's happening there. Can't find the entry in question in the spreadsheet (anymore?) and I don't get what the entry looked like.


leonhandreke:
I made the following changes:
- Named the lookup `exclusion_period` otherwise it's a bit confusing that we have only the start date under a lookup that's called `end_date`
- Made the lookup trigger in all cases, so in the future if they add something that doesn't have any dashes, we can use the same lookup
- Use `h.is_active` to set the topic (I guess this was broken before?)

nvmbrasserie (approved):

PR #2114, opened 2025-04-04
Allow uploading to Google Cloud Storage with gzip compression

leonhandreke:
Gzipped FtM files are roughly 75% smaller. I checked yente and httpx already sets `Accept-Encoding: gzip`, so we could be getting fairly quick results on both storage and networking costs. Let's save some cash! 

The whole thing is hidden behind a feature flag.

I tested this on a test bucket. Uploading seems to work fine, and getting the object back from the bucket using the archive backend works fine in a REPL. So I'm fairly confident this won't blow up.


leonhandreke:
https://github.com/opensanctions/operations/pull/1171 has a little change that flips this only for ru_egrul. Seems like a good place to start that's not too mission-critical.

jbothma:
nice research!
Sheesh, that's a bit scary - now you mention it I remember reading about checksums not being useful when compressing.
As you found, I think by default writing to GCS is atomic, as in either the write is complete, or the write is invisible, but never partial. I imagine S3 is the same. If we're relying on this, we should possibly make it explicit that we rely on this - IIRC it's an option or policy somewhere so we want to be sure we don't change from that without realising it affects this.

I wanna come back to this but those are my first thoughts.

jbothma:
I think this is the sort of thing I had in mind, but I'm not sure if all our uploads would automatically be multipart. If this is it, it'd be worth adding a comment to say that we rely on this for integrity because we're copying with --gcs-decompress 
https://github.com/opensanctions/operations/blob/main/tf/backup/bucket.tf#L39

jbothma:
An idea - on the one hand I worry this is overkill, on the other hand this doesn't check integrity either, so doesn't really get t the crux of what we give up with --decompress - but here it is anyway:

What do you think of an end-to-end/smoke test of the backed up data - something that runs once a day, and does some loosey goosey checks like

- we can read a recent statements.csv - it has data and it's not surprisingly compressed
- it's the sort of size we expect

leonhandreke:
I sketched something out here that gets a file from the backup and runs some verification on it: https://github.com/opensanctions/operations/pull/1231

leonhandreke:
We have decided to deprioritize this for now, so closing for now until we dig it out again someday.

PR #2122, opened 2025-04-08
Fixes-8-apr

nvmbrasserie:


datasets/us/wv/med_exclusions/crawler.py
```diff
@@ -31,6 +31,8 @@ def crawl_item(row: Dict[str, str], context: Context):
     npi = row.pop("npi")
     first_name = row.pop("first_name")
     last_name = row.pop("last_name")
+    if "last name" in last_name.lower():
```
leonhandreke, line 34: Hm, what's this?

nvmbrasserie: there is an issue with some pdf pages, where header is on the second row. Setting `headers_per_page=False` allows us to take the headers only from the first page, and then for each page we skip the header row with` if "last name" in last_name.lower():`

PR #2126, opened 2025-04-09
foreign terrorist orgs

nvmbrasserie:
opensanctions/crawlers#322

### DIFF
--- a/datasets/us/state_terrorist_orgs/crawler.py
+++ b/datasets/us/state_terrorist_orgs/crawler.py
@@ -15,7 +15,7 @@ def split_clean_name(context, name):
             context.log.warning("Name override is not found", name=name)
             return name_clean, name_former, alias
         for name in result.names:
-            name_clean = name.get("name")
+            name_clean = name.get("name_clean")
             name_former = name.get("name_former")
             alias = name.get("alias")
     # Parse inline alias format: e.g., "Organization Name (Alias)"
--- a/datasets/us/state_terrorist_orgs/crawler.py
+++ b/datasets/us/state_terrorist_orgs/crawler.py
@@ -6,8 +6,8 @@ PROGRAM = "Foreign Terrorist Organizations designated under section 219 of the I
 def split_clean_name(context, name):
     # Defaults
     name_clean = name.strip()
-    name_former = ""
-    alias = ""
+    name_former = None
+    alias = None
     # Lookup override if name contains a dash
     if any(dash in name for dash in ["—", "–"]):
         result = context.lookup("names", name)
@@ -23,7 +23,7 @@ def split_clean_name(context, name):
         base, alias_part = name.rsplit(" (", 1)
         name_clean = base.strip()
         alias = alias_part.rstrip(")").strip()
-        name_former = ""
+        name_former = None
 
     return name_clean, name_former, alias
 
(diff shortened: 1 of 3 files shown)


datasets/us/state_terrorist_orgs/us_state_terrorist_orgs.yml
```diff
@@ -0,0 +1,104 @@
...
+description: |
+  Foreign Terrorist Organizations (FTOs) are entities designated by the U.S. Secretary of State 
+  under Section 219 of the Immigration and Nationality Act (INA). It is unlawful to knowingly 
+  provide these organizations with material support or resources. Non-citizen members of FTOs 
+  may be denied entry into, or removed from, the United States. Additionally, U.S. financial 
+  institutions are required to block and report any assets associated with these organizations 
+  to the U.S. Department of the Treasury.
+publisher:
+  name: U.S. Department of State
+  acronym: DOS
```
nvmbrasserie, line 20: I'm hesitant regarding this acronym. In a similar dataset we have `acronym: State`. Apparently both of them work, but I think `DOS` is closer to what we define as acronym

datasets/us/state_terrorist_orgs/crawler.py
```diff
@@ -0,0 +1,67 @@
...
+
+def crawl_row(context, row):
+    name = row.pop("name")
+    start_date = row.pop("date_designated", None) or row.pop(
+        "date_originally_designated", None
+    )
+    end_date = row.pop("date_removed", None)
+
+    name_clean, name_former, alias = split_clean_name(context, name)
+    entity = context.make("LegalEntity")
```
leonhandreke, line 39: Naive question, why don't we go for Organization here?

datasets/us/state_terrorist_orgs/crawler.py
```diff
@@ -0,0 +1,67 @@
...
+        result = context.lookup("names", name)
+        if not result or not result.names:
+            context.log.warning("Name override is not found", name=name)
+            return name_clean, name_former, alias
+        for name in result.names:
+            name_clean = name.get("name_clean")
+            name_former = name.get("name_former")
+            alias = name.get("alias")
+    # Parse inline alias format: e.g., "Organization Name (Alias)"
+    elif len(name.split(" (")) == 2:
```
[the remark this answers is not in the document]
nvmbrasserie: we generally try to avoid using regex when simpler solutions are available. as a rule of thumb, if something can be solved clearly without regex, that's usually preferred

datasets/us/state_terrorist_orgs/crawler.py
nvmbrasserie: not at all, you're right

datasets/us/state_terrorist_orgs/crawler.py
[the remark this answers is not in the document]
leonhandreke: Thinking a bit more about this, I'm wondering if we can find a general regex (or other rule) that matches the "normal" case so that we can warn in all other cases. I worry the current thing might not be super robust, like "Org (Alias1) (Alias2)" will not be touched at all by the function I think, and if they ever decide to use `|` instead of `—`, we also won't notice.

So maybe something like `r"[\w\s'-]+(\([\w\s'-]+\)){0,1}"` (to also capture Al-quaeda and Al'quaeda))

```python
SIMPLE_NAME_RE = r"[\w\s'-]+"
BORING_CASE_RE = rf"(?P<name>{SIMPLE_NAME_RE})(\((?P<alias>{SIMPLE_NAME_RE})\)){0,1}"
if match =| re.match(BORING_CASE_RE, name):
   return match.group() ...
else:
   result = context.lookup
   if result:
     return

context.warn("failed to parse")
return the unsplit name
```

leonhandreke (approved):
Just a little comment about the name parsing, but I'll let you decide --- I think this is good as well. Nice little crawler :)

nvmbrasserie:
you're right, this way we capture more edge cases

Issue #2121, opened 2025-04-08
Move programs info into etl db


leonhandreke:
- [ ] Write a little tool to copy sanction program info from Directus to the etl database
    - [ ] Productionize & set up cronjob
- [ ] Let `make_sanction` verify and set metadata on the Sanction object
- [ ] Make sure a couple of crawlers (eu?) are properly using this new infrastructure

Optional stretch goals in order of importance
- [ ] Build a custom metadata exporter (I don't really know what this is yet) to let the website consume this information from zavod instead of from Directus
- [ ] Allow search by program key. Currently unclear to me how to do this, either denormalize in yente
- [ ] Pull this into zavod-ui instead of Directus (I don't think this is a good idea yet, Directus seems a great place for it for now?)

cc @nvmbrasserie Let's work on this together :)

leonhandreke:
@nvmbrasserie I created https://github.com/opensanctions/opensanctions/pull/2128 with the basic infrastructure

So I think what needs to happen next is:

- Modify `be_fod` and `eu_fsf` (and a few other) crawlers to do the lookup from local program ID -> OpenSanctions program ID themselves.
- Let a few interesting crawlers pass in program IDs

pudo:
So https://github.com/opensanctions/opensanctions/pull/2269 does the backend part of making sanctioned entities searchable by program. We still need to upgrade yente to make the new field available in the API.

PR #2128, opened 2025-04-10
Load sanctions programs from etl-db

leonhandreke:
Add a little tool to load them from Directus and from a JSON (for development).

`cat contrib/load_programs_from_cms/fixture.json | python contrib/load_programs_from_cms load_programs_from_cms.py load-fixture`

See https://github.com/opensanctions/opensanctions/issues/2121

Discussion items:

- Do we want to use the SQLAlchemy ORM? I think it's sort of cool to have a free dataclass with type safety but @pudo had reservations against using it?

### DIFF
--- a/datasets/gb/hmt/crawler.py
+++ b/datasets/gb/hmt/crawler.py
@@ -126,7 +126,7 @@ def parse_row(context: Context, row: Dict[str, Any]):
     sanction = h.make_sanction(
         context,
         entity,
-        program=regime_name,
+        program_name=regime_name,
         source_program_key=regime_name,
         program_key=h.lookup_sanction_program_key(context, regime_name),
         start_date=designated_date,
--- a/datasets/gb/hmt/crawler.py
+++ b/datasets/gb/hmt/crawler.py
@@ -126,7 +126,7 @@ def parse_row(context: Context, row: Dict[str, Any]):
     sanction = h.make_sanction(
         context,
         entity,
-        program_key=context.lookup_value("sanction.programs", regime_name, None),
+        program_key=h.lookup_sanction_program_key(regime_name),
         start_date=designated_date,
     )
     sanction.add("authority", row.pop("ListingType", None))
--- a/datasets/gb/hmt/crawler.py
+++ b/datasets/gb/hmt/crawler.py
@@ -126,7 +126,7 @@ def parse_row(context: Context, row: Dict[str, Any]):
     sanction = h.make_sanction(
         context,
         entity,
-        program=regime_name,
+        program_name=regime_name,
         program_key=regime_name,
         start_date=designated_date,
     )
--- a/datasets/gb/hmt/crawler.py
+++ b/datasets/gb/hmt/crawler.py
@@ -126,7 +126,8 @@ def parse_row(context: Context, row: Dict[str, Any]):
     sanction = h.make_sanction(
         context,
         entity,
-        program_key=h.lookup_sanction_program_key(regime_name),
+        source_program_key=regime_name,
+        program_key=h.lookup_sanction_program_key(context, regime_name),
         start_date=designated_date,
     )
     sanction.add("authority", row.pop("ListingType", None))
--- a/datasets/gb/hmt/crawler.py
+++ b/datasets/gb/hmt/crawler.py
@@ -126,6 +126,7 @@ def parse_row(context: Context, row: Dict[str, Any]):
     sanction = h.make_sanction(
         context,
         entity,
+        program=regime_name,
         source_program_key=regime_name,
         program_key=h.lookup_sanction_program_key(context, regime_name),
         start_date=designated_date,
--- a/datasets/gb/hmt/crawler.py
+++ b/datasets/gb/hmt/crawler.py
@@ -126,8 +126,7 @@ def parse_row(context: Context, row: Dict[str, Any]):
     sanction = h.make_sanction(
         context,
         entity,
-        program_name=regime_name,
-        program_key=regime_name,
+        program_key=context.lookup_value("sanction.programs", regime_name, None),
         start_date=designated_date,
     )
     sanction.add("authority", row.pop("ListingType", None))
--- a/zavod/zavod/helpers/sanctions.py
+++ b/zavod/zavod/helpers/sanctions.py
@@ -12,7 +12,7 @@ def make_sanction(
     context: Context,
     entity: Entity,
     key: Optional[str] = None,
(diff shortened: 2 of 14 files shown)


leonhandreke:
Typechecks need https://github.com/alephdata/followthemoney/pull/1725 to go in.

zavod/zavod/helpers/sanctions.py
```diff
@@ -45,16 +56,25 @@ def make_sanction(
...
+            if program.url:
+                sanction.add("programUrl", program.url)
+            else:
+                sanction.add(
+                    "programUrl",
+                    f"https://www.opensanctions.org/programs/{program_key}",
+                )
         else:
-            context.log.warn(f"Program key {program_key!r} not found.", program=program)
+            context.log.warn(f"Program key {program_key!r} not found.")
```
jbothma, line 77: so

- if program_name is provided, `Sanction:program` is set to that
- if program_key is provided, `Sanction:program` is set to `program.title` unless `program_name` was provided

This feels like the sort of finicky but important logic where a test might save someone from an unintended mistake

zavod/zavod/shed/fsf.py
```diff
@@ -58,12 +58,11 @@ def parse_sanctions(context: Context, entity: Entity, entry: Element) -> None:
     for regulation in regulations:
         url = regulation.findtext("./publicationUrl")
         assert url is not None, etree.tostring(regulation)
-        program = regulation.get("programme")
+        source_program_key = regulation.get("programme")
         sanction = h.make_sanction(
             context,
             entity,
-            program=program,
-            program_key=program,
```
jbothma, line 66: is replacing the source-provided program string we've been publishing in `Sanction:program` for some time with our `programId` and `program.title` something we need to take through a scheduled and pre-announced release process? In case customers have logic on this on their end

zavod/zavod/stateful/programs.py
```diff
@@ -0,0 +1,20 @@
...
+# Since we'll only ever have a few positions, it's cheaper to just read them all once.
+@functools.cache
+def get_all_positions_by_key(context: Context) -> dict[str, Program]:
+    session = Session(context.conn)
+    programs = session.scalars(select(Program)).all()
+    return {p.key: p for p in programs}
+
+
+def get_program_by_key(context: Context, program_key: str) -> Optional[Program]:
+    return get_all_positions_by_key(context).get(program_key, None)
```
jbothma, line 20: s/position/program/

and lol, only a few positions

zavod/zavod/shed/fsf.py
jbothma: Do we want to keep the source-provided keys somewhere, e.g. Sanction:unscId. Should we introduce a sanctionTitle or sanctionLabel property instead?

zavod/zavod/shed/fsf.py
```diff
@@ -58,12 +58,11 @@ def parse_sanctions(context: Context, entity: Entity, entry: Element) -> None:
...
+            program_key=h.lookup_sanction_program_key(source_program_key),
```
jbothma, line 65: does this need a context?

jbothma (commented):
I like the separation of the lookup and make_sanction helper.

My main query is about 

a. dropping the source's string identifying the program entirely (once a crawler is migrated to this)
b. emitting our data under the source's name rather than our own using an analyzer (special crawler) - did you guys discuss the trade-offs and opt for this? 

I see you discussed not using their keys as our keys, it's mainly just confirming the provenance thing and dropping source data we've previously published.

As for the ORM, I'd also prefer not to mix them in one application. I also don't recall particularly tough times, but I probably haven't used it in anger like Friedrich has. I'm up for trying it in one full application (e.g. all of etl is ORM, but all of saas is table) - then we might learn from the experiment.

zavod/zavod/helpers/sanctions.py
leonhandreke: I simplified the logic. Now we do this:

- If `program_key` is given, check if it's valid. If it is:
    - Stuff it in `Sanction:programId`
    - Stuff the `Program.url` into `Sanction:programUrl`

I also added a test for this.

zavod/zavod/stateful/programs.py
leonhandreke: Whoops, done.

zavod/zavod/shed/fsf.py
leonhandreke: I added a little comment.

zavod/zavod/shed/fsf.py
leonhandreke: I simplified this, we're now no longer putting in our title.

Do you think it's worth keeping around the original value that we generated our `program_key` from?

leonhandreke: Well, I think you had a strong-ish opinion about this and I don't, so I'll put it in there.

leonhandreke:
So, about the ORM: I think it's cool to have free typesafe wrappers here and I don't think the mix is too bad. I'm up for converting Positions as well, but we'll still have the statements table, which is pulled in from nomenklatura.

I'm not too attached to it but a) I've already done it and b) I have a slight aversion to doing it that way just because we've always done it that way :)

pudo:
It's not entirely true that zavod and nomenklatura can do their own thing here, unfortunately: sqlalchemy.core uses explicit connections, whereas the ORM uses a session for transaction management and caching that's tied to a connection so we'll need to instantiate that on the `Context` and make sure its lifecycle is aligned with the runtime of the crawler, and doesn't clash with the nomenklatura bits using the connection directly. All doable, but an extra layer of state to keep track of.  

Definitely not a fan of this migration, but seems like Leon's set his mind to it.

datasets/gb/hmt/crawler.py
```diff
@@ -126,8 +126,8 @@ def parse_row(context: Context, row: Dict[str, Any]):
     sanction = h.make_sanction(
         context,
         entity,
-        program=regime_name,
```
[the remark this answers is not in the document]
leonhandreke: Done.

leonhandreke:
I see the argument about transaction management, and it convinced me that I don't understand enough about this to judge whether this is a good idea. So I transitioned to using the tables API again. Sorry for the back and forth...

pudo:
No I apologize for the "grandpa always likes his tea this way" push-back. Thank for you for being flexible.

jbothma (approved):
Nice, and thanks for the cleanup of the mappings!

PR #2136, opened 2025-04-14
[gb_coh_disqualified] Emit things without DoB as LegalEntity

leonhandreke:
It seems they added a bunch of companies to the list of Persons.

### DIFF
--- a/datasets/gb/coh/disqualified.py
+++ b/datasets/gb/coh/disqualified.py
@@ -69,19 +69,26 @@ def crawl_item(context: Context, listing: Dict[str, Any]) -> None:
     person.add("sourceUrl", source_url)
 
     # TODO(Leon Handreke): Clean this up, it seems they added lots of Companies to this list of persons.
+    nationality = data.pop("nationality", "")
     person.add_cast("Person", "birthDate", data.pop("date_of_birth", None))
-    h.apply_name(
-        person,
-        first_name=data.pop("forename", None),
-        last_name=data.pop("surname", None),
-        middle_name=data.pop("other_forenames", None),
-        lang="eng",
-    )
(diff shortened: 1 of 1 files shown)


datasets/gb/coh/disqualified.py
```diff
@@ -55,7 +55,14 @@ def crawl_item(context: Context, listing: Dict[str, Any]) -> None:
...
     if data is None:
         return
-    person = context.make("Person")
+    dob = data.pop("date_of_birth", None)
+
+    if dob:
+        person = context.make("Person")
+    else:
+        # TODO(Leon Handreke): Clean this up, it seems they added lots of Companies to this list of persons.
+        person = context.make("LegalEntity")
```
jbothma, line 64: a common pattern we have

```
entity = context.make("LegalEntity")
entity.add_cast("Person", "birthDate", dob)  # if dob is truthy, cast to Person and add birthDate
```

jbothma: maybe worth renaming person to entity here?

datasets/gb/coh/disqualified.py
```diff
@@ -65,19 +72,20 @@ def crawl_item(context: Context, listing: Dict[str, Any]) -> None:
...
+        )
+        person.add("title", data.pop("title", None))
 
-    nationality = data.pop("nationality", None)
-    if nationality is not None:
-        person.add("nationality", nationality.split(","))
-    person.add("birthDate", data.pop("date_of_birth", None))
+        nationality = data.pop("nationality", None)
+        if nationality is not None:
+            person.add("nationality", nationality.split(","))
```
jbothma, line 87: let's put this into `country` if it's not a Person

PR #2137, opened 2025-04-14
Issues 2025-04-14

leonhandreke:
Mostly low-hanging easy stuff, but:

- **[cia_world_factbook] Add lookups**: Please review, some entries are split so that the context required to infer the meaning is actually in the statement before it.
- **[il_mod_crypto] Update hash**: Couldn't figure out what changed, turns out saving the past HTML was smart after alll...
- **[gb_coh_disqualified] Some easy lookups & type.country value should be ISO**: Did this ever work? Why weren't we warned about this?


datasets/za/mm_officials/za_mm_officials.yaml
```diff
@@ -32,10 +32,10 @@ data:
 assertions:
   min:
     schema_entities:
-      Person: 750
+      Person: 1500
   max:
     schema_entities:
-      Person: 800
```
jbothma, line 38: I wonder where this is going wrong.
If I visit https://municipaldata.treasury.gov.za/api/cubes/officials/facts I get
- 396 instances of `"Mayor`
- 292 instances of `"municipal manager`
- 292 instances of `"chief fina`

So I don't know how we're ending up at 2134 Persons

jbothma (commented):
Nice - I'd just query the ZA one - something's fishy

datasets/za/mm_officials/za_mm_officials.yaml
jbothma: for context, South Africa has 257 municipalities. Disestablished municipalities don't exist any more and their area has been taken over by new/existing municipalities. So even 292 is a lot of managers and CFOs

jbothma (approved):
> [gb_coh_disqualified] Some easy lookups & type.country value should be ISO: Did this ever work? Why weren't we warned about this?

What do you mean? Mapping Ukranian to Ukraine results in a statement `nationality, ua, ..., Ukrainian`

Anyway, ship it

PR #2151, opened 2025-04-17
Inline version comment allows dependabot to update the comment too

jbothma:
A friend pointed out this functionality

https://github.com/dependabot/dependabot-core/pull/5951#issuecomment-1298147084

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
 
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4.1.7
+      - uses: actions/checkout@v4
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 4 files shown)


jbothma:
Just need to investigate this caveat and/or pin the other actions in the repo 

> BUT, if you do that, you have to do it on **every** action in the repo, or dependabot fails

leonhandreke:
This is cool (but lame documentation by the dependabot people). Is this caveat actually documented anywhere? Do we lose anything in just trying this and seeing what happens? Ship it I say :)

jbothma:
The risk is other things not being updated, so we can ship this once we
know what dependabot failure looks like, and if we can e.g. pin the
github-provided actions back a version and verify that dependabot actually
does update them.

jbothma:
so the caveat seems to have been referring to all the actions in all the workflows being consistent, otherwise dependabot resolves one of those as the version, and if it thinks that's up to date, it won't update the others and they might become stale.

PR #2159, opened 2025-04-22
[regulatory] Make frequency metadata match reality (weekly -> daily)

leonhandreke:
schedule trumps frequency anyway, but we show frequency in the overview
table


leonhandreke:
This was dumb, sorry.

PR #2161, opened 2025-04-22
[exporters] Record count by Sanction.programId in stats

leonhandreke:
This will allow us to fill the relevant tables on the website (datasets
table on program page, program table in dataset page) from the exported
metadata instead of keeping them manually up to date from Directus.

See https://github.com/opensanctions/opensanctions/issues/2121

This adds an entry like this to the stats:

```json
"sanctions": {
  "programs": {
    "id": "EU-UKR",
    "count": 212,
  }
}
```

Making it a dict like this potentially allows us to fill in more program metadata in the future (like title, issuer, whatever) and generate everything in the website from that. But I'm not sure we even want that? Or should there be some other export of this database that the site reads?

### DIFF
--- a/zavod/zavod/exporters/statistics.py
+++ b/zavod/zavod/exporters/statistics.py
@@ -37,6 +37,16 @@ def get_country_facets(countries: Dict[str, int]) -> List[Any]:
     return facets
 
 
+def get_sanctions_programs_facets(sanctions_programs: Dict[str, int]) -> List[Any]:
+    return [
+        {
+            "id": program_id,
+            "count": count,
+        }
+        for program_id, count in sanctions_programs.items()
+    ]
+
+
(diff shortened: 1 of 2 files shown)


pudo:
First off: I love that this exists now! We already list associated programs on the dataset page, perhaps we can drive this from that data now to include counts. 

Regarding moving off of Directus: that would only make sense once we have a `zavod-ui` in place where we edit programs. Then we could dump the programs data maintained in that UI to `data.opensanctions.org/meta` and drive the website from here. But I also take your point about Directus being a really, really good UI for editing programs, so this isn't a priority.

pudo (approved):
Love it :)

leonhandreke:
Thanks for taking a look and giving this the green light!

I'll migrate the website to this in the coming weeks.

PR #2167, opened 2025-04-23
ignore fileds for context.audit_data expanded

nvmbrasserie:


datasets/sk/rpvs/crawler.py
```diff
@@ -158,6 +159,7 @@ def process_entry(context, entry):
     entity_data = context.fetch_json(details_url, cache_days=3)
     entity_data = rename_headers(context, entity_data)
     legal_entity_name = entity_data.get("trading_name", "")
+    entity_type = entity_data.pop("entity_type", "")
```
nvmbrasserie, line 162: we want to `.pop` it for all entities but add it as a `legalForm` only for entities (not persons)

leonhandreke: What does it contain for persons? Might not hurt to add a comment about this!

leonhandreke (approved):
Nice, thanks for expanding this!

PR #2173, opened 2025-04-24
initial commit

nvmbrasserie:


datasets/cz/terrorists/cz_terrorists.yml
```diff
@@ -0,0 +1,441 @@
...
+data:
+  url: https://www.e-sbirka.cz/sb/2008/210/2009-04-01?zalozka=text
+  format: HTML
+dates:
+  formats: ["%d.%m.%Y"]
+
+assertions:
+  min:
+    schema_entities:
+      Person: 30
```
leonhandreke, line 32: Since this is static, let's assert on the exact number?

leonhandreke:
Did you do any research if the Czech government has more of these laws? I tried to look and it seems they have a concept of "related laws", with this one relating to a more general "sanctions" law, but in the related laws of that one, there are many law, so not really crawlable... Friedrich suggested maybe finding a good filter and then setting up an email subscription? Maybe you could send a couple minutes looking into this.

And I guess we could also keep this crawler completely static in a Google Sheet? What are your thoughts on this?

Just added two commits, do you agree with these changes? I felt like the paths for Person/Organization are different, so decided to separate them more.

nvmbrasserie:
I looked into it and couldn't find any related Regulations. The one we're working with was updated once in 2009 and we're crawling already amended regulation. Other than that and the main Czech sanctions list, I haven't found any lists.  Could you point me to other lists you've seen?

I think since it's crawlable and clean from the original source, there's no need to switch it to the Google sheet. It's going to be harder to maintain and whenever possible we prefer to stick with the original source.

leonhandreke (approved):

leonhandreke:
Nice, thanks for following up so quickly, let's ship it :)

PR #2177, opened 2025-04-25
[afdb_sanctions] Aaaand we're back to datatable

jbothma:
Revert to the previous implementation - looks like they just wanted to play with server-side pagination for a day.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,4 +19,5 @@ repos:
     - id: mypy
       # Don't run isolated but use the activated venv. Not perfect but better than re-declaring all deps here
       language: system
-      args: ['--strict', '--exclude', 'zavod/zavod/tests']
+      exclude: '^datasets|^zavod/zavod/tests'
+      args: ['--strict']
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -19,4 +19,5 @@ repos:
     - id: mypy
       # Don't run isolated but use the activated venv. Not perfect but better than re-declaring all deps here
       language: system
-      args: ['--strict', '--exclude', 'zavod/zavod/tests']
+      exclude: '^datasets|^zavod/zavod/tests'
+      args: ['--strict']
```
jbothma, line 23: @leonhandreke it looks like exclude doesn't work when the file is explicitly specified, e.g. when it's staged for commit but pre-commit also has an exclude option per-hook. Does this look ok to you?

```
mypy --strict --exclude zavod/zavod/tests --exclude datasets datasets/_global/afdb_sanctions/crawler.py
```

leonhandreke: If it works, looks good to me! Another way I just found is https://stackoverflow.com/questions/71657355/run-mypy-from-pre-commit-for-different-directories

leonhandreke (approved):

PR #2178, opened 2025-04-25
Warn to avoid duplication between FSF and Journal Sheet

pudo:
This is so that we don't end up with lots of duplicate but unmanaged information in the journal sheet.

### DIFF
--- a/datasets/eu/journal_sanctions/crawler.py
+++ b/datasets/eu/journal_sanctions/crawler.py
@@ -1,11 +1,15 @@
 import csv
-from typing import Dict, Iterable
+from typing import Dict
 
+from zavod import Context, Entity
 import zavod.helpers as h
-from zavod import Context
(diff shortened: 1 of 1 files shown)


jbothma (approved):
The only issue I have with this is that someone might be listed for multiple programs. So we might remove someone when they're in FSF for one program, but the journal that added them to the sheet is another regime not yet in FSF.

They might also be in the resolver but not in the dataset.

But since the point of this crawler is to ensure they get added sooner than the person gets added to FSF, I think the chance of removing someone "too soon" is very low.

pudo:
The weird thing about that is that we don't actually store the relevant program in the journal sheet, only the amendment notice. I share that concern of multiple programs designating an entity, but then perhaps we need to bring programs into the journal sheet?

I'm a bit surprised by how little overlap there is between the FSF XML and the journal sheet, to be honest. We may need to investigate that.

PR #2181, opened 2025-04-25
lookups for sanction programs in ofac

nvmbrasserie:
We need to merge the `IGNORE` solution for the programs lookups before merging


datasets/us/ofac/us_ofac_sdn.yml
```diff
@@ -63,7 +63,147 @@ assertions:
...
+        value: US-TCO
+      - match:
+          - Non-SDN Palestinian Legislative Council List
+          - CAPTA List
+          - FSE List
+          - Non-SDN CMIC List
+          - Non-SDN Menu-Based Sanctions List
+          - Non-SDN Palestinian Legislative Council List
+          - Sectoral Sanctions Identifications List
+        value: null
```
nvmbrasserie, line 206: we need to find the best way to ignore these in `us_ofac_sdn`, the same functionality is needed for `us_ofac_cons`

nvmbrasserie: I implemented it in https://github.com/opensanctions/opensanctions/pull/2258

leonhandreke (dismissed):
Looks good in general, added some comments where I think we should add some comments for the next person that doesn't know the data as well :)

@pudo since this dataset is so core to our business, maybe you could take a quick look here whether this change is sane? The core change is that for entries on the "Consolidated List", we've so far been emitting at least two `Sanction`s: one with program "Consolidated List" and at least one more specific e.g. "CATA List". After this change, the "Consolidated List" `Sanction` will no longer be emitted.

The reason is that I feel this is more "correct", cause the "Consolidate List" is really just a list of lists and we know the exact list an item is on in every case. But I don't have a super strong opinion on this, so would also be fine to keep the old behavior around. WDYT?

datasets/us/ofac/ofac_advanced.py
```diff
@@ -302,22 +302,16 @@ def parse_distinct_party(
 
     # Sanctions designations
     entry_path = f"SanctionsEntries/SanctionsEntry[@ProfileID='{profile_id}']"
-    # Assert that each entity has at least one sanction linked to it
+    # We skip emitting entries with ListId "Consolidated List" because they always have a second,
+    # more specific ListId attached (such as "CAPTA List"). This assert ensures that this assumption
+    # always holds so that we don't skip emitting a Sanction for every entity on the list.
     sanction_found = False
```
leonhandreke, line 308: (nit) we can remove this declaration

PR #2185, opened 2025-04-28
US Washington DC Excluded Parties List

nvmbrasserie:
opensanctions/crawlers#329


datasets/us/dc/med_exclusions/us_dc_med_exclusions.yml
```diff
@@ -0,0 +1,104 @@
+title: US DC Medicaid Sanctioned Providers
```
jbothma, line 1:
```suggestion
title: US Washington DC Excluded Parties List
```

datasets/us/dc/med_exclusions/us_dc_med_exclusions.yml
```diff
@@ -0,0 +1,104 @@
...
+entry_point: crawler.py
+prefix: us-meddc
+coverage:
+  frequency: daily
+  start: "2025-04-28"
+load_statements: true
+summary: >
+  List of providers excluded from participating in the District of Columbia Medicaid program.
```
jbothma, line 9: It's not just medicaid - I phoned and medicaid is included, but this is all providers barred from doing business with the district.

So let's make sure the metadata, prefix, and the dataset name are in line with their naming and not medicaid specific.

datasets/us/dc/med_exclusions/crawler.py
```diff
@@ -0,0 +1,85 @@
...
+    company_name = row.pop("name_of_company", None)
+    schema = "Person" if name else "Company"
+
+    entity = context.make(schema)
+    entity.id = context.make_id(name, company_name)
+    if entity.schema.is_a("Person"):
+        entity.add("name", name)
+        log_messy_names(context, entity.get("name")[0])
+    else:
+        entity.add("name", company_name)
```
jbothma, line 49: It feels like the code isn't really reflecting how there's only one name and it's either an individual name, or a company name. It's a bit of a nitpick but I think the code would be a bit clearer if we end up assigning to `name_raw`.

Part of what I think makes the current approach a bit weird how we're making an id where either of those are expected to have a value. It might actually be good to add an assertion that you end up with _some_ value in name_raw.

Let's use name_raw and address for id just in case there are two different John Smiths getting banned.

datasets/us/dc/med_exclusions/crawler.py
```diff
@@ -0,0 +1,85 @@
...
+    entity.add("address", row.pop("principal_address"))
+
+    sanction = h.make_sanction(context, entity)
+    sanction.set("authority", row.pop("agency_instituting_the_action", None))
+    sanction.add("reason", row.pop("reason_for_the_action", None))
+    h.apply_date(sanction, "startDate", row.pop("action_date"))
+    h.apply_date(
+        sanction,
+        "endDate",
+        row.pop("expiration_date", row.pop("termination_date", None)),
```
jbothma, line 65: maybe worth just having an assert somehow that we ended up with some value for this

jbothma (commented):
Thanks! Some smallish changes then we're done, I think

jbothma (commented):
Thanks for the quick turnaround! Just see the rename comment, then you can ship

datasets/us/dc/med_exclusions/us_dc_med_exclusions.yml
```diff
@@ -0,0 +1,105 @@
+title: US Washington DC Excluded Parties List
+entry_point: crawler.py
+prefix: us-meddc
```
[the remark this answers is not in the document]
nvmbrasserie: I'd suggest `datasets/us/dc/us_dc_excluded_contractors.yml`, what do you think?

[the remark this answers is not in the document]
jbothma: contractors I think sounds a lot like "people who are gonna renovate your house"in US speak so I think either `us_dc_excluded_parties` (their words) or `us_dc_exclusions` - general term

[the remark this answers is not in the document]
nvmbrasserie: will the same one work as the prefix? `us-dc-exclusions`? or do you think it would be better to cut it?

PR #2188, opened 2025-04-29
[pl_mswia_sanctions] Better name parsing

leonhandreke:
- Don't add names twice for Person entities (once as name, once as
  alias)
- Parse some more other-language prefixes for companies
- Parse previous name
- clean up some quotes around aliases


nvmbrasserie:
Added a couple of things, everything else looks super clean!

nvmbrasserie (approved):

PR #2195, opened 2025-04-30
Fixes-30-04

nvmbrasserie:


datasets/ua/war_sanctions/crawler.py
```diff
@@ -423,6 +429,43 @@ def emit_relation(
...
+    # `asset` refers to the previous entity in the chain, starting with the first entity
+    asset = legal_entity
+
+    # Iterate over the remaining entities in the rostec_ownership
+    for entity in rostec_ownership[1:]:
+        # Create the new entity for each linked company
+        owner = context.make("LegalEntity")
+        owner.id = context.make_id(entity, "affiliated with Rostec")
+        # owner.add("name", rostec_ownership[i])
+        # context.emit(owner)
```
nvmbrasserie, line 456: there's a major [update](https://war-sanctions.gur.gov.ua/en/rostec?kc%5Bh%5D=&kc%5Bsearch%5D=&page=2&per-page=12) on the rostec's holdings and entities owned by them. the issue is that `rostec` itself is not on the list, but it [exists](https://war-sanctions.gur.gov.ua/en/rostec/16) on the website. Currently, I decided not to emit owner's, since all of them will be emitted in `crawl_legal_entity` anyway. But in this case we don't emit rostec, which is a key owner for all of them

datasets/ua/war_sanctions/crawler.py
```diff
@@ -423,6 +429,43 @@ def emit_relation(
...
+    Args:
+        context: Context
+        legal_entity: The initial legal entity (the first in the ownership chain).
+        rostec_ownership: A list of entities that form the ownership chain, from the legal entity to Rostec.
+    """
+    # `asset` refers to the previous entity in the chain, starting with the first entity
+    asset = legal_entity
+
+    # Iterate over the remaining entities in the rostec_ownership
+    for entity in rostec_ownership[1:]:
```
jbothma, line 451: Since it's so common in our code that `entity` is an `Entity` type, perhaps we can have a name that's a bit different from that here

```suggestion
    for abbrev_name in rostec_ownership[1:]:
```

datasets/ua/war_sanctions/crawler.py
```diff
@@ -509,9 +552,16 @@ def crawl_legal_entity(context: Context, link, program):
...
 
     legal_entity = context.make("LegalEntity")
-    legal_entity.id = context.make_id(name, name_abbr, reg_num)
+    # For companies owned by Rostec, we want to create simplified IDs
+    # to tie them to the parent company.
+    legal_entity.id = (
+        context.make_id(name_abbr, "affiliated with Rostec")
+        if rostec_ownership
+        else context.make_id(name, name_abbr, reg_num)
+    )
```
jbothma, line 564: since this is breaking onto multiple lines, perhaps we can allow ourselves a fully-fledged `if...else` as opposed to a ternary - that would be easier to read.

datasets/ua/war_sanctions/crawler.py
jbothma: why don't we just emit them here too, just in case? It's not so many that emitting a small entity would be a big pain space-wise.

Do you mean rostec currently doesn't get crawled? You could make it crawl the last URL in the list each time so that rostec gets crawled even if its ID changes. The http cache will mean it isn't that slow.

datasets/ua/war_sanctions/crawler.py
```diff
@@ -423,6 +429,43 @@ def emit_relation(
...
+        rostec_ownership: A list of entities that form the ownership chain, from the legal entity to Rostec.
+    """
+    # `asset` refers to the previous entity in the chain, starting with the first entity
+    asset = legal_entity
+
+    # Iterate over the remaining entities in the rostec_ownership
+    for entity in rostec_ownership[1:]:
+        # Create the new entity for each linked company
+        owner = context.make("LegalEntity")
+        owner.id = context.make_id(entity, "affiliated with Rostec")
```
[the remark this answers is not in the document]
jbothma: what if you did the IDs of the rostec entities based on the bit after /en/ of their URL from the start? `en/rostec/144`
Then at least they don't get re-keyed if we do the russian thing later

jbothma (approved):
Thanks for thiis!

PR #2205, opened 2025-05-01
[eu journal updates] Move git stuff to script for cron job to call

jbothma:


contrib/eu_journal/check_updates.sh
```diff
@@ -0,0 +1,27 @@
...
+export EU_JOURNAL_SEEN_PATH=/data/eu_journal/seen.txt
+
+python3 ${SCRIPT_PATH}/fetch_updates.py --slack --update-seen
+
+cd $EU_JOURNAL_STATE_PATH
+git diff
+git config user.name "EU Journal Bot"
+git config user.email "184417265+eu-journal-bot@users.noreply.github.com"
+git update-index --refresh || (git add seen.txt && git commit -m "Update seen.txt")
+git push --verbose origin main:test-eu-journal
```
leonhandreke, line 27: This still pushes to the test branch.

leonhandreke:
Nice! Two things:

1. Maybe this should actually go in `operations/etl`? It's pretty specific to us. But I see how having it next to the actual workhorse in `contrib/` also makes sense.
2. The heartbeat is called in the python script, and things can still go wrong after that. Would it make sense to move that to a curl at the end of a script? Alternatively, https://blog.sentry.io/sentry-bash/ looks like the integration would only be three lines and could then file GitHub issues for us.

jbothma:
> The heartbeat is called in the python script, and things can still go wrong after that. Would it make sense to move that to a curl at the end of a script? Alternatively, https://blog.sentry.io/sentry-bash/ looks like the integration would only be three lines and could then file GitHub issues for us.

aaa nice catch - I had it in the github actions workflow and moved it to the python when I added the explicit failure, but the success call should be after the git work yeah. Thanks!

leonhandreke (approved):

PR #2219, opened 2025-05-06
[us_bis_denied] switched to crawl xls from the new website

nvmbrasserie:


pudo:
Nice work! This may be a larger chunk of work: there's four lists on BIS (https://www.bis.gov/licensing/guidance-on-end-user-and-end-use-controls-and-us-person-controls), all of which are also mirrored in `us_trade_csl`. I'm wondering a) if we can gain more insight from crawling the BIS site ourselves - looks like there's some company hierarchy info, and links to FR notices, and b) if we, alternatively, also get the sublist ID (Denied, UVL, EL, ...) from the CSL list?

nvmbrasserie:
a) None of the BIS-hosted lists appear to be broader than the CSL JSON data. The only notable exception is that the new XLS file we're switching to in this PR includes more entities than the older Denied Persons List — likely because some entries haven’t been fully migrated to the CSL yet. I don’t think we’d gain much from crawling the BIS site directly, though I’ll take another close look to see if there are any hierarchy relationships (I haven’t seen any so far). Regarding the Federal Register links, they seem to be similar across both sources.

b) CSL is far more coherent structurally. In some cases, we can even infer relationships just by splitting strings (I’ve started doing that here: https://github.com/opensanctions/opensanctions/pull/2241). As for sublist IDs like Denied, UVL, EL, etc., we already have their full names in `source = result.pop("source", "")`, and could just remap those to shorter identifiers.

pudo:
Oh so this is a lovely outcome: we don't have a reason to crawl BIS directly, let's not do it. I feel like maybe doing a full lookup in `us_trade_csl` could be a wonderful improvement: split out a clean version of the `Authority`, and map to a `programId` at the same time. Would you be up for taking that on?

nvmbrasserie:
Sure thing, will do

datasets/us/bis_denied/crawler.py
```diff
@@ -1,43 +1,56 @@
...
-    path = context.fetch_resource("source.tsv", context.data_url)
+    doc = context.fetch_html(context.data_url, cache_days=1)
+    url = doc.xpath(".//a[contains(normalize-space(.), 'Export as CSV')]/@href")
+    assert len(url) == 1, "Expected exactly one URL"
+    path = context.fetch_resource("source.xls", url[0])
     context.export_resource(path, "text/tsv", title=context.SOURCE_TITLE)
-    with open(path, "r") as csvfile:
-        for row in csv.DictReader(csvfile, delimiter="\t"):
-            parse_row(context, row)
+    for row in h.parse_xls_sheet(context, xlrd.open_workbook(path)["dpl"]):
```
jbothma, line 55: when we crawl spreadsheets, it's a good idea to ensure somehow that if worksheets get added, we either automatically crawl them too, or get a warning that there might be unhandled data - whichever is appropriate.

jbothma (commented):
I'm not as clued in on what the plan is here so maybe I'm missing something.

>  we don't have a reason to crawl BIS directly, let's not do it

It looks like we're still crawling DPL from BIS in this PR, but it sounds like the plan is to crawl CSL instead?

> maybe doing a full lookup in us_trade_csl could be a wonderful improvement: split out a clean version of the Authority, and map to a programId at the same time

we've got a program name now but not a programId yet. Is that still coming?

pudo:
Honestly let's keep it if for no other reason than that otherwise we need announce we're killing BIS and it'll confuse the living crap out of 75 partners :/

pudo (approved):

PR #2241, opened 2025-05-09
[us_trade_csl] relationships establishment

nvmbrasserie:
@pudo:
"I feel like maybe doing a full lookup in us_trade_csl could be a wonderful improvement: split out a clean version of the Authority, and map to a programId at the same time."


nvmbrasserie:
most programs are not found: `[warning  ] Program with key 'EU-AFG' not found.`. Taking a closer look

datasets/us/trade_csl/us_trade_csl.yml
```diff
@@ -100,6 +114,277 @@ lookups:
...
+        properties:
+          name: Huawei Technologies Co., Ltd. (Huawei)
+          related: Shenzhen Huawei Anjiexin Electricity Co., Ltd
+          notes: "Affiliated Entity:  Shenzhen Huawei Anjiexin Electricity Co., Ltd."
+      - match: "China Aerospace Science and Industry Corporation Second Academy.  Subordinate Institution -  Second Design Department"
+        properties:
+          name: China Aerospace Science and Industry Corporation Second Academy
+          related: Second Design Department
+          notes: "Subordinate Institution -  Second Design Department"
+      - match: "China Aerospace Science and Industry Corporation Second Academy.  Subordinate Institution - 23rd Research Institute"
```
leonhandreke, line 235: Hm, for these branches I feel it makes sense to keep around the full name or put the branch name in parentheses or with a comma (like for the Huawei departments above). I think "23rd Research Institute" on its own isn't really enough to make sense.

datasets/us/trade_csl/us_trade_csl.yml
```diff
@@ -100,6 +114,277 @@ lookups:
...
+        properties:
+          name: China Aerospace Science and Industry Corporation Second Academy
+          related: 23rd Research Institute
+          notes: "Subordinate Institution - 23rd Research Institute"
```
leonhandreke, line 239: I don't think putting this in the notes adds any value here, let's try to put it in `original_value` somewhere? Or do you think it adds value here?

datasets/us/trade_csl/crawler.py
```diff
@@ -110,10 +140,15 @@ def parse_result(context: Context, result: Dict[str, Any]):
     name_with_information_res = context.lookup("name_with_information", name)
     if name_with_information_res is not None:
         entity.add("name", name_with_information_res.properties["name"])
```
leonhandreke, line 142: can we add the `original_value` here?

datasets/us/trade_csl/us_trade_csl.yml
```diff
@@ -100,6 +114,277 @@ lookups:
...
+          related: Hangzhou Huawei Communication Technology Co., Ltd
+          notes: "Affiliated Entity:  Hangzhou Huawei Communication Technology Co., Ltd."
+      - match: "Huawei Technologies Co., Ltd. (Huawei).  Affiliated Entity:  Huawei Digital Technologies (Suzhou) Co., Ltd."
+        properties:
+          name: Huawei Technologies Co., Ltd. (Huawei)
+          related: Huawei Digital Technologies (Suzhou) Co., Ltd
+          notes: "Affiliated Entity:  Huawei Digital Technologies (Suzhou) Co., Ltd."
+      - match: "Huawei Technologies Co., Ltd. (Huawei).  Affiliated Entity:  Hangzhou Huawei Enterprises"
+        properties:
+          name: Huawei Technologies Co., Ltd. (Huawei)
```
leonhandreke, line 139: I wonder whether it should actually be the other way round: "Hangzhou Huawei Enterprises" is the entity being targeted by this entry, and they are related to "Huawei Technologies Co., Ltd. (Huawei)". What do you think?

datasets/us/trade_csl/crawler.py
```diff
@@ -8,10 +8,40 @@
 from zavod import Context, Entity
 from zavod import helpers as h
 
+#     Programs falling under the 'is_ofac=TRUE' check:
```
leonhandreke, line 11: I see the `is_ofac` check is somewhere below, maybe move that documentation there?

leonhandreke: I just looked into what this actually means. IIUC, this means that we emit these entities with the `ofac-<entity_number>` ID so that they match without a merge? Let's add a sentence about this here to explain this speciality.

datasets/us/trade_csl/crawler.py
```diff
@@ -8,10 +8,83 @@
...
+    related_name: str,
+    list_entry: Dict[str, Any],
+    source_program: str
+):
+    related_entity = context.make("LegalEntity")
+    related_entity.id = context.make_id(related_name)
+    related_entity.add("name", related_name)
+    context.emit(related_entity)
+
+    make_and_emit_sanction(
```
leonhandreke, line 44: Do we need to emit a sanction here? If I understand this correctly, this is mostly "Huawei Inc. Affiliated Entity: Special Lab Ltd.". This sanction means that we not only sanction Special Lab Ltd (which happens in the main function) but also "Huawei Inc." Is this what we want, conceptually? Don't they have their own "you are sanctioned" list entry?

leonhandreke:
I pushed a little commit with some renames, type fixes, etc, hope that's alright.

Only one question remaining to me about whether to add a sanction object for both sides of the relationship. Otherwise, looks good!

datasets/us/trade_csl/crawler.py
leonhandreke: I'm a bit worried that after we merge these into Huawei Inc., we'll have 12 Sanctions objects on Huawei that we then need to dedupe.

leonhandreke: Acutally, looking at https://www.opensanctions.org/statements/NK-P3xHZasb3VZasvJLDwiGHC/?prop=name, it seems like they (always?) have the parent company with the same name. 

I guess we could build a dict of name → entity ID where we then look up the parent... but that seems a bit overkill, we can let the dedupe process take care of it I guess.

PR #2254, opened 2025-05-12
[ky_parliament] 2025-2029

nvmbrasserie:


datasets/ky/parliament/crawler.py
```diff
@@ -139,13 +137,19 @@ def crawl(context: Context):
...
         context.log.warning(
             f"Expected at least {expected_current_member_count} current members but found {current_member_count}"
         )
-
+    # Former members were added to the historical Google Sheet
+    # based on https://parliament.ky/members/former-members/.
+    # Since many individuals in Parliament rotate through different roles,
+    # some entries are duplicated to reflect multiple positions held over
+    # time. Where exact dates are provided (e.g., “November 23, 2023–end of term”),
+    # I used 2025 as the assumed end date. For positions without specific timeframes,
```
leonhandreke, line 151: (nit) "The Parliament was dissolved on 1st March 2025 through Proclamation No. 3 of 2025. " so maybe use that?

leonhandreke (approved):
Nice, happy this transferred so cleanly and nice that you added more documentation. To an uneventful parliamentary term, at least in terms of this crawler 🥂

PR #2255, opened 2025-05-13
topic for forced labor 

nvmbrasserie:


pudo:
Renamed this into `forced.labor` in 3.8.4 :) Keep it simple

PR #2256, opened 2025-05-13
Fixes-13-05

nvmbrasserie:


datasets/fr/amf_regulatory_sanctions/fr_amf_regulatory_sanctions.yaml
```diff
@@ -997,9 +996,7 @@ lookups:
         relationships:
           - schema: Succession
             from_prop: predecessor
-            from:
-              - "Caceis Investor Services Bank France"
-              - "RBC Investor Services Bank France"
+            from: ["Caceis Investor Services Bank France", "RBC Investor Services Bank France"]
```
leonhandreke, line 999: (nit) Maybe a quick comment here to explain what this does? I guess it searches for an entity that has both names, but unsure.

leonhandreke (approved):

PR #2258, opened 2025-05-13
Mappings of AU sanctions  & lookup_sanction_program_key update

nvmbrasserie:


datasets/au/dfat_sanctions/au_dfat_sanctions.yml
```diff
@@ -71,6 +71,40 @@ dates:
...
+        value: AU-CRP
+      - match: "Autonomous (Thematic - Human Rights)"
+        value: AU-HUMAN
+      - match: "Autonomous (Thematic - Cyber)"
+        value: AU-CYB
+      # Only one key in the source data but two different regimes: 
+      # https://www.dfat.gov.au/international-relations/security/sanctions/sanctions-regimes/specified-ukraine-regions-sanctions-regime
+      # https://www.dfat.gov.au/international-relations/security/sanctions/sanctions-regimes/ukraine-sanctions-regime
+      # - match: "Autonomous (Ukraine)"
+      #   value: ["AU-UKRAINE", "AU-UKRAINE-REG"]
```
nvmbrasserie, line 103: I'm not sure how to proceed here

zavod/zavod/helpers/sanctions.py
```diff
@@ -10,12 +10,17 @@
...
-    """Lookup the sanction program key based on the source key."""
+    """Lookup the sanction program key based on the source key, unless it is ignored."""
+    if ignore_source_key and source_key in ignore_source_key:
+        return None
+
     program_key = context.lookup_value("sanction.program", source_key)
     if program_key is None:
-        context.log.warn(f"Program key {program_key!r} not found.")
+        context.log.warn(f"Program key for {source_key!r} not found.")
     return program_key
```
jbothma, line 24: another way to do this might be to include the ignored programs in the `sanction.programs` lookup with value `null`.

If we change this to `context.lookup` instead of `context.lookup_value` we can see that the lookup was None for the warning, but treat a value of `None` as _ignore_

jbothma: Yeah that also works.

The value of doing this via the lookup is to be a bit more explicit about data mapping in the metadata file which is a nice property.

leonhandreke: Ah, that's a good point actually. Yeah, I like your suggestion!

datasets/au/dfat_sanctions/au_dfat_sanctions.yml
jbothma: in the branch as it stands, we'll make all DFAT ukraine sanctions AU-UKRAINE - shall we leave it as a warning instead of picking one until we've got clarification from them?

nvmbrasserie: just updated it, should I merge it now?

zavod/zavod/helpers/sanctions.py
```diff
@@ -13,10 +13,10 @@ def lookup_sanction_program_key(
...
 ) -> Optional[str]:
     """Lookup the sanction program key based on the source key."""
-    program_key = context.lookup_value("sanction.program", source_key)
-    if program_key is None:
-        context.log.warn(f"Program key {program_key!r} not found.")
-    return program_key
+    res = context.lookup("sanction.program", source_key)
+    if res is None:
+        context.log.warn(f"Program key for {source_key!r} not found.")
+    return res.value if res is not None else None
```
jbothma, line 19: I think this is a touch easier to read

```suggestion
    if res is None:
        context.log.warn(f"Program key for {source_key!r} not found.")
        return None
    return res.value
```

jbothma (approved):

PR #2260, opened 2025-05-14
Issues-14-may

nvmbrasserie:


datasets/si/conflicts/crawler.py
```diff
@@ -1,57 +1,91 @@
...
+        country=country,
     )
+    h.copy_address(legal_entity, address)
 
-    context.emit(legal_entity)
-    context.audit_data(record, ignore=["omejitev_do", "st_transakcij"])
-
+    sanction = h.make_sanction(context, legal_entity)
+    h.apply_date(sanction, "startDate", start_date)
+    h.apply_date(sanction, "endDate", end_date)
```
jbothma, line 70: let's not make Sanction entities for these - this dataset should be handled a bit carefully because it's preventative, not punitive. While we use Sanction entities to describe other debarments, we know some users sometimes over-react when they see them and these are merely public officials and SOE employees probably doing their job.

jbothma: Perhaps add a comment in the code explaining exactly this decision for a year from now when we've forgotten

datasets/si/conflicts/crawler.py
```diff
@@ -1,57 +1,91 @@
...
-    legal_entity.add(
-        "program",
-        f"{organization_name} is restricted from procurement from {subject_name} from {start_date} until {end_date} due to ownership or management role of a public official in {organization_name} or their family member. This is a preventative restriction by Komisija za preprečevanje korupcije and implies no wrongdoing.",
+    address = h.make_address(
+        context,
+        street=record.pop("entity_address"),
+        place=record.pop("entity_place"),
+        country=country,
     )
+    h.copy_address(legal_entity, address)
```
jbothma, line 66: I think we can leave the address out for their privacy - I'd say this falls more under the data priorities of PEPs even though this is technically debarment https://zavod.opensanctions.org/best_practices/priorities/#politically-exposed-persons

I think with their name and ID they can be identified within the organisation they work in.

zavod/zavod/shed/fsf.py
```diff
@@ -84,7 +84,8 @@ def parse_sanctions(context: Context, entity: Entity, entry: Element) -> None:
             entry.get("designationDate"),
             regulation.get("entryIntoForceDate"),
         ]
-        start_date = min([d for d in start_dates if d is not None])
+        valid_start_dates = [d for d in start_dates if d is not None]
+        start_date = min(valid_start_dates) if valid_start_dates else None
```
jbothma, line 88: shiiit thanks. I think you can also give min a default
```suggestion
        start_date = min(valid_start_dates, None)
```

datasets/si/conflicts/crawler.py
nvmbrasserie: these are all companies, so these are all companies addresses, should we keep the address in this case?

jbothma: ahhh right yes - in that case I think the address is cool - would be fun if something linked on that.

nvmbrasserie: sorry, not all of them. some are people actually, but acting as a legal entity. I think it's safer to remove the address in this case

jbothma (approved):

PR #2267, opened 2025-05-15
[ca_facfoa] programs mapped

nvmbrasserie:


datasets/ca/dfatd/crawler.py
```diff
@@ -92,7 +92,15 @@ def parse_entry(context: Context, node: _Element):
...
 
-    sanction = h.make_sanction(context, entity)
+    sanction = h.make_sanction(
+        context,
+        entity,
+        program_name=program,
+        source_program_key=program,
+        program_key=(
+            h.lookup_sanction_program_key(context, program) if program else None
+        ),
```
jbothma, line 102: nitpick, but maybe this could be on its own line for a bit better readability

datasets/ca/named_research_orgs/crawler.py
```diff
@@ -65,9 +66,12 @@ def emit_nro(context: Context, nro):
...
-    sanction = h.make_sanction(context, entity)
-    sanction.add(
-        "program", "Policy on Sensitive Technology Research and Affiliations of Concern"
+    sanction = h.make_sanction(
+        context,
+        entity,
+        program_name=PROGRAM_NAME,
+        source_program_key=PROGRAM_NAME,
+        program_key=h.lookup_sanction_program_key(context, PROGRAM_NAME),
     )
```
jbothma, line 75: have we made some kind of standardising decision to use the lookup even when we're hardcoding the source program text? I think there are arguments for/against this, but if we've made a decision on that, it's nice to document it in the zavod docs and share the decision in #data so that we're all on the same page.

I ask because I was about to say this is a bit of a workaround - just hardcode full text and key here, but thought it's kinda cute to always have a sanction.programs lookup if the sanctions are connected to programs.

jbothma (approved):
Nice

datasets/ca/named_research_orgs/crawler.py
nvmbrasserie: not yet, but I'm with you here. I think it's good to have it in the lookups too

leonhandreke: I honestly don't see the point of having a lookup here if there is nothing to map. What's wrong with `program_name=PROGRAM_NAME, program_key=PROGRAM_KEY`?

Also, `source_program_key` doesn't really make sense here, it's another name for `programId` `original_value` and there is none here in the source.

jbothma: the upside is consistently pushing mappings into the metadata.
the downside is indirection.

I don't feel strongly either way so my comment was more that if it's been said otherwise that we want it done this way, we should just be consistent about it

nvmbrasserie: I think it's cleaner to treat them all in the same way and use mappings consistently, but I agree that `source_program_key` doesn't make sense here

PR #2285, opened 2025-05-19
[gb_fcdo_sanctions] and [ua_nsdc_sanctions] sanction programs mapped

nvmbrasserie:


datasets/gb/fcdo_sanctions/crawler.py
```diff
@@ -173,8 +173,20 @@ def crawl(context: Context):
...
+                for regime in designation.iterfind(".//RegimeName")
+                if regime.text
+            ]
             # Make a sanctions entity
-            sanction = h.make_sanction(context, entity)
+            sanction = h.make_sanction(
+                context,
+                entity,
+                program_name=regime_name[0],
+                source_program_key=regime_name[0],
```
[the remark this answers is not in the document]
nvmbrasserie: I think especially here it adds a lot of context. And it's already set as `programId.original_value`, isn't it?

datasets/ua/nsdc_sanctions/crawler.py
```diff
@@ -141,7 +142,13 @@ def crawl_common(
         context, f"/subjects/{subject_id}/actions", cache_days=CACHE_LONG
     ):
         action_id = action.pop("aid")
-        sanction = h.make_sanction(context, entity, key=action_id)
+        sanction = h.make_sanction(
+            context,
+            entity,
+            key=action_id,
+            program_name=PROGRAM_NAME,
```
jbothma, line 149: just noting that we'll end up with two values here - `Law of Ukraine "On Sanctions" No. 1644-VII dated 14 Aug. 2014` and the decree, e.g.

```
jq 'select(.schema == "Sanction") | .properties.program' ~/Downloads/entities.ftm\ \(4\).json |sort|uniq -c|sort -rn
31301 ]
31301 [
3761   "726/2022"
3237   "133/2017"
2547   "176/2018"
2506   "126/2018"
```

I think that's fine though. It's all technically correct

@pudo is a publicationReference prop sufficiently interesting when we already have a sourceUrl to add to FTM?

jbothma (approved):

datasets/gb/fcdo_sanctions/crawler.py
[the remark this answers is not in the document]
leonhandreke: That sounds like a good policy!

datasets/ua/nsdc_sanctions/crawler.py
jbothma: @nvmbrasserie either way this thread isn't a blocker for this PR.

`eu_fsf` does `sanction.add("reason", regulation.get("numberTitle"))`
`ch_seco` seems to just have a section in the act of the program, so nothing like the numbers ua_nsdc has.
`eg_terrorists` does this, where I'm the naughty culprit for making key/value notes
```
    gazette_issue = input_dict.pop("issue_in_official_gazette")
    sanction = h.make_sanction(context, entity, case_number + gazette_issue)
    sanction.add("description", f"Case number: {case_number}")
    sanction.add("description", f"Issue in official gazette: {gazette_issue}")
```
`il_mod_crypto` has an order number which we put into authorityId which is probably wrong because it's the same for all entities/addresses in the order

jbothma: `qa_nctc_sanctions` pops `"designationDTO":{"desigId":"2","designNumber":"1988","legalBasisEn":"قرار مجلس الأمن 1988 (2011)","`

PR #2296, opened 2025-05-20
[ir_sanctions] switched to a google sheet

### LINKED ISSUE #2150: [ir_sanctions] Captcha is blocking access
A captcha is blocking access to the page. Geolocating in Iran through Zyte is no longer enough (or they no longer have exit nodes in Iran?).

I have sent them an email requesting a bypass on 2025-04-17.

nvmbrasserie:
fixes opensanctions/opensanctions#2150

### DIFF
--- a/datasets/ir/sanctions/ir_sanctions.yaml
+++ b/datasets/ir/sanctions/ir_sanctions.yaml
@@ -26,6 +26,12 @@ description: |
   **Note:** Due to CAPTCHA restrictions on the website, the crawler cannot access data in 
   real-time. Instead, it relies on a static Google Sheet that contains manually extracted 
   data, which may not reflect the most current updates.
+tags:
+  - list.sanction.counter
+  - target.us
+  - target.eu
+  - target.gb
+  - target.ca
 publisher:
   name: Iran Ministry of Foreign Affairs
   acronym: MFA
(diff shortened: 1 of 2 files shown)


datasets/ir/sanctions/ir_sanctions.yaml
```diff
@@ -22,6 +22,16 @@ description: |
 
   As of April 2024, the persons sanctioned in this list are generally current and former
   members of the US and UK governments, militaries, as well as journalists.
+
+  **Note:** Due to CAPTCHA restrictions on the website, the crawler cannot access data in 
+  real-time. Instead, it relies on a static Google Sheet that contains manually extracted 
+  data, which may not reflect the most current updates.
```
jbothma, line 28: Is there any plan on maintaining the data manually? Otherwise this could be in the contib directory of the data bucket. Either way, we can skip mentioning the product we're using and just say it's static data.

image



```suggestion
  **Note:** Due to CAPTCHA restrictions on the website, the crawler cannot access data automatically.
  Instead, it relies on a static extract of the
  data, which may not reflect the most current updates.
```

jbothma (commented):
Sorry I missed this earlier - datasets that aren't automatically updated get frequency=never, an end date, and a schedule under deploy so it still gets rerun monthly for FTM updates https://zavod.opensanctions.org/metadata/#data-coverage

jbothma (approved):

PR #2305, opened 2025-05-21
helper to schedule manual check for updates

nvmbrasserie:


leonhandreke:
I like that idea! But how does `last_change` happen? Isn't that only present/usable after export?

Also, I'd prefer something based on `datetime.timedelta`:

```python
h.schedule_alert(entity, after=timedelta(months=1))
h.schedule_alert("2025-10-10")
```

Seomething like that?

pudo:
I have an alternative proposal: What if we made a small GitHub actions script that read the YAML files and created GitHub issues on a regular basis? Like this sort of thing? https://dailystuff.nl/blog/2023/create-github-issue-on-schedule

leonhandreke:
Another data point: `at_meine_abgeordneten` already has such a check: https://github.com/opensanctions/opensanctions/pull/2311

Another implementation idea: don't the assertions kick after the export? Could they support a `max_age` assertion?

jbothma:
looping over datasets https://github.com/opensanctions/opensanctions/blob/main/contrib/check_hierarchy.py#L19

jbothma:
I just remembered, datasets that aren't automatically updated get a coverage end date - if we bump that, we have a date we can use to silence this. Not sure it's strictly needed if the thing runs monthly, but pointing out the option

pudo:
I feel like the coverage end date has been a major source of errors, where people think we haven't updated data that we did, just because we forget to move the value there. if we can come up with a better mechanism that doesn't require adhering to a checklist we all forget, which involves updating this field after updating a google doc....

nvmbrasserie:
superseded by https://github.com/opensanctions/opensanctions/pull/2313, closing this one

PR #2306, opened 2025-05-21
[cn_sanctions] programs mapped

nvmbrasserie:


datasets/cn/sanctions/crawler.py
```diff
@@ -14,6 +14,7 @@ def crawl(context: Context) -> None:
             if name is None:
                 continue
             qid = row.pop("QID", None)
+            program = row.pop("List", None)
```
jbothma, line 17: it's usually better to assign a value in the context where it's used. Since it isn't used early in this function, and it's probably more readable/maintainable if it's assigned just before it's used the first time.

jbothma (approved):

PR #2312, opened 2025-05-22
Make duckdb memory and threads configurable

jbothma:
- make duckdb memory and threads configurable
- document duckdb memory debugging fundamentals
- split the pairs query to make it easier to see where it exceeds memory limits

when deploying this, we'll probably want to set the environment variables for production deployments:

- enrichers
- xref chron job
- workspace


jbothma:
https://claude.ai/share/72e6ab29-5be9-499c-a0c1-391e964ba8cb

PR #2313, opened 2025-05-22
schedule maintenance action

nvmbrasserie:
from https://github.com/opensanctions/opensanctions/pull/2305#issuecomment-2901891968:

#
it sounds like we want something like this for almost all, but not necessarily exactly all datatasets where frequency=never

so how about a metadata field

`manual_updates: never | @monthly | @quarterly`
or simpler

`manual_update_months: null | 1 | 2 ...`
then whatever does the checking can turn that into a days ago or timedelta or whatever.

Never is useful because things that shouldn't see changes like everypolitician won't get fake updates. Other things might.

I don't love fake edits to silence this, and cleaning code changes might silence this even when we haven't checked it because it'd bump last_modified. So I think simply something that runs monthly as a github action and has basic scheduling logic will do the trick.

I think the first step is a list of datasets and what the most sensible frequency is (including explicit nevers).
second step is a little python script that's called by the github action and outputs a list for the template

here's an example of an action calling some python https://github.com/opensanctions/opensanctions/pull/2084/files
here's an example for how a list of datasets from your script can go into the issue template https://stackoverflow.com/questions/59191913/how-do-i-get-the-output-of-a-specific-step-in-github-actions

looping over datasets https://github.com/opensanctions/opensanctions/blob/main/contrib/check_hierarchy.py#L19

### DIFF
--- a/contrib/schedule_maintenance.py
+++ b/contrib/schedule_maintenance.py
@@ -42,7 +42,7 @@ def is_due_for_manual_check(manual_check: dict) -> bool:
     try:
         last_checked = datetime.strptime(last_checked_str, "%Y-%m-%d")
     except ValueError:
-        log.info("Couldn't parse last_checked %s in %s" % (last_checked_str))
+        log.info("Couldn't parse last_checked %s" % (last_checked_str))
         return True  # Malformed date, better to check
 
     return datetime.today() >= (last_checked + timedelta(days=interval_days))


nvmbrasserie:
'frequency: never' list: 
```

ext_ge_company_registry
ext_ua_edr
sy_obsalytics_opensyr
us_special_leg
us_cuba_sanctions
us_nteea_sec_353
us_fcc_covered_list
ua_edr
iadb_sanctions
everypolitician
gem_energy_ownership
un_1718_vessels
cn_sanctions
ru_navalny35
ru_egrul
ru_fedsfm_terror
ru_fedsfm_wmd
ru_billionaires_2021
ru_myrotvorets_wagner
ro_onpcsb_sanctions
ps_ohchr_settlement
ph_most_wanted
si_zvezoskop
ge_company_registry
ge_declarations
at_nbter_sanctions
np_mha_sanctions
de_offeneregister
az_laundromat
id_regional_2018
jp_meti_ru
il_mod_crypto
eu_journal_sanctions
```

nvmbrasserie:
The initial suggestion from @pudo was to make it commit-free. This would prevent us from creating additional manual work. I don't have a strong opinion on that

.github/workflows/schedule_maintenance.yml
```diff
@@ -0,0 +1,64 @@
...
+          python-version: "3.12"
+          cache: "pip"
+
+      - name: Install dependencies
+        run: |
+          pip install ruamel.yaml
+        # If there is any output (non-empty), set should_create_issue to true.
+        # Otherwise, set it to false.
+        # Then export the DATASETS variable to the GitHub Actions environment
+        # so it can be accessed later, when creating the issue.
```
jbothma, line 31:
```suggestion
```

.github/workflows/schedule_maintenance.yml
```diff
@@ -0,0 +1,64 @@
...
+          DATASETS=$(python contrib/schedule_maintenance.py)
+          # the heredoc (<<EOF) syntax is to set a multiline value as env var
+          echo "DATASETS<<EOF" >> $GITHUB_ENV
+          echo "$DATASETS" >> $GITHUB_ENV
+          echo "EOF" >> $GITHUB_ENV
+
+      - name: Commit and push changes
+        run: |
+          git config --global user.name 'github-actions[bot]'
+          git config --global user.email 'github-actions[bot]@users.noreply.github.com'
```
jbothma, line 44: I think the numeric ID is supposed to be part of the email address https://github.com/actions/checkout?tab=readme-ov-file#Push-a-commit-using-the-built-in-token

jbothma (commented):
Can ruamel maintain wrapping and explicit nulls?

```diff
index d3734997f..a4330426d 100644
--- a/datasets/_global/gem_energy_ownership/gem_energy_ownership.yml
+++ b/datasets/_global/gem_energy_ownership/gem_energy_ownership.yml
@@ -5,14 +5,15 @@ coverage:
   frequency: never
   start: 2024-09-23
   end: 2024-06-30
+manual_check:
+  last_checked: "2025-05-29"
 deploy:
   schedule: "@monthly"
 exports:
   - statistics.json
 load_statements: true
 summary: >
-  Information on the chain of ownership for various energy projects. The data maps
-  each level of the chain from the direct owner up to their highest-level ultimate parents.
+  Information on the chain of ownership for various energy projects. The data maps  each level of the chain from the direct owner up to their highest-level ultimate parents.
 description: |
   The Global Energy Ownership Tracker provides information on the chain of ownership for
   various energy projects. The data maps each level of the chain from the direct owner
@@ -181,7 +182,7 @@ lookups:
           - Borås Energy and Environment
           - nan
           - nan,nan
-        value: null
+        value:
   type.identifier:
     options:
       - match:
@@ -231,4 +232,4 @@ lookups:
           - 3442123614
         prop: registrationNumber
       - match: not found
-        value: null
+        value:
```

Also, are you sure it's supposed to auto-bump the dates on issue creation, and not wait until the issue is closed? No idea how it'd check if the issue is closed, just checking :)

pudo:
I have an annoying ask: can this generate an issue per dataset, rather than a list of all? I know there's some sort of fan-out in github actions which may be able to do this. The concern I have otherwise is that we end up with all these long-open issues that are about a month, when we can probably close may of them quickly if they're just about one of the datasets....

jbothma:
ruaml issues:

- explicit nulls replaced with blank
- wrapped multiline text got unwrapped

way forward:

- creative patching solution by Ivan
- make PRs rather than commit/push directly to main - we have a chance to check

.github/workflows/schedule_maintenance.yml
```diff
@@ -0,0 +1,69 @@
+name: schedule_maintenance
+
+on:
+  push:
+    branches:
+      - github-action-datasets-check
+      - main
```
jbothma: let's remove this - I don't think it should run on every push to main

.github/workflows/schedule_maintenance.yml
```diff
@@ -0,0 +1,69 @@
...
+          git config --global user.email '41898282+github-actions[bot]@users.noreply.github.com'
+  
+      - name: Create commit and PR
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          BRANCH_NAME="auto/update-dataset-metadata-$(date +%Y%m%d%H%M%S)"
+          git checkout -b $BRANCH_NAME
+
+          git add datasets/*.yml
```
jbothma: you might need a backslash before the * 
https://git-scm.com/docs/git-add#_examples

.github/workflows/schedule_maintenance.yml
```diff
@@ -0,0 +1,69 @@
...
+          if git diff --cached --quiet; then
+            echo "No changes to commit. Skipping PR."
+            exit 0
+          fi
+
+          git commit -m "chore: update last_checked dates in the metadata files"
+          git push origin $BRANCH_NAME
+
+          gh pr create \
+            --title "[Auto] Update dataset metadata" \
```
jbothma:
```suggestion
            --title "[Auto] Update manually-checked sources" \
```

contrib/schedule_maintenance.py
```diff
@@ -0,0 +1,104 @@
+import sys
+import ruamel.yaml
```
jbothma: if we're not using ruamel for writing any more, can we just use the standard yaml library?

contrib/schedule_maintenance.py
```diff
@@ -0,0 +1,104 @@
...
+    today = datetime.today().strftime("%Y-%m-%d")
+    with path.open("r", encoding="utf-8") as f:
+        lines = f.readlines()
+    lines = [
+        (
+            f'  last_checked: "{today}"\n'
+            if line.lstrip().startswith("last_checked:") and line.startswith("  ")
+            else line
+        )
+        for line in lines
```
jbothma: what do you think of re.sub instead to we're not indentation-dependent?

contrib/schedule_maintenance.py
```diff
@@ -0,0 +1,104 @@
...
+    last_checked_str = manual_check.get("last_checked")
+    interval_days = manual_check.get("interval", 90)
+
+    if not last_checked_str:
+        return False  # No last check date — skip
+
+    try:
+        last_checked = datetime.strptime(last_checked_str, "%Y-%m-%d")
+    except ValueError:
+        return True  # Malformed date, better to check
```
jbothma: you should basically always log when you catch an exception, even if it's only info level. Otherwise it can take painful debugging to figure out why something is behaving in an unexpected way
```suggestion
    except ValueError:
        log.info("Couldn't parse last_checked %s in %s" % (last_checked_str, path))
        return True  # Malformed date, better to check
```

.github/workflows/schedule_maintenance.yml
```diff
@@ -0,0 +1,69 @@
...
+  push:
+    branches:
+      - github-action-datasets-check
+      - main
+  schedule:
+    - cron: '0 9 1 * *'  # Every month on the first day at 09:00 UTC
+
+jobs:
+  create-pr:
+    name: Create a PR
```
jbothma:
```suggestion
  check-schedule:
    name: Check for scheduled updates
```

.github/workflows/schedule_maintenance.yml
nvmbrasserie: There is also a case without expanding: 

> Because this example lets the shell expand the asterisk (i.e. you are listing the files explicitly), it does not consider subdir/git-foo.sh.

contrib/schedule_maintenance.py
```diff
@@ -0,0 +1,98 @@
...
+    last_checked_str = manual_check.get("last_checked")
+    interval_days = manual_check.get("interval")
+
+    if not last_checked_str:
+        return False  # No last check date — skip
+
+    try:
+        last_checked = datetime.strptime(last_checked_str, "%Y-%m-%d")
+    except ValueError:
+        log.info("Couldn't parse last_checked %s in %s" % (last_checked_str))
```
jbothma:
```suggestion
        log.info("Couldn't parse last_checked %s" % (last_checked_str))
```

jbothma (approved):
wooo ship it

PR #2315, opened 2025-05-23
[ua_nsdc_sanctions] Revamp name field assignments

leonhandreke:
- "alias" contains Russian an English versions (that are not just direct
  transliterations of the Ukrainian) which are better for captions.
- "translit" is just a direct transliteration of the Ukrainian. Demote
  that to alias (or should we kick it out?)

### DIFF
--- a/datasets/ua/nsdc_sanctions/crawler.py
+++ b/datasets/ua/nsdc_sanctions/crawler.py
@@ -65,12 +65,16 @@ def crawl_common(
         entity.add("name", match.groups())
     else:
         entity.add("name", name, lang="ukr")
+    # Aliases contains a name in Russian and English which is high-enough value for us to put it in the name field
+    # (and allows us to have a good latin caption)
+    entity.add("name", item.pop("aliases"))
 
+    # The transliteration is just a literal transliteration of the Ukrainian name
     name_translit = item.pop("translit")
     if match := REGEX_NAME_3_PARTS.match(name_translit):
-        entity.add("name", match.groups())
+        entity.add("alias", match.groups())
     else:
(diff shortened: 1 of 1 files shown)


leonhandreke:
@nvmbrasserie maybe you can take a look at a few entities (maybe both a company and a person) and also weigh in with your knowledge on how Cyrillic names are usually translated to sanity check this? Thanks :)

jbothma:
Are we sure this is generally an improvement?

Here we're losing the last name from the caption:

```
{
  "op": "MOD",
  "entity": {
    "id": "NK-2JgPTFBX7pHzbPhDcMPSpP",
    "caption": "Zinko Viktor",
    "schema": "Person",
    "properties": {
      "name": [
        "Зінько Віктор Миколайович",
        "Зинько Виктор Николаевич",
        "Zinko Viktor"
      ],
      "alias": [
        "Zinko Viktor Mykolaiovych"
      ],
```

Seems to be a few of those

nvmbrasserie:
I think it's a nice improvement. In the name `Zinko Viktor` the first part is the surname and the second part is the first name. `Mykolaiovych` is a patronymic. This way, the main entity name will be just the name and surname, and the transliterated Ukrainian version will be an alias.

nvmbrasserie (approved):

jbothma (approved):
local knowledge is good!

jbothma:
transliteration in alias currently has no impact on matching in logic-v1. All name features use all name-type props. name and alias are both 'name' type.

But yeah let's keep the transliterated - it might still be useful, and I think their transliteration is better than our python-based transliteration.

PR #2337, opened 2025-05-30
uk fca firds

nvmbrasserie:


pudo:
Yo! I'm super impressed that you refactored this into a shed module. Very near. Perhaps we could also add `ext_gb_fca_firds` already?

nvmbrasserie:
sure, I'm on it

datasets/_externals/ext_gb_fca_firds.yml
```diff
@@ -0,0 +1,83 @@
...
+summary: >-
+  List of financial instruments and their reference data, as published by the
+  Financial Conduct Authority (FCA) in the UK.
+description: |
+  This is a dataset of financial instruments and their reference data, as published
+  by the Financial Conduct Authority (FCA) in the UK. The data establishes
+  connections between financial instruments and their issuers, and includes
+  information such as the instrument's name, ISIN, and LEI.
+url: https://data.fca.org.uk/#/download
+publisher:
```
jbothma, line 29: maybe nice to duplicate the related dataset data section for now to make it easy to find. I see ext_eu_esma_firds does that.

datasets/_externals/ext_gb_fca_firds.yml
```diff
@@ -0,0 +1,83 @@
+type: external
+title: UK Financial Instruments Reference Data System (FIRDS)
+entry_point: zavod.runner.local_enricher:enrich
+full_dataset: gb_fca_firds
+disabled: false
+hidden: false
```
jbothma: I think we can leave these out normally and only add them when we're setting them to something other than defaut . That way most crawlers have slimmer metadata and less noise

datasets/_externals/ext_gb_fca_firds.yml
```diff
@@ -0,0 +1,83 @@
...
+  - ext_gb_fca_firds
+  - ext_gleif
+  - ext_icij_offshoreleaks
+  - ext_lv_business_register
+  - ext_md_companies
+  - ext_ru_egrul
+  - ext_ua_edr
+  - opencorporates
+  - sanctions
+  - special_interest
```
jbothma: I see ext_eu_esma_firds has securities and regulatory too, which makes sense for financial instruments - perhaps worth adding here too

datasets/_externals/ext_gb_fca_firds.yml
```diff
@@ -0,0 +1,83 @@
...
+
+config:
+  limit: 100
+  max_bin: 14
+  index_options:
+    max_candidates: 200
+    match_batch: 300
```
jbothma: let's try it without index_options, limit and max_bin for now, and tweak from defaults if needed

datasets/_externals/ext_gb_fca_firds.yml
```diff
@@ -0,0 +1,83 @@
...
+    - sanction.linked
+    - debarment
+    - asset.frozen
+    - poi
+    - gov.soe
+  schemata:
+    - Company
+    - Organization
+    - LegalEntity
+    - Person
```
jbothma: it looks like we can skip Person, if the target dataset won't contain any

datasets/gb/fca_firds/crawler.py
```diff
@@ -0,0 +1,30 @@
+from zavod import Context
+from zavod.shed.firds import parse_xml_file
+
+
+def get_full_dumps_index(context: Context):
+    page = 0
+    size = 100
+    total = None
+    while True:
+        url = f"{context.data_url}?q=file_type:FULINS&from={page * size}&size={size}&pretty=true"
```
jbothma: at some point it's nice to make the querystring with urlencode
https://github.com/opensanctions/opensanctions/blob/uk-fca-firds/datasets/_wikidata/categories/crawler.py#L172
Perhaps the point where we're tempating two values and doing maths in the template string is that point

jbothma (commented):
Very nice - just some small tweaks needed I think

Are there any changes to the stuff moved to shed? It doesn't look like it but I can't seem to run eu_esma_firds right now to check that it's still working as expected so just want to be sure.

PR #2339, opened 2025-05-30
[paris_mou] initial commit

nvmbrasserie:


datasets/_global/paris_mou_banned/paris_mou_banned.yml
```diff
@@ -0,0 +1,52 @@
+title: Paris MoU Banned List
+entry_point: crawler.py
+prefix: mou-ban-list
```
jbothma: let's make this more concise, and I think mou is a very generic term so getting paris in there (even shortened) helps.

```suggestion
prefix: par-mou
```

datasets/_global/paris_mou_banned/paris_mou_banned.yml
```diff
@@ -0,0 +1,52 @@
...
+coverage:
+  frequency: weekly
```
jbothma: feels like something we can run daily, or is there a reason for running it less often?

datasets/_global/paris_mou_banned/crawler.py
```diff
@@ -0,0 +1,99 @@
...
+
+from zavod import Context, helpers as h
+
+# Full URL for the Paris MoU Banned Ships
+# https://portal.emsa.europa.eu/o/portlet-public/rest/ban/getBanShips.json?banSituationIds=VALIDATED&page=1&start=0&limit=200
+
+
+def clean(value: Optional[str]) -> Optional[str]:
+    if not isinstance(value, str):
+        return None
```
jbothma: what's the chance that this is given something interesting like an array of strings, or something? then we turn all of that to None.
What do you think of using normality.stringify on this instead?

datasets/_global/paris_mou_banned/crawler.py
```diff
@@ -0,0 +1,99 @@
...
+    sanction.add("authority", authority.get("description"))
+
+    context.emit(vessel)
+    context.emit(sanction)
+
+    context.audit_data(
+        item,
+        [
+            "banningAuthority",
+            "minimumDuration",
```
jbothma, line 66: how about we put this in Sanction:duration?

datasets/_global/paris_mou_banned/crawler.py
```diff
@@ -0,0 +1,100 @@
...
+            city=clean(ism.get("city")),
+            country=clean(ism.get("countryDescription")),
+        )
+        h.copy_address(company, address)
+        context.emit(company)
+        vessel.add("owner", company.id)
+
+    ban_status = item.get("banOrderStatus", {}).get("active")
+    if ban_status:
+        vessel.add("topics", "sanction")
```
jbothma:
```suggestion
        vessel.add("topics", "reg.warn")
```

PR #2340, opened 2025-05-31
Cache graph context in entity exporters

pudo:
This seems to cache 20% off the export time by avoiding multiple queries to the store for retrieving the same adjacent entities.


jbothma (approved):
Very cool. What i'd expect to be there, is there.
Cached nested inverted iteration is... good mind gym. But it makes sense.
Seems to produce consistent results on eu_journal_sanctions which has two types of relations

pudo:
pre-merge default export time: 

start: 8:54
store complete: 10:35
export complete: 12:28
export time: 115 min

Issue #582, opened 2024-02-19
Complex CSV exporter


pudo:
Columns:

* EntityID 
* Featured Properties
* Sources
* Temporal Extent
* Ignore? Address

Keep non-matchable fields as an array?
Persons

jbothma:
Only sharing because the name is amazing in british accents, but https://docs.flatterer.dev/
By documentation alone, flatten-tool looks more complete

jbothma:
jd can add notes from recent customer's tool

leonhandreke:
We talked about this today in the Engineering Meeting.

Where to do this
1. Delivery Service
    a) Streamed
    b) "Pizza-style" (request with a ticket that gets processed in the background). Some thoughts around that are in https://github.com/opensanctions/operations/issues/1420
2. zavod/etl with publishing to a bucket with quick expiry

We are now leaning towards keeping it out of Delivery service, in a top-level directory in operations instead.

- Default could be to have metadata at the top of the CSV as a convencience feature
- What bucket? data.opensanctions.org is fine for now
- index.json that looks a lot like our index.json. Should contain last updated, last updated source data
- Served by Delivery Service.
- Should generate from `entities.nested.json`
- Ideally, our system should be expressive enough to generate our securities and maritime exports

PR #2354, opened 2025-06-03
Play with a JMESPath-based JSON to CSV mapper

pudo:
Refs #582


pudo:
Merging this because there's no real reason it shouldn't be merged. Probably going to eventually end up in the delivery system.

PR #2355, opened 2025-06-03
Tokyo MoU PSC Database

nvmbrasserie:
opensanctions/crawlers#328

Potential sources to expand the coverage: 
https://bsmou.org/database/inspections/
https://abujamou.org/en/inspection-database/


datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,141 @@
...
+MAX_RETRIES = 5
+RETRY_DELAY = 5  # seconds
+
+
+def solve_arithmetic(expression: str) -> str:
+    match = re.search(r"(\d+)\s*([+\-*/])\s*(\d+)", expression)
+    if not match:
+        raise ValueError(f"Invalid CAPTCHA expression: {expression}")
+    a, op, b = match.groups()
+    return str(eval(f"{a}{op}{b}"))
```
jbothma: Let's add a comment to make it super super clear how dangerous this can be.

Can we make the regex more strict by matching start/end with ^ and $?

```suggestion
    # WARNING!! WARNING!!
    # This is validating external input to eval() - review EXTREMELY carefully.
    # Match an expression like 123 + 456
    match = re.search(r"^(\d+)\s*([+\-*/])\s*(\d+)$", expression)
    if not match:
        raise ValueError(f"Invalid CAPTCHA expression: {expression}")
    a, op, b = match.groups()
    return str(eval(f"{a}{op}{b}"))
```

datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def crawl(context: Context):
+    with httpx.Client(http2=False, follow_redirects=True) as client:
```
jbothma: any reason why you need to use httpx rather than the usual context.fetch_* or if they don't work, the requests session via `context.http.*`?

jbothma (commented):
so eval() with a response from another site is pretty scary, but the regex seems to do the right sort of validation. If we go down this road we need to make it incredibly clear that this is a danger area.

The only other approach that comes to mind right away would be some maths library that does the parsing, but it feels like that's a much bigger attack surface area than very strictly validated input as done here.

datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,141 @@
...
+            "type": "0",
+            "result": "0",
+            "insptype": "-1",
+            "sort1": "0",
+            "sort2": "DESC",
+            "sort3": "0",
+            "sort4": "DESC",
+        }
+
+        for attempt in range(MAX_RETRIES):
```
leonhandreke: Can we get away with reconfiguring the standard retry strategy here? We have a retry methods thing in the YAML (the more I think about it, the more I feel the YAML is not the right place to put this, but whatever).

If not, can we configure a `httpx.RetryStrategy` instead of doing it manually? Or is this a special case?

datasets/_global/tokyo_mou/crawler.py
leonhandreke: I feel like an `if op == "+" ... elif op == "-"` wouldn't hurt that much and make it feel a bit safer :)

nvmbrasserie:
Potential sources to expand the coverage: https://bsmou.org/database/inspections/

datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,207 @@
...
+        "captcha": answer,
+    }
+    login_url = "https://apcis.tmou.org/public/?action=login"
+    login_resp = context.http.post(login_url, data=login_data, headers=HEADERS)
+    print("Login status:", login_resp.status_code)
+
+    # Step 2: Load PSC inspection page
+    # First, fetch page 0 to get total_pages
+    search_data = make_search_data(0)
+    search_resp = context.http.post(
```
jbothma: nitpick: this can be context.fetch_html, can't it?

datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,207 @@
...
+        "password": "mgrsicpa11",
+        "captcha": answer,
+    }
+    login_url = "https://apcis.tmou.org/public/?action=login"
+    login_resp = context.http.post(login_url, data=login_data, headers=HEADERS)
+    print("Login status:", login_resp.status_code)
+
+    # Step 2: Load PSC inspection page
+    # First, fetch page 0 to get total_pages
+    search_data = make_search_data(0)
```
jbothma: Another nitpick: how about 

```suggestion
    search_req = make_search_req(0)
```

as parallel naming to `search_resp`

The client calls it data because to the client it's just data and request_body is kinda long. But in the context of our crawler, naming a variable `data` can probably be made a little clearer

datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,207 @@
...
+        "UID": f"{shipuid}",
+        "initiator": "insp",
+    }
+
+    # Step 4: POST to get full ship profile using shipuid
+    detail_resp = context.http.post(
+        "https://apcis.tmou.org/public/?action=getshipinsp",
+        data=detail_data,
+        headers=HEADERS,
+    )
```
jbothma: with context.fetch_html we can cache the detail pages, which is probably useful with such a big crawl

datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,207 @@
...
+    detail_resp.raise_for_status()
+    tree = html.fromstring(detail_resp.text)
+    tables = tree.xpath("//table[@class='table']")
+    assert len(tables) >= 3, "Expected at least 3 tables in the response"
+    ship_data = tables[1]
```
jbothma: can we select these with something like `//h2[contains(text(), "Ship Data")]/following-sibling::table[1]`?

datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,207 @@
...
+}
+SEARCH_DATA = {
+    "Param": "0",
+    "callsign": "",
+    "name": "",
+    "compimo": "",
+    "compname": "",
+    "From": f"{TODAY}",
+    # Go back ~6 months (approximate as 182 days)
+    "Till": f"{TODAY - timedelta(days=182)}",
```
jbothma: shall we format these like this?

```diff
-    "From": f"{TODAY}",
+    # Format dates as DD.MM.YYYY
+    "From": TODAY.strftime("%d.%m.%Y"),
     # Go back ~6 months (approximate as 182 days)
-    "Till": f"{TODAY - timedelta(days=182)}",
+    "Till": (TODAY - timedelta(days=30)).strftime("%d.%m.%Y"),
```

datasets/_global/tokyo_mou/tokyo_mou_psc.yml
```diff
@@ -0,0 +1,50 @@
...
+url: https://www.tokyo-mou.org/inspections-detentions/psc-database/
+data:
+  url: https://www.tokyo-mou.org/inspections-detentions/psc-database/
+  format: HTML
+  lang: eng
+http:
+  retry_statuses: [500]
+  total_retries: 5
+  backoff_factor: 2
+  retry_methods: POST
```
jbothma:
```suggestion
  retry_methods: [POST, GET]
```

datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,207 @@
...
+    print("Login status:", login_resp.status_code)
+
+    # Step 2: Load PSC inspection page
+    # First, fetch page 0 to get total_pages
+    search_data = make_search_data(0)
+    search_resp = context.http.post(
+        "https://apcis.tmou.org/public/?action=getinspections",
+        data=search_data,
+        headers=HEADERS,
+    )
```
jbothma: this can simply be done in the loop if crawl_page returns total_pages and the loop looks like

```python
total_pages = None
page = 0
while total_pages is None or page <= total_pages:
    total_pages = crawl_page
```

then we don't have this additional fetch code and page 0 special case in crawl_page.

datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,207 @@
...
+    answer = solve_arithmetic(question)
+
+    login_data = {
+        "user": "apcismgr",
+        "password": "mgrsicpa11",
+        "captcha": answer,
+    }
+    login_url = "https://apcis.tmou.org/public/?action=login"
+    login_resp = context.http.post(login_url, data=login_data, headers=HEADERS)
+    print("Login status:", login_resp.status_code)
```
jbothma: this could just be a context.fetch_html(method=POST), right?

datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,207 @@
...
+    tree = html.fromstring(login_page.text)
+    # Solve the arithmetic CAPTCHA
+    question = tree.xpath("string(//span[contains(text(), '=')])").strip(" =")
+    answer = solve_arithmetic(question)
+
+    login_data = {
+        "user": "apcismgr",
+        "password": "mgrsicpa11",
+        "captcha": answer,
+    }
```
jbothma: what login details are these? It doesn't prompt me for any. Are they publicly available?

datasets/_global/tokyo_mou/crawler.py
jbothma: and maybe put the past date in `from` :)

datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,200 @@
...
+    a, b = int(a), int(b)
+    if op == "+":
+        return str(a + b)
+    if op == "-":
+        return str(a - b)
+    raise ValueError(f"Unknown op: {op}")
+
+
+def make_search_data(page):
+    data = dict(SEARCH_DATA)
```
leonhandreke: (nit) totally optional, but this could be a
```python
return {**SEARCH_DATA, "Page": str(page)}
```

Does exactly the same thing though, so who cares, it just looks different :P

datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,200 @@
...
+    imo = str_row.pop("imo_number")
+    vessel = context.make("Vessel")
+    vessel.id = context.make_id(ship_name, imo)
+    vessel.add("name", ship_name)
+    vessel.add("imoNumber", imo)
+    vessel.add("type", str_row.pop("type"))
+    vessel.add("callSign", str_row.pop("callsign"))
+    vessel.add("mmsi", str_row.pop("mmsi"))
+    vessel.add("grossRegisteredTonnage", str_row.pop("tonnage"))
+    # TODO: map the 'deadweight' once we have a property for it
```
leonhandreke, line 89: Is there an issue link we can put in the code here so that a future person can pick this up?

datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,200 @@
...
+    # TODO: add the topic (most likely 'mar.detained') once we have it
```
leonhandreke: Same here, is there an issue for this?

datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,200 @@
...
+    company.add("email", str_row.pop("email"))
+    company.add("phone", str_row.pop("phone"))
+    context.emit(company)
+
+    emit_unknown_link(context, vessel_id, company.id, None)
+    context.audit_data(str_row, ["fax"])
+
+
+def crawl_vessel(context: Context, shipuid: str):
+    print(f"Processing shipuid: {shipuid}")
```
leonhandreke: (nit) context.log.debug?

datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,200 @@
...
+        "https://apcis.tmou.org/public/?action=getinspections",
+        data=make_search_data(page),
+        headers=HEADERS,
+        method="POST",
+    )
+    # Parse the response to find shipuids
+    shipuids = doc.xpath(
+        "///tr[contains(@class, 'even') or contains(@class, 'odd')]//input[@type='hidden']/@value"
+    )
+    print(f"Found {len(shipuids)} shipuids in the search response")
```
leonhandreke: (nit) Does this want to be a context.log?

datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,200 @@
...
+    if len(shipuids) < 15:
```
leonhandreke: You might have already run this through locally, but what happens on the last page? Can this trigger there?

datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,200 @@
...
+    company.id = context.make_slug(company_name, company_imo)
+    company.add("name", company_name)
+    company.add("imoNumber", company_imo)
+    company.add("mainCountry", str_row.pop("registered"))
+    company.add("jurisdiction", str_row.pop("residence"))
+    company.add("email", str_row.pop("email"))
+    company.add("phone", str_row.pop("phone"))
+    context.emit(company)
+
+    emit_unknown_link(context, vessel_id, company.id, None)
```
leonhandreke: Bit of a structure elegance nit, but this function could also return a `company.id` and then the link emitting could happen at the call site. I think this slices the concerns more neatly and has more symmetry :)

leonhandreke: Can we say something about the link here? Operator, Owner, something like that? Also totally fine if not, just wanted to check :)

datasets/_global/tokyo_mou/tokyo_mou_psc.yml
```diff
@@ -0,0 +1,50 @@
...
+  start: "2025-06-26"
+load_statements: true
+summary: >
+  Records detailing inspections of ships by member Authorities under the Tokyo MoU.
+description: |
+  The dataset contains inspection records for ships visiting ports under the jurisdiction 
+  of Tokyo MoU member Authorities. Each record includes, but is not limited to, the ship’s 
+  identity, flag, and associated company.
+
+  **Note:** This dataset is updated regularly and typically contains records from the last 6 months.
```
leonhandreke, line 15: Does this mean that the records also drop away after six months? I think this is the relevant information here -- the dataset updating regularly is sort of the default for us and also shown in the `coverage` metadatum above.

datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,200 @@
...
+    print(f"Processed page {page}, total pages: {total_pages}")
+    return total_pages
+
+
+def crawl(context: Context):
+    # Submit login form
+    login_page = context.fetch_html("https://apcis.tmou.org/public/")
+    # Solve the arithmetic CAPTCHA
+    question = login_page.xpath("string(//span[contains(text(), '=')])").strip(" =")
+    answer = solve_arithmetic(question)
```
leonhandreke, line 190: 💖 love the vibes of this source :)

datasets/_global/tokyo_mou/crawler.py
nvmbrasserie: let's see in prod :)  It depends on the overall count, so it's hard to predict. I'll bump it up then if it's not happy

datasets/_global/tokyo_mou/crawler.py
nvmbrasserie: no, not really, the link is not specified

nvmbrasserie: I feel that it's been an established pattern for some time now:
- https://github.com/opensanctions/opensanctions/blob/main/datasets/ua/war_sanctions/crawler.py#L406C5-L406C18
- https://github.com/opensanctions/opensanctions/blob/main/datasets/sg/mas_investor_alert/crawler.py#L52

datasets/_global/tokyo_mou/crawler.py
jbothma: sometimes this will succeed, but sometimes this will fail, depending on number of results modulo 25, right? I can make it fail today 

image

Is the goal here just to see if we're finding shipuids? Doesn't the minimum entities assertion do the same job?

Perhaps we can assert that we've found at least 1 if your concern is that their "security features" might resut in successfully finding the table but leave it empty? Maybe on this page we can assert that some other content is present like `INSPECTIONS SEARCH RESULTS` or the results count (regardless of the numbers - its presence means results exist

image

datasets/_global/tokyo_mou/crawler.py
jbothma: wil we add this topic? does this data source tell us that it's detained?

datasets/_global/tokyo_mou/crawler.py
leonhandreke: What I'm noting is not that there is a helper function to build the link, it's about where and how it's called. I think your two examples are quite different also.

In `mas_investor_alert`, we first build then entities and then call `emit_relationship(entity, other_entities)` (or a variation thereof)

I think `emit_relation` in `ua_war_sanctions` is pretty confusing. It takes a few strings that are used to build a LegalEntity, there is a weird way `from_prop` and `to_prop` to reverse the whole thing (which wouldn't be needed at all if you could just pass in two entities) and in the end it returns the LegalEntity, which totally isn't obvious from the name `emit_entity`. But that's a different issue.

I feel here the most readable thing would be, similar to what's going on in `mas_investor_alert`:

```python
def crawl_vessel(...):
    ...
    company = crawl_company_details(...)
    emit_unknown_link(context, from=vessel, to=company, role="Classification society")
```

datasets/_global/tokyo_mou/crawler.py
```diff
@@ -0,0 +1,199 @@
...
+    company_data = detail_doc.xpath(
+        "//h2[text()='Company details']/following-sibling::table[1]"
+    )
+    assert len(company_data) == 1, "Expected exactly one company data table"
+    for row in h.parse_html_table(company_data[0]):
+        str_row = h.cells_to_str(row)
+        crawl_company_details(context, str_row, vessel_id)
+
+
+def crawl_page(context: Context, page: int):
```
jbothma: I think I didn't submit this comment previously
```suggestion
def crawl_list_page(context: Context, page: int):
```

datasets/_global/tokyo_mou/crawler.py
jbothma: so much nicer

jbothma (commented):
I think this is really close - well done

jbothma (commented):
what do you think of putting this in `datasets/_global/tokyo_mou/psc/tokyo_mou_psc.yml` and putting the detentions in `datasets/_global/tokyo_mou/detention/tokyo_mou_detention.yml`?

datasets/_global/tokyo_mou/crawler.py
jbothma: the two returns in crawl_ship_data is a bit confusing, and a couple of naming things make it a bit hard to follow. What do you thin of

- renaming crawl_vessel to crawl_vessel_page
- renaming crawl_ship_data to crawl_vessel_row
- assert that there's exactly one vessel row and don't call it in a loop - we're using the last-assigned vessel id so this is an assumption we're making
- if we don't want to call emit_unknown_link from crawl_ship_data, let's make the returns a bit clearer. I think one option is
  - assign a default None to class_soc_id
  - if it exists, assign the id
  - return vessel_id, class_soc_id at the end

However, I think the pattern is usually that crawl_ functions kinda deal with their everything they've crawled as far as possible, and only return things if they're needed for things crawled elsewhere. It's not a hard rule and it's good to discuss where change is needed.

leonhandreke (approved):

datasets/_global/tokyo_mou/crawler.py
leonhandreke: Just talked about this offline.

I think `crawl_ship_data` should emit the classification society & the link inside. The caller doesn't have to care.

`crawl_company_details` should not take a vessel as an argument. The caller has to get involved anyway, so it's more obvious to do the linking at the caller.

Sorry, I didn't realize there were two uses of `emit_unknown_link` and therefore wrote a dumb comment above. I just stumbled upon the function signature of `crawl_company_data`, like, why does it take a vessel? Now it all makes sense :) thanks for bearing with me!

jbothma (approved):
🚢

PR #2358, opened 2025-06-04
[tokyo_mou_detention] initial commit

nvmbrasserie:


datasets/_global/tokyo_mou_detention/crawler.py
```diff
@@ -0,0 +1,94 @@
+from datetime import datetime
+
+from zavod import Context, helpers as h
+
+YEARS = [2024, 2025]
+MONTHS = range(1, 13)
+
```
nvmbrasserie: The data is available starting from 2001, but that may be more than we need

datasets/_global/tokyo_mou_detention/crawler.py
```diff
@@ -0,0 +1,94 @@
...
+    vessel = context.make("Vessel")
+    vessel.id = context.make_id(ship_name, imo)
+    vessel.add("name", ship_name)
+    vessel.add("imoNumber", imo)
+    vessel.add("flag", clean_row.pop("ship_flag"))
+    vessel.add("buildDate", clean_row.pop("year_of_build"))
+    vessel.add("grossRegisteredTonnage", clean_row.pop("gross_tonnage"))
+    vessel.add("type", clean_row.pop("ship_type"))
+    vessel.add("topics", "reg.warn")
```
nvmbrasserie: re: `topics` Should we apply this to all ships, or just the ones that are still detained?

datasets/_global/tokyo_mou_detention/crawler.py
jbothma: can you make this try up to the current year minus one or two so we don't have to remember and come and edit this?

datasets/_global/tokyo_mou_detention/tokyo_mou_detention.yml
```diff
@@ -0,0 +1,66 @@
+title: Tokyo MoU Detention List
+entry_point: crawler.py
+prefix: tok-mou-det
+coverage:
+  frequency: monthly
```
jbothma: let's run weekly so we're not up to 1 month out of date with them. https://en.wikipedia.org/wiki/Nyquist%E2%80%93Shannon_sampling_theorem

datasets/_global/tokyo_mou_detention/crawler.py
jbothma: just detained ones - if using release date, the pattern would be

```python
if h.is_active(sanction)
    vessel.add("topic", topic)
```

datasets/_global/tokyo_mou_detention/crawler.py
```diff
@@ -0,0 +1,94 @@
...
+        link.add("object", vessel_id)
+        context.emit(link)
+
+
+def crawl_row(context: Context, clean_row: dict):
+    ship_name = clean_row.pop("ship_name")
+    imo = clean_row.pop("imo_no")
+    company_name = clean_row.pop("company")
+    ros = clean_row.pop("related_ros")
+    cs = clean_row.pop("classification_society")
```
jbothma: declare variables where you use them. It makes it easier to understand code, and reduces the chance of bugs.

In some really old languages it's either required or historically preferred to declare variables at the top. But with dynamic languages, we get to declare and assign variables close to where they're used, which means we don't have to scan up and down the page to figure out what's what. it makes code review easier and maintenance less buggy.

jbothma: try and name variables so that someone reading it can have a fairly good guess what it refers to. I need a reference with a better author, but this works for now https://gist.github.com/klaaspieter/12cd68f54bb71a3940eae5cdd4ea1764

```suggestion
    related_ros = clean_row.pop("related_ros")
    class_soc = clean_row.pop("classification_society")
```

nvmbrasserie: I know we've discussed it several times recently, but I thought it would make the code cleaner because I like the pattern of putting everything at the top and then only mapping the variables with `entity.add.` I also assumed that this function was simple and short enough to stick to this pattern. However, that's not the issue. I'll move them to where they are used.

jbothma: Thanks. Yeah - when most of your time in spent reading code, not writing code, I think the scale tips more obviosly toward code locality.

jbothma: I think in this case the fact that the variable names were acronyms added to the layers to be kept in mind before it meant something. When you can make something mean something without any mind "hops", it's ideal.

jbothma (approved):

PR #2369, opened 2025-06-06
Black Sea MoU Detention List

nvmbrasserie:


datasets/_global/black_sea_mou_detention/crawler.py
```diff
@@ -0,0 +1,123 @@
...
+            )
+            table = doc.xpath("//table[@id='dvData']")
+            assert len(table) == 1, "Expected one table in the document"
+            table = table[0]
+            for row in h.parse_html_table(table):  # , header_tag="th", skiprows=1):
+                str_row = h.cells_to_str(row)
+                clean_row = {k: v for k, v in str_row.items() if k is not None}
+                crawl_row(context, clean_row)
+        except Exception as e:
+            context.log.error(f"Skipping {year}-{month:02} due to fetch failure: {e}")
```
nvmbrasserie, line 106: Once in a while we're getting 500 Server Error:
Screenshot 2025-06-06 at 21 32 50

I tried to handle it using `time.sleep(random.uniform(1.5, 3.0))`, but that didn't solve the issue. Sometimes it works fine, but at other times it pops up in a different place every time. We could switch to `Zyte`, but there might be other ways to make the request itself more robust.

leonhandreke: Does it help if we add 500 to the retry codes? 

If `time.sleep` makes retries less likely, we should also keep that around.

datasets/_global/black_sea_mou_detention/crawler.py
```diff
@@ -0,0 +1,123 @@
...
+    context.audit_data(clean_row, ["place"])
+
+
+def crawl(context: Context):
+    now = datetime.utcnow()
+    year = START_YEAR
+    month = START_MONTH
+    while (year, month) <= (now.year, now.month):
+        # Break if the month is in the future
+        if is_future_month(year, month, now):
```
leonhandreke: Why would it be in the future? or is this more of a safety check?

datasets/_global/black_sea_mou_detention/crawler.py
```diff
@@ -0,0 +1,123 @@
...
+    )
+    sanction.add("reason", clean_row.pop("nature_of_deficiencies"))
+
+    if h.is_active(sanction):
+        vessel.add("topics", "reg.warn")
+
+    context.emit(vessel)
+    context.emit(sanction)
+
+    context.audit_data(clean_row, ["place"])
```
leonhandreke, line 76: place is where the inspection happened IIUC? Let's add that to a comment, just for the future dev wondering what this is/was :)

datasets/_global/black_sea_mou_detention/crawler.py
```diff
@@ -0,0 +1,123 @@
...
+    sanction = h.make_sanction(
+        context,
+        vessel,
+        start_date=clean_row.pop("date_of_detention"),
+        end_date=clean_row.pop("date_of_release", None),
+    )
+    sanction.add("reason", clean_row.pop("nature_of_deficiencies"))
+
+    if h.is_active(sanction):
+        vessel.add("topics", "reg.warn")
```
leonhandreke, line 71: I wonder if we should be a bit more lenient here? It seems that the ships are all released within a day or two, so this will rarely trigger. But maybe we want to keep them around for longer? Maybe this is a topic for team discussion?

nvmbrasserie: Only the topic will be removed if the vessel is not currently detained. I don't think we should keep it after the vessel's release. We have a similar logic in the paris_mou_banned https://github.com/opensanctions/opensanctions/blob/main/datasets/_global/paris_mou_banned/crawler.py#L48

leonhandreke: Makes sense!

leonhandreke:
Ship it (badum-tsh)!

leonhandreke (approved):

Issue #2291, opened 2025-05-20
Clean up and deduplicate values on exported entities


leonhandreke:
- Dedupe the name a bit, we don't need every casing of a name.
- Names that are already in name don't have to be in alias again
- Date of birth could be cleaned/deduped a bit. It doesn't have the full complexity of addresses, so maybe Geonames or Wikidata
- Date of birth: omit less specific (years) if we have the full date.

- Put a `registrationNumber` that looks like an OGRN on an entity with `country: ru` into the more specific field, same with INN. Question: should this actually happen in `Entity.add` or yet another `apply` helper?

PR #2392, opened 2025-06-16
ViewFragment & Export consolidation

### LINKED ISSUE #2341: Move golden entity consolidation from store to exporters
Right now, `store.assemble()` does some clean-up on entities that makes them more presentable (eg. when stated birthdays include 1985 and 1985-05-31, it will remove the former). This is logically a part of the exports and doesn't need to be run when we do internal activities like running an xref. Let's move it to `zavod.exporters.common`, and call it in the fragment view proposed in #2340.

pudo:
Adds name deduplication to the output filter

Fixes #2341, refs #2291.

### DIFF
--- a/zavod/zavod/exporters/__init__.py
+++ b/zavod/zavod/exporters/__init__.py
@@ -16,7 +16,7 @@ from zavod.exporters.statements import StatementsCSVExporter
 from zavod.exporters.maritime import MaritimeExporter
 from zavod.exporters.delta import DeltaExporter
 
-# from zavod.exporters.fragment import ViewFragment
+from zavod.exporters.fragment import ViewFragment
 from zavod.exporters.metadata import write_dataset_index, write_issues
 from zavod.exporters.metadata import write_catalog, write_delta_index
 
@@ -70,11 +70,11 @@ def export_data(context: Context, view: View) -> None:
 
     for idx, entity in enumerate(view.entities()):
         # Use it once we figure memory explosion out
-        # fragment = ViewFragment(view, entity)
+        fragment = ViewFragment(view, entity)
         if idx > 0 and idx % 10000 == 0:
             log.info("Exported %s entities..." % idx, dataset=context.dataset.name)
         for exporter in exporters:
-            exporter.feed(entity, view)
+            exporter.feed(entity, fragment)
 
     for exporter in exporters:
         exporter.finish(view)
--- a/zavod/zavod/exporters/__init__.py
+++ b/zavod/zavod/exporters/__init__.py
@@ -1,5 +1,6 @@
 from typing import List, Dict, Type, Set
 
+from zavod.exporters.consolidate import consolidate_entity
 from zavod.logs import get_logger
 from zavod.store import View
 from zavod.context import Context
@@ -70,6 +71,7 @@ def export_data(context: Context, view: View) -> None:
 
     for idx, entity in enumerate(view.entities()):
         # Use it once we figure memory explosion out
+        entity = consolidate_entity(view.store.linker, entity)
         fragment = ViewFragment(view, entity)
         if idx > 0 and idx % 10000 == 0:
             log.info("Exported %s entities..." % idx, dataset=context.dataset.name)
--- a/zavod/zavod/exporters/consolidate.py
+++ b/zavod/zavod/exporters/consolidate.py
@@ -29,7 +29,7 @@ def simplify_names(entity: Entity) -> Entity:
         names = entity.get(prop)
         reduced = reduce_names(names)
         if len(reduced) < len(names):
-            stmts = entity._statements.get(prop_, [])
+            stmts = entity._statements.get(prop_, set())
             for stmt in stmts:
                 if stmt.value not in reduced:
                     entity._statements[prop_].remove(stmt)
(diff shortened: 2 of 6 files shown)


zavod/zavod/exporters/fragment.py
```diff
@@ -5,17 +5,20 @@
 from zavod.meta import Dataset
 from zavod.entity import Entity
 from zavod.store import View as ZavodView
+from zavod.exporters.consolidate import consolidate_entity
 
 
 class ViewFragment(View[Dataset, Entity]):
```
leonhandreke, line 14: This is a naming nit, but why is it called `ViewFragment`? Could it be called `EntityCachingView` or something like that?

zavod/zavod/exporters/fragment.py
```diff
@@ -28,7 +31,9 @@ def get_entity(self, id: str) -> Optional[Entity]:
             return self._entities[id]
         entity = self.view.get_entity(id)
         if entity is not None:
-            self._entities[id] = entity
+            entity = consolidate_entity(self.view.store.linker, entity)
+            if len(self._entities) < self.MAX_BUFFER:
```
leonhandreke, line 38: Is there a reason this is not an LRU cache but rather a cache that fills up once?

zavod/zavod/exporters/__init__.py
```diff
@@ -70,11 +71,12 @@ def export_data(context: Context, view: View) -> None:
 
     for idx, entity in enumerate(view.entities()):
         # Use it once we figure memory explosion out
```
jbothma, line 73:
```suggestion
```

zavod/zavod/exporters/consolidate.py
```diff
@@ -0,0 +1,49 @@
...
+        prop = entity.schema.get(prop_)
+        if prop is None:
+            continue
+        names = entity.get(prop)
+        reduced = reduce_names(names)
+        if len(reduced) < len(names):
+            stmts = entity._statements.get(prop_, set())
+            for stmt in stmts:
+                if stmt.value not in reduced:
+                    entity._statements[prop_].remove(stmt)
```
jbothma, line 35: bit of a nit, but it feels like this is maybe something that can be pushed into `zavod.entity.Entity` like e.g. `entity.keep_prop_values(prop, values)`. At least at a glance it doesn't look like we're caching any value accessors at the moment so it isn't critical now, just a thought as I read it.

zavod/zavod/runtime/cleaning.py
```diff
@@ -62,6 +63,14 @@ def value_clean(
                     fuzzy=fuzzy,
                     format=format,
                 )
+        if prop_.type == registry.name and clean is not None and not is_name(clean):
+            log.warning(
+                f"Property {prop_.name!r} is not a valid name: {value}",
```
jbothma: Just to be clearer it's the value, not the property

```suggestion
                f"Property value {prop_.name!r} is not a valid name: {value}",
```

zavod/zavod/exporters/fragment.py
jbothma: The intended use case for this is e.g. when we put the same entity through multiple exporters, each fetching the same set of adjacent entities for each entity.

Am I right in thinking this implementation is better than simply slapping @lru_cache on View.get_entity for this use case because it allows View.get_inverted to be a generator and can be cached?

jbothma: and worth talking through this - so we cache all the keys found via get_inverted without limit, but we only cache up to MAX_BUFFER entities - if an entity has more than MAX_BUFFER adjacent entities, we will skip the leveldb query for related keys, we'll yield related entities from the cache up to MAX_BUFFER, then we'll fetch the remaining entities from leveldb.

If we wanted to be even more careful about memory, we could make sure we don't cache the keys if there are more than MAX_BUFFER. That's maybe not a terrible idea, just because this was a bit of a pain when it turned out we did have horrid corner cases that blow up the memory. And these aren't particularly detailed entities taking a ton of memory each.

zavod/zavod/exporters/fragment.py
```diff
@@ -49,7 +54,5 @@ def get_inverted(self, id: str) -> Generator[Tuple[Property, Entity], None, None
                 yield prop, entity
 
     def entities(self) -> Generator[Entity, None, None]:
-        for entity in self.view.entities():
-            if entity.id is not None:
-                self._entities[entity.id] = entity
-            yield entity
+        # Don't cache entities here
+        yield from self.view.entities()
```
jbothma: is it maybe even worth making this an exception?

jbothma (approved):
Looks good!

It might be worth not caching adjacent ids for entities with more than MAX_BUFFER adjacents, just because if that is a problem, it's quite a shitty one. But it does add complexity which is also less fun.

zavod/zavod/exporters/fragment.py
pudo: Fair point, it could be a `CachedView`! The current name is a bit of a head nod to: https://linkeddatafragments.org/concept/

zavod/zavod/exporters/consolidate.py
pudo: We're doing it two more times in `nomenklatura` already, so it definitely wants to be a method. More generally, I wonder if `nomenklatura.publish` as a module shouldn't start moving into zavod - it's very specific, and we want to be able to tune and test it here.

zavod/zavod/exporters/fragment.py
pudo: Fair, yes it should be

Issue #2172, opened 2025-04-24
blank positions


jbothma:
We have some blank positions not connected to any occupancies.

I suspect they're not dropping out because the peps collection is input to ann_pep_positions and ann_pep_positions is part of the peps collection.

I propose we skip positions in ann_pep_positions wose only dataset is ann_pep_positions.

![Image](https://github.com/user-attachments/assets/0e64af89-792e-40c3-a9fd-3d9e258174bf)

pudo:
This should now be handled, can re-open.

PR #2409, opened 2025-06-19
indicate status and influence for all PEPs with position indicating influence

### LINKED ISSUE #2124: Move PEP descriptive text generation from wd_peps to ann_pep
We're generating text there. This should happen in the annotator for all datasets. Things to include:

* Level of government
* Status - current / past / unknown

- [x] add status
- [ ] announce big delta on release

jbothma:
Fixes #2124, #2172

Ready to review. Release 1 July as per changelog

image

### DIFF
--- a/datasets/_wikidata/peps/crawler.py
+++ b/datasets/_wikidata/peps/crawler.py
@@ -213,6 +213,8 @@ def query_positions(
             position_countries[bind.plain("position")].add(picked_country)
 
     # b) Positions held by politicans from that country
+    # occupation (P106) == politician (Q82955)
+    # country of citizenship (P27) == country.qid
     politician_query = f"""
         SELECT ?position ?positionLabel ?jurisdiction ?country ?abolished
         WHERE {{


datasets/_analysis/ann_pep_positions/analyzer.py
```diff
@@ -1,41 +1,128 @@
...
+                continue
+            seen_status = self.topic_status.get(topic, None)
+            if seen_status == OccupancyStatus.CURRENT.value:
+                continue  # Current trumps all
+
+            if (
+                seen_status == OccupancyStatus.UNKNOWN.value
+                and OccupancyStatus.CURRENT.value not in statuses
+            ):
+                continue  # Let current replace unknown
```
leonhandreke: I think I have a hard time grokking the logic here. I think what I understand: If we've seen a CURRENT, that trumps all. If not, we prefer UNKNOWN over ENDED to play it safe. We emit ENDED only if we're totally sure (I.e. no other value). But maybe I'm missing something? 

So would this be equivalent maybe? https://github.com/opensanctions/opensanctions/commit/4b193d5650e5376c3101054177c19e7b91a7d9f4

jbothma: there's one case we do need to handle a bit differently - currently an occupancy can have CURRENT and UNKNOWN or ENDED and UNKNOWN. In the latter case, we want to prefer ENDED because that's a case where two occupancies are considered to refer to the same term, they're merged, so we can make a stronger statement about the status. I added a function for that.

It might be simpler to handle once we're able to remove UNKNOWN.

Other than that, thanks - this is nice and clean so I went with it.

jbothma:
#2172 is fixed implicitly by driving this from PEPs, but I've added explicit checks to help ensure this property remains true if we refactor in future.

datasets/_analysis/ann_pep_positions/analyzer.py
leonhandreke: Nice, I think I get it now. I think the more comments here, the better, cause this is tricky stuff :)

pudo:
Does this require any additional changes/polish for its production debut tomorrow?

PR #2416, opened 2025-06-20
[ph_most_wanted] switch to the new website

nvmbrasserie:
fixes opensanctions/crawler-planning#346

### DIFF
--- a/datasets/ph/most_wanted/crawler.py
+++ b/datasets/ph/most_wanted/crawler.py
@@ -1,9 +1,5 @@
 import csv
-from pathlib import Path
 from typing import Dict
-import re
-
-import pdfplumber
 
 from zavod import Context, helpers as h
 
@@ -21,7 +17,6 @@ PDFs = {
         "3a34d71972f2c67b14a1db66d324f88451851e40",
     ),
 }
-TRAILING_WHITESPACE_PATTERN = re.compile(r"\s+$", re.MULTILINE)
 
 
 def parse_name(context, name):
@@ -84,17 +79,6 @@ def crawl_row(context: Context, row: Dict[str, str]):
     context.audit_data(row, ignore=["jor-no", "reward"])
 
 
-def local_save_pdf_text(context: Context, name: str, pdf_path: Path):
-    text = ""
-    with pdfplumber.open(pdf_path) as pdf:
-        for page in pdf.pages:
-            text += page.extract_text()
-    text = TRAILING_WHITESPACE_PATTERN.sub("", text)
-    pdf_text_path = context.dataset.base_path / f"{name}.txt"
-    with open(pdf_text_path, "w") as fh:
-        fh.write(text)
-
-
 def crawl(context: Context):
     path = context.fetch_resource("source.csv", context.data_url)
     with open(path, "r", encoding="utf-8") as fh:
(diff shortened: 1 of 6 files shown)


datasets/ph/most_wanted/crawler.py
```diff
@@ -1,22 +1,67 @@
...
+    """
+    parts = [p.strip() for p in name.split(",")]
+
+    if len(parts) == 3:
+        last_name, first_name, suffix = parts
+        full_name = f"{first_name} {last_name} {suffix}".strip()
+    elif len(parts) == 2:
+        last_name, first_name = parts
+        suffix = ""
+        full_name = f"{first_name} {last_name}".strip()
```
leonhandreke: I feel we should return `None` for the non-available components here. I think it should be `h.apply_name`'s job to generate the full name.

datasets/ph/most_wanted/ph_most_wanted.yml
```diff
@@ -1,37 +1,48 @@
...
 entry_point: crawler.py
-prefix: ph-crw
+prefix: ph-mwp
 coverage:
   frequency: never
   start: "2024-07-23"
+manual_check:
+  last_checked: "2025-06-20"
+  interval: 90
+  message: Please review all three source links and update the sheet if any changes are found. Source PDFs are located in this repository.
```
leonhandreke: I'm only seeing one source link that leads to a single PDF? Also, maybe you could elaborate on how you got items from the PDF into the sheet?

nvmbrasserie:
I don’t think ChatGPT can reliably handle this type of PDF. Despite trying different prompts, I still saw hallucinations and inaccuracies. That said, I don’t think we’ll need to redo this process every 90 days. The source files haven’t been updated since 2022 and may remain unchanged for a while.

My suggestion is to merge the current version as is, using the data I’ve already reviewed. We can always drop the updates and make it static, but I don't think that now is a good time.

nvmbrasserie:
not yet, I'll take a look at it once I wrap up a couple of other things

jbothma:
just noting that we think the zavod ui MVP should be ready in a week or so which seems early enough to do this through that.

jbothma:
@pudo I think we have a couple of options here:

1. stick the [PDF from the new URL](https://pnp.gov.ph/most-wanted-persons-with-reward/) through gpt with reviews per page, re-review modified pages on edit
2. hash the pdf, dump the full pdf text (collpasing whitespace other than newlines) and add to git and see if we can get a decent diff when there's a change, to update the spreadsheet manually

I'm kinda leaning to 2. Do you have strong feelings?

pudo:
So this hasn't changed since october 2022? I feel then a google sheet is totally fine, and we can rely on the date published on the website

datasets/ph/most_wanted/ph_most_wanted.yml
```diff
@@ -1,39 +1,67 @@
...
+
+    Source PDFs can be found in this repository.
+    You can use ChatGPT OCR (max 2 pages at a time), Claude, or any reliable OCR tool,
+    but be aware of their limits—Claude may crash on large files. Manual, page-by-page
+    review is essential to catch OCR errors and maintain data quality.
+    How-To:
+    Split large PDFs if needed.
+    OCR 1–2 pages at a time.
+    Compare results with the original PDF and correct errors.
+    Update the Google Sheet only after thorough review.
```
jbothma, line 24: maybe we can do a little script that outputs with the columns we have in the google sheet to make copy/pasting in rows easier.

jbothma: and maybe these URLs should be moved to comments in the PDF URL dict or we should scrape those from these URLs

PR #2426, opened 2025-06-25
Only crawl the set of files making up the latest full dataset release

jbothma:


pudo:
Weird question: are we 100% sure we need to process all of these files? I had the feeling previously like they were sets, and we may be processing *all historical releases* of the data. Can we double check?

jbothma:
will do!

PR #2454, opened 2025-07-01
[us_ofac_sdn] Promoting Accountability for Assad and Regional Stabilization Sanctions (PAARSS)

nvmbrasserie:


datasets/us/ofac/us_ofac_sdn.yml
```diff
@@ -132,8 +132,8 @@ lookups:
           - BPI-Sudan-14098
           - SUDAN-EO14098
         value: US-DARFUR
-      - match: SYRIA-EO13894
-        value: US-SYR-REL
+      - match: PAARSSR-EO13894
+        value: US-PAARSS
```
jbothma: As discussed, would it make sense to keep the code the same on our side so that we don't break existing data or links to the program page? Is this fundamentally still the same program, they just changed their code, and added some updates to the page? 

Looking at the program page, it looks like they've just added an executive order delisting some people

nvmbrasserie: I believe it's still the same program, so I agree

nvmbrasserie: Screenshot 2025-07-02 at 16 38 14
I also just added more details to description, kept everything else as it was.

PR #2458, opened 2025-07-02
allow numbers to be the name of the vessel

nvmbrasserie:


zavod/zavod/runtime/cleaning.py
```diff
@@ -63,7 +63,12 @@ def value_clean(
                     fuzzy=fuzzy,
                     format=format,
                 )
-        if prop_.type == registry.name and clean is not None and not is_name(clean):
+        if (
+            prop_.type == registry.name
+            and prop_.schema != "Vessel"
```
pudo: This doesn't work because prop_.schema is the schema where the prop is defined (Thing). Maybe it makes sense to first check if something is a name value, then make a separate check `if entity.schema.is_a("LegalEntity") and not is_name()`.

PR #2465, opened 2025-07-04
fixes 04 07

nvmbrasserie:


datasets/cn/peoples_congress_wikipedia/crawler.py
```diff
@@ -183,8 +183,9 @@ def crawl(context: Context):
             table = h3.getparent().getnext()
         else:
             continue
-
-        if table.tag != "table":
+        # There are cases where the table is not immediately after the <h3>
+        # so we look until we get the next <table> element
+        while table is not None and table.tag != "table":
```
jbothma, line 188: What do you think of this? Basically if we encounter a heading, something went wrong - the current h3_text doesn't apply to `table` any more.

```suggestion
        while table is not None and table.tag != "table":
            assert not table.tag.startswith("h"), table
```

datasets/lt/illegal_gambling/lt_illegal_gambling.yml
```diff
@@ -32,6 +32,13 @@ data:
   format: HTML
 
 lookups:
+  type.name:
+    options:
+      - match: 
+          - ","
+          - "("
+          - "),"
```
jbothma: A better fix for this _might_ be to add e.g. `(„` as a split option before `„` then it's all in one place

nvmbrasserie:
@jbothma I think it should be ready to go, the only thing is that I don't want to break `jp_mof_sanctions` splits

datasets/mc/fund_freezes/mc_fund_freezes.yml
```diff
@@ -68,6 +68,113 @@ lookups:
         value: Organization
       - match: Navire
         value: Vessel
+  aliases:
+    options:
+      - match: Hussam/Hossam Ahmed/Mohammed/Muhammad al-Katerji
+        aliases:
+          - Hussam al-Katerji
+          - Hossam al-Katerji
+          - Ahmed al-Katerji
```
jbothma:
```suggestion
          - Hossam Ahmed al-Katerji
```

datasets/mc/fund_freezes/crawler.py
```diff
@@ -45,8 +45,21 @@ def crawl_entity(context: Context, data: Dict[str, Any]):
     entity.id = context.make_id("mc-freezes", entity_id)
     entity.add("topics", "sanction")
     details = data.pop("mesureDetails")
-    alias = details.pop("alias")
-    entity.add("alias", re.split(r"[;”]", alias))
```
jbothma, line 49: nice to stop stripping the trailing " off the end

jbothma (approved):
Could you see if you're happy with my changes in jp_mof? Shout if anything's unclear

PR #2467, opened 2025-07-04
Fixes-04-07-part-2

nvmbrasserie:


datasets/za/fic_sanctions/crawler.py
```diff
@@ -92,6 +92,8 @@ def crawl_row(context: Context, data: Dict[str, str]):
 
         aliases = data.pop("IndividualAlias", None)
         for alias in h.multi_split(aliases, [", ", "Good", "Low"]):
+            if "?" in alias:
+                continue
```
jbothma: possibly a touch broad - anything with ? gets skipped.

you could do something like

```suggestion
            if all(c in {"?", " "} for c in alias)
                continue
```

jbothma (approved):

PR #2470, opened 2025-07-07
Fixes-07-07-part-2

nvmbrasserie:


datasets/us/fbi_most_wanted/crawler.py
```diff
@@ -99,10 +99,11 @@ def crawl_person(context: Context, url: str) -> None:
 
     aliases = doc.findtext('.//div[@class="wanted-person-aliases"]/p')
     if aliases is not None:
-        for alias in aliases.split(","):
-            alias = alias.strip()
-            prop = "alias" if " " in alias else "weakAlias"
-            person.add(prop, alias)
+        for raw_alias in aliases.split(","):
```
jbothma: what do you think of replacing `,”, “` with `”, “`?
image

datasets/us/fbi_most_wanted/us_fbi_most_wanted.yml
```diff
@@ -50,6 +50,14 @@ assertions:
...
 
 lookups:
+  type.name:
+    # These are the real aliases of
+    # https://www.fbi.gov/wanted/cyber/yasar-balaghi
+    # https://www.fbi.gov/wanted/cyber/masoud-jalili
+    - match: "18"
+      value: 18
+    - match: "1028"
+      value: 1028
```
jbothma: Let's map these to weakAlias with a note that having these as names would make for quite rubbish matching results in a lot of cases.

jbothma (commented):
I'll check plum later tonight

datasets/us/plum_book/crawler.py
```diff
@@ -20,7 +20,13 @@ def crawl(context: Context):
             agency_name = row.pop("AgencyName")
             position_title = row.pop("PositionTitle")
             appointment_type = row.pop("AppointmentTypeDescription")
+            # 'Expiration date (for term and time-limited appointments)'
+            # The assumption here is that it's the date when the position expires
             expiration_date = row.pop("ExpirationDate")
+            # 'IncumbentBeginDate' is the date when the person started
+            start_date = row.pop("IncumbentBeginDate")
+            # 'IncumbentVacateDate' is the date when the person vacated the position
+            end_date = row.pop("IncumbentVacateDate")
```
jbothma: The way I understand this, it's when the person has to vacate the position - term-limited usually means one term can only be that long. So it applies to the occupancy (aka term) rather than the position.  Either the person then has to be re-appointed, or that person is not allowed to hold that position again for some time or ever. Someone else will be `ADMINISTRATIVE CONFERENCE OF THE UNITED STATES`, `CHAIRMAN` in 2027.

So I think the correctly logic for end date is something like 

    end_date = row.pop("IncumbentVacateDate", row.pop("ExpirationDate"))`

image

https://chcoc.gov/sites/default/files/Periodically%20Listing%20Updates%20to%20Management%20%28PLUM%29%20Guidance%20March%202024.pdf

PR #2471, opened 2025-07-08
Fixes-08-07

nvmbrasserie:


datasets/us/finra_actions/crawler.py
```diff
@@ -55,6 +55,9 @@ def crawl_item(input_dict: dict, context: Context):
     date = input_dict.pop("action-date-sort-ascending")[0].strip()
 
     for name in names:
+        # Skip invalid names
```
leonhandreke: Is there a reason we don't do this as a datapatch to null?

nvmbrasserie: if we skip it in the code we won't end up with an empty entity https://www.opensanctions.org/entities/us-finra-1/

leonhandreke: Ah, that makes sense - or does it? Anyway, let's discuss tomorrow and merge this now :)

PR #2474, opened 2025-07-09
Fixes-09-07

nvmbrasserie:


datasets/at/meine_abgeordneten/at_meine_abgeordneten.yml
```diff
@@ -79,7 +79,7 @@ lookups:
         value: null
   expected_503:
     options:
-      - match: https://www.meineabgeordneten.at/Abgeordnete/karl.weber
+      - match: https://www.meineabgeordneten.at/Abgeordnete/gunther.baschinger
```
jbothma, line 82: This is fine but I think next time we can just info log these and then rely on assertions to catch if more than a fraction of the site is giving 503s. I'm not sure if this was built before data assertions or why I did it this way but we now have better tools.

datasets/sg/gov_dir/crawler.py
```diff
@@ -161,7 +161,7 @@ def crawl_person(
     rank = official.find(".//div[@class='rank']").text_content().strip()
     full_name = official.find(".//div[@class='name']").text_content().strip()
     # Skip empty names
-    if full_name.strip() == "-":
+    if not full_name.strip("-"):
```
jbothma: should this also strip whitespace (as before)? A string with a space isn't falsy.

datasets/us/state_terrorist_orgs/us_state_terrorist_orgs.yml
```diff
@@ -126,3 +126,7 @@ lookups:
         names:
           - name_clean: Shining Path
             alias: ["Sendero Luminoso", "SL"]
+      - match: al-Nusrah Front, aka Hay’at Tahrir al-Sham (ANF/HTS)
+        names:
+          - name_clean: al-Nusrah Front
+            alias: ["Hay’at Tahrir al-Sham", "ANF/HTS"]
```
jbothma: Ideally we'd have ANF and HTS as distinct weakAlias values

jbothma: I'm really sorry - I misremembered the conversation about this - we want the acryonumns in alias, not weakAlias - so all that was needed was splitting.

nvmbrasserie: no worries, I just fixed it

PR #2478, opened 2025-07-10
Moldova State Register of Non-Profit Entities

nvmbrasserie:


datasets/md/nonprofits/crawler.py
```diff
@@ -0,0 +1,88 @@
...
+    for link in links:
+        href = link.get("href")
+        # Exclude a known outdated file and any from 2019
+        # It's safe bacause we assert that the latest link is from the last 30 days
+        if "17.06.2024" in href or re.search(r"resources/2019-\d{2}", href):
+            continue
+        # Full date: YYYY.MM.DD or YYYY-MM-DD
+        match = re.search(r"(\d{4})[.\-](\d{2})[.\-](\d{2})", href)
+        if match:
+            date_obj = datetime.strptime(match.group(0), "%Y.%m.%d")
```
leonhandreke: Why does this work if it can also be dashes?

datasets/md/nonprofits/crawler.py
```diff
@@ -0,0 +1,88 @@
...
+
+    return latest_link[1]
+
+
+def crawl_row(context, row):
+    tax_number = row.pop("tax_number")
+    name = row.pop("name")
+    director = row.pop("director")
+    if not tax_number and not name:
+        context.log.warning("Row is missing 'tax_number' and 'name', skipping")
```
leonhandreke: Does this happen? This will show up in the issues log and I wonder what the resolution strategy is in this case.

datasets/md/nonprofits/crawler.py
```diff
@@ -0,0 +1,88 @@
...
+    links = doc.xpath(
+        "//li[contains(@class, 'resource-item')]//a[contains(@class, 'resource-url-analytics') and contains(@href, '.xlsx')]"
+    )
+    dated_links = []
+
+    for link in links:
+        href = link.get("href")
+        # Exclude a known outdated file and any from 2019
+        # It's safe bacause we assert that the latest link is from the last 30 days
+        if "17.06.2024" in href or re.search(r"resources/2019-\d{2}", href):
```
leonhandreke: I don't understand, why do we need to exclude them and what does the 30 day assertion have to do with it?

datasets/md/nonprofits/crawler.py
```diff
@@ -0,0 +1,88 @@
...
+
+def crawl_row(context, row):
+    tax_number = row.pop("tax_number")
+    name = row.pop("name")
+    director = row.pop("director")
+    if not tax_number and not name:
+        context.log.warning("Row is missing 'tax_number' and 'name', skipping")
+        return
+    entity = context.make("Organization")
+    entity.id = context.make_id(tax_number, name)
```
leonhandreke: In ru_egrul, we only use the INN - maybe `make_slug` with only the tax number is a better option?

leonhandreke (commented):
looks like a nice clean data source :)

datasets/md/nonprofits/crawler.py
nvmbrasserie: not really, I just wanted to check that we'd always have these values for ids.

datasets/md/nonprofits/crawler.py
nvmbrasserie: We get the links for the last 7 or 8 years. The newer ones are easier to parse since they all follow the same pattern. The older ones are messier, so I decided to exclude them upfront

datasets/md/nonprofits/crawler.py
nvmbrasserie: some of the orgs don't have the tax number, so here it's better to rely on both of them

leonhandreke: Alright!

nvmbrasserie:
merge with the companies dataset

datasets/md/companies/md_companies.yml
```diff
@@ -16,21 +16,66 @@ exports:
...
+      Person: 15000
+      Companies: 250000
+  max:
+    schema_entities:
+      Organization: 25000
+      Person: 25000
+      Companies: 500000
 
 lookups:
+  columns:
```
leonhandreke: maybe call this nonprofit_headers?

leonhandreke (approved):
nice!

PR #2482, opened 2025-07-11
Fixes-11-07

nvmbrasserie:


datasets/gb/coh/gb_coh_disqualified.yml
```diff
@@ -118,3 +117,8 @@ lookups:
         value: "2031-07-16"
       - match: "9999-12-31"
         value: null
+  type.name:
+    options:
+      # as per https://find-and-update.company-information.service.gov.uk/disqualified-officers/natural/nPada61zCehhR2wnbTffM3PbVXA
+      - match: '5191/2015'
+        value: GSE LONDON LIMITED
```
jbothma: so is the name of the company `5191/2015` in the gb_coh data? and we're inferring that the company's name is `GSE LONDON LIMITED`?

when I search for the company I don't find it, so It isn't clear what we're mapping to this name
https://find-and-update.company-information.service.gov.uk/search/companies?q=5191

datasets/za/fic_sanctions/crawler.py
```diff
@@ -94,10 +94,9 @@ def crawl_row(context: Context, data: Dict[str, str]):
         for alias in h.multi_split(aliases, [", ", "Good", "Low"]):
             if all(c in {"?", " "} for c in alias):
                 continue
-            if "?" in alias:
-                context.log.warning("Alias contains '?', skipping", alias=alias)
-                continue
             entity.add("alias", alias)
+            if "?" in entity.get("alias"):
+                context.log.warning("Alias contains '?'", alias=alias)
```
jbothma: this is saying "if any entire alias in the alias list is `?`"
I think what you want is this:
```suggestion
            if any(["?" in a for a in entity.get("alias")]):
                context.log.warning("Alias contains '?'", alias=alias)
```

datasets/za/fic_sanctions/crawler.py
```diff
@@ -111,12 +110,11 @@ def crawl_row(context: Context, data: Dict[str, str]):
...
             # if we split on a comma, we will separate ", LTD" from the name
             alias = alias.rstrip(",")
-            if "?" in alias:
-                context.log.warning("Alias contains '?', skipping", alias=alias)
-                continue
             if all(c in {"?", " "} for c in alias):
                 continue
             entity.add("alias", alias)
+            if "?" in entity.get("alias"):
+                context.log.warning("Alias contains '?'", alias=alias)
```
jbothma: ditto

datasets/za/fic_sanctions/za_fic_sanctions.yml
```diff
@@ -173,6 +173,12 @@ lookups:
...
   type.name:
     options:
+      - match: Jin Tiesan (???)
+        value: Jin Tiesan
+      - match: ??? ?????? (Abdullah Sindhi)
+        value: Abdullah Sindhi
       - match:
           - None
+          - ????? ??? ????? ?????? ???????
+          - ?????????????
```
jbothma: our logic should mean that these get skipped, right?

PR #2483, opened 2025-07-11
Export a useful issues log for collections

leonhandreke:
Issues are currently one of the main ways of observability into our ETL
pipeline. As more logic moves into the export stage, it makes sense to
export issues from there. This commit bundles the following changes:

- As a general pattern: catch exceptions in the
  {crawl,validate,export}_dataset functions, write to issues
  (because we have a Context there) and raise
  RunFailedException from there to catch that in CLI. Publish doesn't
  have the Context, so no issues logging from there. But if the
  publishing is broken, the issues file would never see the light of day
  anyway (If a log falls in a forest and no one is there to hear it,
  does it make a sound?)
- Write issue_levels and issue_count in collections metadata. So far,
  they would always just be empty for collections. This will make the
  issues show up in our site.
- When writing the issues for collections, don't aggregate issues from
  sub-datasets. Yeah, this didn't make sense, we always showed 0 issues
  for collections but the issues file would be filled with all the
  issues in the world...
- Reset the issues log in export_dataset if it's a collection. I know
  this is a weird place to do it, but for non-collections this is done
  in crawl_dataset. There isn't really a good place that covers the
  whole "scope lifecycle" (which is different for collections, since
  they don't have a crawl/validate stage), and the CLI seemed like a bad
  place to do it.

Please review: Why did publish_failure `assert not
dataset.is_collection`?

I guess there is the larger question here of is our issues log really
the right place to catch programming errors deep in our application
code. One could argue that job failures are something different from
data/data source issues and they should be dealt with at a different
level and with a different system. But this is it for now.

### DIFF
--- a/zavod/zavod/exporters/__init__.py
+++ b/zavod/zavod/exporters/__init__.py
@@ -18,7 +18,7 @@ from zavod.exporters.maritime import MaritimeExporter
 from zavod.exporters.delta import DeltaExporter
 
 from zavod.exporters.fragment import ViewFragment
-from zavod.exporters.metadata import write_dataset_index, write_issues
+from zavod.exporters.metadata import write_dataset_index
 from zavod.exporters.metadata import write_catalog, write_delta_index
 
 log = get_logger(__name__)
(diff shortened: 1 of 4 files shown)


leonhandreke:
`DatasetIssues.export` basically does a `cat issues.log | ... > issues.json` (I put this in the commit message and just realized it did not show up in the PR)

PR #2489, opened 2025-07-14
Black Sea MoU PSC Inspections

nvmbrasserie:
fixes opensanctions/crawler-planning#357

Currently blocked because of the captcha. 
captcha


pudo:
Ok - if we can't solve it using gpt, then this one is inaccessible. Let's close.

PR #2490, opened 2025-07-14
Abuja MoU PSC Inspections

nvmbrasserie:
fixes https://github.com/opensanctions/crawler-planning/issues/357

### DIFF
--- a/datasets/_global/tokyo_mou/psc/crawler.py
+++ b/datasets/_global/tokyo_mou/psc/crawler.py
@@ -169,6 +169,8 @@ def crawl_list_page(context: Context, page: int):
         "///tr[contains(@class, 'even') or contains(@class, 'odd')]//input[@type='hidden']/@value"
     )
     context.log.info(f"Found {len(shipuids)} shipuids in the search response")
+    if len(shipuids) < 2:
+        context.log.warn("Not enough shipuids found, double check the logic.")
     for shipuid in shipuids:
         crawl_vessel_page(context, shipuid)
     # Extract and return total pages


datasets/_global/abuja_mou/psc/abuja_mou_psc.yml
```diff
@@ -0,0 +1,46 @@
+title: Abuja MoU PSC Database
+entry_point: crawler.py
+prefix: abuja-mou-psc
+coverage:
+  frequency: weekly
+  start: "2025-07-14"
+load_statements: true
```
jbothma: let's make this false for the inspections (this) dataset. We'll make it true for the enricher on this dataset.

datasets/_global/abuja_mou/psc/abuja_mou_psc.yml
```diff
@@ -0,0 +1,46 @@
...
+summary: >
+  Records detailing inspections of ships by member Authorities under the Abuja MoU.
+description: |
+  The dataset contains inspection records for ships visiting ports under the jurisdiction 
+  of Abuja MoU member Authorities. Each record includes, but is not limited to, the ship’s 
+  identity, flag, and associated company.
+
+  **Note:** This dataset is updated regularly and typically contains records from the last 6 months.
+  Older entries are removed in each update.
```
jbothma: let's make this 3 years, and the same for tokyo PSC

datasets/_global/abuja_mou/psc/abuja_mou_psc.yml
jbothma: same for tokyo PSC

datasets/_global/abuja_mou/psc/abuja_mou_psc.yml
```diff
@@ -0,0 +1,46 @@
+title: Abuja MoU PSC Database
```
jbothma:
```suggestion
title: Abuja MoU Port State Control Inspections
```

datasets/_global/abuja_mou/psc/abuja_mou_psc.yml
```diff
@@ -0,0 +1,46 @@
...
+    > Abuja MoU was established as an inter-governmental organization comprising the Maritime 
+    > Administrations of twenty-two Countries abutting the Atlantic Coast of Africa.
+  url: https://abujamou.org/en/
+  official: true
+  country: "zz"
+tags:
+  - sector.maritime
+url: https://abujamou.org/en/inspection-database/
+data:
+  url: https://abujamou.org/en/inspection-database/
```
jbothma:
```suggestion
  url: https://abuja.marinet.ru/public.php
```

datasets/_global/abuja_mou/psc/crawler.py
```diff
@@ -0,0 +1,107 @@
...
+HEADERS = {
+    "Content-Type": "application/x-www-form-urlencoded",
+    "Referer": "https://abuja.marinet.ru/",
+    "User-Agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.2 Safari/605.1.15",
+    "X-Requested-With": "XMLHttpRequest",
+    "Origin": "https://abuja.marinet.ru",
+}
+SEARCH_DATA = {
+    # Go back ~6 months (approximate as 182 days)
+    "From": f"{(TODAY - timedelta(days=182)).strftime("%d.%m.%Y")}",
```
jbothma: let's make this 3 years

datasets/_global/abuja_mou/psc/crawler.py
```diff
@@ -0,0 +1,107 @@
...
+        "UID": f"{shipuid}",
+    }
+
+    # POST to get full ship profile using shipuid
+    detail_doc = context.fetch_html(
+        "https://abuja.marinet.ru/public.php/?action=getinsppublic",
+        data=detail_data,
+        headers=HEADERS,
+        method="POST",
+        cache_days=30,
```
jbothma: let's make this 6 months

datasets/_global/abuja_mou/psc/crawler.py
```diff
@@ -0,0 +1,107 @@
...
+    assert len(ship_data) == 1, "Expected exactly one ship data table"
+    row = list(h.parse_html_table(ship_data[0]))
+    assert len(row) == 1, "Expected exactly one row in ship data table"
+    str_row = h.cells_to_str(row[0])
+    crawl_vessel_row(context, str_row)
+
+
+def crawl(context: Context):
+    doc = context.fetch_html(
+        "https://abuja.marinet.ru/public.php/?action=getinsppublicall",
```
jbothma: this could then be built from context.data_url

datasets/_global/abuja_mou/psc/crawler.py
```diff
@@ -0,0 +1,107 @@
...
+    # POST to get full ship profile using shipuid
+    detail_doc = context.fetch_html(
+        "https://abuja.marinet.ru/public.php/?action=getinsppublic",
+        data=detail_data,
+        headers=HEADERS,
+        method="POST",
+        cache_days=30,
+    )
+    tables = detail_doc.xpath("//table")
+    assert len(tables) >= 2, "Expected at least 2 tables in the response"
```
jbothma: no need for this I think - the next queries will fail if what we want isn't there

datasets/_global/abuja_mou/psc/crawler.py
```diff
@@ -0,0 +1,107 @@
...
+        data=SEARCH_DATA,
+        headers=HEADERS,
+        method="POST",
+    )
+    shipuids = doc.xpath(
+        "///tr[contains(@class, 'even') or contains(@class, 'odd')]//input[@type='hidden']/@value"
+    )
+    context.log.info(f"Found {len(shipuids)} shipuids in the search response")
+    if len(shipuids) < 2:
+        context.log.warn("Not enough shipuids found, double check the logic.")
```
jbothma: no need for this - if we didn't find shipids, the data assertions will fail - there's no pagination so no logic to verify

datasets/_global/abuja_mou/psc/crawler.py
nvmbrasserie: `f"{context.data_url}?{urlencode({'action': 'getinsppublicall'})}"` this is the solution I wanted to suggest here, but it actually doesn't get it right:

datasets/_global/abuja_mou/psc/abuja_mou_psc.yml
```diff
@@ -0,0 +1,49 @@
...
+  country: "zz"
+tags:
+  - sector.maritime
+url: https://abujamou.org/en/inspection-database/
+data:
+  url: https://abuja.marinet.ru/public.php
+  format: HTML
+  lang: eng
+http:
+  retry_statuses: [500]
```
leonhandreke: This will replace all the retry statuses - should we add the default ones here, just to be safe?. @jbothma I wonder whether we should have an `extra_retry_statuses` or something that just adds to the default?

(I'm not a big fan of putting this in the YAML anyway, but that's a different story).

datasets/_global/abuja_mou/psc/abuja_mou_psc.yml
```diff
@@ -0,0 +1,49 @@
...
+  of Abuja MoU member Authorities. Each record includes, but is not limited to, the ship’s
+  identity, flag, and associated company.
+
+  **Note:** This dataset is updated regularly and typically contains records from the last 3 years.
+  Older entries are removed in each update.
+publisher:
+  name: Abuja Memorandum of Understanding
+  acronym: Abuja MoU
+  description: |
+    > The Memorandum of Understanding on Port State Control for West and Central African
```
leonhandreke, line 24: (nit) Do we always use a block quote when we copy-paste from the source?

leonhandreke (commented):
Nice! Maybe we can mention in a comment in the code that these two crawlers are very similar so that they have a chance of not diverging that much?

datasets/_global/abuja_mou/psc/abuja_mou_psc.yml
nvmbrasserie: yes, that's what we usually do

datasets/_global/abuja_mou/psc/abuja_mou_psc.yml
jbothma: I think the thinking behind overwriting was partly to give the option of not retrying for some cases, e.g. il_knesset, but perhaps making the common case a pain to make the rare case work isn't good. And it looks like the knesset crawler would work just fine if it did retry 403s - it'll crash out - just a couple of minutes later than it currently does 

```
  # 503 is a ban for a few seconds.
  # 403 seems to mean you're banned for hours or more.
  retry_statuses: [503]
```

So yeah we could even replace retry_statuses with extra_retry_statuses in my view. Keen to hear another time why this shouldn't be in the yaml

datasets/_global/abuja_mou/psc/abuja_mou_psc.yml
jbothma: often yeah. A pattern I'm not a fan of is quoting from their site and then citing their site in the description, just above the dataset or publisher url field... to their site. The Source link is to /about - sure - but I think it's a bit redundant when we're being nice enough to indicate it's a quote

image

 In my view this case in fine though, but what are your thoughts?

datasets/_global/abuja_mou/psc/abuja_mou_psc.yml
jbothma: In this case, 500 is already part of the defaults and I don't see a need to prevent retries for other unsuccessful statuses, so let's delete

```suggestion
```

jbothma (approved):

datasets/_global/tokyo_mou/psc/crawler.py
```diff
@@ -171,8 +169,6 @@ def crawl_list_page(context: Context, page: int):
         "///tr[contains(@class, 'even') or contains(@class, 'odd')]//input[@type='hidden']/@value"
     )
     context.log.info(f"Found {len(shipuids)} shipuids in the search response")
-    if len(shipuids) < 2:
-        context.log.warn("Not enough shipuids found, double check the logic.")
```
jbothma: checking for exactly 1 is not unreasonable, since we don't expect to land on a page of results with zero items. But there's a chance of a page having 1 result, then a test for <2 would fail incorrectly.

long story short, let's keep this, but it didn't gain anything for abuja.

PR #2502, opened 2025-07-17
Clean up issue export & don't export union of all leave issues for collections

leonhandreke:
Don't aggregate issues from leaves of collections, we hide collections from the overview page anyway, it's just not useful.

Remove zavod.exporters.metadata.write_issues. `DatasetIssues.export()`, called in `Context.close()` does the same thing.


pudo (approved):
lol pretty embarrassing stuff. let's get it out.

PR #2506, opened 2025-07-17
Zavod UI

jbothma:
This adds the following based on https://github.com/opensanctions/operations/issues/1459

- the `zavod.stateful.review` module for crawlers to add automated extraction we'd like humans to review
- the `ui` directory housing the Zavod Review UI where humans do those reviews

This is ready to take a look at. The main things to address before I click Ready to review:

- [x] no idea why there are type errors in CI. I'm not getting that locally
- [x] jest tests aren't running - not sure if that's because this PR is adding the workflow.

I'll add comments to other things I spot with a self-review.

## In this change:

- source html snippets are also formatted as highlighed markdown to make review easier
- authentication using IAP with some excessive logging because we'll probably learn a lot about it once it's deployed
- review updates are versioned except for updating the last seen version of a crawl when the extracted data hash matches
  -  yes this adds complexity, but I see surprise hash changes in our future, and it'd suck to lose 1000 human verifications because of that

## Future work

- keyboard shortcuts
- yaml editor
- evidence upload to bucket
- image viewer support served from the evidence bucket


datasets/us/cftc_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,118 @@
...
+        description=(
+            "The relationship of the associate to the defendant e.g. owner, officer, etc."
+        ),
+    )
+
+
+# Heads-up associates seems to be a bit different each time making potentially
+# unnecessary invalidations when the cache expires.
+class Defendant(BaseModel):
+    schema: Literal["Person", "Company", "LegalEntity"]
```
jbothma:
```
/Users/jdb/projects/opensanctions/opensanctions/env/lib/python3.12/site-packages/pydantic/_internal/_fields.py:198: UserWarning: Field name "schema" in "Defendant" shadows an attribute in parent "BaseModel"
  warnings.warn(
/Users/jdb/projects/opensanctions/opensanctions/env/lib/python3.12/site-packages/pydantic/json_schema.py:2324: PydanticJsonSchemaWarning: Default value <bound method BaseModel.schema of <class '_dataset_mod_crawler.Defendant'>> is not JSON serializable; excluding default from JSON schema [non-serializable-default]
  warnings.warn(message, PydanticJsonSchemaWarning)
```

zavod/zavod/shed/gpt.py
```diff
@@ -111,7 +140,7 @@ def run_text_prompt(
                 ],
             }
         ],
-        response_format={"type": "json_object"},
+        response_format=response_type or NotGiven(),
```
jbothma:
```
  File "/Users/jdb/projects/opensanctions/opensanctions/zavod/zavod/crawl.py", line 35, in crawl_dataset
    entry_point(context)
  File "/Users/jdb/projects/opensanctions/opensanctions/datasets/ge/declarations/crawler.py", line 387, in crawl
    crawl_declaration(context, declaration, year == current_year)
  File "/Users/jdb/projects/opensanctions/opensanctions/datasets/ge/declarations/crawler.py", line 246, in crawl_declaration
    apply_translit_names(context, person, "kat", first_name, last_name, TRANSLIT_OUTPUT)
  File "/Users/jdb/projects/opensanctions/opensanctions/zavod/zavod/shed/trans.py", line 64, in apply_translit_names
    last_response = run_text_prompt(context, prompt, last_name)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jdb/projects/opensanctions/opensanctions/zavod/zavod/shed/gpt.py", line 149, in run_text_prompt
    data = json.loads(response.choices[0].message.content)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.9/Frameworks/Python.framework/Versions/3.12/lib/python3.12/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.9/Frameworks/Python.framework/Versions/3.12/lib/python3.12/json/decoder.py", line 338, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.9/Frameworks/Python.framework/Versions/3.12/lib/python3.12/json/decoder.py", line 356, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

datasets/us/cftc_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,118 @@
...
+    if not result:
+        return
+    for item in result.defendants:
+        entity = context.make(item.entity_schema)
+        entity.id = context.make_id(item.name, item.address, item.country)
+        entity.add("name", item.name)
+        entity.add("address", item.address)
+        entity.add("country", item.country)
+        entity.add("alias", item.aliases)
+        context.emit(entity)
```
jbothma: need to add associate relationships and flesh out crawler before we could ship it but it's something we can play with

datasets/us/cftc_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,118 @@
...
+    relationship: Optional[str] = Field(
+        default=None,
+        description=(
+            "The relationship of the associate to the defendant e.g. owner, officer, etc."
+        ),
+    )
+
+
+# Heads-up associates seems to be a bit different each time making potentially
+# unnecessary invalidations when the cache expires.
```
jbothma: maybe I should remove associate relationships for now so we can get this dataset up without thousands getting invalidated when the cache is cleared?

datasets/us/cftc_enforcement_actions/us_cftc_enforcement_actions.yml
```diff
@@ -0,0 +1,39 @@
...
+  url: https://www.cftc.gov/LawRegulation/EnforcementActions/index.htm
+  format: HTML
+ci_test: false # GPT key not available in CI
+
+assertions:
+  min:
+    schema_entities:
+      Person: 10
+      Company: 10
+    countries: 1
```
jbothma: we'll want to bump this once we get to a decent level of accepted reviews

ui/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## How it works
+
+1. OpenSanctions Crawlers call `zavod.stateful.extraction.get_accepted_data` to add the results of automated data extraction to the review database, which returns the value if a matching review entry is marked `accepted`.
+2. Human reviewers visit the user interface, review any non-accepted reviews
+
+
+## Getting Started
+
+For local development, you'll want to connect to your local etl db using `ZAVOD_DATABASE_URI` and disable authentication using `ZAVOD_IAP_DISABLED`
```
jbothma:
```suggestion
For local development, you'll want to connect to your local etl db using `ZAVOD_DATABASE_URI` and disable authentication using `ZAVOD_UNSAFE_IAP_DISABLED`
```

ui/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+- Next.js for the reviews website
+  - Server side pages directly query the database to show data and handle form posts to update data. There's no API.
+- SQLAlchemy is used for Python-based crawlers to maintain review entries in the reviews table.
+
+
+## Configuration
+
+- `ZAVOD_DATABASE_URI` - (e.g. `postgresql://user:pw@host.com/db`) - only postgres supported.
+- `ZAVOD_IAP_AUDIENCE` - https://cloud.google.com/iap/docs/signed-headers-howto#iap_validate_jwt-nodejs
```
jbothma, line 64: I'm not actually sure what this will be because the docs don't mention cloud run. I'll try it and parse the token.

ui/app/api/extraction/save/route.test.ts
```diff
@@ -0,0 +1,32 @@
...
+      key: 'test',
+      accepted: 'true',
+      extractedData: '{}',
+      accept_and_continue: 'false',
+    });
+    const req = new NextRequest('http://localhost', {
+      method: 'POST',
+      headers: {
+        'content-type': 'application/x-www-form-urlencoded',
+        'x-goog-iap-jwt-assertion': 'valid.jwt.token',
```
jbothma:
```suggestion
        'x-goog-iap-jwt-assertion': 'verify-is-mocked.to.return-null-representing-rejection',
```

ui/app/api/extraction/save/route.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export async function POST(req: NextRequest) {
+  const email = await verify(req);
+  if (!email) {
+    return new Response(
+      '<h1>Error: Unauthorized</h1>',
+      { status: 401, headers: { 'Content-Type': 'text/html' } }
+    );
+  }
```
jbothma: add this to the other pages, or middleware? would've been cool for everything but save to be public but not sure that's possible with IAP and maybe private is best until this is less prototype

ui/lib/db.ts
```diff
@@ -0,0 +1,213 @@
...
+    dialect = new PostgresDialect({
+      pool: new Pool({
+        connectionString: dbUrl,
+        max: 10,
+      }),
+    })
+  } else {
+    throw new Error(`Unsupported database type ${dbUrl}`);
+  }
+  db = new Kysely<ReviewDatabase>({ dialect, log: ['error'] })
```
jbothma: it'd be nice if this introspected the table and rejected saving if the model doesn't represent all the fields in the table

ui/lib/db.ts
```diff
@@ -0,0 +1,213 @@
...
+export interface ReviewDatabase {
+  review: ReviewTable
+}
+
+
+const dbUrl = process.env.ZAVOD_DATABASE_URI;
+let db: Kysely<ReviewDatabase> | null = null;
+
+if (dbUrl) {
+  const isPostgres = dbUrl.startsWith('postgres');
```
jbothma: no sqlite but you wouldn't want to have to stop the server locally to run a crawler, would you?

zavod/zavod/integration/dedupe.py
```diff
@@ -106,7 +106,7 @@ def merge_entities(
             edge = resolver.get_resolved_edge(canonical_id, other_id)
             if edge is not None:
                 if force is True or edge.judgement == Judgement.UNSURE:
-                    log.warn("Removing existing edge", edge=edge)
+                    log.warn("Removing existing edge", edge=str(edge))
```
jbothma: this showed up maybe because of test order rejiggle causing this to run after something that starts logging but doesn't stop it?

either way the true fix is probably https://github.com/opensanctions/opensanctions/issues/2513

zavod/zavod/shed/gpt.py
```diff
@@ -82,6 +86,50 @@ def run_image_prompt(
     return data
 
 
+def run_typed_image_prompt(
```
jbothma, line 91: these differ from the untyped versions on:

- including the schema in the cache hash
- calling `client.chat.completions.parse` not `create`
- returning `response.choices[0].message.parsed` not `response.choices[0].message.content` (but still setting the latter in the cache)

I couldn't figure out the right types for the response format argument to `client.chat.completions.parse` with json_object within a couple of hours so.... giving up for now

zavod/zavod/shed/gpt.py
```diff
@@ -82,6 +86,50 @@ def run_image_prompt(
...
+        ],
+        response_format=response_type,
+        max_tokens=max_tokens,
+    )
+    assert len(response.choices) > 0
+    assert response.choices[0].message is not None
+    assert response.choices[0].message.parsed is not None
+    data = response.choices[0].message.parsed
+    context.cache.set_json(cache_key, data)
+    return data
```
jbothma: @jbothma fix this to set content in cache

zavod/zavod/stateful/review.py
```diff
@@ -0,0 +1,272 @@
...
+    ) -> Dict[str, Any]:
+        json_schema = super().generate(schema, mode=mode)
+        json_schema["$schema"] = self.schema_dialect
+        return json_schema
+
+
+class Review(BaseModel, Generic[ModelType]):
+    """
+    A review is the smallest unit of data that's convenient to extract data from,
+    as well as information about the source data, whether it's been accepted, and by whom.
```
jbothma:
```suggestion
    A review is the smallest unit of data that's convenient to extract data from and review,
    along with info about the source data, whether it's been accepted, and by whom.
```

zavod/zavod/stateful/review.py
```diff
@@ -0,0 +1,272 @@
...
+                "Existing data doesn't validate against current model. key=%s version=%s",
+                review.key,
+                review.last_seen_version,
+            )
+            conn.execute(
+                update(review_table)
+                .where(review_table.c.id == review.id)
+                .values(deleted_at=datetime.now(timezone.utc))
+            )
+            return None
```
jbothma: we might want better schema change handling than this - but at least we keep the old version with a deletion timestamp and go ahead with the crawl for now.

datasets/us/cftc_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,118 @@
...
+        url = urls[0]
+        crawl_enforcement_action(context, date, url)
+
+
+def crawl(context: Context) -> None:
+    next_url: Optional[str] = context.data_url
+    while next_url:
+        # Dev
+        if "1" in next_url:
+            break
```
jbothma: @jbothma TODO

ui/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+1. OpenSanctions Crawlers call `zavod.stateful.extraction.get_accepted_data` to add the results of automated data extraction to the review database, which returns the value if a matching review entry is marked `accepted`.
+2. Human reviewers visit the user interface, review any non-accepted reviews
+
+
+## Getting Started
+
+For local development, you'll want to
+
+1. connect to your local etl db using `ZAVOD_DATABASE_URI`
```
leonhandreke: How do you feel about putting the dev postgres in `docker-compose.yml` in the root? Then we can have instructions like this:

```bash
docker compose -f ../docker-compose.yml up -d db # Bring up a dev database
export ZAVOD_DATABASE_URI="postgresql://postgres:password@localhost/etldev"
export  ZAVOD_ALLOW_UNAUTHENTICATED="true" # UNSAFE, don't do this in production!
export NEXT_PUBLIC_BASE_URL="http://localhost:3000"
npm install && npm run dev
```

zavod/zavod/shed/gpt.py
```diff
@@ -120,3 +170,48 @@ def run_text_prompt(
...
+    cache_hash.update(prompt.encode("utf-8"))
+    json_schema = response_type.model_json_schema()
+    cache_hash.update(json.dumps(json_schema, sort_keys=True).encode("utf-8"))
+    cache_key = cache_hash.hexdigest()
+    cached_data = context.cache.get_json(cache_key, max_age=cache_days)
+    if cached_data is not None:
+        log.info("GPT cache hit: %s" % string[:50])
+        return response_type.model_validate(cached_data)
+    log.info("Prompting %r for: %s" % (model, string[:50]))
+    response = client.chat.completions.parse(
```
leonhandreke, line 198: Had to switch this to `client.beta.chat.completions.parse` to get it to work for some reason?

jbothma: does

```
pip install -U  --upgrade-strategy=eager -e './zavod[dev]'
``` 

bring you a compliant version?

leonhandreke: Yeah, I forgot the upgrade eager. I pushed a commit to up the version.

ui/README.md
jbothma: I've added it as docker-compose.dev.yml - it can then be overlaid on the first if needed, but it keeps the dev/test stuff a bit separate - does that feel good to you? 

Should I add an app stub in the dev docker-compose with the db url env var? Or maybe just a reminder in the zavod ui readme to configure zavod to use that db to load some data?

datasets/us/cftc_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,147 @@
...
+    defendants: List[Defendant]
+
+
+PROMPT = """
+Extract the defendants or entities added to the Red List in the attached article.
+Leave out any relief defendants. Leave fields null or lists empty if values are not
+present in the source text.
+
+Trading/D.B.A. names looking like company rather than person names should be extracted
+as companies, not aliases of a person.
```
jbothma: actually apparently the decision is to let these be aliases of a person in medicaid examples

datasets/us/cftc_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,147 @@
...
+        entity.add("name", item.name)
+        entity.add("address", item.address)
+        entity.add("country", item.country)
+        entity.add("alias", item.aliases)
+        if item.status != "Dismissed":
+            entity.add("topics", "reg.action")
+
+        press_release_num = doc.xpath(".//h1[contains(@class, 'press-release-title')]")[
+            0
+        ].text_content()
```
jbothma: yucky wrapping

datasets/us/cftc_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,147 @@
...
+        return
+    for item in result.defendants:
+        entity = context.make(item.entity_schema)
+        entity.id = context.make_id(item.name, item.address, item.country)
+        entity.add("name", item.name)
+        entity.add("address", item.address)
+        entity.add("country", item.country)
+        entity.add("alias", item.aliases)
+        if item.status != "Dismissed":
+            entity.add("topics", "reg.action")
```
jbothma: Not really any point in this condition - the filing notice will still have emitted the topic.

datasets/us/cftc_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,147 @@
...
+
+        press_release_num = doc.xpath(".//h1[contains(@class, 'press-release-title')]")[
+            0
+        ].text_content()
+        press_release_num = press_release_num.replace("Release Number", "").strip()
+        # We try to link press releases that refer to an original press release number
+        # back to the original press release.
```
jbothma: there can be more than one but maybe best effort is sufficient

datasets/us/fed_enforcements/crawler.py
```diff
@@ -33,26 +45,30 @@ def crawl_item(input_dict: Dict[str, str], context: Context):
...
-        entities = [{"name": name, "locality": None, "country": None} for name in names]
+        entities = [
+            BankOrgEntity(name=name, locality=None, country=None) for name in names
+        ]
     else:
         schema = "Company"
         affiliation = None
         party_name = input_dict.pop("Banking Organization")
 
-        result = context.lookup("bank_orgs", party_name)
```
jbothma, line 42: TODO: maybe on startup we can load al the items in the lookup as accepted=True

datasets/us/fed_enforcements/crawler.py
```diff
@@ -33,26 +45,30 @@ def crawl_item(input_dict: Dict[str, str], context: Context):
...
-                match=party_name,
-                entities=entities,
-            )
-            NEW_BANK_ORGS[party_name] = entities
-        else:
-            entities = result.entities
+        prompt_result = run_typed_text_prompt(
+            context,
+            prompt=ORG_PARSE_PROMPT,
+            string=party_name,
```
leonhandreke: (nit) how do you feel about calling this param `input_text`?

leonhandreke:
Built a little crawler for the SEC, which turned out to be very easy since it's almost the same as the CFTC page. I haven't really nailed down the data structure and the prompt yet. Overall, great experience. I've not really reviewed the code in depth yet, but everywhere I clicked through in my IDE, things made perfect sense.

One thing I'm wondering is how to make changes to the schema. Right now, IIUC, a change to the extraction data structure will trigger a cache invalidation for the GPT cache, a new `original_extraction` being created, which in turn triggers the creation of a new unaccepted review.

I wonder if we can in the future easily support the usecase where we want to add a field, but not re-GPT and re-review the entire dataset? I've thought about this alone for a while and the best thing I've come up with is to add a version and min_version param to `run_text_prompt` and `get_accepted_review`. Version becomes part of the the cache key (instead of `json_schema`).

We would then document on fields and in the code at which version they were introduced, and if we eventually want to drop fields or or code handling old versions, we up the min_version requirement to drop the old versions.

Would this work, in principle? Is there a much easier way? Do you see a case for this?

init-test-db.sql
```diff
@@ -0,0 +1,3 @@
+CREATE USER testuser WITH PASSWORD 'testpass';
```
leonhandreke, line 1: Why not use the user and db created on container init?

ui/lib/db.ts
leonhandreke: Yeah I'm totally okay with that.

ui/app/dataset/[dataset]/[key]/page.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { BoxArrowUpRight } from 'react-bootstrap-icons';
+import SourceView from './SourceView';
+import { PageProps } from '@/lib/pageProps';
+
+export default async function EntryPage({ params }: PageProps) {
+  // Check authorization
+  const headersList = await headers();
+  const email = await verify(headersList);
+
+  if (!email) {
```
Comment (leonhandreke) [inline ui/app/dataset/[dataset]/[key]/page.tsx]: Does this want to be a `middleware.js`?

datasets/us/cftc_enforcement_actions/crawler.py
leonhandreke: I don't understand? Does every dismissed case have a "Filed"? Can we somehow match them?

init-test-db.sql
jbothma: it's nice to use a different db for tests, so that tests aren't affected by messing about in dev, and tests don't mess with potentially carefully constructed dev data.

leonhandreke:
Nice, I like the pattern!

leonhandreke:
I think maybe the way to think about versions is not only in terms of raw schema compatibility but also just "version of our code". If we change the prompt, we might also want to bump the version, to eventually be able to invalidate records based on the old prompt.

datasets/us/cftc_enforcement_actions/crawler.py
jbothma: I could actually only find a couple of examples where there was a notice where the ultimate result was dismissal. In once case because the person is deceased, in one case where some charges were dismissed and others were settled on.

datasets/us/cftc_enforcement_actions/crawler.py
jbothma: not sure how much it can be guaranteed, but it can exist yeah. And we merge the sanctions on a best effort basis.

So it's in the data model and I use that as part of the key for the sanction. So the sanction key ends up being made from (defendant_key, previous_press_release). If the defendant key is the the same for different press releases referring back to the same release, they're merged, so the sanction would see different statuses, dates, and source URLs in one sanction entity.

Often the address info is stated slightly differently in different releases, so the defendant Ids differ. So when the defendant entities get merged, they'd have different sanction entities relating to the same case.

regarding dismissed, in the end I only found a couple of examples with dismissed status - it doesn't look like they're making press releases for that often.

PR #2507, opened 2025-07-17
Taiwan Strategic High-Tech Commodities Entity List

nvmbrasserie:
opensanctions/crawler-planning#360

### DIFF
--- a/datasets/tw/shtc/crawler.py
+++ b/datasets/tw/shtc/crawler.py
@@ -1,7 +1,9 @@
 import csv
+
+import datapatch
 from rigour.mime.types import CSV
 
-from zavod import Context, helpers as h
+from zavod import Context, helpers as h, Entity
 from zavod.shed.zyte_api import fetch_html
 
 ADDRESS_SPLITS = [
@@ -33,7 +35,22 @@ ADDRESS_SPLITS = [
 ]
 
 
-def crawl_row(context, row):
+def apply_details_override(
+    context: Context, entity: Entity, lookup_result: datapatch.Result
+):
+    details = lookup_result.details[0]
+
+    entity.add("name", details.get("name"))
+    entity.add_cast("Person", "birthDate", details.get("dob"))
+    entity.add_cast("Person", "birthPlace", details.get("pob"))
+    entity.add("notes", details.get("notes"))
+    entity.add("idNumber", details.get("id_num"))
+    entity.add("address", details.get("address"))
+    entity.add("email", details.get("email"))
+    entity.add("phone", details.get("telephone"))
+
+
+def crawl_row(context: Context, row):
     row = {
         # BOM in the middle of a file, probably a Microsoft artifact, should be ignored,
         # see https://www.unicode.org/faq/utf_bom.html#bom6
@@ -50,28 +67,20 @@ def crawl_row(context, row):
     entity.id = context.make_id(names, aliases)
 
     for address in h.multi_split(addresses, ADDRESS_SPLITS):
-        result = context.lookup("details", address)
-        if result is not None:
-            override = result.details[0]
-            if override:
-                entity.add("address", override.get("address"))
-                entity.add("email", override.get("email"))
-                entity.add("phone", override.get("telephone"))
+        # Generic override to map more details in the address field
+        details_lookup_result = context.lookup("details", address)
+        if details_lookup_result is not None:
+            apply_details_override(context, entity, details_lookup_result)
         else:
             entity.add("address", address)
 
-    for id in ids.split(";"):
-        result = context.lookup("details", id)
-        if result is not None:
-            override = result.details[0]
-            if override:
-                entity.add("name", override.get("name"))
-                entity.add_cast("Person", "birthDate", override.get("dob"))
-                entity.add_cast("Person", "birthPlace", override.get("pob"))
-                entity.add("notes", override.get("notes"))
-                entity.add("idNumber", override.get("id_num"))
+    for id_number in ids.split(";"):
+        # Generic override to map more details in the ID number field
+        details_lookup_result = context.lookup("details", id_number)
+        if details_lookup_result is not None:
+            apply_details_override(context, entity, details_lookup_result)
         else:
-            entity.add("idNumber", id)
+            entity.add("idNumber", id_number)
 
     for name in names.split(";"):
         entity.add("name", name)
--- a/datasets/tw/shtc/crawler.py
+++ b/datasets/tw/shtc/crawler.py
@@ -1,4 +1,5 @@
 import csv
+from rigour.mime.types import CSV
 
 from zavod import Context, helpers as h
 from zavod.shed.zyte_api import fetch_html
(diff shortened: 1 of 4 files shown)


datasets/tw/shtc_list/crawler.py
```diff
@@ -0,0 +1,83 @@
...
+        url_xpath,
+        cache_days=1,
+    )
+    url = doc.xpath(url_xpath)
+    assert len(url) == 1, "Expected exactly 1 URL in the document"
+    h.assert_url_hash(context, url[0], "d046359c5be70faccb040a94035bba54faff6e80")
+    # 2025-03-04	SHTC Entity List
+
+    # Crawl the CSV file
+    path = context.fetch_resource("shtc_list.csv", context.data_url)
```
leonhandreke: Call `export_resource` on this?

datasets/tw/shtc_list/crawler.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def crawl_row(context, row):
+    row = {
+        # BOM in the middle of a file, probably a Microsoft artifact, should be ignored,
+        # see https://www.unicode.org/faq/utf_bom.html#bom6
+        k.lstrip("\ufeff")
+        .strip()
+        .lower()
+        .replace(" ", "_"): v.strip() if isinstance(v, str) else v
```
leonhandreke: Hm, this is more of a nit, but I feel we should try to use the keys in their most original form instead of processing them here.

datasets/tw/shtc_list/crawler.py
```diff
@@ -0,0 +1,92 @@
...
+    "Branch Office 15:",
+    "Branch Office 16:",
+]
+
+
+def crawl_row(context, row):
+    row = {
+        # BOM in the middle of a file, probably a Microsoft artifact, should be ignored,
+        # see https://www.unicode.org/faq/utf_bom.html#bom6
+        k.lstrip("\ufeff")
```
leonhandreke: This the BOM something that can also happen to values?

datasets/tw/shtc_list/tw_shtc_list.yml
```diff
@@ -0,0 +1,56 @@
+title: Taiwan Strategic High-Tech Commodities Entity List
+entry_point: crawler.py
+prefix: tw-shtc
+coverage:
+  frequency: weekly
+  start: 2025-07-17
+load_statements: true
+ci_test: false
+summary: >
+  Entities involved in the production, sale, or distribution of strategic high-tech
```
leonhandreke: So after reading the documentation on their website (which isn't in super-clear English) and [asking Claude](https://claude.ai/share/d9972587-6fe0-4e97-a5b4-02f549a5d49d), my impression is the following:

- ALL exports to the entities on this list require explicit approval. The document "Types of Strategic High-tech Commodities, Specific Strategic High-tech Commodities, and Exportation to Restricted Regions" sort of hints that these "their end-use or end-users are suspected of developing or manufacturing nuclear, chemical or biochemical weapons, missiles or any other weapons of mass destruction (WMDs)."
- This list is part of the legal SHTC legal framework, which restricts exports based on product type and export region. But this export control isn't limited to high-tech goods, nor are the entities on it always involved in the production of high-tech goods.

The list seems to be a standard baddies list, not all of these people are directly involved in manufacturing WMD. So maybe let's say something like "Restricts all exports to these entities, as the end-use or end-users are suspected of...". I like the "end-use or end-users".

datasets/tw/shtc_list/tw_shtc_list.yml
```diff
@@ -0,0 +1,56 @@
+title: Taiwan Strategic High-Tech Commodities Entity List
+entry_point: crawler.py
+prefix: tw-shtc
+coverage:
+  frequency: weekly
+  start: 2025-07-17
+load_statements: true
+ci_test: false
```
leonhandreke: Why?

datasets/tw/shtc_list/crawler.py
```diff
@@ -0,0 +1,92 @@
...
+    entity.id = context.make_id(names, aliases)
+    for name in names.split(";"):
+        entity.add("name", name)
+    for alias in aliases.split(";"):
+        entity.add("alias", alias)
+    for country in row.pop("國家代碼country_code").split(";"):
+        entity.add("country", country)
+    for address in h.multi_split(row.pop("地址address"), ADDRESS_SPLITS):
+        entity.add("address", address)
+    for id in row.pop("護照號碼id_number").split(";"):
```
leonhandreke: > Debarred entities: Companies that have been excluded from public procurement, often because they previously committed fraud in their execution of a government contract.

I would suggest `export.control` or even `sanction`, given that all export is subject to explicit approval?

datasets/tw/shtc_list/tw_shtc_list.yml
```diff
@@ -0,0 +1,56 @@
...
+    schema_entities:
+      LegalEntity: 9000
+  max:
+    schema_entities:
+      LegalEntity: 15000
+
+lookups:
+  type.identifier:
+    options:
+      - match: 'Counterfeit Danish driving licence number 20645897 note:made out to Ali Barkani, born on 22 Aug. 1973 in Morocco'
```
leonhandreke: Alternatively, we could have a generic lookup that maps to properties that kicks in before adding them in their "normal" fields? There is some more info here. WDYT?

leonhandreke (commented):
Nice, looks like it's a chill crawler despite the weird PDF. Maybe strip the `_list` from the ID? We don't usually have that as a suffix.

datasets/tw/shtc_list/crawler.py
nvmbrasserie: nope, all the values are clean

leonhandreke: Nice!

datasets/tw/shtc_list/tw_shtc_list.yml
nvmbrasserie: working on it

datasets/tw/shtc_list/tw_shtc_list.yml
nvmbrasserie: because of zyte, since it uses the API key we skip the ci

leonhandreke: Ah, didn't realize we used Zyte here. Do we have to?

nvmbrasserie: yup, it doesn't load the html properly without it and then the hash check fails:` Runner failed: Expected exactly one document called "SHTC Entity List" [tw_shtc] dataset=tw_shtc`

datasets/tw/shtc/crawler.py
```diff
@@ -0,0 +1,123 @@
...
+    entity.add("address", details.get("address"))
+    entity.add("email", details.get("email"))
+    entity.add("phone", details.get("telephone"))
+
+
+def crawl_row(context: Context, row):
+    row = {
+        # BOM in the middle of a file, probably a Microsoft artifact, should be ignored,
+        # see https://www.unicode.org/faq/utf_bom.html#bom6
+        k.lstrip("\ufeff"): v.strip() if isinstance(v, str) else v
```
leonhandreke, line 57: I just found `normality.cleaning.remove_byte_order_mark`. Could use that instead :)

PR #2514, opened 2025-07-18
[at_meine_abgeordneten] Limitation or bug in date parsing

nvmbrasserie:
opensanctions/crawler-planning#323

### DIFF
--- a/datasets/at/meine_abgeordneten/crawler.py
+++ b/datasets/at/meine_abgeordneten/crawler.py
@@ -17,10 +17,8 @@ MAX_POSITION_NAME_LENGTH = 120
 
 def extract_dates(context: Context, url, el):
     active_date_el = el.find('.//span[@class="aktiv"]')
-    inactive_dates_el = [
-        el.find('.//span[@class="inaktiv"]'),
-        el.find('.//div[@class="text-right ml-auto flex-grow-1"]'),
-    ]
+    inactive_dates_el = el.find('.//span[@class="inaktiv"]')
+    # el.find('.//div[@class="text-right ml-auto flex-grow-1"]'),
     start_date = None
     end_date = None
     assume_current = False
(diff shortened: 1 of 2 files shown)


datasets/at/meine_abgeordneten/crawler.py
```diff
@@ -17,27 +17,34 @@
 
 def extract_dates(context: Context, url, el):
     active_date_el = el.find('.//span[@class="aktiv"]')
-    inactive_dates_el = el.find('.//span[@class="inaktiv"]')
+    inactive_dates_el = [
+        el.find('.//span[@class="inaktiv"]'),
+        el.find('.//div[@class="text-right ml-auto flex-grow-1"]'),
```
jbothma: I'm trying to think if there's a better way than optionally matching `class="text-right ml-auto flex-grow-1"`. The problem with this is that we won't know when this breaks, at which point, what did we fix? All it takes is for someone to feel like they want a little more padding or margin in there for it to break.

I know it adds complexity, but maybe somehting we can do is this logic:

- if we didn't find an aktiv or an inaktiv
  - assert that el (which is funktionseile) has two children
  - take the first child (this has sources and dates)
  - assert it has three children
  - take the third child (this has dates)
  - take its contents as the date string

if we run that logic on the whole dataset, does it work out ok?

jbothma (commented):
testing so far:

it looks like

image

for https://www.meineabgeordneten.at/Abgeordnete/susanne.rosenkranz disappears because once we parse the end date, we know it falls outside our after-office time for non-national position - so that's good

nvmbrasserie:
https://www.meineabgeordneten.at/Abgeordnete/aslihan.bozatemur
https://www.meineabgeordneten.at/Abgeordnete/Safak.Akcay
https://www.meineabgeordneten.at/Abgeordnete/omar.alrawi
https://www.meineabgeordneten.at/Abgeordnete/aslihan.bozatemur
https://www.meineabgeordneten.at/Abgeordnete/helmut.hofer-gruber

datasets/at/meine_abgeordneten/crawler.py
```diff
@@ -181,8 +210,10 @@ def crawl_item(url_info_page: str, context: Context):
     )
 
     parsed_some_mandatee_date = False
+    # Only parse the mandate rows that are not hidden
+    # (i.e. not containing the class "d-block d-lg-none")
```
jbothma:
```suggestion
    # Only parse the mandate rows that are shown on desktop ("d-block d-lg-none")
    # i.e. skip the rows with same content but different layout for mobile
```

jbothma (approved):
Thanks! Looks good!

I made a suggestion for tightening up the hidden content comment, other than that 🚀

PR #2519, opened 2025-07-21
Fixes-21-07

nvmbrasserie:


datasets/us/cbp_forced_labor/crawler.py
```diff
@@ -64,89 +23,46 @@ def crawl_company(context: Context, row: Dict[str, Any], country: str):
         if not match_entity.get("name"):
             context.log.warning("No name found for a company", entity=match_entity)
             continue
-        entity = context.make("LegalEntity")  # create multiple entries for each entity
-        entity.id = context.make_id(match_entity.get("name"), internal_id)
+        schema = "Vessel" if "Fishing Vessels" in country else "LegalEntity"
+        entity = context.make(schema)
+        entity.id = context.make_id(match_entity.get("name"))
```
jbothma: let's add country to the id arguments just in case some very generic name pops up

```suggestion
        entity.id = context.make_id(match_entity.get("name"), country)
```

jbothma (approved):
Nice! just see the one comment

PR #2524, opened 2025-07-22
[afdb_sanctions] cleaned up names

nvmbrasserie:


datasets/_global/afdb_sanctions/crawler.py
```diff
@@ -1,7 +1,45 @@
...
+    "f/k/a",
+    "d/b/a",
+    "dba",
+    "d.b.a.",
+    "a/k/a",
+    "operating",
+    "formerly",
+    "previously",
+]
+RE_ALIAS = re.compile("|".join(NAME_SPLITS), re.IGNORECASE)
```
jbothma: maybe re.escape these so that `.` doesn't mean "match anything"

datasets/_global/afdb_sanctions/crawler.py
```diff
@@ -1,7 +1,45 @@
...
+NAME_REGEX = re.compile(r"^(?P<name>[^()]+?)\s*\(\s*(?P<alias>[^()]+?)\s*\)$")
+
+
+def apply_clean_name(context: Context, entity, name):
+    if "(" in name and NAME_REGEX.match(name) and not RE_ALIAS.search(name):
```
leonhandreke: (nit) I have a suggestion for how to restructure this cascading `if`, but use what you prefer

https://github.com/opensanctions/opensanctions/commit/1f538a297e0dd8ad63ca553ffa86b88c592803e6

datasets/_global/afdb_sanctions/crawler.py
```diff
@@ -1,7 +1,45 @@
...
+        entity.add("name", name)
+    else:
+        names_lookup_result = context.lookup("names", name)
+        if names_lookup_result is not None:
+            details = names_lookup_result.names[0]
+            entity.add("name", details.get("name"))
+            entity.add("alias", details.get("alias"))
+            entity.add("previousName", details.get("previous_name"))
+        else:
+            context.log.warn("No lookup result for name", name=name)
```
leonhandreke: What happens if we don't apply a name? Won't the name be empty in this case? Shouldn't we apply the name anyway in this case? The ID won't change, so it will just show up as a name change.

leonhandreke (commented):
Always love a good data cleanup :)

datasets/_global/afdb_sanctions/afdb_sanctions.yml
```diff
@@ -61,6 +61,218 @@ assertions:
...
+        names:
+          - name: Elmcrest Group Limited
+            previous_name: "Landmarc Limited"
+      - match: 'Private Enterprise (Pe) Rabmer'
+        names:
+          - name: Private Enterprise (Pe) Rabmer
+      - match: 'Laxminarayan Mallick Also Doing Business Under Laxmi Narayan Construction Or Laxminarayan Construction'
+        names:
+          - name: Laxminarayan Mallick
+            alias: ["Laxmi Narayan Construction", "Laxminarayan Construction"]
```
jbothma, line 190: this is a person and two company names. I think we need to somehow end up with these company names in a different but connected entity

nvmbrasserie: we had this discussion with @pudo at some point regarding the medicaid crawlers. the final decision was to keep all `dba`s in the alias field, even though most of them sound like companies

jbothma: oh cool - I'd love for us to capture more decisions like this somehow. Anyway eventually we're in sync :)

PR #2526, opened 2025-07-22
Fixes-22-07-part-2

nvmbrasserie:


datasets/gb/hmt/gb_hmt_sanctions.yml
```diff
@@ -813,11 +813,20 @@ lookups:
           - 苏州佰昇国际贸易有限公司 - SUZHOU BAISHENG INTERNATIONAL TRADE CO., LTD
       - match: INITSIAVIVA-23
         value: INITSIAVIVA-23
+      - match: "(1) Kyrgz BIK number: 136001 (2) TIN: 02012201010017 (3) OKPO: 27231453 (4) SWIFT: RINBKG22"
+        values:
+          - "Kyrgz BIK number: 136001"
+          - "TIN: 02012201010017"
+          - "OKPO: 27231453"
+          - "SWIFT: RINBKG22"
```
jbothma, line 826: doesn't this mean we're treating these as company names? it feels like we need to take a step back here - is there a problem in the crawler?

nvmbrasserie: It's a mistake in the source data: `<Entity_Subsidiaries>(1) BIC: 136001 (2) TIN: 02012201010017 (3) OKPO: 27231453 (4) SWIFT: RINBKG22</Entity_Subsidiaries>`

I saw that you solved the same case like that (and it's exactly the same lookup), so just following the best practice here :)

jbothma: ahhh that's what I get for being fancy and not explaining myself.

```suggestion
      # These appear in <Entity_Subsidiaries>(1) Kyrgz BIK number: 136001...
      # Confirmed these are theirs and not subsidiaries:
      # bik via ofac_sdn
      # TIN, OKPO via https://record.minjust.gov.kg/view/02012201010017
      # SWIFT via https://wise.com/gb/swift-codes/bic-swift-code-checker?code=RINBKG22
      - match: "(1) Kyrgz BIK number: 136001 (2) TIN: 02012201010017 (3) OKPO: 27231453 (4) SWIFT: RINBKG22"
        values:
          - "Kyrgz BIK number: 136001"
          - "TIN: 02012201010017"
          - "OKPO: 27231453"
          - "SWIFT: RINBKG22"
```

PR #2527, opened 2025-07-22
Add additional_retry_statuses HTTP config parameter

leonhandreke:
Useful if you only want to add to the defaults, which is most of the
time.


jbothma (approved):
nice

PR #2528, opened 2025-07-22
[tw_shtc] cleaned up names

nvmbrasserie:


datasets/tw/shtc/crawler.py
```diff
@@ -82,10 +84,13 @@ def crawl_row(context: Context, row):
...
             entity.add("idNumber", id_number)
 
-    for name in names.split(";"):
+    for name in h.multi_split(names, NAME_SPLITS):
         entity.add("name", name)
-    for alias in aliases.split(";"):
-        entity.add("alias", alias)
+    for alias in h.multi_split(aliases, NAME_SPLITS):
+        if len(alias.split()) == 1:
+            entity.add("weakAlias", alias)
```
jbothma: This looks fine for cases like

image

but there are cases like 

image

where I think we're losing information (weakAlias isn't matchable in dedupe/yente)

I think we need to make this a bit more strict - maybe only move it to weakAlias if it's titlecase and shorter than 6 chars?

jbothma (commented):
anoher one I spotted was `永久參考號：KPe.074）` (Permanent reference number: KPe.074)`

datasets/tw/shtc/crawler.py
nvmbrasserie: I'd move short aliases in both title- and lowercase to the weakAlias, I think titlecase is too tight

nvmbrasserie: most of them end up being titlecase anyway, but then we also cover cases like: 
```
weakAlias,Artur,
weakAlias,Kus,
```

pudo: FWIW `weakAlias` is matched in yente, but only if it's a perfect literal match. So we can be a bit aggressive in moving things over there :)

datasets/tw/shtc/crawler.py
```diff
@@ -33,6 +33,12 @@
     "Branch Office 15:",
     "Branch Office 16:",
 ]
+NAME_SPLITS = [
+    ";",
+    "繁體中文：",  # Traditional Chinese:
+    "簡體中文：",  # Simplified Chinese:
+]
+PERMANENT_ID_RE = re.compile(r"^(?P<name>.+?)（永久參考號：(?P<reg_number>.+?)）")
```
jbothma:
```suggestion
PERMANENT_ID_RE = re.compile(r"^(?P<name>.+?)（永久參考號：(?P<unsc_num>.+?)）")
```

datasets/tw/shtc/crawler.py
```diff
@@ -82,10 +88,18 @@ def crawl_row(context: Context, row):
...
             entity.add("idNumber", id_number)
 
-    for name in names.split(";"):
-        entity.add("name", name)
-    for alias in aliases.split(";"):
-        entity.add("alias", alias)
+    match = PERMANENT_ID_RE.match(names)
+    if match:
+        entity.add("name", match.group("name").strip())
+        entity.add("registrationNumber", match.group("reg_number").strip())
```
jbothma: this goes into `Sanction:unscId`

Issue #2459, opened 2025-07-02
[zavod] Make sure that warnings and errors from validation and export are in issues log


pudo:
We need to make sure that a run failure - no matter at which stage - uploads an `issues.json` documenting the error.

leonhandreke:
I think we're getting all stages (need to check for collections), but we're not showing collections.

pudo:
Is there a next action on this?

goose-life:
@leonhandreke can you take a look please

leonhandreke:
I took a look at this:
- only `run` archives failure - this has not been an issue so far
- data sources: export stage in `run` currently does not archive `issues.log`. https://github.com/opensanctions/opensanctions/pull/4550 fixes this.
- collections: currently does not archive failure at all (I believe it will publish `issues.log` in case of a successful run).
  - We have been writing useful issues logs for collections since https://github.com/opensanctions/opensanctions/pull/2502, i.e. only the `export` errors and not errors from all leaf datasets
  - `archive_failure` currently doesn't like being called with a collection. But that may be outdated: I believe the reason for this is that we never wanted to publish a failed `default` to `/datasets`. Since https://github.com/opensanctions/opensanctions/commit/476dcbc0088d5f92b9258244644e61754e85ffdb, we no longer publish failed `index.json` to `/datasets`, and since https://github.com/opensanctions/opensanctions/commit/ff9c602c66668393b66e79850fc1fb8810b899fa we have a mechanism to indicate success/failure in `index.json`. 
  - Conclusion: I think it's safe to start calling `archive_failure` for collections as well. If `default` fails, I think we will end up with a `result: failure` version in `/artifacts`, and that's what we want I think.

leonhandreke:
Opened https://github.com/opensanctions/opensanctions/pull/4576 for the second bullet — letting `archive_failure` be called for collections too, so a failed `default` lands a `result: failure` version in `/artifacts` instead of disappearing.

We agreed in sprint planning that @jbothma will give this a second pair of brains — he's touching this area anyway as part of his work to move all resources to `/artifacts`.

PR #2529, opened 2025-07-22
Close context before writing delta/index/catalog

leonhandreke:
I think this may fix the issue where issues in the index are 0 even
though the exporter logged some and they're in the issues.json. My guess
is that they're still in the write buffer of Context at the time that
`write_dataset_index` is called

See https://github.com/opensanctions/opensanctions/issues/2459

`eu_fsf` is a good candidate to verify this, currently two issues but
doesn't show up in our overview, only on the actual issues page.


jbothma (approved):
very nice find

pudo:
Thank you, great find!

PR #2536, opened 2025-07-23
UscCode-mappings

nvmbrasserie:


leonhandreke:
Super cool! Did you grep for identifiers of the right length in the codebase/data or how did you find them?

nvmbrasserie:
I matched them with the regex in VScode

pudo:
Quick thought: I would like to propose that we wait with merging this until we have ftm 4.0 behind the yente in our SaaS.

PR #2538, opened 2025-07-23
Emit all uscCodes also as regNr

pudo:


leonhandreke:
Nice, didn't think of this. Just for my understanding, doesn't the `continue` mean we'll *only* emit it as regNr? Or am I missing something? 

> For users of the [on-premise API](https://www.opensanctions.org/docs/on-premise/), each major version of the yente software has a service life of one year. After that period, the metadata and data structure of newly published data versions may no longer be compatible with the import mechanism in old software releases.

Does this mean we should give it a year?

PR #2544, opened 2025-07-24
fixes 24 07

nvmbrasserie:


datasets/de/abgeordnetenwatch/de_abgeordnetenwatch.yml
```diff
@@ -45,3 +45,9 @@ assertions:
       # inexplicably so we want it a bit tighter than usual
       Person: 3000
       Position: 30
+
+lookups:
+  type.name:
+    options:
+      - match: '04.12.'
```
jbothma, line 54: it could be nice for all of us to put a little more info in about how we verified that it's safe to map things to null

```suggestion
      # Geburtsname: 04.12
      # https://www.abgeordnetenwatch.de/profile/anne-koenig
      - match: '04.12.'
```

PR #2545, opened 2025-07-24
Spanish Councillors and Mayors

nvmbrasserie:
opensanctions/crawlers#362

### DIFF
--- a/datasets/es/mayors_councillors/es_mayors_councillors.yml
+++ b/datasets/es/mayors_councillors/es_mayors_councillors.yml
@@ -3,7 +3,7 @@ entry_point: crawler.py
 prefix: es-mc
 load_statements: true
 coverage:
-  frequency: monthly
+  frequency: weekly
   start: 2025-07-24
 summary: >
   Individuals holding mayoral or councillor positions in Spain.


nvmbrasserie:
Hey @santialbo! Thank you for sharing a link

datasets/es/mayors_councillors/es_mayors_councillors.yml
```diff
@@ -0,0 +1,86 @@
...
+    The Ministry of Territorial Policy and Democratic Memory is the government department 
+    responsible for proposing and executing policy on relations and cooperation with 
+    autonomous communities and local administrations.
+  country: es
+  url: https://mpt.gob.es/en/ministerio.html
+tags:
+  - list.pep
+url: https://mpt.gob.es/en/politica-territorial/local/sistema_de_informacion_local_-SIL-/alcaldes_y_concejales.html
+data:
+  url: https://concejales.redsara.es/consulta/
```
jbothma, line 30: have you seen any suggestion that this data gets updated with by-elections or is it just the original election result? I've spent way too much time trying to figure this but it's tricky.

datasets/es/mayors_councillors/crawler.py
```diff
@@ -0,0 +1,118 @@
...
+    if first_last_name or second_last_name:
+        last_name = f"{first_last_name} {second_last_name}"
+        h.apply_name(pep, first_name=name, last_name=last_name)
+    else:
+        pep.add("name", name)
+    pep.add("political", row.pop("party"))
+    position = h.make_position(
+        context,
+        # Positions are available for the current officials; historical data lists only mayors
+        position if position else "Mayor",
```
jbothma: we usually do something like `f"{role} of {municipalities[municipality_code].get('municipality.long_name')}"` or `f"Mayor of {municipality}"`

So that'd be the muni and province in the position name, in case munis in different provinces have the same name.

Are there any positions other than mayor or councillor? Otherwise I think you can just translate both of them

jbothma: you can also add gov.muni to everything and add gov.had to mayor and gov.legislative to councillors if we're already doing logic on the position name.

datasets/es/mayors_councillors/crawler.py
```diff
@@ -0,0 +1,118 @@
...
+        context,
+        pep,
+        position,
+        start_date=row.pop("start_date"),
+        end_date=row.pop("end_date", None),
+        categorisation=categorisation,
+    )
+    if occupancy:
+        context.emit(occupancy)
+        context.emit(pep)
```
jbothma: also emit the position

datasets/es/mayors_councillors/crawler.py
```diff
@@ -0,0 +1,118 @@
...
+    first_last_name = row.pop("last_name", None)
+    second_last_name = row.pop("second_last_name", None)
+    position = row.pop("position", None)
+
+    if not name or not province or not municipality:
+        context.log.warning("Missing required fields", row=row)
+        return
+
+    pep = context.make("Person")
+    pep.id = context.make_id(name, province, municipality)
```
jbothma: let's include the last names in the id

jbothma (commented):
nice! Just a couple of small things I think

datasets/es/mayors_councillors/es_mayors_councillors.yml
nvmbrasserie: I don't think so. I checked with the archive and it seems like it's just the original election result

santialbo: ~~I've checked a few news about resigning councillors, and some of them have been replaced in the generated excel file but mostly during 2023. The list doesn't seem to have been updated after that. This is the most up to date list we have found.~~

This was incorrect.

jbothma: amazing, thanks @santialbo. Did the dates reflect the the correct (later) start date there?

datasets/es/mayors_councillors/es_mayors_councillors.yml
```diff
@@ -0,0 +1,100 @@
+title: Spain Mayors and Councillors
+entry_point: crawler.py
+prefix: es-mc
+load_statements: true
+coverage:
+  frequency: daily
```
nvmbrasserie:
```suggestion
  frequency: monthly
```

datasets/es/mayors_councillors/es_mayors_councillors.yml
santialbo: Ignore what I said before, I had a different version, the list is indeed updated and mantained. Yes, the dates for the replacements of the councillors that resign are the correct dates. Resigned councillors don't appear on the list after resignation.

datasets/es/mayors_councillors/crawler.py
```diff
@@ -0,0 +1,141 @@
...
+    pep.id = context.make_id(
+        name, province, municipality, first_last_name, second_last_name
+    )
+    if first_last_name or second_last_name:
+        last_name = f"{first_last_name} {second_last_name}"
+        h.apply_name(pep, first_name=name, last_name=last_name)
+    else:
+        pep.add("name", name)
+    pep.add("political", row.pop("party"))
+    # Positions are available for the current officials; historical data lists only mayors
```
nvmbrasserie:
```suggestion
    pep.add("political", row.pop("party"))
    pep.add("topics", "role.pep")
    # Positions are available for the current officials; historical data lists only mayors
```

datasets/es/mayors_councillors/es_mayors_councillors.yml
jbothma: How about weekly? Since it seems to get updates
image

datasets/es/mayors_councillors/crawler.py
jbothma: the pep topic is added automatically if the end date is within the conditions to be considered a PEP. We then only emit if an occupancy is returned.  https://zavod.opensanctions.org/peps/#only-emit-if-the-person-is-a-pep

datasets/es/mayors_councillors/es_mayors_councillors.yml
jbothma: it's useful to know, because with the heading on the page where we get the data including the term of office (2023), it seemed like it might be static, then the metadata should explain that.

Since it gets updates, this is awesome data.

jbothma (approved):
cool! could do weekly, otherwise ship it

datasets/es/mayors_councillors/es_mayors_councillors.yml
```diff
@@ -0,0 +1,100 @@
+title: Spain Mayors and Councillors
+entry_point: crawler.py
+prefix: es-mc
+load_statements: true
+coverage:
+  frequency: monthly
```
nvmbrasserie:
```suggestion
  frequency: weekly
```

PR #2549, opened 2025-07-25
Jp meti track changes

nvmbrasserie:


jbothma:
I redid the files so that the diff showed archive-org html cleaned down to text content only, diffed with current html cleaned down to text content only. There was no diff, meaning there's no content change. Must have been a markup change.

The text content is useful since the japanese gets escaped to html entities which would be a pain to throw through google translate.

Also made sure we're working from archive.org versions of the PDFs and didn't see a diff in their text content.

PR #2552, opened 2025-07-27
Begin to adopt normality 3, nk 4.1

pudo:


datasets/_global/un_sc_sanctions/crawler.py
```diff
@@ -24,14 +24,14 @@ def values(node):
 
 def parse_alias(entity: Entity, node: Element):
     names = node.findtext("./ALIAS_NAME")
-    quality = node.findtext("./QUALITY")
+    quality = node.findtext("./QUALITY") or "Good"
```
leonhandreke: Now the `"": None` will never match. Intended? Was there an issue with this previously?

datasets/_global/un_sc_sanctions/crawler.py
```diff
@@ -24,14 +24,14 @@ def values(node):
...
     name_prop = NAME_QUALITY[quality]
     if names is None or name_prop is None:
         return
 
     for name in names.split("; "):
-        name = collapse_spaces(name)
-        if not len(name):
+        name = squash_spaces(name)
+        if len(name) == 0:
```
leonhandreke: I went down a rabbit hole today, `Entity.add` will filter out empty strings. I think that makes sense - `Entity.add` is sort of our throw-anything-at-it thing, no? I feel we should use this `len` pattern as little as possible, it's just cruft in the code (the `is None` pattern also was)

leonhandreke (approved):

datasets/_global/un_sc_sanctions/crawler.py
pudo: You're right - I'll remove this.

leonhandreke: Done

PR #2557, opened 2025-07-28
Run is_name validation for Person:{first,last,...}Name as well

leonhandreke:
Right now, we only get these warnings in the export stage, and that's not great.

I know this isn't pretty, and I welcome better suggestions.


pudo (approved):
This function is becoming more and more ripe for a round of refactoring, but I'm happy for this to be merged!

PR #2562, opened 2025-07-28
Make parse_xlsx_sheet take a datapath.Lookup instead of a str

leonhandreke:
Typesafe all the things!


pudo:
I like this a whole lot. Can we get it merged?

PR #2564, opened 2025-07-28
[gleif] Emit everything owned as a Company

### LINKED ISSUE #2551: [gleif] Emit companies as Company, not Organization
Organization means that they're not an `Asset`, so Ownerships for which they're an `ownershipAsset` will not show up on the website. Sad!

A good measure for whether they are a company is whether they have an owner

leonhandreke:
Fixes https://github.com/opensanctions/opensanctions/issues/2551

### DIFF
--- a/datasets/_global/gleif/crawler.py
+++ b/datasets/_global/gleif/crawler.py
@@ -292,6 +292,7 @@ def parse_rr_file(context: Context, fh: BinaryIO):
 
         if rel_schema == "Ownership":
             # Organization entities cannot be assets, so we re-emit the asset entity as a Company
+            assert start_prop == "asset"
             asset = context.make("Company")
             asset.id = lei_id(start_lei)
             context.emit(asset)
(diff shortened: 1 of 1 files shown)


datasets/_global/gleif/crawler.py
```diff
@@ -290,6 +290,12 @@ def parse_rr_file(context: Context, fh: BinaryIO):
         proxy.add("role", rel_type.replace("_", " "))
         proxy.add("status", rel.findtext("RelationshipStatus"))
 
+        if rel_schema == "Ownership":
+            # Organization entities cannot be assets, so we re-emit the asset entity as a Company
+            asset = context.make("Company")
+            asset.id = lei_id(start_lei)
```
jbothma, line 297: I wonder if it's worth adding an assertion or something that `start_prop == "asset"` - basically - if we edit `RELATIONSHIPS`, will we think to check what makes assumptions about the ordering of the props?

jbothma (approved):
I _think_ it'd work! https://github.com/opensanctions/followthemoney/blob/main/followthemoney/model.py#L128
Cool fix.

datasets/_global/gleif/crawler.py
leonhandreke: I think a typesafe FtM-python would be really nice.

PR #2565, opened 2025-07-29
Spain Parliament Members

nvmbrasserie:
opensanctions/crawlers#365


jbothma:
Currently the XML endpoint for members is returning 503
Also the member listing for e.g. J is returning 503

jbothma:
this stuff is nice to have, but fundamentally we need something like `Member of the Senate` (country=es)

image

datasets/es/parliament/crawler.py
```diff
@@ -0,0 +1,167 @@
...
+        first_name=item.pop("first_name"),
+        last_name=item.pop("last_name"),
+    )
+    pep.add("political", party)
+    pep.add("political", item.pop("parliamentary_group"))
+    pep.add("topics", "role.pep")
+    emit_pep_position(
+        context,
+        pep,
+        "Member of Parliament",
```
jbothma: We can specify senate or congress of deputies here

datasets/es/parliament/crawler.py
```diff
@@ -0,0 +1,167 @@
...
+    pep.id = context.make_id(id, name, party)
+    h.apply_name(
+        pep,
+        full=name,
+        first_name=item.pop("first_name"),
+        last_name=item.pop("last_name"),
+    )
+    pep.add("political", party)
+    pep.add("political", item.pop("parliamentary_group"))
+    pep.add("topics", "role.pep")
```
jbothma: you can leave it to make_occupancy to apply the pep topic then it's automatically part of the 

```
if position.is_pep:
    make_occupancy(...)
```

pattern

datasets/es/parliament/crawler.py
```diff
@@ -0,0 +1,167 @@
...
+    emit_pep_position(
+        context,
+        pep,
+        "Member of Parliament",
+        item.pop("start_date"),
+        item.pop("end_date", None),
+        item.pop("constituency_name"),
```
jbothma: the idea with subnational area is to identify the jurisdiction of the role, rather than the constituency the person represents. But since it doesn't make sense for country because that's already in the country prop, it's subnational areas only.
What happens when we put constituency in position, we end up with lots of positions with the same name, but different IDs, and that's not really useful modelling.

```
{
  "id": "es-par-07f49454e02aacf3b1dea914b7519dd8ae2d02bb",
  "caption": "Member of Parliament",
  "schema": "Position",
  "referents": [],
  "datasets": [
    "es_parliament"
  ],
  "first_seen": "2025-07-30T14:08:41",
  "last_seen": "2025-07-30T14:08:41",
  "last_change": "2025-07-30T14:08:41",
  "properties": {
    "name": [
      "Member of Parliament"
    ],
    "country": [
      "es"
    ],
    "subnationalArea": [
      "Almería"
    ],
    "topics": [
      "gov.national",
      "gov.legislative"
    ]
  },
  "target": false
}
```

jbothma (commented):
a few small things - I'm making these changes as I go.

PR #2577, opened 2025-07-31
assign concatenated dates sequentially

nvmbrasserie:


leonhandreke:
This only contains the docs for me - is something still missing here?

jbothma:
Yeah I was getting date warnings in spite of the prompt change.

very good point on the date string length and being assigned to the correct person

image

We could try altering the prompt not to parse the date better, but to tell us whether this is the first or second person in a section. If the date is long enough to be two people, we then do the splitting programmatically and pick the first or second part dependent on the person's arrangement, instead of both dates to one person.

nvmbrasserie:
This is a fully programmatic fix that relies solely on the raw parsed dates. I experimented with several approaches, including prompting ChatGPT to identify whether the date corresponds to the first or second person in a section. None of those methods were robust enough. After multiple iterations and running the solution locally, this approach is the most stable one.

datasets/_global/un_ga_protocol/crawler.py
```diff
@@ -69,11 +99,21 @@ def crawl(context: Context):
...
+                    last_concatenated_date,
+                    person_count_for_date,
+                )
+            )
             occupancy = h.make_occupancy(
-                context, entity, position, start_date=holder.get("date_of_appointment")
+                context, entity, position, start_date=start_date
             )
+            if occupancy is not None:
+                context.emit(occupancy)
```
jbothma: not sure what happened here, but if there's no occupancy, we don't want to emit the person or position either. Could you move them into the condition?

datasets/_global/un_ga_protocol/crawler.py
```diff
@@ -24,10 +28,36 @@ def crawl_pdf_url(context: Context) -> str:
...
+        return start_dates, None, 0
+    # Handle concatenated dates (length >= 18)
+    if start_dates != last_concatenated_date:
+        # New concatenated date - first person gets first 9 chars
+        return start_dates[:9], start_dates, 1  # "09-Sep-22"
+    # Same concatenated date as previous
+    person_count_for_date += 1
+    if person_count_for_date == 2:
+        # Second person gets last 9 chars
+        return start_dates[-9:], start_dates, person_count_for_date  # "07-May-13"
```
jbothma: i'm struggling a little to understand this logic quickly. I'll try again tomorrow and come up with concrete suggestions.

- perhaps we can name things differently to be a bit more obvious?
- perhaps we can just drop start dates if the string is too long
-

jbothma:
Hi there, just to confirm the lookups wouldn't have interfered.
Getting the date wrong is much worse than not having a date for some items. The state machine idea is very cool, but I think let's back away from this:

- if date is longer than a single date, drop it
- just a straight apply_date for the rest

without putting this through data reviews, I'm not sure how we can get confident about the the extraction. And we have a few bits to do before we can sustainably put it through data reviews:

- image checksums that aren't bothered by small pixel shifts but do detect real changes in the image
- identifying the grid of data
- cut up the pdf page into an image per state
- archive the image as evidence
- submit for review

PR #2596, opened 2025-08-04
Fixes-04-08

nvmbrasserie:


datasets/_global/black_sea_mou_detention/crawler.py
```diff
@@ -15,6 +18,37 @@
...
+        raise ValueError("No <tr> elements found in table")
+
+    header_cells = rows[0].findall(f".//{header_tag}")
+    if not header_cells:
+        raise ValueError("No header cells found")
+
+    headers = []
+    for el in header_cells:
+        text = el.text_content().strip()
+        header_text = slugify(text, sep="_") or text.replace(" ", "_")
```
jbothma: an alternative to this function is to add an argument `slugify_header: bool` to the parse_html_table arguments and then make header_text `slugify(text, sep="_") if slugify_header else text`, but I think this is fine.

The main reason I'm thinking about it is that it's a sane argument to add to the helper, and it took me a little bit to figure out which part of this is specific to this crawler

datasets/_global/tokyo_mou/detention/crawler.py
```diff
@@ -17,6 +20,40 @@ def is_future_month(year: int, month: int, now: datetime) -> bool:
...
+def parse_html_table(
+    table: HtmlElement,
+    header_tag: str = "th",
+    skiprows: int = 0,
+) -> Generator[Dict[str, HtmlElement], None, None]:
+    """
+    Parse an HTML table using the first row as headers.
+
+    Custom implementation to handle edge cases like "�" headers that cause
+    issues with the standard h.parse_html_table function.
```
jbothma: this would be covered by the suggestion above, right? Or are there multiple columns with the same header?

nvmbrasserie: yeah, that's all we need, only to make `slugify` optional

nvmbrasserie: do you think it'd be better to adjust the helper function instead?

jbothma: I've adjusted the helper.
A third approach would have been to adjust the slugification to allow values with a single punctuation char, although that's a bit more involved because the current slugify doesn't support that, and conflicts on column name are then more likely.

PR #2610, opened 2025-08-06
[contrib/egrul] Ingest new format that seems to be the same

leonhandreke:
Their official docs at
https://www.nalog.gov.ru/rn77/service/egrip2/egrip_vzayim/ don't warn
about any differences either. Who knows?

### DIFF
--- a/contrib/egrul/egrul_xml.py
+++ b/contrib/egrul/egrul_xml.py
@@ -1,5 +1,5 @@
 from datetime import date
-from typing import Optional, Dict, IO
+from typing import Generator, Optional, Dict, IO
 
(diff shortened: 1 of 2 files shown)


leonhandreke:
Official Documentation on EGRUL 4.07 and EGRIP 4.06: https://www.nalog.gov.ru/rn77/about_fts/docs/13673441/
Official Documentation on EGRUL 4.06 and EGRIP 4.05: https://www.nalog.gov.ru/rn77/about_fts/docs/10428863/

Got confirmation that the changes in the new format shouldn't touch anything we process.

PR #2641, opened 2025-08-11
more categories for belarus

nvmbrasserie:


pudo (commented):
I'm happy with either naming scheme, let's keep it close to https://en.wikipedia.org/wiki/Government_of_Belarus#Composition

PR #2647, opened 2025-08-12
[eu_sanctions_map] map sanction programs

nvmbrasserie:
[eu_journal_sanctions]

- fetch sanctions program titles directly from EUR-Lex pages.
- extract the first EU act code from program titles (e.g., 2024/254, 833/2014) using regex.
- consistently map programs via the `sanction.program` lookup.

[eu_sanctions_map]

- extract and map the regimes reusing the lookups from `eu_fsf`.

### DIFF
--- a/datasets/eu/journal_sanctions/crawler.py
+++ b/datasets/eu/journal_sanctions/crawler.py
@@ -1,6 +1,7 @@
 import csv
 from typing import Dict
 from requests.exceptions import HTTPError
+from normality import squash_spaces
 
 from zavod import Context, Entity
 import zavod.helpers as h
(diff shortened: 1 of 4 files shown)


datasets/eu/journal_sanctions/crawler.py
```diff
@@ -1,11 +1,46 @@
...
+SPECIAL_CASE_URL = (
+    "https://eur-lex.europa.eu/legal-content/EN/TXT/HTML/?uri=CELEX:02014R0833-20240625"
+)
+# year/number or number/year with optional suffix
+FIRST_CODE_RE = re.compile(
+    r"\b(?:No\s+)?(\d{1,4}/\d{1,4})(?:/[A-Z]{2,5})?\b", re.IGNORECASE
+)
+
+
+def extract_program_code(context, source_url):
```
pudo, line 24: Add @cache (from functools) to this so the document is only fetched once).

pudo (approved):
Thanks :)

PR #2672, opened 2025-08-15
US OFAC Enforcement Actions

nvmbrasserie:


datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -0,0 +1,295 @@
...
+        doc = context.fetch_html(base_url, cache_days=1)
+        doc.make_links_absolute(context.data_url)
+        links = doc.xpath(
+            "//div[@class='view-content']//a[contains(@href, 'recent-actions') and not(contains(@href, 'enforcement-actions'))]/@href"
+        )
+        enforcement_dates = doc.xpath(
+            "//div[contains(@class,'margin-top-1') and contains(., 'Enforcement Actions')]/text()[normalize-space()]"
+        )
+        if break_the_loop(context, enforcement_dates):
+            break
```
jbothma: if a dedicated function makes sense, it's nice to name it after what it's doing, rather than what its result is used for. e.g. could rather name it `within_max_age` - it's basically the same as the the function from `shed`, it's just a more specialised version doing some cleaning and parsing first.

But I wonder if we really need a dedicated function?

I think we can make two changes here:

1. loop over children of `.//div[contains(@class, "search-results")`
2. for each search result
    - check if the date for just that search result is within scope, otherwise break
    - just pass the extracted date string to enforcements.within_max_age - it does parsing with the dataset's formats

maybe the code is then slim enough not to need a function for it, but it's cool if that makes it more readable

Checking per search result is nice because we don't stop crawling weirdly a few items short of our date threshold just because something further down the page is beyond the limit. It's also semantically closer to the goal which might make the code less surprising and more straight forward (test one date, not many dates - I was wondering why there are multiple dates to check)

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -0,0 +1,295 @@
...
+    assert all(
+        [article_name, article_pdf, article_html, date]
+    ), "One or more fields are empty"
+
+    review = get_review(context, Defendants, article_name, MIN_MODEL_VERSION)
+    if review is None:
+        prompt_result = run_typed_text_prompt(context, PROMPT, article_html, Defendants)
+        review = request_review(
+            context,
+            article_name,
```
jbothma: I think the url is better than the article title as a unique key.
I can totally imagine in 10 years, they've had two items like `The U.S. Department of the Treasury's Office of Foreign Assets Control Issues a Penalty Notice to an Individual`

While the URL has the risk mentioned in the best practices, that's less scary than incorrectly missing someone becore of a key conflict.

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -0,0 +1,295 @@
...
+            article_html,
+            "text",
```
jbothma: if you tell it source_mime_type=`rigour.mime.types.HTML` you'll get a sexy html and html->markdown viewer in addition to the raw html for review

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -0,0 +1,295 @@
...
+NEVER infer, assume, or generate values that are not directly stated in the source text.
+
+Trading/D.B.A. names which follow a person name but look like company can just be
+aliases of the person. If the name is a person name, use `Person` as the entity_schema.
+
+Specific fields:
+
+- entity_schema: {schema_field.description}
+- address: {address_field.description}
+- country: Any countries explicitly associated with the defendant in the text. Leave empty if not explicitly stated.
```
jbothma: looks like this is a bit broad

image

```suggestion
- country: Any countries the entity is indicated to reside, operate, or have been born or registered in. Leave empty if not explicitly stated.
```

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -0,0 +1,295 @@
...
+
+
+class Defendant(BaseModel):
+    entity_schema: Schema = schema_field
+    name: str
+    aliases: List[str] = []
+    address: str | List[str] | None = address_field
+    country: str | List[str] | None = []
+    status: Status = status_field
+    notes: Optional[str] = notes_field
```
jbothma:
```suggestion
    address: List[str] = address_field
    country: List[str] = []
    status: Status = status_field
    notes: Optional[str] = notes_field
```

jbothma: these can be stricter - I just made it a list after we had reviews done in prod already

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -0,0 +1,295 @@
...
+Extract the defendants or entities subject to OFAC enforcement actions in the attached article.
+NEVER include relief defendants.
+NEVER infer, assume, or generate values that are not directly stated in the source text.
+
+Trading/D.B.A. names which follow a person name but look like company can just be
+aliases of the person. If the name is a person name, use `Person` as the entity_schema.
+
+Specific fields:
+
+- entity_schema: {schema_field.description}
```
jbothma: Shall we try and save ourselves some effort?

image


```suggestion

- name: The null name of the entity precisely as expressed in the text. If an acronym follows the name in parentheses, include it as an alias and not as part of the name.
- entity_schema: {schema_field.description}
```

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -0,0 +1,295 @@
...
+    return False
+
+
+def crawl_enforcement_action(context: Context, url: str) -> None:
+    article = context.fetch_html(url, cache_days=1)
+    article.make_links_absolute(context.data_url)
+    if article is None:
+        return
+    article_name = article.xpath(NAME_XPATH)[0]
+    article_element = article.xpath(BODY_XPATH)[0]
```
jbothma: are we sure there's only one of these? People use `p` tags like they're candy. We wouldn't want to miss out on important content just because they decided to use two `p`'s one time.

Perhaps `div[@class='content']` is a safer bet with an assertion that there's exactly one.

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -0,0 +1,295 @@
...
+    request_review,
+    get_review,
+    model_hash,
+    html_to_text_hash,
+)
+
+NAME_XPATH = "//span[@class='treas-page-title']/text()"
+BODY_XPATH = "//div[@id='block-ofac-content']//div[@class='field__item']/p"
+DATE_XPATH = "//div[@id='block-ofac-content']//div[@class='field__item']/text()"
+PDF_XPATH = "//div[@id='block-ofac-content']//a[@data-entity-type='media']/@href"
```
jbothma: this fails on https://ofac.treasury.gov/recent-actions/20190502, and sometimes there's more than one PDF linked https://ofac.treasury.gov/recent-actions/20190415

I'm not sure it's that nice when there are links for specific entities if we add all links to all entities. Wanna try asking gpt to associate links to specific entities when possible? If the results suck, we better not. But could be worth trying. Any other ideas?

jbothma (commented):
nice! some tweaks we can make, I think, but really nice start.

was it `check_something_changed` that you thought might fit under `shed`, or what are your thoughts so far?

datasets/us/ofac/enforcement_actions/crawler.py
jbothma: then you also don't have to parse the date out of the article - you already have it and can just pass it as an argument to crawl_enforcement_action

nvmbrasserie: I feel that parsing it directly from article wouldn't hurt either

datasets/us/ofac/enforcement_actions/crawler.py
nvmbrasserie: It works very well. I suggest we proceed with it

datasets/us/ofac/enforcement_actions/crawler.py
jbothma: I think it would be better if this loop
```
for result in doc.xpath(".//div[contains(@class, 'search-result')]"):
```
is in `crawl()`

That would address this:

> Checking per search result is nice because we don't stop crawling weirdly a few items short of our date threshold just because something further down the page is beyond the limit. It's also semantically closer to the goal which might make the code less surprising and more straight forward (test one date, not many dates 

Basically it's nice if the logic is like

- for each index page
  - for each search result on the index page
    - is this result and everything after it too old? bail out
    - crawl this search result
 
Right now the logic is

- for each index page
  - if anything on this page is too old, bail out
  - for each search result on the index page
    - crawl this search result
 
It's not the end of the word to miss a few of the things that are on the same page as something more than 10 years old, but it's so easy to make the logic more natural that it feels worth doing

datasets/us/ofac/enforcement_actions/crawler.py
jbothma: it doesn't look like this change has gone in

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -0,0 +1,291 @@
...
+Extract the defendants or entities subject to OFAC enforcement actions in the attached article.
+NEVER include relief defendants.
+NEVER infer, assume, or generate values that are not directly stated in the source text.
+
+Trading/D.B.A. names which follow a person name but look like company can just be
+aliases of the person. If the name is a person name, use `Person` as the entity_schema.
+
+Specific fields:
+
+- name: The null name of the entity precisely as expressed in the text. If an acronym follows the name in parentheses, include it as an alias and not as part of the name.
```
jbothma: sorry, not sure how this `null` snuck in (actually it could well be cursor)

```suggestion
- name: The name of the entity precisely as expressed in the text. If an acronym follows the name in parentheses, include it as an alias and not as part of the name.
```

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -0,0 +1,291 @@
...
+    request_review,
+    get_review,
+    model_hash,
+    html_to_text_hash,
+)
+
+NAME_XPATH = "//span[@class='treas-page-title']/text()"
+CONTENT_XPATH = "//div[@id='block-ofac-content']//div[@class='content']"
+DATE_XPATH = "//div[@id='block-ofac-content']//div[@class='field__item']/text()"
+# Traveling back to summer of 2016 (the first full pdf enforcement was issued on July 5, 2016)
```
jbothma: could you add a note why this date? it was because older notices were PDF, right?

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -0,0 +1,291 @@
...
+class Defendant(BaseModel):
+    entity_schema: Schema = schema_field
+    name: str
+    aliases: List[str] = []
+    address: str | List[str] | None = address_field
+    country: str | List[str] | None = []
+    status: Status = status_field
+    notes: Optional[str] = notes_field
+    related_companies: List[RelatedCompany] = []
+    pdf_url: Optional[str] = None
```
jbothma: nice - could make it a list defaulting to [] from the start (I'm finding more and more things that want to be a list in my crawlers). Then I imagine we can expect to see both links on https://ofac.treasury.gov/recent-actions/20241203_33 added, and just need to check that the result for unicredit is still good

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -0,0 +1,291 @@
...
+        return
+    article_name = article.xpath(NAME_XPATH)[0]
+    article_content = article.xpath(CONTENT_XPATH)
+    assert len(article_content) == 1
+    article_element = article_content[0]
+    date = article.xpath(DATE_XPATH)[0]
+    article_html = tostring(article_element, pretty_print=True).decode("utf-8")
+    assert all([article_name, article_html, date]), "One or more fields are empty"
+
+    review = get_review(context, Defendants, article_name, MIN_MODEL_VERSION)
```
jbothma: we should update the key here too - currently each run re-requests a review for each notice because we're fetching a different key than what we're using when requesting the review

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -0,0 +1,291 @@
...
+        entity.add("country", item.country, origin=DEFAULT_MODEL)
+        entity.add("alias", item.aliases, origin=DEFAULT_MODEL)
+        entity.add("topics", "reg.action")
+
+        # We try to link press releases that refer to an original press release number
+        # back to the original press release by using that in the sanction key.
+
+        # In practice often the entity ID and thus sanction ID differs because of
+        # different levels of address details in the press releases.
+        sanction = h.make_sanction(context, entity)
```
jbothma: we don't have any grouping key for this dataset so let's use the date to make sure notices about separate actions (and in fact all actions) are separate sanction entities. Let's update the comment accordingly.

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -0,0 +1,291 @@
...
+    defendants: List[Defendant]
+
+
+PROMPT = f"""
+Extract the defendants or entities subject to OFAC enforcement actions in the attached article.
+NEVER include relief defendants.
+NEVER infer, assume, or generate values that are not directly stated in the source text.
+
+Trading/D.B.A. names which follow a person name but look like company can just be
+aliases of the person. If the name is a person name, use `Person` as the entity_schema.
```
jbothma: let's remove this based on https://opensanctions-corp.slack.com/archives/C07C5G9KZED/p1755695976669969 - I'll add a separate comment about clarifying the prompt for related entities

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -0,0 +1,291 @@
...
+
+Specific fields:
+
+- name: The null name of the entity precisely as expressed in the text. If an acronym follows the name in parentheses, include it as an alias and not as part of the name.
+- entity_schema: {schema_field.description}
+- address: {address_field.description}
+- country: Any countries the entity is indicated to reside, operate, or have been born or registered in. Leave empty if not explicitly stated.
+- status: {status_field.description}
+- notes: {notes_field.description}
+- related_companies: If the defendant is a person and a related company is mentioned in the source text, add it here.
```
jbothma: let's try and clarify this so that we only get related entities where the entity has an ownership or controlling relationship with the related entity. e.g. on https://ofac.treasury.gov/recent-actions/20190425 we got rosneft as related which is cool, but I'm not sure quite whether we want to represent basic business dealings like this. With `transacted or otherwise dealt in new debt` it's also not clear if the business relationship was with rosneft or just underwriting the debt for another provider.

See also https://opensanctions-corp.slack.com/archives/C07C5G9KZED/p1755704163127739 but I think if we can make it fairly sane we can keep it in.

datasets/us/ofac/enforcement_actions/us_ofac_enforcement_actions.yml
```diff
@@ -0,0 +1,64 @@
+title: US OFAC Enforcement Actions
+entry_point: crawler.py
+prefix: ofac-ea
+load_db_uri: ${OPENSANCTIONS_DATABASE_URI}
```
jbothma: fogive my copy pasta. I've since fixed it in cftc.  It's now load_statements: true
https://github.com/opensanctions/opensanctions/blame/main/datasets/us/cftc_enforcement_actions/us_cftc_enforcement_actions.yml#L3

datasets/us/ofac/enforcement_actions/us_ofac_enforcement_actions.yml
```diff
@@ -0,0 +1,64 @@
...
+  acronym: OFAC
+  description: >
+    OFAC is a divison of the United States Department of the Treasury that administers
+    and enforces economic and trade sanctions based on US foreign policy and national
+    security goals against targeted foreign countries and regimes, terrorists,
+    international narcotics traffickers, those engaged in activities related to the
+    proliferation of weapons of mass destruction, and other threats to the national
+    security, foreign policy or economy of the United States.
+  country: us
+  url: https://www.treasury.gov/
```
jbothma:
```suggestion
  url: https://ofac.treasury.gov/
```

datasets/us/ofac/enforcement_actions/us_ofac_enforcement_actions.yml
```diff
@@ -0,0 +1,64 @@
...
+  official: true
+tags:
+  - list.sanction
```
jbothma: interesting... should there be an enforcements tag?

and should this still be tagged sanction because it's in response to sanction violations?

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -0,0 +1,291 @@
...
+        for related_company in item.related_companies:
+            related_company_entity = make_related_company(context, related_company.name)
+            link = make_company_link(
+                context, entity, related_company_entity, related_company.relationship
+            )
+            context.emit(related_company_entity)
+            context.emit(link)
+
+        article = h.make_article(context, url, title=article_name, published_at=date)
+        documentation = h.make_documentation(context, entity, article, date=date)
```
jbothma: It can now take the date from the article

```suggestion
        documentation = h.make_documentation(context, entity, article)
```

datasets/us/ofac/enforcement_actions/crawler.py
jbothma: Just remember to update the prompt too

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -0,0 +1,291 @@
...
+
+# Settlement Agreement
+# Penalty Notice to an Individual
+# Finding of Violation
+# Sanctions Compliance Guidance
+# Issuance of Sanctions Regulations
+# Civil Monetary Penalty
+# Release of OFAC Enforcement Information
+# Release​ of Civil Penalties Information
+# Recent OFAC Actions
```
jbothma, line 46: should this be worked into Status options? Do we still need it? It looks interesting but if we wanna keep it we might wanna have a comment explaining what it means.

datasets/us/ofac/enforcement_actions/crawler.py
nvmbrasserie: I removed it because it always crashed with this change:
```
2025-08-20 14:38:47 [info     ] Prompting 'gpt-4o' for: <div class="content">


<article id="node-10 [zavod.shed.gpt] dataset=us_ofac_enforcement_actions
2025-08-20 14:38:51 [error    ] Runner failed: 4 validation errors for Defendants
defendants.1.address
  Input should be a valid list [type=list_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.11/v/list_type
defendants.1.country
  Input should be a valid list [type=list_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.11/v/list_type
defendants.2.address
  Input should be a valid list [type=list_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.11/v/list_type
defendants.2.country
  Input should be a valid list [type=list_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.11/v/list_type [us_ofac_enforcement_actions] dataset=us_ofac_enforcement_actions
```

jbothma: ah - you'll have to delete the data in your db because it isn't compatible with the model change. But since it isn't in prod yet, we can make the model nice and tidy before we ship it

datasets/us/ofac/enforcement_actions/us_ofac_enforcement_actions.yml
nvmbrasserie: I think it can just become list.enforcement

datasets/us/ofac/enforcement_actions/crawler.py
nvmbrasserie: how should the prompt be updated in this case?

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -0,0 +1,284 @@
...
+        for idx, result in enumerate(search_results):
+            enforcement_date = result.xpath(
+                ".//div[contains(@class,'margin-top-1') and contains(., 'Enforcement Actions')]/text()[normalize-space()]"
+            )
+            assert len(enforcement_date) == 1, "Expected exactly one enforcement date"
+            clean_date = enforcement_date[0].strip().removesuffix(" -")
+            if not enforcements.within_max_age(context, clean_date, MAX_AGE_DAYS):
+                stop_crawl = True
+                break
+            link = links[idx]
```
jbothma: I think a more natural way to get the link for a given search result at this point is to scope the xpath to the search result, so something like 

```
link = result.xpath(".//a[contains(@href, 'recent-actions') and not(contains(@href, 'enforcement-actions'))]/@href")[0]
```

datasets/us/ofac/enforcement_actions/crawler.py
jbothma: I thought it'd need adjustment when it changes to a list, but the text works well enough for a list.

jbothma: it did a nice job here https://ofac.treasury.gov/recent-actions/20230921
looks like entirely unrelated cases, just on the same notice because same date.

datasets/us/ofac/enforcement_actions/us_ofac_enforcement_actions.yml
jbothma: let's remove this until we've confirmed - @pudo should this be `list.regulatory` or do we want to introduce `list.enforcement`?

PR #2680, opened 2025-08-18
[exporters] Put Thing.programId into targets.simple.csv

### LINKED ISSUE #2654: [targets.simple.csv] Add column for programID
We have quite a few people using our (very awkward) `targets.simple.csv` format. I wonder if we could add a column to it, called `program_ids` (or similar), that lists out the programs under which an entity is designated?

leonhandreke:
One slightly awkward design choice here: The `identifiers` column pulls out all values with type identifier, of which `programId` is one. But the documentation for `identifiers` states:

> Identifiers such as corporate registrations, passport numbers or tax identifiers linked to this sanctions target.

So I decided to exclude `programId` from that column. More complexity on our side, but less surprising for users I feel? Comments welcome.

Fixes https://github.com/opensanctions/opensanctions/issues/2654

### DIFF
--- a/zavod/zavod/exporters/simplecsv.py
+++ b/zavod/zavod/exporters/simplecsv.py
@@ -26,6 +26,7 @@ class SimpleCSVExporter(Exporter):
         "sanctions",
         "phones",
         "emails",
+        "program_ids",
         "dataset",
         "first_seen",
         "last_seen",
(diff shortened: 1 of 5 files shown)


pudo (approved):
Really nicely done, and I agree with splitting the sets!

PR #2681, opened 2025-08-18
Fed enforcements articles

jbothma:
- Take Documentation date from article - the date is in Documentation to make listing articles on the website easier
- Add articles linked from the us_fed_enforcements CSV. Skip title and pub date for PDFs and ancient pages

### DIFF
--- a/zavod/zavod/helpers/articles.py
+++ b/zavod/zavod/helpers/articles.py
@@ -37,10 +37,12 @@ def make_documentation(
     entity: Entity,
     article: Entity,
     key_extra: Optional[str] = None,
+    date: Optional[str] = None,
 ) -> Entity:
     """
     Creates a documentation entity to link an article to a related entity.
-    The article's publishedAt date is added to the Documentation date property.
+    The article's publishedAt date is added to the Documentation date property
+    unless the date argument is provided.
 
     This is useful to link one or more entities to an article they were mentioned in.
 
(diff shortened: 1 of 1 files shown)


datasets/us/fed_enforcements/crawler.py
```diff
@@ -119,8 +148,11 @@ def crawl_item(context: Context, original_filename: str, input_dict: Dict[str, s
         if is_active:
             entity.add("topics", "reg.action")
 
+        documentation = h.make_documentation(context, entity, article)
```
pudo, line 151: Can the documentation also have the date?

pudo: Could be `article.get`

pudo: Perhaps the helper could even default that in when present

zavod/zavod/helpers/articles.py
```diff
@@ -62,5 +61,5 @@ def make_documentation(
     )
     documentation.add("entity", entity)
     documentation.add("document", article)
-    h.apply_date(documentation, "date", date)
+    documentation.set("date", article.get("publishedAt"))
```
pudo: lol lol ok ... maybe we leave the override as an option? I can't see a reason to.

datasets/us/fed_enforcements/crawler.py
jbothma: I love how obvious a refactor becomes the second time I use a thing

PR #2682, opened 2025-08-18
Fixes-18-08

nvmbrasserie:


datasets/md/companies/parse.py
```diff
@@ -56,7 +56,7 @@ def parse_directors(
         except ValueError:
             pass
 
-        director = director.strip()
+        director = director.strip(".").strip()
```
jbothma: let's move this to after the make_id, before dir.add("name" to prevent a re-key of entities with .'s at the end of the name

I'll do it cos I'm already adding something similar for non-profits

PR #2686, opened 2025-08-19
Fixes-19-08

nvmbrasserie:


datasets/ua/war_sanctions/ua_war_sanctions.yml
```diff
@@ -499,6 +512,8 @@ lookups:
           - 04.01.20205)
           - 01.11.22023)
           - 17.03.20205)
+          - 1804807)
+          - 6299069)
```
jbothma, line 516: can you tell where these are coming from? Looks like our date/regnum splitting is a bit confused

nvmbrasserie: it's a typo on their side
Screenshot 2025-08-19 at 15 54 04

PR #2690, opened 2025-08-20
Fixes 20 08

nvmbrasserie:


datasets/ua/war_sanctions/crawler.py
```diff
@@ -117,6 +122,7 @@ def rostec_affiliate_id_parts(abbrev_name):
 
 
 def crawl_index_page(context: Context, index_page, data_type, program):
+    print(index_page)
```
leonhandreke: Stray print

datasets/us/fcc_covered_list/us_fcc_covered_list.yml
```diff
@@ -53,6 +53,8 @@ data:
   lang: eng
 dates:
   formats: "%B %d, %Y"
+http:
+  user_agent: "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.2 Safari/605.1.15"
```
leonhandreke: We usually try to append "zavod/opensanctions.org" at the end, just to identify ourselves. Does that not work here? If so, let's quickly mention it. I don't think we want to piss them off.

datasets/us/mi/med_exclusions/us_mi_med_exclusions.yml
```diff
@@ -38,6 +38,8 @@ data:
   format: HTML
 dates:
   formats: ["%m/%d/%Y"]
+http:
```
leonhandreke, line 41: Same here, can we append zavod/opensanctions.org to this normal-looking UA?

PR #2692, opened 2025-08-20
[un_sc_sanctions] Fix mapping of name list to name parts

leonhandreke:
Previously, using `h.apply_name`, the family name would end up in `secondName` or `middleName`. I've observed names that are not the first or last in the list to be either more first names, patronymics or last names. Because we can't say, don't assign them to any of the semantic fields in the FtM entity, just put them in the name.


pudo (approved):
Yep, that makes sense and is more defensive.

PR #2702, opened 2025-08-22
US OFAC Press Releases

nvmbrasserie:
opensanctions/crawler-planning#367

### DIFF
--- a/datasets/us/ofac/press_releases/crawler.py
+++ b/datasets/us/ofac/press_releases/crawler.py
@@ -15,14 +15,8 @@ from zavod.stateful.review import (
     html_to_text_hash,
 )
 
-# Rough conversion: 1 GPT token ≈ 4 chars English text
-TOKEN_CHAR_RATIO = 4
-MAX_TOKENS = 3000
-MAX_CHARS = MAX_TOKENS * TOKEN_CHAR_RATIO
-
-
 Schema = Literal["Person", "Company", "LegalEntity", "Vessel"]
-NAME_XPATH = "//h2[@class='uswds-page-title']/text()"
+NAME_XPATH = "//h2[@class='uswds-page-title']"
 CONTENT_XPATH = "//article[@class='entity--type-node']"
 DATE_XPATH = "//article[@class='entity--type-node']//time[@class='datetime']/@datetime"
 
@@ -62,7 +56,7 @@ Output each entity with these fields:
 - name: Exact name as written in the article. If followed by an acronym in parentheses, store that acronym as an alias, not in the name.
 - entity_schema: {schema_field.description}
 - aliases: Other names or acronyms the entity is referred to in the article.
-- nationality: Nationality of the designee is an individual and it is stated.
+- nationality: Nationality of the designee if they are an individual and it is stated.
 - country: Countries explicitly mentioned as residence, registration, or operation. Leave empty if not stated.
 - related_url: URLs mentioned in the article specifically associated with the entity.  
   • If multiple URLs are present, link each one only to the entity it is associated with.  
@@ -156,10 +150,13 @@ def crawl_item(context, item, date, url, article_name):
 def crawl_press_release(context: Context, url: str) -> None:
     article = context.fetch_html(url, cache_days=7)
     article.make_links_absolute(context.data_url)
-    if article is None:
-        return
-    article_name = article.xpath(NAME_XPATH)[0]
+    names = article.xpath(NAME_XPATH)
+    assert len(names) == 1, f"Expected 1 title, got {len(names)}"
+    article_name = names[0].text_content().strip()
     article_content = article.xpath(CONTENT_XPATH)
+    for img in article.xpath(".//img"):
+        if img.get("src").startswith("data:image"):
+            img.getparent().remove(img)
     assert len(article_content) == 1
     article_element = article_content[0]
     date = article.xpath(DATE_XPATH)[0]
--- a/datasets/us/ofac/press_releases/crawler.py
+++ b/datasets/us/ofac/press_releases/crawler.py
@@ -64,8 +64,8 @@ Output each entity with these fields:
 - aliases: Other names or acronyms the entity is referred to in the article.
 - nationality: Nationality of the designee if stated.
 - country: Countries explicitly mentioned as residence, registration, or operation. Leave empty if not stated.
-- related_url: URLs mentioned in the article that directly reference the entity.  
-  • If multiple URLs are present, link each one only to the entity it explicitly refers to.  
+- related_url: URLs mentioned in the article specifically associated with the entity.  
+  • If multiple URLs are present, link each one only to the entity it it is associated with.  
   • If no URL is provided for an entity, leave this field empty.  
   • Do not invent, infer, or alter URLs.
 """
--- a/datasets/us/ofac/press_releases/crawler.py
+++ b/datasets/us/ofac/press_releases/crawler.py
@@ -190,6 +190,43 @@ def check_something_changed(
         return False
 
 
+def crawl_item(context, item, date, url, article_name):
+    entity = context.make(item.entity_schema)
+    entity.id = context.make_id(item.name, item.address, item.country)
+    entity.add("name", item.name, origin=DEFAULT_MODEL)
+    if item.address != item.country:
+        entity.add("address", item.address, origin=DEFAULT_MODEL)
+    entity.add("country", item.country, origin=DEFAULT_MODEL)
+    entity.add("alias", item.aliases, origin=DEFAULT_MODEL)
+    entity.add("topics", "reg.action")
+
+    # We use the date as a key to make sure notices about separate actions are separate sanction entities
+    sanction = h.make_sanction(context, entity, date)
+    h.apply_date(sanction, "date", date)
+    sanction.set("sourceUrl", url)
+    sanction.add("sourceUrl", item.pdf_url, origin=DEFAULT_MODEL)
+    sanction.add("summary", item.notes, origin=DEFAULT_MODEL)
+
+    for related_company in item.related_companies:
+        related_company_entity = make_related_company(context, related_company.name)
+        link = make_company_link(
+            context,
+            entity,
+            related_company_entity,
+            related_company.relationship,
+        )
+        context.emit(related_company_entity)
+        context.emit(link)
+
+    article = h.make_article(context, url, title=article_name, published_at=date)
+    documentation = h.make_documentation(context, entity, article)
+
+    context.emit(entity)
+    context.emit(sanction)
+    context.emit(article)
+    context.emit(documentation)
+
+
 def crawl_enforcement_action(context: Context, url: str) -> None:
     article = context.fetch_html(url, cache_days=1)
     article.make_links_absolute(context.data_url)
@@ -233,44 +270,7 @@ def crawl_enforcement_action(context: Context, url: str) -> None:
             return
 
         for item in review.extracted_data.defendants:
-            entity = context.make(item.entity_schema)
-            entity.id = context.make_id(item.name, item.address, item.country)
-            entity.add("name", item.name, origin=DEFAULT_MODEL)
-            if item.address != item.country:
-                entity.add("address", item.address, origin=DEFAULT_MODEL)
-            entity.add("country", item.country, origin=DEFAULT_MODEL)
-            entity.add("alias", item.aliases, origin=DEFAULT_MODEL)
-            entity.add("topics", "reg.action")
-
-            # We use the date as a key to make sure notices about separate actions are separate sanction entities
-            sanction = h.make_sanction(context, entity, date)
-            h.apply_date(sanction, "date", date)
-            sanction.set("sourceUrl", url)
-            sanction.add("sourceUrl", item.pdf_url, origin=DEFAULT_MODEL)
-            sanction.add("summary", item.notes, origin=DEFAULT_MODEL)
-
-            for related_company in item.related_companies:
-                related_company_entity = make_related_company(
-                    context, related_company.name
-                )
-                link = make_company_link(
-                    context,
-                    entity,
-                    related_company_entity,
-                    related_company.relationship,
-                )
-                context.emit(related_company_entity)
-                context.emit(link)
-
-            article = h.make_article(
-                context, url, title=article_name, published_at=date
-            )
-            documentation = h.make_documentation(context, entity, article)
-
-            context.emit(entity)
-            context.emit(sanction)
-            context.emit(article)
-            context.emit(documentation)
+            crawl_item(context, item, date, url, article_name)
 
 
 def crawl(context: Context):
--- a/datasets/us/ofac/press_releases/crawler.py
+++ b/datasets/us/ofac/press_releases/crawler.py
@@ -16,6 +16,11 @@ from zavod.stateful.review import (
     html_to_text_hash,
 )
 
+# Rough conversion: 1 GPT token ≈ 4 chars English text
+TOKEN_CHAR_RATIO = 4
+MAX_TOKENS = 3000
+MAX_CHARS = MAX_TOKENS * TOKEN_CHAR_RATIO
+
 
 Schema = Literal["Person", "Company", "LegalEntity", "Vessel"]
 NAME_XPATH = "//h2[@class='uswds-page-title']/text()"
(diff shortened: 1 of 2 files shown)


datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,300 @@
...
+    related_url: List[str] = []
+
+
+class Defendants(BaseModel):
+    defendants: List[Defendant]
+
+
+PROMPT = f"""
+Extract the defendants, entities and vessels mentioned in OFAC press release in the attached article.
+NEVER include relief defendants.
```
pudo: not relevant

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,300 @@
...
+)
+notes_field = Field(default=None, description=("Only used if `status` is `Other`."))
+
+
+class RelatedCompany(BaseModel):
+    name: str
+    relationship: str
+
+
+class Defendant(BaseModel):
```
pudo: Can we rename this to "Target" or "Designee" or "Sanctioned"?

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,300 @@
...
+        "The status of the enforcement action notice."
+        " If `Other`, add the text used as the status in the source to `notes`."
+    )
+)
+notes_field = Field(default=None, description=("Only used if `status` is `Other`."))
+
+
+class RelatedCompany(BaseModel):
+    name: str
+    relationship: str
```
pudo: Don't really need this.

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,300 @@
...
+        "- 'Person', if the name refers to an individual."
+        "- 'Vessel', if the name refers to a ship or vessel."
+        "- 'LegalEntity', for companies, organizations, or when unclear if the entity is a person or company."
+        "Never invent new schema labels."
+    )
+)
+address_field = Field(
+    default=[],
+    description=("The addresses or even just the districts/states of the defendant."),
+)
```
pudo: Let's do nationality instead?

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,300 @@
...
+    entity.add("country", item.country, origin=DEFAULT_MODEL)
+    entity.add("alias", item.aliases, origin=DEFAULT_MODEL)
+    entity.add("topics", "reg.action")
+
+    # We use the date as a key to make sure notices about separate actions are separate sanction entities
+    sanction = h.make_sanction(context, entity, date)
+    h.apply_date(sanction, "date", date)
+    sanction.set("sourceUrl", url)
+    sanction.add("sourceUrl", item.related_url, origin=DEFAULT_MODEL)
+    sanction.add("summary", item.notes, origin=DEFAULT_MODEL)
```
pudo: All these entities will have specific sanctions objects from OFAC, let's not make this messy with this

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,300 @@
...
+
+def crawl_item(context, item, date, url, article_name):
+    entity = context.make(item.entity_schema)
+    entity.id = context.make_id(item.name, item.address, item.country)
+    entity.add("name", item.name, origin=DEFAULT_MODEL)
+    if item.address != item.country:
+        entity.add("address", item.address, origin=DEFAULT_MODEL)
+    entity.add("country", item.country, origin=DEFAULT_MODEL)
+    entity.add("alias", item.aliases, origin=DEFAULT_MODEL)
+    entity.add("topics", "reg.action")
```
pudo: Does not apply

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,300 @@
...
+    article = h.make_article(context, url, title=article_name, published_at=date)
+    documentation = h.make_documentation(context, entity, article)
+
+    context.emit(entity)
+    context.emit(sanction)
+    context.emit(article)
+    context.emit(documentation)
+
+
+def crawl_enforcement_action(context: Context, url: str) -> None:
```
pudo: Rename

pudo (changes requested):
The copy from CFTC is a bit direct, lets make this more applicable to the genre it is.

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,240 @@
...
+    article = context.fetch_html(url, cache_days=1)
+    article.make_links_absolute(context.data_url)
+    if article is None:
+        return
+    article_name = article.xpath(NAME_XPATH)[0]
+    article_content = article.xpath(CONTENT_XPATH)
+    assert len(article_content) == 1
+    article_element = article_content[0]
+    date = article.xpath(DATE_XPATH)[0]
+    article_html = tostring(article_element, pretty_print=True).decode("utf-8")
```
jbothma: I've since learnt this is better https://github.com/opensanctions/opensanctions/pull/2749

```suggestion
    article_html = tostring(article_element, pretty_print=True, encoding="unicode")
```

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,240 @@
...
+NEVER infer, assume, or generate values that are not directly stated in the source text.
+If the name is a person name, use `Person` as the entity_schema.
+Output each entity with these fields:
+- name: Exact name as written in the article. If followed by an acronym in parentheses, store that acronym as an alias, not in the name.
+- entity_schema: {schema_field.description}
+- aliases: Other names or acronyms the entity is referred to in the article.
+- nationality: Nationality of the designee if stated.
+- country: Countries explicitly mentioned as residence, registration, or operation. Leave empty if not stated.
+- related_url: URLs mentioned in the article that directly reference the entity.  
+  • If multiple URLs are present, link each one only to the entity it explicitly refers to.
```
jbothma:
```suggestion
- related_url: URLs mentioned in the article specifically associated with the entity.  
  • If multiple URLs are present, link each one only to the entity it it is associated with.  
```

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,240 @@
...
+            return
+
+        for item in review.extracted_data.designees:
+            crawl_item(context, item, date, url, article_name)
+
+
+def crawl(context: Context):
+    page = 0
+    while True:
+        base_url = f"https://ofac.treasury.gov/press-releases?page={page}"
```
jbothma, line 193: a common bug in websites is to just produce a page for a given url, even when the page number in the query string is out of range. if they introduce that bug in the future, this would just keep running until we notice that it hasn't exported in a while.

Thats why I prefer building the pagination in a way that takes something very explicit about there being another page of results into account to continue looping. e.g. in this case on the last page, the next link is entirely gone. And it's hard to imagine a selector like `.//a[contains(@class, 'usa-pagination__next-page')]` matching anything unless their pagination is functionally broken (as opposed to functional but technically broken above)

As an alternative, we've done that assertion like `assert page < 1000`

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,240 @@
...
+    article = h.make_article(context, url, title=article_name, published_at=date)
+    documentation = h.make_documentation(context, entity, article)
+
+    context.emit(entity)
+    context.emit(article)
+    context.emit(documentation)
+
+
+def crawl_press_release(context: Context, url: str) -> None:
+    article = context.fetch_html(url, cache_days=1)
```
jbothma: we could cache this for a week and run the crawler daily. Then we'd quickly get new entities without fetching them all every day.

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,240 @@
...
+    article.make_links_absolute(context.data_url)
+    if article is None:
+        return
```
jbothma: in what circumstance might this be needed?

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,240 @@
...
+        if len(article_html) > MAX_CHARS
+        else [article_html]
+    )
+    context.log.info(
+        f"Article length {len(article_html)} chars — {'splitting into sections' if len(sections) > 1 else 'no split needed'}."
+    )
+
+    for i, section_html in enumerate(sections, 1):
+        section_url = f"{url}#section-{i}"  # distinguish reviews per section
+        review = get_or_request_review(context, section_html, section_url, url)
```
jbothma: We're kinda making up a URL here, but there's no need for it to be a valid URL - you just want a unique key per section.

```suggestion
        section_key = slugify(url, "section", i)  # distinguish reviews per section
        review = get_or_request_review(context, section_key, section_url, url)
```

then also update the variable name in get_or_request_review

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,240 @@
...
+    request_review,
+    get_review,
+    model_hash,
+    html_to_text_hash,
+)
+
+# Rough conversion: 1 GPT token ≈ 4 chars English text
+TOKEN_CHAR_RATIO = 4
+MAX_TOKENS = 3000
+MAX_CHARS = MAX_TOKENS * TOKEN_CHAR_RATIO
```
jbothma: is it still too big and requires splitting we prompt with text_content() instead of the html?

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,240 @@
...
+    sections = (
+        split_article_by_headers(article_element)
+        if len(article_html) > MAX_CHARS
+        else [article_html]
+    )
+    context.log.info(
+        f"Article length {len(article_html)} chars — {'splitting into sections' if len(sections) > 1 else 'no split needed'}."
+    )
+
+    for i, section_html in enumerate(sections, 1):
```
jbothma: what problem are we solving by splitting? It'd be nice if we can avoid it.

if it's this:
```
openai.LengthFinishReasonError: Could not parse response content as the length limit was reached - CompletionUsage(completion_tokens=3000, prompt_tokens=8804, total_tokens=11804, completion_tokens_details=CompletionTokensDetails(accepted_prediction_tokens=0, audio_tokens=0, reasoning_tokens=0, rejected_prediction_tokens=0), prompt_tokens_details=PromptTokensDetails(audio_tokens=0, cached_tokens=0))
```

does `run_types_text_prompt(..., max_tokens=20_000)` solve it? It worked as far as I ran it and encountered this issue. The re-prompt upon changes will need the same argument.

If it's the source_value db field being too small, https://github.com/opensanctions/opensanctions/pull/2752 should fix that.

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,240 @@
...
+                prompt_result=prompt_result.model_dump(),
+            )
+        return True
+    else:
+        return False
+
+
+def crawl_item(context, item, date, url, article_name):
+    entity = context.make(item.entity_schema)
+    entity.id = context.make_id(item.name, item.address, item.country)
```
jbothma: .address doesn't exist any more

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,240 @@
...
+
+PROMPT = f"""
+Extract the designees, entities and vessels mentioned in OFAC press release in the attached article.
+NEVER infer, assume, or generate values that are not directly stated in the source text.
+If the name is a person name, use `Person` as the entity_schema.
+Output each entity with these fields:
+- name: Exact name as written in the article. If followed by an acronym in parentheses, store that acronym as an alias, not in the name.
+- entity_schema: {schema_field.description}
+- aliases: Other names or acronyms the entity is referred to in the article.
+- nationality: Nationality of the designee if stated.
```
jbothma:
```suggestion
- nationality: Nationality of the designee if they are an individual and it is stated.
```

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,240 @@
...
+    entity = context.make(item.entity_schema)
+    entity.id = context.make_id(item.name, item.address, item.country)
+    entity.add("name", item.name, origin=DEFAULT_MODEL)
+    entity.add("nationality", item.nationality, origin=DEFAULT_MODEL)
+    entity.add("country", item.country, origin=DEFAULT_MODEL)
+    entity.add("alias", item.aliases, origin=DEFAULT_MODEL)
+    entity.add("sourceUrl", item.related_url, origin=DEFAULT_MODEL)
+    entity.add("sourceUrl", url)
+
+    article = h.make_article(context, url, title=article_name, published_at=date)
```
jbothma, line 144: in the exported data the title seems to be blank

datasets/us/ofac/press_releases/crawler.py
jbothma: ignore this one sorry - I think sending the html is nice to get the links

datasets/us/ofac/press_releases/crawler.py
jbothma: With a bigger max_tokens I get to https-home-treasury-gov-news-press-releases-jy1507 and get 

```
openai.BadRequestError: Error code: 400 - {'error': {'message': "This model's maximum context length is 128000 tokens. However, your messages resulted in 1788216 tokens (including 239 in the response_format schemas.). Please reduce the length of the messages or schemas.", 'type': 'invalid_request_error', 'param': 'messages', 'code': 'context_length_exceeded'}}
```

which is a bit surprising because it doesn't look like such a long article.

but it's because they're embedding images with data URLs:

```
<img alt id="Picture_x0020_5" src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAPkAAAD5CAIAAABqLQTHAAAAAXNSR0IArs4c6QA
```

I think it's safe te delete those from the article before prompting/saving.

jbothma: e.g. 
```diff
     article_element = article_content[0]
+    for img in article.xpath(".//img"):
+        if img.get("src").startswith("data:image"):
+            img.getparent().remove(img)
```

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,212 @@
...
+        assert len(table) == 1, "Expected exactly one table in the document"
+        for row in h.parse_html_table(table[0]):
+            links = h.links_to_dict(row.pop("press_release_link"))
+            url = next(iter(links.values()))
+            # Filter out unwanted download/media links
+            if "/news/press-releases/" not in url:
+                continue  # skip this row
+            crawl_press_release(context, url)
+        page += 1
+    assert page < 200
```
jbothma: this has to be inside the loop to be effective at detecting an infinite loop. So just indent one level.

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,212 @@
...
+
+def get_or_request_review(context, html_part, article_key, url):
+    review = get_review(context, Designees, article_key, MIN_MODEL_VERSION)
+    if review is None:
+        prompt_result = run_typed_text_prompt(
+            context,
+            PROMPT,
+            html_part,
+            Designees,
+            16384,  # This model supports at most 16384 completion tokens
```
jbothma: check_something_changed needs the same arguments so let's factor this out to a constant and update the comment to say `gpt-4o` instead of `This model` since the limit is specific to it.

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,212 @@
...
+        return True
+    else:
+        return False
+
+
+def crawl_item(context, item, date, url, article_name):
+    entity = context.make(item.entity_schema)
+    entity.id = context.make_id(item.name, item.country)
+    entity.add("name", item.name, origin=DEFAULT_MODEL)
+    entity.add("nationality", item.nationality, origin=DEFAULT_MODEL)
```
jbothma:
```
    entity.add("nationality", item.nationality, origin=DEFAULT_MODEL)
  File "/Users/jdb/projects/opensanctions/opensanctions/env/lib/python3.12/site-packages/followthemoney/statement/entity.py", line 222, in add
    prop_name = self._prop_name(prop, quiet=quiet)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jdb/projects/opensanctions/opensanctions/env/lib/python3.12/site-packages/followthemoney/proxy.py", line 107, in _prop_name
    raise InvalidData(msg % (self.schema, prop))
followthemoney.exc.InvalidData: Unknown property (<Schema('Company')>): nationality
```

I'm surprised this is failing when nationality is empty, but it looks like this wants an `if item.nationality` so we don't even try to set when it's not there (to hopefully always avoid setting it on companies.

Or just make the prop "country" when the schema isn't Person

datasets/us/ofac/press_releases/crawler.py
```diff
@@ -0,0 +1,213 @@
...
+        return False
+
+
+def crawl_item(context, item, date, url, article_name):
+    entity = context.make(item.entity_schema)
+    entity.id = context.make_id(item.name, item.country)
+    entity.add("name", item.name, origin=DEFAULT_MODEL)
+    if item.nationality:
+        # Add nationality only for 'Person' schema
+        entity.add("nationality", item.nationality, origin=DEFAULT_MODEL)
```
jbothma: The comment didn't quite match the implementation

```suggestion
    nationality_prop = "nationality"
    if item.entity_schema != "Person":
        nationality_prop = "country"
    entity.add(nationality_prop, item.nationality, origin=DEFAULT_MODEL)
```

jbothma (approved):

PR #2704, opened 2025-08-22
Daily fixes 22 08

nvmbrasserie:


zavod/zavod/shed/un_sc.py
```diff
@@ -95,7 +95,7 @@ def make_entity(context: Context, prefix: str, schema: str, node: Element) -> En
         entity.add("firstName", names[0])
         entity.add("lastName", names[-1])
         # make_name (which is just a fancy wrapper around " ".join) to generate the full name.
-        name_args = {f"name{i}": name for i, name in enumerate(names)}
+        name_args = {f"name{i+1}": name for i, name in enumerate(names)}
```
leonhandreke, line 98: Sorry :(

datasets/us/mi/med_exclusions/crawler.py
```diff
@@ -4,6 +4,16 @@
 
 from zavod import Context, helpers as h
 
+HEADERS = {
+    "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
```
leonhandreke: Does it need all of them? Could we strip them down to a smaller sufficient set? Like, maybe the Accept-Header is enough...? (same for FCC above)

nvmbrasserie: the thing is that it fails only in prod, and I think it's easier to push them all in. locally it's happy with user_agent

leonhandreke: Ah, that makes sense.

datasets/us/mi/med_exclusions/crawler.py
```diff
@@ -3,16 +3,9 @@
...
 
-HEADERS = {
-    "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
-    "Accept-Encoding": "gzip, deflate, br",
-    "Accept-Language": "en-GB,en;q=0.9",
-    "Priority": "u=0, i",
-    "Sec-Fetch-Dest": "document",
-    "Sec-Fetch-Mode": "navigate",
-    "Sec-Fetch-Site": "none",
-}
```
nvmbrasserie: it was failing in prod and locally with both headers and data; switched it to zyte to make it happy.

PR #2721, opened 2025-08-25
UA War Sanctions API

nvmbrasserie:


jbothma:
The problem with that approach is that it's an enormous amount of dedupe to
do by hand, and secondly they'll show up as new matches and again as
deletions for customers, while a more graceful migration where both IDs are
available at the same time could mean we can give a graceful changeover, at
least for the users that implement our recommended identifier handling.

1545 "ADD Company"
 743 "ADD LegalEntity"
  12 "ADD Organization"
1664 "ADD Ownership"
1270 "ADD Person"
4625 "ADD Sanction"
2907 "ADD UnknownLink"
1067 "ADD Vessel"
11191 "DEL "

So we might then have to run all the old code and all the new code once, do
the dedupe, then remove the old code.

But are you sure we won't be able to get at least _most_ of the old IDs
right? Presumably they're generating the site and the API data from the
same database.

captain.id = context.make_id(name, dob_pob, tax_number)
vessel.id = context.make_id(vessel_name, vessel_category, vessel_url)
vessel.id = context.make_id(name, imo_num)
other.id = context.make_id(name, country)
person.id = context.make_id(names, positions)
legal_entity.id = context.make_id(name, name_abbr, reg_num)

I'm not super fussed which approach we use, but I don't think we should
release this as a blank slate.

We don't care about migrating interstitial entities (unknown link,
ownership).

pudo:
FWIW I'm fully anticipating having the re-dedupe this, that's just another day in paradise. But my hope would be that we can first run the API crawler from a fake/stub dataset (`ua_war_sanctions_api.yml`), then dedupe it by manually pushing it into the scope, and then swap out the current crawler code for the new entities. Can we do that?

datasets/ua/war_sanctions/ua_war_sanctions.yml
```diff
@@ -66,11 +66,11 @@ tags:
   - sector.maritime
 url: https://war-sanctions.gur.gov.ua/en
 data:
-  url: https://war-sanctions.gur.gov.ua/en
-  format: HTML
+  url: "https://war-sanctions.gur.gov.ua/api/v1/"
+  format: JSON
```
jbothma: noting here for lack of a better place so we remember the first version of this crawler should be an entirely separate crawler. If it was only a handful of entities we could maybe have just set a more specific dataset name attribute but since it's so many, it might take a couple of days to dedupe and we might want updates in the meantime so let's just copy this to `ua_war_sanctions_api` and then make `ua_war_sanctions` become this implementation after dedupe.

datasets/ua/war_sanctions/crawler.py
```diff
@@ -1,706 +1,479 @@
...
-from normality import collapse_spaces
 
 from zavod import Context, helpers as h
-from zavod.entity import Entity
-from zavod.shed.zyte_api import fetch_html
 
+# Make sure variables don't get interpolated by the shell.
+WS_CID = env.get("OPENSANCTIONS_WS_CID")
+WS_API_KEY = env.get("OPENSANCTIONS_WS_API_KEY")
+WS_API_DOCS = env.get("OPENSANCTIONS_WS_API_DOCS")
```
jbothma: let's make this a bit more specific, e.g. `OPENSANCTIONS_UA_WS_...`

datasets/ua/war_sanctions/crawler.py
```diff
@@ -1,706 +1,479 @@
...
-    other.add_cast("Organization", "imoNumber", imo_number)
-    other.add("country", country)
-    context.emit(other)
-
+):
     relation = context.make(rel_schema)
-    relation.id = context.make_id(entity.id, rel_role, other.id)
-    relation.add(from_prop, entity.id)
-    relation.add(to_prop, other.id)
+    relation.id = context.make_id(object_id, rel_role, subject_id)
```
jbothma: we need more info in the relationship id. Perhaps the schema? Or if there's a start/end date, that could help?

```
2025-09-02 11:57:26 [error    ] Assemble error: ua-ws-0aec7fb4364d5c1ae9bf6ba94307317d9d388144: No common schema: <Schema('Ownership')> and UnknownLink [zavod.store] dataset=ua_war_sanctions statements=[['ua_war_sanctions', 'ua-ws-0aec7fb4364d5c1ae9bf6ba94307317d9d388144', 'Ownership', 'owner', 'ua-ws-entity-835'], ['ua_war_sanctions', 'ua-ws-0aec7fb4364d5c1ae9bf6ba94307317d9d388144', 'Ownership', 'role', 'c/o'], ['ua_war_sanctions', 'ua-ws-0aec7fb4364d5c1ae9bf6ba94307317d9d388144', 'Ownership', 'asset', 'ua-ws-entity-836'], ['ua_war_sanctions', 'ua-ws-0aec7fb4364d5c1ae9bf6ba94307317d9d388144', 'Ownership', 'id', 'f4d093db80b187c7ae7a00562e97ab57af92a70b'], ['ua_war_sanctions', 'ua-ws-0aec7fb4364d5c1ae9bf6ba94307317d9d388144', 'UnknownLink', 'subject', 'ua-ws-entity-835'], ['ua_war_sanctions', 'ua-ws-0aec7fb4364d5c1ae9bf6ba94307317d9d388144', 'UnknownLink', 'id', 'ebae81a76b07f7c72c2569225afa2a418ba245ea'], ['ua_war_sanctions', 'ua-ws-0aec7fb4364d5c1ae9bf6ba94307317d9d388144', 'UnknownLink', 'object', 'ua-ws-entity-836'], ['ua_war_sanctions', 'ua-ws-0aec7fb4364d5c1ae9bf6ba94307317d9d388144', 'UnknownLink', 'role', 'c/o']]
```

datasets/ua/war_sanctions/crawler.py
```diff
@@ -1,706 +1,479 @@
...
-        return next_link.get("href")
+def crawl_manager(context: Context, management_data, program, entity_type: str):
+    manager = context.make("Company")
+    manager.id = make_id(context, entity_type, management_data.pop("id"))
+    manager.add("name", management_data.pop("name"))
+    # We null falsy names via the lookups (and we end up with some loose ends because of that)
+    # Linked companies may not exist:
+    # 'ua-ws-entity-22',Company,'Company'
+    # 'ua-ws-entity-32',Company,'Company'
+    # 'ua-ws-entity-238',Company,'Company'
```
nvmbrasserie: @leonhandreke  could you take a look at it? what would be a good way of fixing it? that's one of the reasons validation is not being happy

datasets/ua/war_sanctions/api/ua_war_sanctions_api.yml
```diff
@@ -0,0 +1,201 @@
...
+    options:
+      - match: 
+          - "-"
+          - "Unknown"
+        value: null
+      - regex: 
+          - "Unknown (.+)"
+          # Rptd -> Reported https://www.insurancejournal.com/news/international/2024/03/20/765613.htm
+          - "Rptd Sold (.+)"
+        value: null
```
nvmbrasserie: since we nullify some names here, it messes up the relationship establishment links, and we end up with empty owners.

datasets/ua/war_sanctions/api/crawler.py
```diff
@@ -0,0 +1,481 @@
...
+        token = generate_token(WS_CID, WS_API_KEY)
+        headers = {"Authorization": token}
+        endpoint = link_info["endpoint"]
+        data_type = link_info["type"]
+        program = link_info["program"]
+
+        url = f"{context.data_url}{endpoint}"
+        response = context.fetch_json(url, headers=headers, cache_days=1)
+        if not response or response.get("code") != 0:
+            context.log.warn("No valid data to parse")
```
leonhandreke: Should this be an error? And a `continue`?

leonhandreke:
I pushed the required env vars to prod, see https://github.com/opensanctions/operations/commit/db63f4dfaad672a4e9bf076221b5797f5d2dbd28

datasets/ua/war_sanctions/api/crawler.py
```diff
@@ -0,0 +1,489 @@
...
+        person_data, ["sanctions", "documents", "category", "sport", "places"]
+    )
+
+
+def crawl_legal_entity(context: Context, company_data, program, entity_type: str):
+    legal_entity = context.make("LegalEntity")
+    legal_entity.id = make_id(context, entity_type, company_data.pop("id"))
+    legal_entity.add("name", company_data.pop("name"))
+    name_abbr = company_data.pop("short")
+    if len(name_abbr) < 11:
```
leonhandreke, line 293: Could you add a comment about this magic number?

datasets/ua/war_sanctions/api/crawler.py
```diff
@@ -0,0 +1,489 @@
...
+    vessel.add("type", vessel_data.pop("type"))
+    vessel.add("description", vessel_data.pop("info"))
+    vessel.add("callSign", vessel_data.pop("callsign"))
+    vessel.add("flag", vessel_data.pop("flag"))
+    vessel.add("mmsi", vessel_data.pop("mmsi"))
+    vessel.add("buildDate", vessel_data.pop("year"))
+    vessel.add("grossRegisteredTonnage", vessel_data.pop("weight"))
+    vessel.add("deadweightTonnage", vessel_data.pop("dwt"))
+    photo_url = vessel_data.pop("photo")
+    if "no-ship-photo" not in photo_url:
```
leonhandreke, line 364: I think it's cool that they have a photo for it, but a) it's not really a source for the data and b) how comfortable are we publishing these URLs? Are they publicly accessible?

I don't really have a good idea for where to put them though :( But maybe it's okay to drop it?

datasets/ua/war_sanctions/api/crawler.py
```diff
@@ -0,0 +1,489 @@
...
+    sanction = h.make_sanction(context, vessel)
+    sanction.add("program", program)
+    sanction.add("sourceUrl", vessel_data.pop("links"))
+
+    context.emit(vessel)
+    context.emit(sanction)
+
+    for role in ["commerce_manager", "security_manager", "owner"]:
+        party_info = vessel_data.pop(role, None)
+        crawl_ship_relation(context, party_info, vessel.id, role)
```
leonhandreke, line 386: (nit) small  style thing, but I think it's more elegant to check here if party_info is None and don't call crawl_ship_relation if None.

leonhandreke:
Sorry, that was a dumb comment, now I found some of them.

leonhandreke:
OK, so upon further investigation, 238 and 32 are there but have no properties - unclear what we do with them? I'd be okay with not nulling their names for now, but don't really know what a "good" solution would look like. Maybe just special case a `country: ru`?

1272 and 1425 still missing, have we seen them anywhere?

datasets/ua/war_sanctions/api/crawler.py
```diff
@@ -0,0 +1,494 @@
...
+        )
+
+
+def check_updates(context: Context):
+    doc = context.fetch_html(WS_API_DOCS_URL)
+    # Have any new sections been added?
+    change_log = doc.xpath(".//main[@class='relative']")
+    assert len(change_log) == 1, change_log
+    h.assert_dom_hash(change_log[0], "18328f23fa487338c88a408674d1c4682a1f32c6")
+    # Kidnappers:
```
leonhandreke, line 444: They seem to have a changelog in the documentation, so maybe copying this list here is not needed?

datasets/ua/war_sanctions/api/crawler.py
```diff
@@ -0,0 +1,494 @@
...
+WS_API_CLIENT_ID = env.get("OPENSANCTIONS_UA_WS_API_CLIENT_ID")
+WS_API_KEY = env.get("OPENSANCTIONS_UA_WS_API_KEY")
+# We keep these two secret because they were shared with us confidentially
+WS_API_DOCS_URL = env.get("OPENSANCTIONS_UA_WS_API_DOCS_URL")
+WS_API_BASE_URL = env.get("OPENSANCTIONS_UA_WS_API_BASE_URL")
+
+LINKS = [
+    {  # child kidnappers
+        "endpoint": "kidnappers/persons",
+        "type": "person",
```
leonhandreke: (nit) more of a style thing, but instead of using strings here, we could create an `WSAPIDataType(Enum)` that has all the data types. Then we're throwing around fewer strings :)

datasets/ua/war_sanctions/api/crawler.py
```diff
@@ -0,0 +1,494 @@
...
+    # Linked companies may not exist:
+    # 'ua-ws-entity-22': {'name': 'Unknown', 'country': None, 'imo': ''}
+    # 'ua-ws-entity-32': {'name': 'Rptd Sold Undisclosed Interest', 'country': None, 'imo': ''}
+    # 'ua-ws-entity-238': {'name': 'Rptd Sold Russia', 'country': None, 'imo': ''}
+    # 'ua-ws-entity-1272': {'name': 'Rptd Sold Undisclosed Interest', 'country': None, 'imo': '9991942'}
+    # 'ua-ws-entity-1425': {'name': 'Rptd Sold Russia', 'country': None, 'imo': '9992075'}
+    if not manager.get("name"):
+        manager.add("topics", "poi")
+        context.emit(manager)
+        return
```
leonhandreke: So IIUC correctly, adding the topic will make it not empty? That's cool. But why do we return here?

datasets/ua/war_sanctions/api/crawler.py
```diff
@@ -0,0 +1,494 @@
...
+    to_prop: str = "object",
+):
+    if not party_info:
+        return
+    company_id_raw = party_info.pop("id")
+    start_date = party_info.pop("date")
+    care_of_id_raw = party_info.pop("co_id", None)
+
+    if rel_role == "owner":
+        rel_schema = "Ownership"
```
leonhandreke: Do we have to set role (in the Ownership entity) to "owner" if we're using an Ownership relation? Maybe we can just set it to None?

datasets/ua/war_sanctions/api/crawler.py
```diff
@@ -0,0 +1,494 @@
...
+    dod = parts[1] if len(parts) > 1 and parts[1] else None
+    return dob, dod
+
+
+def crawl_ship_relation(
+    context: Context,
+    party_info,
+    vessel_id_slug,
+    rel_role: Optional[str] = None,
+    rel_schema: str = "UnknownLink",
```
leonhandreke: We generate rel_schema, from_prop, to_prop from rel_role anyway. So let's remove these arguments.

Maybe we can make it clear that role is a field in the API response that we're matching on.

datasets/ua/war_sanctions/api/crawler.py
nvmbrasserie: they are publicly accessible too, I don't think they'd hurt. It's nice to put a face to the vessel if anyone would be interested in it.

datasets/ua/war_sanctions/api/crawler.py
nvmbrasserie: there is no magic about it, it's about the right length for a long abbreviation. If it's longer, it's usually just a little shortened version of name

nvmbrasserie: I will add this comment to the code :)

datasets/ua/war_sanctions/api/crawler.py
nvmbrasserie: we also use it for id creation, I don't think it hurts

leonhandreke: Fair enough.

PR #2726, opened 2025-08-26
Fixes 26 08

nvmbrasserie:


datasets/jp/meti_ru/jp_meti_ru.yml
```diff
@@ -39,6 +39,12 @@ data:
   format: CSV
 dates:
   formats: ["%Y年%m月%d日", "%Y 年%m 月%d 日"]
+http:
+  user_agent: "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.2 Safari/605.1.15 (zavod; opensanctions.org)"
+  total_retries: 3
+  backoff_factor: 2
+  retry_statuses: [443]
```
leonhandreke: I added `additional_retry_statuses`, so in case you explicitly don't want the other statuses to be retried anymore, use that maybe?

datasets/jp/meti_ru/jp_meti_ru.yml
```diff
@@ -39,6 +39,12 @@ data:
...
+  retry_methods: [GET]
```
leonhandreke: I think GET is the default.

datasets/jp/meti_ru/jp_meti_ru.yml
```diff
@@ -39,6 +39,12 @@ data:
   format: CSV
 dates:
   formats: ["%Y年%m月%d日", "%Y 年%m 月%d 日"]
+http:
+  user_agent: "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.2 Safari/605.1.15 (zavod; opensanctions.org)"
+  total_retries: 3
```
leonhandreke: (nit) 3 is the default, so no need to re-state that here.

PR #2732, opened 2025-08-27
Fixes 27 08

nvmbrasserie:


datasets/sg/gov_dir/crawler.py
```diff
@@ -229,6 +229,10 @@ def crawl_body(context: Context, state: CrawlState, link) -> None:
     official_count = 0
     pep_count = 0
     board_doc = context.fetch_html(link, cache_days=1)
+    board_doc_html = etree.tostring(board_doc, encoding="unicode")
+    if "(SGDI) is temporarily unavailable" in board_doc_html:
+        context.log.warning(f"SGDI failover page encountered at {link}, skipping.")
+        return
```
jbothma: The effect of this would be skipping (dropping) all the entities on these pages when the site is under maintenance.

We normally prefer not doing this, so the simplest is to raise an exception. For massive sites like this it can be nice to update a gobal variable like an error count to and continue the crawl to cache any pages that we can crawl, then raise at the end if error count is > 0 or whatever seems tolerable.

In this case, it looks like it's planned temporary maintenance and maybe not something we need to fix because the message should go away in a couple of days?

image

datasets/ua/nsdc_sanctions/ua_nsdc_sanctions.yml
```diff
@@ -539,6 +539,10 @@ lookups:
         values:
           - Российская Федерация, г. Санкт-Петербург, ул. Большая Пушкарская, д. 20, лит. А, люкс 110-Н (Люкс часть 17), офис 1
           - Office 1, Suite 110-N (Suite Part 17), Letter A, 20 Bolshaya Pushrarskaya Street, Kronverkskoye Municipal Distinct, Saint Peterburg 197198, Russian Federation
+      - match: 'Люкс частина 17), офіс 1, (Российская Федерация, г. Санкт-Петербург, ул. Большая Пушкарская, д. 20, лит. А, люкс 110-Н (Люкс часть 17), офис 1; Office 1, Suite 110-N (Suite Part 17), Letter A, 20 Bolshaya Pushrarskaya Street, Kronverkskoye Municipal Distinct, Saint Peterburg 197198, Russian Federation'
+        values:
+          - Российская Федерация, г. Санкт-Петербург, ул. Большая Пушкарская, д. 20, лит. А, люкс 110-Н (Люкс часть 17), офис 1
+          - Office 1, Suite 110-N (Suite Part 17), Letter A, 20 Bolshaya Pushrarskaya Street, Kronverkskoye Municipal Distinct, Saint Peterburg 197198, Russian Federation
```
jbothma: can you tell what the difference is between this entry and the one just above it? It looks like they changed the comma to semicolon. So maybe we can try to delete the previous entry 
image

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -274,8 +274,15 @@ def crawl(context: Context):
                 within_age_limit = False
                 break
             link = result.xpath(
-                ".//a[contains(@href, 'recent-actions') and not(contains(@href, 'enforcement-actions'))]/@href"
+                ".//a[contains(@href, 'recent-actions') or contains(@href, 'recent-issues') and not(contains(@href, 'enforcement-actions'))]/@href"
             )[0]
+            # Skipping one known duplicate of "https://ofac.treasury.gov/recent-actions/20181127_33"
+            if link == "https://ofac.treasury.gov/recent-issues/20181127_33":
+                continue
+            if "recent-issues" in link:
```
leonhandreke, line 284: I don't 100% understand what's happening here, why are we adding these to then skip them again? Maybe a comment would help :)

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -274,8 +274,15 @@ def crawl(context: Context):
...
+                context.log.warn(
+                    f"Double check recent-issues link for possible duplicates: {link}"
+                )
```
jbothma: cool

datasets/us/trade_csl/us_trade_csl.yml
```diff
@@ -336,6 +336,202 @@ lookups:
         properties:
           name: SGM Most OOO
           previous_name: Obshchestvo S Ogranichennoi Otvetstvennostyu SGM Most
+      - match: "China Electronics Technology Group Corporation 13th Research Institute (CETC 13). Subordinate Institution - Hebei Medicines Health"
+        properties:
+          name: Hebei Medicines Health
+          related: China Electronics Technology Group Corporation 13th Research Institute (CETC 13)
```
jbothma, line 342: why are we putting the subordinate org into name and the first name into related?

jbothma (commented):
Thanks for these

datasets/us/trade_csl/us_trade_csl.yml
nvmbrasserie: https://github.com/opensanctions/opensanctions/pull/2241#discussion_r2089015168

PR #2747, opened 2025-08-29
Check changes helper

jbothma:


leonhandreke:
Some thoughts on this:

We have two paths of global state here. `Context.assert_all_accepted` and `ChangeTracker`/`something_changed`. They essentially do the same thing, fail the crawl under some circumstances of unaccepted reviews. I feel we should strive to unify them.

Regarding the question where to keep state, I'm a fan of just returning `Result` dataclasses or enums from crawl functions:

```python
def crawl(context):

    results = [crawl_page(page) for page in pages)]
    if not all([result == CrawlResult.REVIEW_ACCEPTED for result in results]):
        context.log.error(...)
        exit(1)
````

Maybe this is something that each crawler should develop its policy to depending on context, but we should have standard ways to do it. Here are some standard ways I can think of.

Most simple case - just re-prompt

```python
review = get_review(release_id)
if review is None or review.source_data != fetched_source_data:
    # Optional: we could warning-log a source diff here if we want
    review = Review(source_data=fetched_source_data, extracted_data=run_typed_text_prompt(...))
    review.save()

return CrawlResult.REVIEW_ACCEPTED if review.accepted else CrawlResult.REVIEW_UNACCEPTED
```

Bit smarter


```python
review = get_most_recent_review(release_id)
if review is None or review.source_data != fetched_source_data:
    new_review = Review(source_data=fetched_source_data, extracted_data=run_typed_text_prompt(...))

    # If the extracted data is the same, we can 
    if review.orig_extraction_data == new_review.orig_extraction_data:
        # Also transfer over the corrected result
        new_review.extracted_data = review.extracted_data
        # This is just pseudocode, but we should have some sort of pointer to the old review
        # We could even show the diff in the web UI instead of in the issues log in that case?
        new_review.previous_source_version_review = review.id 
        # If we're feeling lucky, just mark it as accepted?
        new_review.accepted = True

    # If the review is not accepted, this will prompt a new review
    new_review.save()
    review = new_review

return CrawlResult.REVIEW_ACCEPTED if review.accepted else CrawlResult.REVIEW_UNACCEPTED
````

@jbothma let's talk about this to reality-check my ramblings :)

PR #2749, opened 2025-09-01
Skip superfluous (breaking) encoding/decoding

jbothma:
Encoding broken encoding, decoding as utf-8, saving that to db, and reading that back, gives a slightly different flavour of broken encoding than just the first two steps. Funnily enough that's a bit broken.

e.g.

```
-       In December 1, 2020, the Securities and Exchange Commission charged an unregistered investment adviser based in Orlando, Florida for defrauding the Municipality of Mayagƒ¼ez, Puerto Rico and misappropriating $7.1 million of taxpayer funds.

?                                                                                                                                                                                ^

+       In December 1, 2020, the Securities and Exchange Commission charged an unregistered investment adviser based in Orlando, Florida for defrauding the Municipality of Mayag¼ez, Puerto Rico and misappropriating $7.1 million of taxpayer funds.

?
```

Just stay in the safe home of code points and nobody gets hurt.

### DIFF
--- a/datasets/us/cftc_enforcement_actions/crawler.py
+++ b/datasets/us/cftc_enforcement_actions/crawler.py
@@ -212,7 +212,7 @@ def crawl_enforcement_action(context: Context, date: str, url: str) -> None:
     article_element = fetch_article(context, url)
     if article_element is None:
         return
-    article_html = tostring(article_element, pretty_print=True).decode("utf-8")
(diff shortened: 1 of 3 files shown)


jbothma:
What's been saved to db still compares differently when read back, so I'll be deleting the broken review entries from the db to re-crawl them.

PR #2751, opened 2025-09-01
Fixes-01-09

nvmbrasserie:


datasets/us/bis_denied/crawler.py
```diff
@@ -62,9 +62,9 @@ def crawl(context: Context):
     doc = context.fetch_html(context.data_url, cache_days=1)
     url = doc.xpath(".//a[contains(normalize-space(.), 'Export as CSV')]/@href")
     assert len(url) == 1, "Expected exactly one URL"
-    path = context.fetch_resource("source.xlsx", url[0])
-    context.export_resource(path, XLSX, title=context.SOURCE_TITLE)
-    workbook: openpyxl.Workbook = openpyxl.load_workbook(path, read_only=True)
-    assert workbook.sheetnames == ["dpl", "Sheet2", "Sheet3"]
-    for row in h.parse_xlsx_sheet(context, sheet=workbook["dpl"]):
+    path = context.fetch_resource("source.xls", url[0])
+    context.export_resource(path, XLS, title=context.SOURCE_TITLE)
```
jbothma, line 66: I see this was steve's month to make the file. I wish steve and paul could agree on a file extension.
Maybe next time we do this switch we make it support either (although I see they recently also changed column headings. But this is fine for now.

datasets/us/ma/med_exclusions/crawler.py
```diff
@@ -4,6 +4,17 @@
 from zavod import Context, helpers as h
 
 
+HEADERS = {
+    "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
+    "Accept-Encoding": "gzip, deflate, br",
```
jbothma, line 9: watch out - these headers are kinda saying what formats the client supports. So it's probably best not to send custom Accept-headers - I've probably done so in the past as well - just be aware if we get errors about being unable to parse the response, it might be because we're saying `requests` supports e.g. `deflate` or something when I don't know whether it does.

PR #2753, opened 2025-09-01
[lt_fiu_freezes] add program mappings

nvmbrasserie:


datasets/lt/fiu_freezes/lt_fiu_freezes.yml
```diff
@@ -43,6 +43,12 @@ assertions:
       LegalEntity: 100
 
 lookups:
+  sanction.program:
+    options:
+      - match: Lithuania FNTT – List of legal persons or other unincorporated organisations owned or controlled by a sanctioned entity
+        value: LT-SL
+      - match: Lithuania FNTT – National UNSCR 1373 Terrorist Assets Freeze List
+        value: LT-AF
```
pudo, line 51: Can we maybe start having a micro-convention here? `LT-UNSCR1373`? We've got a few of these, I think.

datasets/lt/fiu_freezes/crawler.py
```diff
@@ -6,10 +6,17 @@
 from zavod.shed.zyte_api import fetch_html
 
 
+SANCTION_PROGRAM = "Lithuania FNTT – List of legal persons or other unincorporated organisations owned or controlled by a sanctioned entity"
+ASSET_FREEZE_PROGRAM = (
+    "Lithuania FNTT – National UNSCR 1373 Terrorist Assets Freeze List"
```
pudo, line 11: Using these strings as a picker for the `sanctions.program` is a really unstable ... we need a better pattern some time, but perhaps not right now.

Issue #2750, opened 2025-09-01
Delta exporter writes original, not consolidated, entity


jbothma:
because it fetches MOD and ADD entities afresh in finish() -> generate()

That's the prod index based on delta updates believed it was up to date but entities whose names/dates should have been consolidated, were not until we did a full index rebuild.

This can still be an issue post-consolidate-release if the name in one dupe is to be consolidated into the name of another dupe when they get merged.

PR #2757, opened 2025-09-01
Consolidate delta export

jbothma:
Fixes https://github.com/opensanctions/opensanctions/issues/2750


pudo:
Brilliant catch. Looks like it's doing the thing, too.

PR #2765, opened 2025-09-02
Let contrib/load_resolver.sh use $ZAVOD_DATABASE_URI if set

leonhandreke:


leonhandreke:
@jbothma adding you not for your bash expertise but to ask whether there is any reason you did it like this? I assume you also you a postgres for development.

PR #2770, opened 2025-09-03
[my_moha_sanctions] crawl pdf

nvmbrasserie:
entity ids were not rekeyed

### DIFF
--- a/datasets/my/moha_sanctions/crawler.py
+++ b/datasets/my/moha_sanctions/crawler.py
@@ -1,10 +1,11 @@
 import string
 
-from urllib.parse import urljoin
 from datapatch import Lookup
+from normality import squash_spaces
+from pdfplumber.page import Page
 from rigour.mime.types import PDF
 from typing import Tuple, Dict, Any
-from pdfplumber.page import Page
+from urllib.parse import urljoin
 
 from zavod import Context
 from zavod import helpers as h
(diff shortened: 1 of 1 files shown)


datasets/my/moha_sanctions/crawler.py
```diff
@@ -1,91 +1,124 @@
...
-        sanction.add("authorityId", reference)
-        sanction.add("program", row.pop("designation", None))
-
-        context.emit(entity)
-        context.emit(sanction)
-        context.audit_data(row, ignore=["no"])
-
-
-def crawl_html_url(context: Context) -> str:
+def clean_value(v):
```
leonhandreke, line 32: (nit) can we use `squash_spaces` here? Or do we want to do something different?

datasets/my/moha_sanctions/crawler.py
```diff
@@ -1,91 +1,124 @@
...
+    if not v:
+        return v
+    v = v.replace("\n", " ").strip()
+    return "" if v == "-" else v
+
+
+def crawl_row(context: Context, row, schema, key) -> None:
+    names = row.pop("name")
```
leonhandreke, line 40: If "Nama" is a special case, let's put it in an `INVALID_ROWS` constant.

leonhandreke (approved):
That's a super clean PDF parser!

I pushed a single commit that makes `rename_headers` a bit more type-ey.

PR #2778, opened 2025-09-05
Fixes-05-09

nvmbrasserie:


datasets/jp/meti_ru/crawler.py
```diff
@@ -137,7 +137,9 @@ def crawl_row(context, row):
 
 
 def crawl(context: Context):
-    html_path = context.fetch_resource("source.html", SOURCE_URL)
+    _, _, _, html_path = fetch_resource(
```
leonhandreke: Why not sure `fetch_html` here? That way, the cache is invalidated if the html does not parse

(nit) I like to `from zavod.shed import zyte_api` because then at the callsite it's immediately obvious that we're using zyte. But maybe this is a personal preference.

datasets/qa/nctc_sanctions/crawler.py
```diff
@@ -95,5 +95,6 @@ def crawl(context: Context):
                 "birthdateStatus",
                 "aliasesStatus",
                 "moiListType",
+                "fullNameAr",
```
leonhandreke, line 99:
```suggestion
                # Sometimes left over because we only use the name1,2,3,4 items for Companies
                "fullNameAr",
```

PR #2780, opened 2025-09-05
fix hardcoded program mappings

nvmbrasserie:
renamed branch for  opensanctions/opensanctions#2753


nvmbrasserie:
@pudo could you please take a quick look? does it make sense?

Issue #2656, opened 2025-08-13
LLM-backed name splitting helper


pudo:
I would like to propose a new helper that works like this: 

* The helper uses a conjunction of entity schema and a regex to see if a name contains atypical characters (eg. `/;()[]<>` for `Person`, `/` for `Organization`. Also, HTML entities). If no character is found, the name is returned as-is. 
* The helper then upserts a review task for the name. The review task is the result of a prompt-based split attempt which has detailed examples of common name corruptions.
* If the review task has not been accepted, the name is returned as-is (for sanctions data SLA!). If the review task is accepted, return the result of that. 

Examples of name splits:

* `EFERMEROV / YEFEMEROV Yulia Alexandrova`
* `Vladimir Putin (Pyutin)`
* `Jeffrey Blankard (Bobo)` (nickname/weak alias)
* `BABIC Josip geb. 06.02.1979 in Deutschland; GOJKOVIC Dragan geb. 08.09.1977 in Berane/Montenegro; GREPO David geb. 02.03.1983; KAROLY Ferenc geb. 31.03.1975; und weitere`
* `Mme Joséphine Solange ANABA MBARGA;Anaba Mbarga joséphine solange`
* `JILL O&#039;CONNOR`

I would propose we roll this out in core sanctions datasets opportunistically first to get a sense of the workload it generates. 

- do we want to use dspy in zavod? Or should the training stuff be optional deps and we just use run_typed_text_prompt?
  - `dspy.Predict` seems to be just a prompt and data structure. The other modules might be more fancy.
- if we use dspy in prod
  - [ ] hook it up to our cache
- if we use run_typed_text_prompt in prod
  - [ ] figure out how to get performance to same level as dspy

jbothma:
https://data.opensanctions.org/datasets/20250804/us_ak_med_exclusions/names.txt is also a nice source of examples

```
MI AE YOUNG (a.k.a SON)
SPINNAKERS LLC dba NORTH COUNTRY DIAGNOSTIC
CHRIS (aka Sa Young) SON
CARONDA PORTERFIELD (aka Smith)
```

pudo:
Oh, nice! Luck would have it, we have this: https://github.com/opensanctions/rigour/blob/main/resources/names/stopwords.yml#L358-L376

jbothma:
non-llm categorisation
is_regular checks - each check might propose a categorisation

current state:
- taiwan still on logic-based
- ru_cbr is reviewed but not applied

[...]

PR #2802, opened 2025-09-10
Fix up ua_war_sanctions_api

nvmbrasserie:


datasets/ua/war_sanctions/api/crawler.py
```diff
@@ -157,6 +166,24 @@ def generate_token(cid: str, pkey: str) -> str:
...
+            res = context.lookup("names", raw_name)
+            if res:
+                person.add("name", res.name, lang=lang)
+                person.add("alias", res.alias, lang=lang)
+            else:
+                context.log.warning(
+                    "Multiple names in a single field, please check",
+                    field=key,
+                    raw_name=raw_name,
+                )
```
pudo, line 184: Can we start making annotations in places where we have name splitting code that can eventually be replaced with https://github.com/opensanctions/opensanctions/issues/2656 ?

datasets/ua/war_sanctions/api/crawler.py
```diff
@@ -236,31 +263,29 @@ def emit_relation(
...
     h.apply_date(person, "birthDate", birth_date)
     h.apply_date(person, "deathDate", death_date)
     person.add("topics", "poi")
     person.add("birthPlace", person_data.pop("city_bd", None))
 
     sanction = h.make_sanction(context, person)
     sanction.add("reason", person_data.pop("reason", None))
-    # sanction.add("sourceUrl", person_data.pop("links", None))
+    sanction.add("sourceUrl", person_data.pop("links", None))
     sanction.add("program", program)
```
pudo: Didn't we want to do a programID lookup here?

nvmbrasserie: I was thinking of adding it after the dedupe stage, but this also works

PR #2835, opened 2025-09-17
add enricher for us_ofac_press_releases

nvmbrasserie:


datasets/_externals/ext_us_ofac_press_releases.yml
```diff
@@ -0,0 +1,71 @@
+type: external
+title: US OFAC Press Releases
+entry_point: zavod.runner.local_enricher:enrich
+full_dataset: us_ofac_press_releases
+disabled: false
+hidden: false
+coverage:
+  schedule: "@weekly"
```
jbothma: this could run daily, I'd say. We often make the enrichers run more regularly than the source. That way any delisted people get dropped quickly, and potential newly listed people are discovered more quickly. Considering its small size, I think daily is cool.

jbothma (approved):
looks good to me! But perhaps bump to daily before merging.

PR #2838, opened 2025-09-17
CH SECO <other-information> using LLM

jbothma:
This is to be merged after we've run the code to load and accept the current data from the current regex-based extraction


datasets/ch/seco_sanctions/crawler.py
```diff
@@ -33,19 +37,129 @@
...
+class OtherInfo(BaseModel):
+    simple_values: List[SimpleValue] = []
+    related_entities: List[RelatedEntity] = []
+    include_in_notes: bool = Field(
+        description="If True, the original value will be included in notes.",
+        default=False,
+    )
+
+
+MIN_VERSION = 1
```
leonhandreke: (nit) Can we prefix those with `REVIEW_`

datasets/ch/seco_sanctions/crawler.py
```diff
@@ -206,6 +320,24 @@ def parse_identity(context: Context, entity: Entity, node: Element, places):
...
+    context: Context, entity: Entity, relationship: RelatedEntity
+) -> List[Entity]:
+    other = context.make("LegalEntity")
+    other.id = context.make_id(relationship.related_entity_name)
+    other.add("name", relationship.related_entity_name)
+
+    res = context.lookup("relations", relationship.relationship_schema)
+
+    rel = context.make(relationship.relationship_schema)
+    rel.id = context.make_id(entity.id, other.id, relationship.relationship)
```
leonhandreke: (nit) Can we use a `make_slug` here? Then we'd get a `<id>-father-<id>` or something?

datasets/ch/seco_sanctions/ch_seco_sanctions.yml
```diff
@@ -127,10 +128,16 @@ lookups:
     options:
       - match:
           - related-to
```
leonhandreke: Can/does related-to this still happen?

datasets/ch/seco_sanctions/crawler.py
```diff
@@ -313,6 +412,50 @@ def parse_entry(context: Context, target: Element, programs, places):
...
+
+        review = get_review(context, OtherInfo, value, MIN_VERSION)
+        if review is None:
+            prompt = PROMPT.format(schema=entity.schema.name)
+            extraction = run_typed_text_prompt(
+                context, prompt, value, OtherInfo, model=LLM_VERSION
+            )
+            review = request_review(
+                context,
+                value,
```
leonhandreke: Some nits here (probably for another PR):

- `model_version` should be called `data_model_version`. `model` is a very overloaded term.
- It's probably worth making everything but context kwargs-required (both `get_review` and `request_review`). Many args, no idea what they mean in a code review.

datasets/ch/seco_sanctions/crawler.py
```diff
@@ -313,6 +412,50 @@ def parse_entry(context: Context, target: Element, programs, places):
...
+                PLAIN,
+                "other-information",
+                None,
+                extraction,
+                VERSION,
+            )
+
+        if review.accepted:
+            for extracted_value in review.extracted_data.simple_values:
+                if "Date" in extracted_value.property:
```
leonhandreke: (nit) that's a bit rough... maybe explicitly list them? Or even ask FtM?

datasets/ch/seco_sanctions/crawler.py
```diff
@@ -33,19 +37,129 @@
...
+
+
+class RelatedEntity(BaseModel):
+    relationship_schema: Literal["Family", "UnknownLink"]
+    related_entity_name: List[str]
+    relationship: Optional[str] = None
+
+
+class OtherInfo(BaseModel):
+    simple_values: List[SimpleValue] = []
```
leonhandreke, line 82: I guess you thought about this, but why no `Dict[Properties, List[str]]`?

datasets/ch/seco_sanctions/crawler.py
```diff
@@ -33,19 +37,129 @@
...
-)
-REGEX_PHONE = re.compile(r"(Tel\.|Telephone)( number)? ?: (\+?[0-9- ()]+)")
-REGEX_INN = re.compile(r"Taxpayer [Ii]dentification [Nn]umber ?: (\d+)\.?")
-REGEX_REGNUM = re.compile(
-    r"(ОГРН/main )?([Ss]tate |Business )?[Rr]egistration number ?: (\d+)\.?"
-)
-REGEX_TAX = re.compile(r"Tax [Rr]egistration [Nn]umber ?: (\d+)\.?")
-REGEX_IMO = re.compile(r"IMO [Nn]umber ?: (\d+)\.?")
+
+Properties = Literal[
```
leonhandreke: (nit) `PropertyName`?

datasets/ch/seco_sanctions/crawler.py
jbothma: how about CRAWLER_? i.e. `CRAWLER_VERSION=2`, `MIN_CRAWLER_VERSION=1`
Because reviews have a last_seen_version which is the crawl/dataset version, but a review already has two notions of version - the soft-delete history and the dataset version.

datasets/ch/seco_sanctions/ch_seco_sanctions.yml
jbothma: yep - related-to is from the xml structure, the schema is from the LLM extraction. Adding comments to this effect.

datasets/ch/seco_sanctions/crawler.py
jbothma: ➕

datasets/ch/seco_sanctions/crawler.py
jbothma: very fair

datasets/ch/seco_sanctions/crawler.py
jbothma: mmm that would make reading and manual edits easier, but I'm not easily winning with pydantic/json schema/gpt
I think it's because currently openai requires all fields to be required by the schema. So while optional fields can be supported as empty lists or null values, the field must always be present, which would be horrible for readability.

```diff
 class OtherInfo(BaseModel):
-    simple_values: List[SimpleValue] = []
+    simple_values: Dict[PropertyName, List[str]] = {}
     related_entities: List[RelatedEntity] = []
```

error:
```
Runner failed: Error code: 400 - {'error': {'message': "Invalid schema for response_format 'OtherInfo': In context=('properties', 'simple_values'), 'propertyNames' is not permitted.", 'type': 'invalid_request_error', 'param': 'response_format', 'code': None}}
```

schema:
```json
'simple_values': {'additionalProperties': {'items': {'type': 'string'}, 'type': 'array'}, 'default': {}, 'propertyNames': {'enum': ['email', 'phone', 'website', 'innCode', 'ogrnCode', 'kppCode', 'taxNumber', 'imoNumber', 'okpoCode', 'passportNumber', 'uscCode', 'swiftBic', 'registrationNumber', 'idNumber', 'incorporationDate', 'dissolutionDate', 'address', 'country', 'position', 'birthPlace', 'birthDate', 'gender', 'legalForm']}, 'title': 'Simple Values', 'type': 'object'}
```

jbothma: I didn't actually try it until now, but yeah, I think the slightly more clunky array of value objects is the way to go for now.

datasets/ch/seco_sanctions/crawler.py
jbothma: that results in ids like `ch-seco-ch-seco-43596-ch-seco-b4465e5d079e7773942fe3a354e4c19c449c904e`
we can do it without the prefix for `entity` but not really easily for `other`

As for precedent of this, we mostly do rel.id = make_id but not always. So I think make_id is best in this case for its simplicity.

datasets/ch/seco_sanctions/crawler.py
leonhandreke: `DATA_MODEL_VERSION`?

datasets/ch/seco_sanctions/crawler.py
```diff
@@ -499,28 +498,44 @@ def parse_entry(context: Context, target: Element, programs, places):
...
+                    entity.add(
+                        extracted_value.property,
+                        extracted_value.value,
+                        origin=LLM_VERSION,
+                        original_value=value,
+                    )
+                else:
+                    if (
+                        extracted_value.property == "imoNumber"
+                        and not entity.schema.is_a("Vessel")
```
leonhandreke: When does this happen? At this point it's not a Vessel because otherwise we wouldn't have ended up in this branch of the if, right?

datasets/ch/seco_sanctions/crawler.py
```diff
@@ -499,28 +498,44 @@ def parse_entry(context: Context, target: Element, programs, places):
...
 
         if review.accepted:
             for extracted_value in review.extracted_data.simple_values:
-                if "Date" in extracted_value.property:
-                    h.apply_date(
-                        entity, extracted_value.property, extracted_value.value
+                prop = entity.schema.get(extracted_value.property)
+                if prop is not None:
+                    if prop.type == registry.date:
+                        h.apply_date(
```
leonhandreke: Not for this PR (it would be a bigger thing), but we have this pattern so often that I'm asking myself whether the zavod `Entity` should maybe have a shortcut for this in `.add`?

datasets/ch/seco_sanctions/crawler.py
leonhandreke: Makes sense.

datasets/ch/seco_sanctions/crawler.py
leonhandreke: Sounds good. Could you add a quick comment about this design choice?

leonhandreke (approved):

datasets/ch/seco_sanctions/crawler.py
jbothma: ah good point. added a comment and removed the vessel condition
```
                    if extracted_value.property == "imoNumber":
                        # If the target wasn't explicitly a vessel, skip the IMO because
                        # it might be of a vessel related to the company, and not the target's IMO.
                        # <justification ssid="10301">IRISL front company...
                        #     It is the registered owner of a vessel owned by IRISL or an IRISL affiliate.
                        # </justification>
                        # <relation ssid="21013" target-id="9702" relation-type="related-to"></relation>
                        # <other-information ssid="10302">No C 38181; IMO number of the vessel: 9387803.</other-information>
                        continue
```

PR #2839, opened 2025-09-17
CH SECO more info

jbothma:
This is to be run once (but more is ok) to load and accept current regex-based extraction.

Once this is done, we can merge https://github.com/opensanctions/opensanctions/pull/2838 to clean up the regexes and loading code.

### DIFF
--- a/datasets/ch/seco_sanctions/crawler.py
+++ b/datasets/ch/seco_sanctions/crawler.py
@@ -1,19 +1,19 @@
 import re
-from itertools import product
 from datetime import datetime
-from prefixdate import parse_parts
-from typing import Dict, Literal, Optional, Tuple, List
+from itertools import product
+from typing import Dict, List, Literal, Optional, Tuple
+
 from followthemoney.types import registry
 from followthemoney.util import join_text
 from lxml.etree import _Element as Element
+from prefixdate import parse_parts
 from pydantic import BaseModel
-
 from rigour.mime.types import PLAIN
+from zavod.shed.gpt import run_typed_text_prompt
+from zavod.stateful.review import assert_all_accepted, get_review, request_review
 
 from zavod import Context, Entity
 from zavod import helpers as h
-from zavod.shed.gpt import run_typed_text_prompt
-from zavod.stateful.review import assert_all_accepted, get_review, request_review
 
 # TODO: sanctions-program full parse
 MayStr = Optional[str]
@@ -74,8 +74,8 @@ Properties = Literal[
     "position",
     "birthPlace",
     "deathDate",
-    "name",
     "gender",
+    "legalForm",
 ]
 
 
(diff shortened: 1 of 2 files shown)


datasets/ch/seco_sanctions/crawler.py
```diff
@@ -313,6 +426,99 @@ def parse_entry(context: Context, target: Element, programs, places):
...
+                item = SimpleValue(property="website", value=website.group(1))
+            elif email := REGEX_EMAIL.fullmatch(value):
+                item = SimpleValue(property="email", value=email.group(2))
+            elif phonenumber := REGEX_PHONE.fullmatch(value):
+                item = SimpleValue(property="phone", value=phonenumber.group(3))
+            elif value == "Registration number: ИНН":
+                pass
+
+            if item is not None:
+                extraction = OtherInfo(simple_values=[item])
```
leonhandreke: So IIUC, we won't be prompting for all the existing values, only for new ones. If so, should we set `include_in_notes` here? Otherwise won't we be losing all the info that wasn't regexed?

datasets/ch/seco_sanctions/crawler.py
```diff
@@ -313,6 +426,99 @@ def parse_entry(context: Context, target: Element, programs, places):
...
+                review = request_review(
+                    context,
+                    value,
+                    value,
+                    PLAIN,
+                    "other-information",
+                    None,
+                    extraction,
+                    VERSION,
```
leonhandreke: Should we bump `VERSION` after this so that we can get rid of these regex extractions eventually if we want to extract all the data eventually?

datasets/ch/seco_sanctions/crawler.py
jbothma: good call - I didn't check this until now. It turns out the regexes are using fullmatch so we shouldn't be leaving any unmatched information out.

Or am I misunderstanding?

datasets/ch/seco_sanctions/crawler.py
jbothma: I think fullmatch also means this isn't needed.

zavod/zavod/shed/gpt.py
```diff
@@ -207,7 +207,7 @@ def run_typed_text_prompt(
             }
         ],
         response_format=response_type,
-        max_tokens=max_tokens,
+        max_completion_tokens=max_tokens,
```
jbothma, line 210: this is needed for gpt-5* and backward compatible to gpt-4o

datasets/ch/seco_sanctions/crawler.py
leonhandreke: No, this is exactly what i meant. TIL about `fullmatch`. I sort of assumed this would be a long-form textfield like the French. That clears it up for me!

leonhandreke (commented):
I'm half asking myself if we should actually clean them up or just keep the regexes around to dampen the review load long-term? What's the coverage on these regexes, more like 30% or 90%?

leonhandreke (approved):

jbothma:
Regarding keeping the regexes, I think let's drop them until we find the review load too painful. I think these are the easiest kinds of reviews, and they should probably arrive in small batches rather than 100s at a time. And maintaining a set of regexes + a megaprompt feels like a bad outcome for the LLM extraction project.

for data:


6164 reviews
4091 unaccepted
so 30% auto-extracted by regex.

PR #2843, opened 2025-09-18
[jp_meti_ru] update 

nvmbrasserie:


datasets/jp/meti_ru/crawler.py
```diff
@@ -27,16 +29,52 @@
...
-    cleaned_address = re.sub(r"^所在地:", "", cleaned_address).strip()
-    return cleaned_address
+    cleaned = re.sub(r"^所在地[:：]", "", raw_address).strip()
+
+    # Split into parts by common delimiters
+    if any(d in cleaned for d in (" and ", ";")):
+        parts = h.multi_split(cleaned, [" and ", ";"])
+        return [p.strip(" ,;") for p in parts if p.strip()]
+
+    return cleaned
```
jbothma: any reason why it won't work to just always do the multi_split and strip whether the splitters occur in the text or not?

Then the code is simpler because we have one return type - a list - and we don't do the condition with the same items as the split.

datasets/jp/meti_ru/crawler.py
```diff
@@ -68,25 +106,23 @@ def clean_name_raw(context, name_jpn, row):
...
     # Check for cases that require manual processing
     if main_name == "" and not aliases:
         # Identify complex cases with certain characters
         if any(char in name_jpn for char in ["（", "、", ")", "）"]):
             main_name = row.pop("name_jpn_cleaned")
             aliases_raw = row.pop("aliases_jpn_cleaned")
-            aliases = [
-                alias.strip() for alias in aliases_raw.split(";") if alias.strip()
-            ]
+            aliases.extend(process_aliases(context, aliases_raw))
```
jbothma: I think split_aliases is a more useful name since it tells us what it's doing. Sure, we're also kinda cleaning and identifying the script, but I think that still falls broadly under splitting.

```suggestion
            aliases.extend(split_aliases(context, aliases_raw))
```

datasets/jp/meti_ru/crawler.py
```diff
@@ -149,10 +186,8 @@ def crawl(context: Context):
...
     assert len(divs) == 1, len(divs)
     # Check hash of the content part of the page
-    h.assert_dom_hash(divs[0], "4aad6cb5239dc452fff0a697d49684aa75eb2901")
-    pdf_xpath = (
-        ".//a[contains(@href, '.pdf') and contains(@href, '17_russia/list')]/@href"
-    )
+    h.assert_dom_hash(divs[0], "982832a856dfe254b6282966ec96cfb58d9464aa")
+    pdf_xpath = ".//a[contains(@href, '.pdf') and contains(@href, 'export/17_russia/') and contains(@href, 'tokutei')]/@href"
     pdf_urls = divs[0].xpath(pdf_xpath)
     assert len(pdf_urls) == 3, "Expected exactly 3 PDFs"
```
jbothma:
```suggestion
    pdf_xpath = ".//a[contains(@href, '.pdf') and contains(@href, 'export/17_russia/') and contains(@href, 'tokutei')]/@href"
    pdf_urls = divs[0].xpath(pdf_xpath)
    assert len(pdf_urls) == 3, len(pdf_urls)
```

datasets/jp/meti_ru/crawler.py
```diff
@@ -27,16 +29,52 @@
...
+        context.log.warning(f"Could not detect script for: {text}")
+        return None
+    dominant = max(set(scripts), key=scripts.count)
+    # Only return a language if confident
+    if dominant in {"Hiragana", "Katakana"}:
+        return "jpn"
+    elif dominant == "Han" and (
+        "Hiragana" not in scripts and "Katakana" not in scripts
+    ):
+        return "zho"
```
jbothma: just chucking this in to be sure we talk about it - not sure if you've had other conversations or research about this.
IIRC Japanese writing can use some Chinese script, but the writing is still Japanese? Is this possibly a case of that?
And is the appropriate 3-letter code then jpn or zho?

nvmbrasserie: let me do some research on it. @leonhandreke already brought it to my attention, I'll investigate

jbothma: @Kaldea might know the answer

nvmbrasserie: I looked into it and there doesn't seem to be an easy solution to it. 95% of names are in Japanese, but we can't be 100% sure if the name, that is written in Han, is Japanese or Chinese. I'd suggest to abandon the language for uncertain cases

Issue #2590, opened 2025-08-03
Bring PEP position classification tool into zavod UI


pudo:
We want to move the classification tool for PEP positions into zavod UI, so that users can tag positions as national/subnational/local and exec/judi/legi in the same tool. 

Desires:
* Bring over the existing tool 
* Make a possibility to classify religious positions as such 
* Search by position name
* Allow for bulk assignments ("mark all search results as gov.muni")
* Option to only show Wikidata items

jbothma:
filter by dataset

num classified/unclassified by dataset

jbothma:
shall we split out a ticket for bulk categorisation?

<img width="1522" height="758" alt="Image" src="https://github.com/user-attachments/assets/4c1a007b-d171-4c19-88ad-678900b5c8c2" />

PR #2846, opened 2025-09-18
[ui] Add positions admin UI

leonhandreke:
https://github.com/opensanctions/opensanctions/issues/2590

### DIFF
--- a/ui/app/api/positions/[entityId]/route.ts
+++ b/ui/app/api/positions/[entityId]/route.ts
@@ -1,0 +1,40 @@
+import { NextRequest, NextResponse } from 'next/server';
+
+import { verify } from '../../../../lib/auth';
+import { softDeleteAndCreatePosition } from '../../../../lib/db';
+
+interface PositionUpdateData {
+  caption: string;
+  countries: string[];
+  is_pep: boolean | null;
+  topics: string[];
+  dataset: string;
+}
+
+export async function PUT(
+  req: NextRequest,
+  { params }: { params: Promise<{ entityId: string }> }
+) {
+  const email = await verify(req.headers);
+  const { entityId } = await params;
+
+  if (!(req.headers.get('content-type') || '').includes('application/json')) {
+    return NextResponse.json(
+      { error: 'Content-Type must be application/json' },
+      { status: 400 }
+    );
+  }
+
+  const data: PositionUpdateData = await req.json();
+
+  // We're quite aggresive here, throwing lots of errors if something goes wrong
+  // but not caching them. Oh well, this is internal, so 500 or 404 doesn't matter
+  // all that much.
+  const updatedPosition = await softDeleteAndCreatePosition({
+    entityId,
+    positionUpdate: data,
+    modifiedBy: email!,
+  });
+
+  return NextResponse.json(updatedPosition);
+}
--- a/ui/app/api/positions/[entityId]/route.ts
+++ b/ui/app/api/positions/[entityId]/route.ts
@@ -28,7 +28,7 @@ export async function PUT(
   const data: PositionUpdateData = await req.json();
 
   // We're quite aggresive here, throwing lots of errors if something goes wrong
-  // but not caching them. Oh well, this is internal, so 500 or 404 doesn't matter
+  // but not catching them. Oh well, this is internal, so 500 or 404 doesn't matter
   // all that much.
   const updatedPosition = await softDeleteAndCreatePosition({
     entityId,
--- a/ui/app/api/positions/[entityId]/route.ts
+++ b/ui/app/api/positions/[entityId]/route.ts
@@ -4,11 +4,8 @@ import { verify } from '../../../../lib/auth';
 import { softDeleteAndCreatePosition } from '../../../../lib/db';
 
 interface PositionUpdateData {
-  caption: string;
-  countries: string[];
   is_pep: boolean | null;
   topics: string[];
-  dataset: string;
 }
 
 export async function PUT(
--- a/ui/app/positions/PositionTagger.tsx
+++ b/ui/app/positions/PositionTagger.tsx
@@ -23,11 +23,20 @@ export default async function PositionTagger({ searchParams }: { searchParams: S
     Object.entries(searchParams).filter(([, value]) => value !== '')
   )
 
+  let isPepSearchParam = undefined;
+  if (filteredSearchParams.is_pep == "null") {
+    isPepSearchParam = null;
+  } else if (filteredSearchParams.is_pep == "true") {
+    isPepSearchParam = true;
+  } else if (filteredSearchParams.is_pep == "false") {
+    isPepSearchParam = false;
+  }
+
   const positions = await getPositions({
     q: filteredSearchParams.q as string,
     dataset: filteredSearchParams.dataset as string,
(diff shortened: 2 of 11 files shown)


ui/app/api/positions/[entityId]/route.ts
```diff
@@ -0,0 +1,40 @@
...
+    return NextResponse.json(
+      { error: 'Content-Type must be application/json' },
+      { status: 400 }
+    );
+  }
+
+  const data: PositionUpdateData = await req.json();
+
+  // We're quite aggresive here, throwing lots of errors if something goes wrong
+  // but not caching them. Oh well, this is internal, so 500 or 404 doesn't matter
```
Comment (jbothma) [inline ui/app/api/positions/[entityId]/route.ts]:
```suggestion
  // but not catching them. Oh well, this is internal, so 500 or 404 doesn't matter
```

ui/app/positions/PositionTaggerForm.tsx
```diff
@@ -0,0 +1,57 @@
...
+          <Form.Select name="dataset" defaultValue={searchParams.dataset || ""} onChange={submitForm}>
+            <option value="">Any dataset</option>
+            {datasets.map((dataset) => {
+              return <option key={dataset} value={dataset}>{dataset}</option>
+            })}
+          </Form.Select>
+        </Col>
+        <Col md={2}>
+          <Form.Select name="is_pep" defaultValue={searchParams.is_pep || ""} onChange={submitForm}>
+            <option value="">Any PEP status</option>
```
jbothma, line 48: it looks like it's currently not showing is_pep=True when this is selected (default)

side note: I'm not averse to Undecided being the default selected value

ui/app/positions/PositionTagger.tsx
```diff
@@ -0,0 +1,96 @@
...
+            })
+        }
+      </tbody>
+    </Table>
+    <div className="d-flex justify-content-center">
+      <ResponsePagination response={positions} searchParams={searchParams} />
+    </div>
+
+    <div className="mb-4">
+      <h5>Keyboard Shortcuts</h5>
```
jbothma: a tiny note like "hover to select a position" would help - it took me a moment to figure out how I make these apply

ui/app/positions/PositionTaggerRow.tsx
```diff
@@ -0,0 +1,197 @@
...
+      updatedTopics.splice(updatedTopics.indexOf(topic), 1);
+    } else {
+      updatedTopics.push(topic);
+    }
+    updatePosition({ topics: updatedTopics });
+  }, [state.topics, updatePosition]);
+
+  // Keyboard shortcuts for when row is hovered
+  useEffect(() => {
+    function handleKeyDown(e: KeyboardEvent) {
```
jbothma, line 89: at the moment, cmd+r when hovering over something is seen as pressing r, rather than reloading the page. is it possible to exclude when ctrl/cmd is pressed?

ui/app/api/positions/[entityId]/route.ts
```diff
@@ -0,0 +1,40 @@
+import { NextRequest, NextResponse } from 'next/server';
+
+import { verify } from '../../../../lib/auth';
+import { softDeleteAndCreatePosition } from '../../../../lib/db';
+
+interface PositionUpdateData {
+  caption: string;
```
Comment (jbothma) [inline ui/app/api/positions/[entityId]/route.ts]: don't think it's the end of the world, but not sure this needs to be allowed at the moment.

ui/app/positions/PositionTaggerRow.tsx
```diff
@@ -0,0 +1,197 @@
...
+import { Position, PositionUpdate } from "@/lib/db";
+
+const SCOPE_ENTRIES = [
+  ["gov.national", "Nat"],
+  ["gov.state", "Subnat"],
+  ["gov.muni", "Local"],
+  ["gov.igo", "IGO"]
+] as const;
+
+const ROLE_ENTRIES = [
```
jbothma, line 15: can add `gov.religion`

ui/lib/db.ts
```diff
@@ -262,3 +317,155 @@ export async function getNextUnacceptedEntryKey(dataset: string): Promise<string
...
+    .where('deleted_at', 'is', null);
+
+  if (entity_id !== undefined) {
+    query = query.where('entity_id', '=', entity_id);
+  }
+  if (caption !== undefined) {
+    query = query.where('caption', '=', caption);
+  }
+  if (country !== undefined) {
+    // TODO(Leon Handreke): How safe is this? I don't know.
```
jbothma, line 380: it's parameterised :) but good question

image

ui/app/positions/PositionTaggerForm.tsx
jbothma: I think blank is ending up querying for null

jbothma (approved):
Very nice! Just tiny comments.

Shall we add a little top nav or something for ` Reviews | Positions ` ?

a surprising thing that I don't think critically needs fixing is that the search results that are edited go to the bottom of the results list after reloading the page. That seems to be because the created_at doesn't have sub-second resolution. So the data is returned in arbitrary order when all the search results were created at the same second, as my test case was.

ui/lib/db.ts
```diff
@@ -102,13 +137,33 @@ async function assertSchemaMatchesExpected(db: Kysely<ReviewDatabase>) {
...
+  const positionTable = tables.find(t => t.name === POSITION_TABLE_NAME)
+  if (!positionTable) {
+    throw new Error(`Position table not found in database`);
+  }
+  const existingPositionColumnNames = new Set<string>();
+  for (const column of positionTable.columns) {
+    existingPositionColumnNames.add(column.name);
+  }
+  const unexpectedPosition = expectedPositionColumns.difference(existingPositionColumnNames);
+  const missingPosition = existingPositionColumnNames.difference(expectedPositionColumns);
```
jbothma, line 158: This is very embarrassing but... I had these the wrong around. Fixed in https://github.com/opensanctions/opensanctions/pull/2900/files#diff-7f8cb697f8ad736066099f06e7ac0700cfde8f6f96c195726f556cb5e0d960a8L102

ui/app/positions/PositionTaggerForm.tsx
leonhandreke: It was even more awfully broken. Fixed now.

ui/app/positions/PositionTagger.tsx
leonhandreke: Done.

ui/app/positions/PositionTaggerRow.tsx
leonhandreke: Done (why was that so hard).

ui/lib/db.ts
leonhandreke: That was my feeling too, but thanks for confirming my suspicion :)

ui/app/positions/PositionTaggerRow.tsx
leonhandreke: Done.

ui/app/api/positions/[entityId]/route.ts
Comment (leonhandreke) [reply ui/app/api/positions/[entityId]/route.ts]: Done.

PR #2847, opened 2025-09-19
Fixes-19-09

nvmbrasserie:


datasets/za/wanted/crawler.py
```diff
@@ -41,10 +39,12 @@ def crawl_detail_page(
         key: (doc.xpath(xpath)[0].strip() if doc.xpath(xpath) else "")
         for key, xpath in details.items()
     }
-    status = doc.xpath("//p[@align='center']/font[@color='blue']/text()")[0]
+    status_list = doc.xpath("//p[@align='center']/font[@color='blue']/text()")
+    status = status_list[0].strip() if status_list else None
+
     if status not in {"Wanted", "Suspect"}:
+        context.log.warning(f"Unknown or missing status: {status}, url: {source_url}")
```
leonhandreke, line 46: (nit) I think a common status is to put it as an argument to the function instead of in the string.

datasets/za/wanted/crawler.py
```diff
@@ -41,10 +39,12 @@ def crawl_detail_page(
         key: (doc.xpath(xpath)[0].strip() if doc.xpath(xpath) else "")
         for key, xpath in details.items()
     }
-    status = doc.xpath("//p[@align='center']/font[@color='blue']/text()")[0]
+    status_list = doc.xpath("//p[@align='center']/font[@color='blue']/text()")
```
leonhandreke, line 42: (nit) Why call this list? Is it a list? Looks like we just `strip` it after, not split.

nvmbrasserie: since xpath() returns a list, I wanted to differentiate it from the unpacked status string, not sure what's the naming convention in this case

PR #2848, opened 2025-09-19
Create a new data review version by default if source data changes

jbothma:
Keeps the current accepted status and extracted_data (potentially modified and accepted by a reviewer) if the old orig_extraction_data matches the new orig_extraction_data unless `keep_if_equals` is set to False.

This also renames the model_version field to crawler_version and moves the responsibility to check if the version has changed to the crawler. This is because building this into `zavod.stateful.review` complicates the notions of whether historical data should be kept, while this is still a hypothetical case.

Only a couple of something_changed crawlers have been updated for discussion, but I'll do the rest in this PR once we settle on a direction.

The cases we want to know we can handle reasonable easily:

- they've made a correction to one page: A relevant name change should appear in the extracted data and trigger a re-review automatically
- [they've changed the article body markup across the board](https://github.com/opensanctions/opensanctions/pull/2849)
- we want to change the crawler in a way that we want to bulk-re-request reviews: bump CRAW:ER_VERSION, compare to get_review().crawler_version, and re-request review with accepted defaulting to true/false as appropriate.

### DIFF
--- a/datasets/us/cftc_enforcement_actions/crawler.py
+++ b/datasets/us/cftc_enforcement_actions/crawler.py
@@ -1,20 +1,21 @@
 from typing import Optional, List, Literal
 from pydantic import BaseModel, Field
 import re
+
+from rigour.mime.types import HTML
 from zavod.entity import Entity
 from zavod.shed import enforcements
 
-from lxml.html import HtmlElement, fromstring, tostring
+from lxml.html import HtmlElement, tostring
 
 from zavod.context import Context
 from zavod import helpers as h
 from zavod.shed.gpt import DEFAULT_MODEL, run_typed_text_prompt
 from zavod.stateful.review import (
-    Review,
     assert_all_accepted,
     request_review,
     get_review,
-    model_hash,
+    source_html_matches,
 )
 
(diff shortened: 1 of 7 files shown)


jbothma:
> What's a case where we'd set keep_if_equal=False - is this when the "media" text has changed but produces the same output from the prompt?

This is for the case where we're not confident the LLM is reliably picking up all the data. e.g. in seco I've seen it skip a bunch of positions - say 10-20% of the time it finds nothing when it should find 2 positions. In seco's case the data is the key so it'd just be a new review, but it's a scenario where polishing the prompt to get better has diminishing returns, but this could happen with a more important property in another crawler where we might not want to trust that a reprompt with the same result was correct.

pudo:
So I've enthusiastically been doing `DELETE FROM review WHERE dataset = 'us_ofac_press_releases' AND accepted = false;` every time I change my prompt - is that the cleaner way of doing that?

jbothma:
in favour of https://github.com/opensanctions/opensanctions/pull/2860 or https://github.com/opensanctions/opensanctions/pull/2900

PR #2860, opened 2025-09-22
Re review factor logic

jbothma:
Based on and takes it a bit further than https://github.com/opensanctions/opensanctions/pull/2848 - see the new stuff used in the CFTC crawler. Tests aren't updated  yet.

Move all the should_extract ~and should_review logic~ into  a function in zavod.stateful.review.

`request_review` still does its own `model_hash(extraction_result) != model_hash(review.orig_extraction_data)` - I think it makes sense not to update source_value until we've decided whether we need to re-review. Even if we're not re-reviewing, we want to update sourve_value.

It's tempting to move 

```python
if observation.should_extract:
        prompt_result = run_typed_text_prompt(
            context,
            extraction_config.prompt,
            source_value.value_string,
            extraction_config.data_model,
            model=extraction_config.llm_model,
        )
        review = request_review(
            context,
            source_value,
            extraction_config,
            orig_extraction_data=prompt_result,
            crawler_version=CRAWLER_VERSION,
        )
```

into observe_source_value - the two kinds of extractions we currently have are once-per-crawler backfilling based on regex/lookups, and LLM. The two options I see for this are:

- logic based on the extraction config type
- passing a callable (extraction_config, source_value_string) -> BaseModel to `observe_source_value`

Should we go that far or is this a good step for now? Is there a better way?

### DIFF
--- a/datasets/us/cftc_enforcement_actions/crawler.py
+++ b/datasets/us/cftc_enforcement_actions/crawler.py
@@ -2,11 +2,10 @@ from typing import Optional, List, Literal
 from pydantic import BaseModel, Field
 import re
 
-from rigour.mime.types import HTML
 from zavod.entity import Entity
 from zavod.shed import enforcements
 
-from lxml.html import HtmlElement, tostring
+from lxml.html import HtmlElement
 
 from zavod.context import Context
 from zavod import helpers as h
@@ -170,27 +169,17 @@ def crawl_enforcement_action(context: Context, date: str, url: str) -> None:
 
     source_value = HtmlSourceValue(
         key_parts=release_id,
-        mime_type=HTML,
         label="Enforcement Action Notice",
         url=url,
-        value_string=tostring(article_element, pretty_print=True, encoding="unicode"),
         element=article_element,
     )
     extraction_config = LLMExtractionConfig(
-        data_model=Defendants,
-        llm_model=DEFAULT_MODEL,
-        prompt=PROMPT,
+        data_model=Defendants, llm_model=DEFAULT_MODEL, prompt=PROMPT
     )
     observation = observe_source_value(context, source_value, extraction_config)
     review = observation.review
     if observation.should_extract:
-        prompt_result = run_typed_text_prompt(
-            context,
-            extraction_config.prompt,
-            source_value.value_string,
-            extraction_config.data_model,
-            model=extraction_config.llm_model,
-        )
+        prompt_result = h.prompt_for_review(context, extraction_config, source_value)
         review = request_review(
             context,
             source_value,
--- a/datasets/us/cftc_enforcement_actions/crawler.py
+++ b/datasets/us/cftc_enforcement_actions/crawler.py
@@ -1,20 +1,21 @@
 from typing import Optional, List, Literal
 from pydantic import BaseModel, Field
 import re
+
+from rigour.mime.types import HTML
 from zavod.entity import Entity
 from zavod.shed import enforcements
 
-from lxml.html import HtmlElement, fromstring, tostring
+from lxml.html import HtmlElement, tostring
 
 from zavod.context import Context
 from zavod import helpers as h
 from zavod.shed.gpt import DEFAULT_MODEL, run_typed_text_prompt
 from zavod.stateful.review import (
-    Review,
     assert_all_accepted,
     request_review,
     get_review,
-    model_hash,
+    source_html_matches,
 )
 
 Schema = Literal["Person", "Company", "LegalEntity"]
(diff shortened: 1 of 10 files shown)


datasets/us/cftc_enforcement_actions/crawler.py
```diff
@@ -207,39 +163,42 @@ def get_title(article_element: HtmlElement) -> str:
...
-    article_html = tostring(article_element, pretty_print=True, encoding="unicode")
-    release_id = get_release_id(url)
-    review = get_review(context, Defendants, release_id, MIN_MODEL_VERSION)
-    if review is None:
-        prompt_result = run_typed_text_prompt(context, PROMPT, article_html, Defendants)
+
+    source_value = HtmlSourceValue(
+        key_parts=release_id,
+        mime_type=HTML,
+        label="Enforcement Action Notice",
```
pudo: What are `label` and `mime_type` used for?

datasets/us/cftc_enforcement_actions/crawler.py
```diff
@@ -207,39 +163,42 @@ def get_title(article_element: HtmlElement) -> str:
...
+        url=url,
+        value_string=tostring(article_element, pretty_print=True, encoding="unicode"),
```
pudo: Should this be derived from element using a validator on the model?

datasets/us/cftc_enforcement_actions/crawler.py
```diff
@@ -207,39 +163,42 @@ def get_title(article_element: HtmlElement) -> str:
...
+    observation = observe_source_value(context, source_value, extraction_config)
+    review = observation.review
+    if observation.should_extract:
+        prompt_result = run_typed_text_prompt(
+            context,
+            extraction_config.prompt,
+            source_value.value_string,
+            extraction_config.data_model,
+            model=extraction_config.llm_model,
+        )
```
pudo: Nice! I'd love to see a helper that just takes `context, LLMExctractionConfig, HtmlSourceValue` ?

datasets/us/ofac/enforcement_actions/crawler.py
```diff
@@ -10,11 +10,10 @@
 from zavod.shed import enforcements
 from zavod.shed.gpt import DEFAULT_MODEL, run_typed_text_prompt
 from zavod.stateful.review import (
-    Review,
     assert_all_accepted,
     request_review,
     get_review,
-    model_hash,
+    source_html_matches,
```
pudo, line 16: Does this want to live in `zavod.helpers.html` ?

pudo (commented):
I feel like it's really starting to be a framework :)

datasets/us/cftc_enforcement_actions/crawler.py
jbothma: viewer tabs shown determined by mimetype. e.g. text/plain gets just one tab. there was briefly a PR with an image viewer but no use case yet (I think ph_sec is a use case now)

label is a tooltip on the viewers to hint to the user where it came from, but I see it doesn't work on the rendered html tab. I see that being useful if it's e.g. an image of a PDF or something

image

datasets/us/cftc_enforcement_actions/crawler.py
jbothma: do you mean value_string?
Would the validator look at mime_type to decide that text/play should just be the string passed to value_string?

pudo: I mean we're instantiating a HTMLSourceValue here, so I wonder if it could have a constructor function that auto-sets the MIME type and accepts and element? I'm basically just chasing for code reductions here because we're going to have all this across like 20 crawlers really quick, and it feels pretty DRY-able.

jbothma: ah I'm a numpty. Of course

datasets/us/cftc_enforcement_actions/crawler.py
jbothma: The main place it's felt useful so far is for fed_enforcements to say that the name of this field is "Banking Organizations" - it felt like for someone who isn't looking at the crawler source and original document, it can be helpful.

An alternative might be a metadata field in the dataset where we just give a dataset-wide burb like 

> The reviews for this dataset come from a column named "Banking Organizations". It can contain one or more organizations. The rough location of the location might be included.

Maybe field-specific things should go in the schema field descriptions like "Don't accept an alias that is merely used as a reference to the full name within the original text. Aliases should be distinct other names, e.g. `Ronnie` for `Ronald McDonald`. And we should then just improve the UI so that documentation is really easily visible.

jbothma: Do you reckon the label is useful at the moment or should I drop it and we see if we feel like bringing the metadata field in separately?

jbothma:
Should we drop Review.crawler_version until something needs it?

datasets/us/ofac/enforcement_actions/crawler.py
jbothma: ah - this is replaced by review.should_extract calling SourceValue.matches()

jbothma:
in favour of https://github.com/opensanctions/opensanctions/pull/2900

PR #2862, opened 2025-09-23
[eu_sanctions_map] Crawl consolidated Vessels sheet

nvmbrasserie:


datasets/eu/sanctions_map/crawler.py
```diff
@@ -1,13 +1,34 @@
...
 DATA_URL = "https://www.sanctionsmap.eu/api/v1/data?"
 REGIME_URL = "https://www.sanctionsmap.eu/api/v1/regime"
 
+# VESSELS_URL originates from the EU Sanctions Map:
+# https://www.sanctionsmap.eu/#/main
+#
+# It is the only file published directly by the EU that provides the
+# consolidated vessels list along with links to the Official Journal.
+#
+# Request Data
```
leonhandreke: What's this request data? My impression is that there is some Javascript on the page that opens this link in a normal GET request.

datasets/eu/sanctions_map/crawler.py
```diff
@@ -74,3 +95,33 @@ def crawl(context: Context):
...
+        vessel = context.make("Vessel")
+        vessel.id = context.make_id(name, imo)
+        vessel.add("name", name)
+        vessel.add("imoNumber", imo)
+        vessel.add("topics", "sanction")
+        sanction = h.make_sanction(context, vessel, key=row.pop("column_0"))
+        sanction.add("sourceUrl", row.pop("link_to_relevant_eu_official_journal"))
+        h.apply_date(sanction, "startDate", row.pop("date_of_application"))
+        # TODO: Add program mappings
+        # sanction.add("program", "EU Consolidated List of Designated Vessels")
```
leonhandreke: (nit) Not really active yet, but will we have any way to figure out the policy program that lead to the sanction?

datasets/eu/sanctions_map/crawler.py
```diff
@@ -1,13 +1,34 @@
...
 from zavod import helpers as h
 
 DATA_URL = "https://www.sanctionsmap.eu/api/v1/data?"
 REGIME_URL = "https://www.sanctionsmap.eu/api/v1/regime"
 
+# VESSELS_URL originates from the EU Sanctions Map:
+# https://www.sanctionsmap.eu/#/main
+#
+# It is the only file published directly by the EU that provides the
+# consolidated vessels list along with links to the Official Journal.
```
leonhandreke: IIUC, this list is compiled from all the amendments to Annex XLII of Council Regulation (EU) 833/2014.

Do we use the links to the journal for dedupe? Maybe we could add a bit more context of how this came into being here or in the dataset metadata.

Why do the individual regimes not contain these vessels, why do we need this separate Excel?

leonhandreke (commented):
- Should we add a 4 month manual check interval (with good instructions how to do it)?
- Should we update the description metadata with this information?

datasets/eu/sanctions_map/crawler.py
```diff
@@ -1,13 +1,34 @@
...
+#   MIME Type: text/plain
+#   Encoding: UTF-8
+#   Request Data: <-- here is the source of the URL below
+#
+# If this URL changes, we can fall back to the Danish Maritime Authority’s list:
+# https://www.dma.dk/growth-and-framework-conditions/maritime-sanctions/sanctions-against-russia-and-belarus/eu-vessel-designations
+# Note: the DMA list does not include links to the Official Journal.
+
+VESSELS_URL = (
+    "https://dk9q89lxhn3e0.cloudfront.net/EU+designated+vessels-+conso+July+2025.xlsx"
```
pudo, line 32: Can we fetch the URL from the home page?

datasets/eu/sanctions_map/crawler.py
nvmbrasserie: it's javascript but that's the only place inside the request where I found a parseable link

datasets/eu/sanctions_map/crawler.py
leonhandreke: It's somewhere deep in a JS pack file with a checksum in the name.

datasets/eu/sanctions_map/crawler.py
nvmbrasserie: yeah, it won't be an issue, I'll add a program code so that we don't forget

PR #2877, opened 2025-09-29
Foreign Representatives in Canada: Heads of Missions

nvmbrasserie:
opensanctions/crawler-planning#326


datasets/ca/foreign_reps/crawler.py
```diff
@@ -0,0 +1,64 @@
...
+    for c in containers:
+        name = c.find(".//div[@title='Salutation and Name']").text_content().strip()
+        title = c.find(".//div[@title='Title']").text_content().strip()
+        spouse = c.find(".//div[@title='Spouse']").text_content().strip()
+        mission = c.find(".//div[@title='Mission Title']").text_content().strip()
+        # Skip entries without a name
+        if not name:
+            continue
+        person = context.make("Person")
+        person.id = context.make_id(name, title)
```
jbothma: name and country feels a bit more like a unique key on this page. Like I think the chance of a duplicate is very low, but it's just the things that are especially distinctive about the individuals.

```suggestion
        person.id = context.make_id(name, country)
```

datasets/ca/foreign_reps/crawler.py
```diff
@@ -0,0 +1,64 @@
...
+
+        for field, xpath in FIELDS:
+            elem = c.find(xpath)
+            assert elem is not None
```
jbothma: ❤️

datasets/ca/foreign_reps/crawler.py
```diff
@@ -0,0 +1,64 @@
...
+        position = h.make_position(
+            context,
+            name=person.get("position"),
+            country=person.get("country"),
+            topics=["gov.national", "role.diplo"],
+            description=mission,
+        )
+        categorisation = categorise(context, position, is_pep=True)
+        if categorisation.is_pep:
+            occupancy = h.make_occupancy(
```
jbothma: it's preferable not to specify status if you don't need to. The default should be to make status current when there's no end date - is that not happening?

datasets/ca/foreign_reps/crawler.py
```diff
@@ -0,0 +1,64 @@
...
+FIELDS = [
+    ("country", ".//h2[@title='Country']"),
+    ("name", ".//div[@title='Salutation and Name']"),
+    ("position", ".//div[@title='Title']"),
+    ("address", ".//div[@title='Address']"),
+    ("address", ".//div[@title='City']"),
+]
+
+
+def emit_relation(context, spouse_name, person_id):
```
jbothma: Since all the code inside is spouse-specific, let's call it emit_spouse. You don't want the name to make it sound like the function is more general than it is.

```suggestion
def emit_spouse(context, spouse_name, person_id):
```

datasets/ca/foreign_reps/crawler.py
```diff
@@ -0,0 +1,64 @@
+from zavod import Context, helpers as h
+from zavod.stateful.positions import categorise, OccupancyStatus
+
+FIELDS = [
+    ("country", ".//h2[@title='Country']"),
+    ("name", ".//div[@title='Salutation and Name']"),
```
jbothma: let's use the name we have.

Alternatively, let's parse these into a dict and do everything from the dict. But the repetition is not so nice.

we should also strip the ginormous titles/salutations from the names otherwise matching on these names will be terrible. How about a lookup of known salutations and a warning of unknown ones, using the name as is when we haven't confirmed a salutation that can be stripped?

datasets/ca/foreign_reps/crawler.py
```diff
@@ -0,0 +1,64 @@
...
+    ("position", ".//div[@title='Title']"),
+    ("address", ".//div[@title='Address']"),
+    ("address", ".//div[@title='City']"),
```
jbothma: we dont normally do addresses for PEPs https://zavod.opensanctions.org/best_practices/priorities/#politically-exposed-persons
So while it looks like these want to be concatenated, I think we're better of just deleting

datasets/ca/foreign_reps/ca_foreign_reps.yml
```diff
@@ -0,0 +1,42 @@
...
+  in Canada, including ambassadors, high commissioners, and other heads of diplomatic
+  missions.
+url: https://w05.international.gc.ca/protocol-protocole/Heads-Chefs.aspx
+tags:
+  - list.pep
+publisher:
+  name: Government of Canada
+  description: |
+    The list is maintained and published by the government of Canada.
+  country: zz
```
jbothma:
```suggestion
  country: ca
```

datasets/ca/foreign_reps/ca_foreign_reps.yml
```diff
@@ -0,0 +1,42 @@
...
+  List of foreign diplomatic representatives and heads of missions accredited to Canada.
+description: |
+  This dataset contains information about foreign diplomatic representatives serving
+  in Canada, including ambassadors, high commissioners, and other heads of diplomatic
+  missions.
+url: https://w05.international.gc.ca/protocol-protocole/Heads-Chefs.aspx
+tags:
+  - list.pep
+publisher:
+  name: Government of Canada
```
jbothma:
```suggestion
  name: Global Affairs Canada
```

datasets/ca/foreign_reps/ca_foreign_reps.yml
```diff
@@ -0,0 +1,42 @@
...
+  description: |
+    The list is maintained and published by the government of Canada.
```
jbothma: let's make this specific to ~Foreign~Global Affairs

datasets/ca/foreign_reps/crawler.py
jbothma: can we give spouse names the same treatment? Could be same lookup.

Perhaps a
```python
def strip_title(name):
    res = context.lookup(...)
    if res
        name = name.replace(...
    ...
   return name
```

datasets/ca/foreign_reps/ca_foreign_reps.yml
```diff
@@ -0,0 +1,42 @@
+title: "Foreign Representatives in Canada: Heads of Missions"
+entry_point: crawler.py
+prefix: ca-missions
```
jbothma: Just an alternative - not really sure what's better. fordip for foreign diplomats. as in, not canadian missions but foreign missions.

```suggestion
prefix: ca-fordip?
```

datasets/ca/foreign_reps/crawler.py
nvmbrasserie: I used regex for the salutations, there are only 5 unique ones. it seems to do the job

PR #2881, opened 2025-10-01
[ps_ohchr_settlement] parse the webpage

nvmbrasserie:


datasets/ps/ochcr_settlement/crawler.py
```diff
@@ -1,62 +1,55 @@
...
-    if section.startswith("A. ") or "no longer involved" in section:
-        context.log.info(
-            f"Skipping company {name} as marked as "
-            f"not involved (Section: {section})"
-        )
-        return
+from zavod import Context, helpers as h
+from zavod.shed import zyte_api
+
+OHCHR_BHR = "OHCHR-BHR"
```
jbothma: let's name the variable so that we know what it represents. My gut feeling is that this doesn't have to be a variable - what was your thinking for making it a variable?

```suggestion
PROGRAM_ID = "OHCHR-BHR"
```

datasets/ps/ochcr_settlement/crawler.py
```diff
@@ -1,62 +1,55 @@
...
+    table_xpath = "//h5[contains(., 'Business enterprises involved in listed activities')]/following-sibling::div//table"
+    doc = zyte_api.fetch_html(
+        context,
+        context.data_url,
+        table_xpath,
+        html_source="httpResponseBody",
+        absolute_links=True,
+        cache_days=1,
+    )
+    assert_database_hash(doc)
```
jbothma, line 60: if the dataset isn't manually updated any more, do we need the hash assertion? Are you concerned that the page changes in a way that we end up not selecting the table? That should be solved with

- asserting that exactly one table was selected by the xpath
- asserting a minimum number of entities were extracted using data assertions

jbothma (commented):
This adds a couple of nice things:
- more countries
- responses from the enterprises

It loses the previous names. scraping them is very possible, but it would be about as many lines of code as there are previous names.

It looks like it works very nicely to just paste the html table into google sheets. It's also been updated 5 times in 7 years.

What do you think of keeping it as a google sheets crawler? Rather than deploying some code that likely needs to be updated in the next 2 years, could we just

1. adding a page hash check
1. paste the data into another sheet
2. format it to be consistent with the current sheet
3. paste that below the current data
4. run the crawler
5. dedupe
6. delete the old rows

Then I think we get all the good stuff, we lose none of the new stuff, and we keep the code trivial.

datasets/ps/ochcr_settlement/crawler.py
```diff
@@ -1,62 +1,55 @@
...
+def crawl_row(context: Context, row):
+    response = h.links_to_dict(row.pop("response_from_business_enterprise"))
+    response_url = response.get("response")
+    str_row = h.cells_to_str(row)
+    name = str_row.pop("business_enterprise")
+    country = str_row.pop("home_state")
+    activities = str_row.pop("listed_activity_subparagraph_of_paragraph_96")
+
     entity = context.make("Company")
     entity.id = context.make_id(name, country)
```
nvmbrasserie, line 30: there is no need to dedupe, most of the ids were preserved

datasets/ps/ochcr_settlement/crawler.py
nvmbrasserie: It became a pattern with the hardcoded program codes, and I'd rather keep it this way

datasets/ps/ochcr_settlement/crawler.py
nvmbrasserie: My concern is that they can split the existing table into several or restructure it otherwise

jbothma: yeah if we're maintaining the sheet manually we definitely want a content assertion. thanks

datasets/ps/ochcr_settlement/ps_ohchr_settlement.yml
```diff
@@ -60,7 +58,8 @@ description: |
   totally or partially by settlers for developing, expanding and
   maintaining the settlements.
 
-  The data has been extracted manually from the [30 June 2023 update PDF.](https://www.ohchr.org/sites/default/files/documents/hrbodies/hrcouncil/sessions-regular/session31/database-hrc3136/23-06-30-Update-israeli-settlement-opt-database-hrc3136.pdf).
+  The data has been extracted manually from the [30 June 2023 update PDF](https://www.ohchr.org/sites/default/files/documents/hrbodies/hrcouncil/sessions-regular/session31/database-hrc3136/23-06-30-Update-israeli-settlement-opt-database-hrc3136.pdf)
+  and the [OHCHR Database of business enterprises involved in the listed activities](https://www.ohchr.org/en/business/bhr-database) on October 1, 2025.
```
jbothma: I think we want to be left with one dataset - just the organisations currently considered to be doing business there.

Since this is quite a sensitive thing and sometimes users don't pay attention to the whole topic implies risk thing, I don't think we should change now to include delisted companies.

datasets/ps/ochcr_settlement/crawler.py
```diff
@@ -1,62 +1,67 @@
...
     entity.add("notes", f"Listed activities: {activities}")
-    entity.add("notes", f"Date of last update: {date}")
-    entity.add("topics", "debarment")
-    if notes:
-        entity.add("notes", notes)
-    entity.add("sourceUrl", source_url)
-    context.emit(entity)
+    # Only entities from the latest update are marked as 'debarred'
+    entity.add("topics", row.pop("Topics"))
+    entity.add("notes", row.pop("Notes").strip())
```
jbothma: The comment says only marked debarred, but the sheet can be any topic. So there's nothing keeping them in sync.

Either way, let's tag everyone in the sheet with debarred (using python as before) and let's remove the previous snapshot from the sheet once we've deduped.

nvmbrasserie: should I merge and dedupe?

PR #2884, opened 2025-10-01
[il_mod_crypto] homoglyphs mappings

nvmbrasserie:


jbothma (approved):
this is....nutty....

I don't think we currently have a library that can do this. I don't think unicode (de)composition solves this problem, becuase it is most or all of these could well be single codepoints that just look like ascii characters. And mapping to potential ascii characters could have more than one possibility https://pypi.org/project/homoglyphs/#:~:text=Converting%20glyphs%20to%20ASCII%20chars

Let's go with this - I think the oh's are all oh's and not zeros. And we can discuss if we want to bring in a library for this elsewhere.

PR #2888, opened 2025-10-01
Daily-fixes-part-2

nvmbrasserie:


datasets/au/dfat_sanctions/crawler.py
```diff
@@ -43,6 +43,7 @@ def clean_date(date):
         "Between",
         "circa",
         "Circa",
+        "_x000D_",
```
jbothma, line 46: I don't even want to know what happened there...
image

PR #2893, opened 2025-10-02
rename hardcoded program keys across codebase

nvmbrasserie:


datasets/lt/fiu_freezes/crawler.py
```diff
@@ -6,8 +6,8 @@
 from zavod.shed.zyte_api import fetch_html
 
 
-LT_SL = "LT-SL"
-LT_UNSCR1373 = "LT-UNSCR1373"
+PROGRAM_KEY_SL = "LT-SL"
+PROGRAM_KEY_UNSCR1373 = "LT-UNSCR1373"
```
jbothma: at this point it can just be inlined, right? Do we gain anything from factoring these out as variables/"constants"?

jbothma: Like, the idea of factoring it out as a constant is kinda nice - we see the program key right at the top of the module. If we move a "dataset default program key" to metadata one day, it's kinda sorta already a thing although it will in really make very little difference to that effort. 

But when this is something we're not regularly tweaking like SLEEP_SECONDS=1 or CACHE_DAYS=7 and we end up naming the variable after the value, let's just use the value. But tell me if I'm missing something.

datasets/us/cuba_sanctions/crawler.py
```diff
@@ -12,8 +12,8 @@
 ACCOMMODATIONS_URL = "https://docs.google.com/spreadsheets/d/e/2PACX-1vQMquWjNWZ09dm9_mu9NKrxR33c6pe4hpiGFeheFT4tDZXwpelLudcYdCdME820aKJJo8TfMKbtoXTh/pub?gid=1890354374&single=true&output=csv"
 ORIGINAL_RESTRICTED_ENTITIES_URL = "https://www.state.gov/division-for-counter-threat-finance-and-sanctions/cuba-restricted-list"
 RESTRICTED_ENTITIES_URL = "https://docs.google.com/spreadsheets/d/e/2PACX-1vQMquWjNWZ09dm9_mu9NKrxR33c6pe4hpiGFeheFT4tDZXwpelLudcYdCdME820aKJJo8TfMKbtoXTh/pub?gid=0&single=true&output=csv"
-US_DOS_CU_REA = "US-DOS-CU-REA"
-US_DOS_CU_PAL = "US-DOS-CU-PAL"
+PROGRAM_KEY_REA = "US-DOS-CU-REA"
+PROGRAM_KEY_PAL = "US-DOS-CU-PAL"
```
jbothma: Same as above

jbothma (approved):
I think the two cases where the variable name is still the value, let's just inline the value. Other than that, thanks for the broad address to my nitpick

datasets/lt/fiu_freezes/crawler.py
nvmbrasserie: do you mean using the value only in those cases when we have more than 1 program_key?

jbothma: exactly

PR #2899, opened 2025-10-03
[us_bis_denied] Handle whatever format they feel like sending on the …

jbothma:
…day.

So much more exciting than just a CSV each time

---

`added=1951 dataset=us_bis_denied deleted=1888 metric=delta_counts modified=431`

the rekey is because the start date is part of the id and they've dropped leading zeros. More fun.

Maybe a good time to replace the start date with something more stable, or to drop the dataset entirely.

### DIFF
--- a/datasets/us/bis_denied_new/crawler.py
+++ b/datasets/us/bis_denied_new/crawler.py
@@ -1,0 +1,109 @@
+import csv
+
+import openpyxl
+import xlrd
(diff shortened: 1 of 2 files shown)


pudo:
The thing is that with the BIS 50% rule, all eyes are on this thing and so double-sourcing it may not be the worst idea. I'll do a little dedupe supervision :)

PR #2900, opened 2025-10-05
Always prompt, usually hit the cache.

jbothma:
If we just always prompt, the crawler can get very very clean. The only example updated is CFTC so see that one.

There's a fair bit happening beneath `review_source_value` but so far I think it covers all our use cases.

Haven't updated the tests beyond https://github.com/opensanctions/opensanctions/pull/2860


zavod/zavod/stateful/review.py
```diff
@@ -170,130 +276,126 @@ def review_key(parts: str | List[str]) -> str:
...
+
+    review = Review.by_key(
+        context.conn, data_model, dataset=context.dataset.name, key=key_slug
+    )
+    save_new_revision = False
     if review is None:
-        # First insert
-        context.log.debug("Requesting review", key=key_slug)
+        context.log.debug("Creating new review", key=key_slug)
         review = Review(
```
jbothma: Any idea why I'm getting
```
zavod/zavod/stateful/review.py:328: error: Call to untyped function "Review" in typed context  [no-untyped-call]
Found 1 error in 1 file (checked 1 source file)
```
on this?
`Review[ModelType]` doesn't fix it, and `Review[data_model]` isn't allowed.

PR #2902, opened 2025-10-06
Add warn flag for common case of warning of missing lookup option

jbothma:
We often want to warn if we don't have a matching lookup for a value. This adds a flag to `context.lookup*` to warn with the lookup name and non-matching value.

Should this rather be a helper?

This looks like this:
```
 [warning  ] No matching lookup found.      [jp_mof_sanctions] dataset=jp_mof_sanctions lookup=schema value=FOOBAR暴力的行為に関与するイスラエルの入植者
```

With the ones I've modified here, at least, I haven't found any that need a custom warning. I think with a well-named lookup, they'll all be self-explanatory.

So far just demoing a few cases, but I'll do the rest if we like the approach. We can find more cases like this:
```
grep -r -A10 context.lookup datasets/|grep -B10 log.warn
```

I think there are over 15 cases that can be simplified with this.

### DIFF
--- a/datasets/fr/tresor/crawler.py
+++ b/datasets/fr/tresor/crawler.py
@@ -248,7 +248,7 @@ def apply_prop(context: Context, entity: Entity, sanction: Entity, field: str, v
         full_address = value.pop("Adresse")
         country = value.pop("Pays")
         if ";" in full_address:
-            res = context.lookup("addresses", full_address, warn=True)
+            res = context.lookup("addresses", full_address, warn_unmatched=True)
             addresses = res.addresses if res else [full_address]
             assert isinstance(addresses, (list, str)), full_address
         else:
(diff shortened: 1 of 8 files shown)


datasets/fr/tresor/crawler.py
```diff
@@ -246,12 +245,16 @@ def apply_prop(context: Context, entity: Entity, sanction: Entity, field: str, v
...
-            context,
-            full=value.pop("Adresse"),
-            country=value.pop("Pays"),
-        )
-        h.copy_address(entity, address)
+        full_address = value.pop("Adresse")
+        country = value.pop("Pays")
+        if ";" in full_address:
+            res = context.lookup("addresses", full_address, warn=True)
+            addresses = res.addresses if res else [full_address]
```
jbothma, line 252:
```
No matching lookup found.      [fr_tresor_gels_avoir] dataset=fr_tresor_gels_avoir lookup=addresses value='Head Office Building, 327 Takeghani (Taleghani) Avenue, Téhéran 15817, Iran; P.O. Box 11365-5964, Téhéran, 15817'
```

Issue #2913, opened 2025-10-07
[ann_graph_topics] Assertion failed for value 424: <Assertion entity_count gte 1200 filter: schema=Organization>


jbothma:
```
Assertion failed for value 424: <Assertion entity_count gte 1200 filter: schema=Organization>
```

locally I got

```
Assertion failed for value 15146: <Assertion entity_count gte 45800 filter: schema=Company> 
Assertion failed for value 9813: <Assertion entity_count gte 12200 filter: schema=Person> 
Assertion failed for value 500: <Assertion entity_count gte 1300 filter: schema=LegalEntity> 
Assertion failed for value 212: <Assertion entity_count gte 1200 filter: schema=Organization> 
Assertion failed for value 5: <Assertion entity_count gte 15 filter: schema=Vessel> 
Assertion failed for value 2: <Assertion entity_count gte 5 filter: schema=PublicBody> 
```

the majority lost are ru-inn

```
diff <(cat ann_graph_topics-statements.pack | csvcut -c1 | sort) <(cat data/datasets/ann_graph_topics/statements.pack | csvcut -c1 | sort)|cut -c1-8|sort | uniq -c | sort -
...
336 < lei-54
 465 ---
 976 < gem-ow
1012 < lei-52
1242 < ua-edr
1254 < icijol
113162 < ru-inn
```

but that could be because most keys in the dataset are of that scheme
```
cut -c1-6 ann_graph_topics-statements.pack | sort | uniq -c | sort -n
...
1026 lei-52
1356 ua-edr
1390 gem-ow
1770 icijol
10440 am-het
145322 ru-inn
```
why did am_het manage not to disappear?

jbothma:
did we update anything relevant to iterating over a store or adjacent entities?

if we do two exports, do we miss the same entities? Is this data related?

here's an entity that dropped out of ann_graph topics in my local run https://www.opensanctions.org/entities/ru-inn-770102822317/ but their sanctioned assets are still in default according to the website.

PR #2917, opened 2025-10-08
[ann_graph_topics] A bunch of entities went from Organization to Company

jbothma:
solves https://github.com/opensanctions/opensanctions/issues/2913

### DIFF
--- a/datasets/_analysis/ann_graph_topics/ann_graph_topics.yml
+++ b/datasets/_analysis/ann_graph_topics/ann_graph_topics.yml
@@ -40,10 +40,10 @@ inputs:
 assertions:
   min:
     schema_entities:
-      Company: 45800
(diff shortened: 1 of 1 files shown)


jbothma:
I used
```
diff <(cat ann_graph_topics-20251007005609-jbm-statements.pack | csvcut -c1-2 | sort) <(cat data/datasets/ann_graph_topics/statements.pack | csvcut -c1-2 | sort)|less
```

but diffing statistics.json would probably be have been just as clear.

Importantly, my local run of this didn't help the first time, because I had some partial runs of relevant datasets that were loaded into the store which wildly messed with the numbers.

A temporary archive of the failed run's data would have helped me debug with what happened in prod, rather than having to run it locally

PR #2921, opened 2025-10-09
disable context.data_time overrides

### LINKED ISSUE #2623: Can coverage.end stop overriding last_seen on statements?
At the moment, the context overwrites the data time based on coverage.end: 

https://github.com/opensanctions/opensanctions/blob/main/zavod/zavod/context.py#L71-L81

While cute to show in the UI, this basically leaks metadata into the dataset operating semantics in ways that might break change detection and cause other unforeseen effects. The initial reason for doing this was to show the end date on individual entities - but that could be done by accessing dataset metadata? 

I propose we remove this and maybe update the website to show coverage end.

pudo:
fixes #2623

### DIFF
--- a/zavod/zavod/shed/wikidata/human.py
+++ b/zavod/zavod/shed/wikidata/human.py
@@ -4,6 +4,7 @@ from datetime import timedelta
 
 from nomenklatura.wikidata import Item, WikidataClient
 
+from zavod import settings
 from zavod import Context, Entity
 from zavod.shed.wikidata.country import is_historical_country, item_countries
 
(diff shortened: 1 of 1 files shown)


datasets/az/fiu_sanctions/crawler.py
```diff
@@ -22,11 +21,11 @@ def crawl(context: Context):
...
-    date = doc.getroot().get("dateGenerated")
-    if date is None:
-        context.log.error("No date found in XML file")
-        return
-    context.data_time = datetime.strptime(date, "%Y-%m-%dT%H:%M:%S")
+    # date = doc.getroot().get("dateGenerated")
+    # if date is None:
+    #     context.log.error("No date found in XML file")
+    #     return
+    # datetime.strptime(date, "%Y-%m-%dT%H:%M:%S")
```
jbothma, line 28: do we still want this around?

pudo: I do want to find a place to store it, eventually. Just don't know where....

pudo: coverage.source_updated_at?

jbothma: sounds good.

Alternatively `coverage.source_published_at` which feels maybe a bit more specifically an explicit publication date, but I think `source_updated_at` is more flexible.

PR #2924, opened 2025-10-10
Parliaments Canada, Switzerland, Malta, Cyprus, Germany, Austria, India, Japan

pudo:


datasets/de/bundestag/crawler.py
```diff
@@ -0,0 +1,75 @@
...
+    )
+    context.emit(position)
+
+    for mdb in doc.findall(".//MDB"):
+        person = context.make("Person")
+        person.id = context.make_slug("mdb", mdb.findtext("./ID"))
+
+        if mdb.findtext(".//STERBEDATUM"):
+            continue
+        # person.add("topics", "role.pep")
```
leonhandreke: Why is this commented out?

datasets/in/sansad/crawler.py
```diff
@@ -0,0 +1,248 @@
...
+def emit_rca(context: Context, person: Entity, name: str | None, role: str) -> None:
+    if not name:
+        return
+    if name.startswith("Late "):
+        return
+    rca = context.make("Person")
+    rca.id = context.make_id(person.id, role, name)
+    rca.add("name", clean_text(name))
+    rca.add("country", "in")
+    # rca.add("topics", "role.rca")
```
leonhandreke: Same here, why not)?

datasets/in/sansad/crawler.py
```diff
@@ -0,0 +1,248 @@
...
+                start_date=start,
+                end_date=end,
+                propagate_country=False,
+            )
+            if occupancy is not None:
+                context.emit(occupancy)
+    except RequestException as exc:
+        context.log.info(f"Broken MP term data for {person.id}", exc=str(exc))
+
+    # if person.has("deathDate"):
```
leonhandreke: We do this in de_bundestag - why don't we do it here? Death date reading above is also commented out.

datasets/de/bundestag/crawler.py
pudo: It's going to be added automatically by the `make_occupancy` code, only for MdBs still inside our time limits

datasets/in/sansad/crawler.py
pudo: Leaving it up to `ann_graph_topics` even though it hurts

datasets/in/sansad/crawler.py
pudo: So I found that `expirationDate` had a lot of dates in the future. Suspecting it's the end of the person's seat, not their life.

datasets/in/sansad/crawler.py
leonhandreke: Ah, that makes sense. Thanks!

PR #2956, opened 2025-10-17
Fixes-17-10

nvmbrasserie:


datasets/gb/coh/gb_coh_psc.yml
```diff
@@ -314,3 +314,77 @@ lookups:
           - Englan
         value: gb
         weight: -1
+  type.name:
```
leonhandreke, line 317: Can we maybe just put a thing in the code to throw out 0-9, ., : ?

nvmbrasserie: I'll add it to the discussion, I think it's a good idea

PR #2961, opened 2025-10-18
fix: ensures DB_BATCH_SIZE is used

kennell:
`DB_BATCH_SIZE` exists in settings.py but is never used, here is a proposed fix.

### DIFF
--- a/zavod/zavod/cli.py
+++ b/zavod/zavod/cli.py
@@ -8,7 +8,6 @@ import click
 from followthemoney.cli.util import InPath, OutPath
 from followthemoney.statement import CSV, FORMATS
 from nomenklatura.matching import DefaultAlgorithm
-from nomenklatura.settings import STATEMENT_BATCH
 from nomenklatura.tui import dedupe_ui
 from zavod import settings
 from zavod.archive import clear_data_path, dataset_state_path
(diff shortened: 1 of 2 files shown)


pudo:
Hey @kennell - welcome, and thanks for contributing this fix. Very much appreciate the sense-making of all the places where defaults are defined here. Looking at it, I'm a bit undecided: why not use `nk_settings.STATEMENT_BATCH` (env: `NOMENKLATURA_STATEMENT_BATCH`) and get rid of the zavod override entirely? My sense is it takes us a bit more in the direction of having one place where something is set, rather than multiple.

kennell:
Hi. I didn't notice `NOMENKLATURA_STATEMENT_BATCH` env exists. So removing `DB_BATCH_SIZE` to avoid confusion would also be fine.

However, i'd argue to change the default `NOMENKLATURA_STATEMENT_BATCH` to 1000. With the current default of 3000, inserts into the `statement` table (14 columns) produce 3000 × 14 = 42,000 bound variables, which exceeds SQLite's default [SQLITE_MAX_VARIABLE_NUMBER of 32,766](https://www.sqlite.org/limits.html). A default of 1000 keeps us at 14,000 variables, safely below that limit.

kennell:
Thank you, will close this PR then.

PR #2976, opened 2025-10-21
German Bundesrat

nvmbrasserie:
opensanctions/crawler-planning#400

### DIFF
--- a/datasets/de/bundesrat/crawler.py
+++ b/datasets/de/bundesrat/crawler.py
@@ -1,20 +1,25 @@
 import re
+from typing import Optional
+
 from datapatch import Lookup
+from lxml.etree import _Element
 from lxml.html import HtmlElement
 from normality import squash_spaces
+from zavod.stateful.positions import OccupancyStatus, categorise
 
-from zavod import Context, helpers as h
-from zavod.stateful.positions import categorise, OccupancyStatus
+from zavod import Context
+from zavod import helpers as h
 
 # Geboren am DD.MM.YYYY
 # Geboren YYYY
 DOB_REGEX = re.compile(r"^Geboren(?: am\s+(\d{1,2}\.\d{2}\.\d{4})| (\d{4}))")
+# Some members have no positons/membership details listed below their name (not the same as the biography)
 NO_DETAILS_LIST = [
     "https://www.bundesrat.de/SharedDocs/personen/DE/laender/nw/feller-dorothee.html"
 ]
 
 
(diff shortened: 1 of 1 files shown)


datasets/de/bundesrat/crawler.py
```diff
@@ -0,0 +1,86 @@
...
+    context.log.warning(f"No {lookup_key} found in biography.", biography=text)
+    return None
+
+
+def crawl_item(context: Context, item: HtmlElement) -> None:
+    url = item.find(".//a").get("href")
+    assert url, "No URL found for member"
+    member_doc = context.fetch_html(url, cache_days=3)
+    details = member_doc.find(".//div[@class='text-box']")
+    if "N. N." in details.text_content().strip():
```
leonhandreke: (nit) Use `h.element_text` to appease the type checker and also do the strip for you. Everywhere, I think it strips a lot of different whitespace.

datasets/de/bundesrat/crawler.py
```diff
@@ -0,0 +1,86 @@
+import re
+from normality import squash_spaces
+from lxml.html import HtmlElement
+
+from zavod import Context, helpers as h
+from zavod.stateful.positions import categorise
+
+START_DATE_REGEX = re.compile(
+    r"(?:Stellvertretendes )?Mitglied des Bundesrates seit (\d{2}\.\d{2}\.\d{4})|,?\s*[Ss]eit (\d{2}\.\d{2}\.\d{4})\s*(?:Stellvertretendes\s+)?Mitglied des Bundesrates"
```
leonhandreke: I trust you it works, but what does this match? Does it do "Mitglied des Bundesrates seit [date1], seit [date2] Stellvertretendes..."

In that case, don't we only care about date1?

Could you add some examples that this matches in a comment? And maybe mention somewhere (metadata? crawl_item?) that we're ignoring deputy designations?

datasets/de/bundesrat/crawler.py
```diff
@@ -0,0 +1,86 @@
...
+)
+DOB_REGEX = re.compile(
+    r"^Geboren(?: am\s+(?:\d{1,2}\.\d{2}\.\d{4}|\d{1,2}\s+\w+\s+\d{4})| (\d{4}))|"
+)
+
+
+def extract_date(context, regex, lookup_key, text):
```
leonhandreke: (nit) let's pass a `Lookup` instead of a `str` here for nicer types :)

PR #2981, opened 2025-10-22
Hellenic Parliament

nvmbrasserie:
opensanctions/crawler-planning#402

### DIFF
--- a/datasets/gr/parliament/crawler.py
+++ b/datasets/gr/parliament/crawler.py
@@ -8,6 +8,12 @@ from zavod.stateful.positions import categorise
 
 HEADERS = {
     "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
+    "Accept-Encoding": "gzip, deflate, br",
+    "Accept-Language": "en-GB,en;q=0.9",
+    "Priority": "u=0, i",
+    "Sec-Fetch-Dest": "document",
+    "Sec-Fetch-Mode": "navigate",
+    "Sec-Fetch-Site": "same-origin",
 }
 # List of IDs to skip:
 # These entries are explicitly skipped because they currently contain no data.
--- a/datasets/gr/parliament/gr_parliament.yml
+++ b/datasets/gr/parliament/gr_parliament.yml
@@ -30,7 +30,7 @@ dates:
   formats: ["%d/%m/%Y", "%b %d %Y"]
 http:
   user_agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.2 Safari/605.1.15 (zavod; opensanctions.org)
-  additional_retry_statuses: [403]
+  additional_retry_statuses: [403, 443]
 
 assertions:
   min:


datasets/gr/parliament/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+            context.emit(occupancy)
+            context.emit(person)
+
+    context.audit_data(str_row, ["date", "costituency"])
+
+
+def crawl(context: Context):
+    doc = context.fetch_html(context.data_url, cache_days=1)
+    # Extract MP IDs and names
+    ids = doc.xpath(f"{MP_XPATH}/@value")
```
leonhandreke: Would something like:

```python
mp_option_els = doc.xpath(...)

for id := el.get("value"), name := h.el_text(el) for el in mp_option_els:
    ...
```
work here?

(Point being: do we need to call `doc.xpath` two times and then `zip`? And then can we git rid of the `MP_XPATH` constant?)

datasets/gr/parliament/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+from typing import Tuple, Optional
+
+from zavod import Context, helpers as h
+from zavod.stateful.positions import categorise
+
+
+# XPath to select all MPs from the dropdown
+MP_XPATH = ".//select[@id='ctl00_ContentPlaceHolder1_dmps_mpsListId']/option"
+# List of IDs to skip:
+# 1. These MPs are not in the dataset
```
leonhandreke: What does this mean, they are not in the dataset?

datasets/gr/parliament/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+    # Skip the first option ('Select Member')
+    mp_ids_names: list[tuple[str, str]] = list(zip(ids[1:], names[1:]))
+    for id, name in mp_ids_names:
+        url = urljoin(context.data_url, f"?MpId={id}")
+        mp_page = context.fetch_html(url, cache_days=1)
+        table = mp_page.xpath(".//table[@class='grid']")
+        # Warn if table is missing for MPs not in skip list
+        if not table and id not in SKIP_LIST:
+            context.log.warning(
+                "No table found for MP not in skip list", id=id, url=url
```
leonhandreke: Could we add very clear documentation here on what to do if this triggers? Something like:

Check the URL and verify that it's only this MP that doesn't have a table, then add them to the KNOWN_BROKEN list.

or something like that.

datasets/gr/parliament/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+            )
+        # Skip processing if table is missing
+        if not table:
+            continue
+        assert len(table) == 1, len(table)
+        table = table[0]
```
leonhandreke: (nit) this likely won't pass the typechecker because the declaration above defines the type of `table` and it's changed here.

datasets/gr/parliament/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+        for row in h.parse_html_table(table, ignore_colspan={"5"}):
```
leonhandreke, line 122: (nit) IIUC, the `ignore_colspan` skips the bottom empty row. Could we add a quick comment about this?

datasets/gr/parliament/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+    "1294b6b9-9a50-4961-b429-b034010b019d",
+    "f8856bfc-54b4-4052-bfe0-b034010b56b3",
+    "c9c199ec-4bcd-4bd3-a545-b034010b0907",
+    "75f888b4-9245-479c-9aaa-b034010b4284",
+    "79bf7490-332e-40c7-a226-b034010b36eb",
+    "591f30ee-64a4-4f03-bfd0-b034010b18f2",
+]
+
+
+def lookup_dates(context: Context, term: str) -> Tuple[Optional[str], Optional[str]]:
```
leonhandreke: (nit) Can we rename this `lookup_term_dates`?

datasets/gr/parliament/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+
+    if categorisation.is_pep:
+        occupancy = h.make_occupancy(
+            context,
+            person,
+            position,
+            start_date=start_date,
+            end_date=end_date,
+            categorisation=categorisation,
+            # Some end dates for previous terms are not available
```
leonhandreke: I wonder if this is something we should look up and add anyway, even though it's missing in the source?

datasets/gr/parliament/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+            no_end_implies_current=False,
+        )
+        if occupancy:
+            context.emit(position)
+            context.emit(occupancy)
+            context.emit(person)
+
+    context.audit_data(str_row, ["date", "costituency"])
```
leonhandreke: Are there any cases where `start_date != row["date"]`? I'm wondering whether maybe some got into office because another person left, so maybe `row["date"]` is better as a start date actually?

nvmbrasserie: I considered it, but the closest `row["date"]` we're interested in is the day of elections, which is not the same. So I think the most accurate way is to use the first day of the term as a `startDate`

leonhandreke (approved):
Nice, let's ship it!

(Let's work through the 403's first)

PR #2984, opened 2025-10-23
Netherlands Senate

nvmbrasserie:
opensanctions/crawler-planning#399

### DIFF
--- a/datasets/nl/senate/crawler.py
+++ b/datasets/nl/senate/crawler.py
@@ -8,41 +8,37 @@ from zavod.stateful.positions import categorise
 from zavod import Context
 from zavod import helpers as h
 
-# Single regex for both fields (covers 95% of cases)
-RESIDENCY_DOB_REGEX = re.compile(r"Woonplaats:\s*(.*?)\s*Geboortedatum:\s*([\d-]+)")
+RESIDENCY_REGEX = re.compile(r"Woonplaats:\s*(.*)")
+DOB_REGEX = re.compile(r"Geboortedatum:\s*([\d-]+)")
 
 
 def get_residency_dob(
     context: Context,
-    member: _Element,
+    member_el: _Element,
 ) -> Tuple[Optional[str], Optional[str]]:
     """
     Extracts residency and date of birth from a member element.
     Tries regex on the combined text of child nodes. Falls back to context.lookup if missing.
     Returns a tuple: (residency, dob)
     """
-    node = member.find(".//div[@class='cell pasfoto_tekst_k2']")
+    node = member_el.find(".//div[@class='cell pasfoto_tekst_k2']")
     if node is None:
         context.log.warning("No node found for residency and date of birth extraction")
         return None, None
-    # Combine all child text
-    full_text = " ".join(child.text.strip() for child in node if child.text)
-    # Attempt regex extraction first
-    match = RESIDENCY_DOB_REGEX.search(full_text)
-    if match:
-        residency = match.group(1).strip()
-        dob = match.group(2).strip()
-        return residency, dob
-    # Fallback to context.lookup
-    details_res = context.lookup("residency_dob", full_text)
-    if details_res:
-        details = details_res.details[0]
-        return details.get("residency"), details.get("dob")
-    context.log.warning(
-        "Could not extract residency and date of birth",
-        full_text=full_text,
-    )
-    return None, None
+    residency, dob = None, None
+    for child in node:
+        if not child.text:
+            continue
+        if m := DOB_REGEX.search(child.text):
+            dob = m.group(1)
+        elif m := RESIDENCY_REGEX.search(child.text):
+            residency = m.group(1)
+        else:
+            context.log.warning(
+                "Could not extract residency or date of birth", text=child.text
+            )
+
+    return residency, dob
 
 
 def get_name_date_party(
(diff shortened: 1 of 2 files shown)


datasets/nl/senate/crawler.py
```diff
@@ -0,0 +1,130 @@
...
+DOB_REGEX = re.compile(r"Woonplaats:\s*(.*?)\s*Geboortedatum:\s*([\d-]+)")
+TENURE_REGEX = re.compile(r"(.*?)\s*Anciënniteit:\s*(.*)")
+
+
+def extract_details(
+    context: Context,
+    member: _Element,
+    xpath: str,
+    regex: re.Pattern[str],
+    lookup_keys: Optional[Dict[int, str]] = None,
```
leonhandreke: I'm not super sure what `lookup_keys` is doing, but it's a bit of a weird mechanism.

I feel maybe returning a `dict[DetailType, str]` with keys that are of a `DetailType(StrEnum)` here would be good?

datasets/nl/senate/crawler.py
```diff
@@ -0,0 +1,130 @@
...
+    node = member.find(xpath)
+    if node is None:
+        context.log.warning(f"No node found for xpath: {xpath}")
+        return None, None
+    full_text = " ".join(child.text.strip() for child in node if child.text)
+    match = regex.search(full_text)
+    if match:
+        return match.group(1), match.group(2)
+    if lookup_keys:
+        details_res = context.lookup("details", full_text)
```
leonhandreke: (nit) `details` is actually two different lookups. Maybe `description_details` and `field_details` or something like that?

datasets/nl/senate/crawler.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    Extracts two fields from a member element. Tries regex first.
+    If regex fails, falls back to context.lookup.
+    """
+    node = member.find(xpath)
+    if node is None:
+        context.log.warning(f"No node found for xpath: {xpath}")
+        return None, None
+    full_text = " ".join(child.text.strip() for child in node if child.text)
+    match = regex.search(full_text)
```
leonhandreke: So first we join them and then we run a regex on them. Couldn't we run a regex on each of the nodes to extract one of the properties? Or is one prop spread across multiple nodes?

datasets/nl/senate/crawler.py
```diff
@@ -0,0 +1,130 @@
...
+    person.add("address", residency)
+    person.add("sourceUrl", url)
+    person.add("topics", "role.pep")
+
+    position = h.make_position(
+        context,
+        name="Member of the Senate of the Netherlands",
+        country="nl",
+        topics=["gov.legislative", "gov.national"],
+        lang="eng",
```
leonhandreke, line 93: is that true here? I don't see any full-text fields here, but the website is in Dutch, no?

datasets/nl/senate/crawler.py
```diff
@@ -0,0 +1,130 @@
...
+        details = details_res.details[0]
+        return details.get("name"), details.get("start_date")
+    context.log.warning(
+        "Could not extract name and start date from details",
+        details=first_sentence,
+    )
+    return None, None
+
+
+def crawl_member(context: Context, member: _Element) -> None:
```
leonhandreke: I think it would be nicer to pass in a URL here, it's a pattern we have in plenty of places.

datasets/nl/senate/crawler.py
```diff
@@ -0,0 +1,130 @@
...
+    party, _ = extract_details(
+        context,
+        member,
+        xpath=".//div[@class='persoon_bijschrift']",
+        regex=TENURE_REGEX,
+    )
+
+    person = context.make("Person")
+    person.id = context.make_id("person", name_clean, dob, party)
+    person.add("name", name_clean, lang="eng")
```
leonhandreke: The website is in dutch, right? So either empty or Dutch I'd say.

datasets/nl/senate/crawler.py
nvmbrasserie: we can but since this regex covers 95% of all cases that we're trying to match, I feel it makes the code more efficient and easier to read

datasets/nl/senate/crawler.py
nvmbrasserie: I decided to discard of the use cases as redundant, so now the function is specific to dob and residency

datasets/nl/senate/nl_senate.yml
```diff
@@ -0,0 +1,421 @@
...
+      - match: 'Rik Janssen (1957) is vanaf 11 juni 2019 lid van de SP-fractie in de Eerste Kamer'
+        details:
+          - name: "Rik Janssen"
+            start_date: "2019-06-11"
+            party: "SP"
+      - match: 'Hetty Janssen-van Helvoort (1954) is vanaf 13 juni 2023 lid van de GroenLinks-PvdA-fractie in de Eerste Kamer'
+        details:
+          - name: "Hetty Janssen-van Helvoort"
+            start_date: "2023-06-13"
+            party: "GroenLinks-PvdA"
```
leonhandreke, line 182: (nit) So they're not a party per se, but a parlamentary group

nvmbrasserie: it goes into the `political` prop so we're good, just discussed it in person

PR #2997, opened 2025-10-27
Fixes-27-10

nvmbrasserie:


datasets/ca/senate/crawler.py
```diff
@@ -43,6 +42,6 @@ def crawl(context: Context) -> None:
             no_end_implies_current=True,
             propagate_country=False,
         )
-        if occupancy is not None:
+        if occupancy:
```
leonhandreke, line 45: (nit) I think we generally prefer `is not None` over truthyness.

datasets/fr/illegal_financial_services/crawler.py
```diff
@@ -1,37 +1,25 @@
...
-        entity.add("website", row.get("nom"))
-    elif re.match(EMAIL_PATTERN, row.get("nom")):
-        entity.add("email", row.get("nom"))
-
-    row.pop("nom")
+    email_clean = registry.email.clean(name)
+    if email_clean:
+        entity.add("email", email_clean)
+    url_clean = registry.url.clean(name)
+    if url_clean:
```
leonhandreke: (nit) again, to avoid the python truthyness, can we have an explicit `is not None` or `!= ""` here?

Also, we could use the new `:=` operator here: `if url_clean := registry.url.clean(url) is not None`

datasets/fr/illegal_financial_services/crawler.py
```diff
@@ -1,37 +1,25 @@
...
+    entity.add("name", name)
     entity.add("topics", "crime.fin")
 
-    if re.match(URL_PATTERN, row.get("nom")):
-        entity.add("website", row.get("nom"))
-    elif re.match(EMAIL_PATTERN, row.get("nom")):
-        entity.add("email", row.get("nom"))
-
-    row.pop("nom")
+    email_clean = registry.email.clean(name)
```
leonhandreke, line 18: Can we add a quick comment what `name` looks like? It's a bit confusing that we are extracting everything from the name.

Just to make sure, did you verify that the diff looks sort of okay, that it does more or less the same thing?

datasets/ca/senate/crawler.py
nvmbrasserie: for occupancy it's an exception where we usually check the truthiness

datasets/fr/illegal_financial_services/crawler.py
nvmbrasserie: I'm not generally against it (`:=`), but I feel in this case the code will lose in readability

datasets/fr/illegal_financial_services/crawler.py
nvmbrasserie: yes, in this case it's even better than regex that we used to have there

PR #2998, opened 2025-10-27
Senate of Poland

nvmbrasserie:
opensanctions/crawlers#406

### DIFF
--- a/datasets/pl/senate/crawler.py
+++ b/datasets/pl/senate/crawler.py
@@ -1,13 +1,15 @@
 import re
-from datapatch import Lookup
-from lxml.html import HtmlElement
-from normality import squash_spaces
 from typing import cast
 from urllib.parse import urljoin
 
-from zavod import Context, helpers as h
+from datapatch import Lookup
+from lxml.html import HtmlElement
+from normality import squash_spaces
 from zavod.shed import zyte_api
 from zavod.stateful.positions import OccupancyStatus, categorise
+
+from zavod import Context
+from zavod import helpers as h
 
 # These entries are explicitly skipped because they currently contain no details.
 # However, the skip list is maintained so we can verify each skipped case,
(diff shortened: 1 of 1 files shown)


datasets/pl/senate/crawler.py
```diff
@@ -0,0 +1,56 @@
...
+            continue
+        name = link.text.strip()
+        href = link.get("href")
+        assert href is not None, "Missing href"
+        assert name, "Missing name for: {}".format(href)
+        url = urljoin(context.data_url, href)
+        entity = context.make("Person")
+        entity.id = context.make_id(name, href)
+        entity.add("citizenship", "pl")
+        entity.add("topics", "role.pep")
```
leonhandreke: This is already being done in `make_occupancy`. So let's either do this in the `if occupancy is not None` to emphasize that they're actually a PEP, or rely on `make_occupancy` to set it. Setting it before in this case is correct. But it's nice to establish the pattern of 1. build the entity 2. PEP decision 3. topic + emit

datasets/pl/senate/pl_senate.yml
```diff
@@ -0,0 +1,41 @@
+title: Poland Members of the Senate
+prefix: pl-senate
+url: https://www.senat.gov.pl/en/senators/lista-senatorow/
+coverage:
+  frequency: monthly
+  start: 2025-10-27
+load_statements: true
+entry_point: crawler.py
+ci_test: false
```
leonhandreke, line 9: Why no CI?

datasets/pl/senate/crawler.py
```diff
@@ -0,0 +1,56 @@
...
+    for row in doc.findall(senator_container):
+        link = row.find(".//a")
+        if link is None:
+            continue
+        name = link.text.strip()
+        href = link.get("href")
+        assert href is not None, "Missing href"
+        assert name, "Missing name for: {}".format(href)
+        url = urljoin(context.data_url, href)
+        entity = context.make("Person")
```
leonhandreke, line 72: Looks like the first sentence in the description always contains the DOB and POB. Is this consistent enough to extract it? And possibly build a nicer ID from that doesn't contain the href? Because if they ever change the website like this, our IDs are toast.

datasets/pl/senate/crawler.py
```diff
@@ -0,0 +1,56 @@
...
+        )
+        categorisation = categorise(context, position, is_pep=True)
+        if not categorisation.is_pep:
+            continue
+
+        occupancy = h.make_occupancy(
+            context,
+            person=entity,
+            position=position,
+            no_end_implies_current=False,
```
leonhandreke, line 95: Aren't they current as of the crawl time? Isn't this the implication of the source? And if it is, what exactly are the implications for the status field (I actually don't know myself - what should it be? Let's talk offline :) )

datasets/pl/senate/pl_senate.yml
nvmbrasserie: because of zyte

datasets/pl/senate/crawler.py
nvmbrasserie: You're right. 

> We therefore only represent people as PEPs if a data source indicates they hold the position now, or they left the position within the past 5 years. In these cases the occupancy status should be current or ended respectively.

Source: https://zavod.opensanctions.org/peps/#creating-occupancies

leonhandreke: So IIUC, it doesn't matter what we set here, since we already set `status` directly, right?

PR #2999, opened 2025-10-27
Sejm of the Republic of Poland

nvmbrasserie:
opensanctions/crawlers#405

### DIFF
--- a/datasets/pl/sejm/crawler.py
+++ b/datasets/pl/sejm/crawler.py
@@ -68,8 +68,7 @@ def crawl(context: Context) -> None:
     doc = context.fetch_html(context.data_url, cache_days=1)
     deputies = doc.findall(".//ul[@class='deputies']/li")
     for deputy in deputies:
-        link = deputy.find(".//a")
-        if link is None:
+        if (link := deputy.find(".//a")) is None:
             continue
         href = link.get("href")
         assert href is not None, "Missing href"


datasets/pl/sejm/crawler.py
```diff
@@ -0,0 +1,77 @@
...
+    name = pep_doc.findtext(".//div[@id='title_content']/h1")
+    if not name:
+        context.log.warning(f"Missing name for: {url}")
+        return
+    start_date = cast(str, pep_doc.xpath(START_DATE_XPATH)[0])
+    if not start_date:
+        context.log.warning(f"Missing start date for: {url}")
+        return
+    dob_pob = pep_doc.findtext(".//p[@id='urodzony']")
+    assert dob_pob is not None, f"Missing date and place of birth for: {url}"
```
leonhandreke, line 30: Would it make sense to check this after the splitting so that we also notice if one of them is empty? Or are we okay with that?

datasets/pl/sejm/crawler.py
```diff
@@ -0,0 +1,77 @@
...
+        context.emit(occupancy)
+        context.emit(position)
+        context.emit(entity)
+
+
+def crawl(context: Context) -> None:
+    doc = context.fetch_html(context.data_url, cache_days=1)
+    deputies = doc.findall(".//ul[@class='deputies']/li")
+    for deputy in deputies:
+        link = deputy.find(".//a")
```
leonhandreke: Feel free to ignore, but we could also do `if (link := deputy.find(".//a")) is None:` to make this a single line :)

datasets/pl/sejm/crawler.py
nvmbrasserie: we're okay, because there is nothing we can do with that. also, both of them are present in 99% of cases

PR #3005, opened 2025-10-30
more categories for wd_categories

nvmbrasserie:


pudo (approved):
Very nice complement to the crawlers, I appreciate that approach!

PR #3010, opened 2025-10-30
[fsf] Heuristic to treat translated names as aliases

leonhandreke:
They are just translations of the primary name, which doesn't have a language tag set. We don't want these translations to clog up the primary name field, so we demote them to alias instead.

This should only be merged once https://www.opensanctions.org/changelog/36/ is effective.


pudo:
Hmm, I appreciate the workaround, but it doesn't feel totally explainable to me after everything is said and done. Would it be more conservative to say: if there's a name with `nameLanguage=""`, use that as a `name`, everything else goes to `alias`.  [Maybe also for `eng`]. In all other cases, keep everything in `name`?

pudo:
That seems fair. The alternative would be to define uninteresting languages, which seems worse. I guess Spanish is a weird edge case of being used by designation targets and being a very useless translation of "ISIS"... But let's go with this for now!

PR #3035, opened 2025-11-06
[au_dfat_sanctions] switch to the new format

nvmbrasserie:


datasets/au/dfat_sanctions/crawler.py
```diff
@@ -143,28 +139,39 @@ def parse_reference(
...
+        if entity.schema.is_a("Vessel"):
+            entity.add("imoNumber", row.pop("imo_number"))
         dates = clean_date(row.pop("date_of_birth"))
-        if dates != ["na"]:
-            h.apply_dates(entity, "birthDate", dates)
+        h.apply_dates(entity, "birthDate", dates)
         entity.add("birthPlace", row.pop("place_of_birth"), quiet=True)
         entity.add("notes", h.clean_note(row.pop("additional_information")))
         listing_info = row.pop("listing_information")
+        designation_instrument = row.pop("instrument_of_designation")
```
jbothma: let's pop this where we use it.

datasets/au/dfat_sanctions/crawler.py
```diff
@@ -107,9 +100,12 @@ def parse_reference(
     primary_name: Optional[str] = None
     names: List[Tuple[str, str]] = []
     for row in rows:
+        alias_strength = row.pop("alias_strength")
```
jbothma: let's assign this where we use it, or even inline it.

datasets/au/dfat_sanctions/crawler.py
```diff
@@ -143,28 +139,39 @@ def parse_reference(
...
             ),
         )
-        country = clean_country(row.pop("citizenship"))
+        country = h.multi_split(row.pop("citizenship"), [","])
         if entity.schema.is_a("Person"):
-            entity.add("nationality", country)
+            entity.add("citizenship", country)
         else:
             entity.add("country", country)
+        if entity.schema.is_a("Vessel"):
```
jbothma: How about this, since Organization can have IMOs, apparently
```suggestion
        if entity.schema.properties.get("imoNumber"):
```

PR #3036, opened 2025-11-06
Check if names need splitting, and optionally use an LLM and Review to do so.

jbothma:
Implements #2656

initially ru_cbr will not have the comma-splits, which are sometimes correct and sometimes wrong:
`Development Bank of production of oil and gas extractive equipment, conversion, shipbuilding and construction (joint-stock company), JSC NOKSSBANK`

```
[ru_cbr_banks] added=0 dataset=ru_cbr_banks deleted=0 metric=delta_counts modified=199
```

- [x] implement a couple more crawlers to see if the interface is sufficiently flexible
  -  iso_bic results in lots and lots of reviews for `Company A/S` - maybe we want to be able to override the default regex in a specific crawler?
  - bg_omnio wants everything to go into alias, except perhaps things that should be weak_alias - maybe a `field_mapping` override arg could be passed to `apply_split_names`?
  - ru_cbr_banks wants to split on some but not all commas for `name_en` - maybe additional regexes could be passed to `should_split`?
- [ ] ~add a separate DSPy "module/program" and helpers for cases where we anticipate distinct entities. I think is some of those cases, there can be distinct entities with aliases - it'd be nice if it could do roughly the right thing for them.~
  - later. maybe.
  - I think the single-entity splitter should be distinct from the multi entity splitter since a lot of aliases look like different entities when we know for a given dataset that a string always pertains to exactly one entity and it's merely very different aliases
    - single: ru_cbr_banks, iso_bic, ca_named_terrorists
    - multi: us_finra, fr_amf_regulatory, medicaid datasets that tend to bundle company and owner in one string
- [x] prompt using run_typed_text_prompt rather than dspy to use our cache and reduce the hackiness of its fight with leveldb.
- [x] CLI to re-optimise with more/better data
- [x] tests
  - [x] CLI smoke test
  - [x] original string is used unless split is needed, and split is accepted
- [x] disable in CI so that the datasets can actually run there
- [ ] ~contextual info to review for the reviewer (e.g. entity id, other names available for the entity)~
  - let's do this separately to manage scope


datasets/_global/iso9362_bic/crawler.py
```diff
@@ -30,9 +30,13 @@ def crawl(context: Context) -> None:
...
         if legal_name == "130" and bic == "FCCCFRP1":
             legal_name = "130 - FCCCFRP1"
+
         entity = context.make("Organization")
         entity.id = f"bic-{bic}"
-        entity.add("name", legal_name)
+        if h.needs_splitting(entity.schema, legal_name):
+            h.split_and_apply_names(context, entity, legal_name)
+        else:
+            entity.add("name", legal_name)
```
jbothma: h.apply_reviewed_names
h.apply_validated_names

jbothma:
This is not very nice.
I'm not sure how to prevent it. It exits successully, but it's not pretty and might be confusing

image

zavod/zavod/meta/names.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Use this to override the default set of characters.
+    There's no need to redefine all the characters handled by an ancestor schema;
+    """
+    dirty_chars_extra: str = ""
+    """
+    Additional characters specific to this schema that suggest a name needs cleaning.
+
+    Use this if you don't want to override the standard characters.
+    """
```
pudo: I think this is a weird split since it sort of presumes the user knows the list of all dirty chars. Would it make sense instead to keep `_dirty_chars` as an internal baseline (which we are free to expand as we mature) and invite users to make contributions using two other variables like:

`block_chars` (same as now `dirty_chars_extra`) 
`allow_chars` (to be removed from the baseline dirty list for this dataset)

zavod/zavod/meta/names.py
```diff
@@ -0,0 +1,56 @@
...
+
```
pudo, line 34: I would really love a few more options here:

* `min_chars` for a minimum number of alphanumeric characters 
* `require_space` for the presence of any `\s` in the string
* `nullwords: false` to check for `rigour.names.check:is_nullword`

zavod/zavod/helpers/names.py
```diff
@@ -230,3 +248,94 @@ def split_comma_names(context: Context, text: str) -> List[str]:
                 return [text]
         else:
             return [text]
+
+
+def name_needs_cleaning(entity: Entity, string: Optional[str]) -> bool:
```
pudo: I anticipate that this will be called all over our codebase, and I'd advocate for a more predicaty name. Ideas: 

`is_name_weak`
`is_name_suspicious`
`is_name_shite`

zavod/zavod/helpers/names.py
```diff
@@ -230,3 +248,94 @@ def split_comma_names(context: Context, text: str) -> List[str]:
...
+        return
+
+    if settings.CI or not name_needs_cleaning(entity, string):
+        prop = "alias" if alias else "name"
+        entity.add(prop, string, lang=lang)
+        return
+
+    names = clean_names(context, string)
+
+    source_value = TextSourceValue(key_parts=string, label="name", text=string)
```
pudo, line 341: We've thought about putting `lang` and the existing names in here, right? but it's not helpful or leads to too much noise?

zavod/zavod/extract/names/clean.py
```diff
@@ -0,0 +1,46 @@
...
+SINGLE_ENTITY_PROGRAM_PATH = Path(__file__).parent / "dspy/single_entity_program.json"
+
+
+class CleanNames(BaseModel):
+    """Names categorised and cleaned of non-name characters."""
+
+    full_name: List[str]
+    alias: List[str]
+    weak_alias: List[str]
+    previous_name: List[str]
```
pudo, line 19: `lang: Optional[str] = None` as an override? I know we talked about it and you came out against, it would just be so damn helpful to be able to set it in the review UI because then we can also work towards skipping out some languages for certain types of matching processes.

pudo: FWIW lang doesn't even need to go through the LLM, it could be overwritten on the result object from the source value if we wanted it to.

zavod/zavod/meta/names.py
jbothma: I prefer that too.

zavod/zavod/helpers/names.py
jbothma: would `not is_name_clean()` be too annoying?

zavod/zavod/helpers/names.py
jbothma: I've seen examples from prompts come up as suggestions, so my inclination is not to do so. But it could be worth a try with clear delineation - DSPY uses stuff like `[[## here comes a thing ##]] a thing [[## that was a thing ##]]` (maybe I'm exaggerating, but we can try that to distinguish _the string to extract from_ from other context that might help the LLM)

zavod/zavod/extract/names/clean.py
jbothma: Let's add to the schema. Would we document the field as "only if all names are in the same langauge"? i.e. it's null if there's a latin and cyrilic name?

zavod/zavod/helpers/names.py
pudo: * irregular
* unusual

jbothma: I think is_name_irregular is the best balance

zavod/zavod/meta/names.py
jbothma: I couldn't quickly figure out how to make pydantic happy to instantiate specific schema defaults with a private field so I've just called it reject_chars_baseline for now

I prefer reject_ over block_ just because I think it's a little more obvious. Happy to make it all block_ if that's preferred.

zavod/zavod/extract/names/clean.py
jbothma: I'm gonna kick this to a next PR. Just adding it to the CleanNames model might well result in the LLM filling it in for us and I want to see how annoying that is or how well we can avoid it. An alternative is to have a subclass with lang for the review but... it's no problem extending the review model with an optional field later.

zavod/zavod/meta/names.py
jbothma: I'm taking a bit long to get this out the door so let's ship and then add

zavod/zavod/extract/names/dspy/clean.py
```diff
@@ -0,0 +1,42 @@
...
+
+class CleanNamesSignature(dspy.Signature):  # type: ignore
+    """Names categorised and cleaned of non-name characters."""
+
+    string: str = dspy.InputField()
+    full_name: List[str] = dspy.OutputField(
+        desc="A list of the names of this entity, potentially in various languages and transliterations."
+    )
+    alias: list[str] = dspy.OutputField(
+        desc="A list of alternative names or nicknames for this entity."
```
pudo: I would move "nicknames" to weak, maybe "A list of alternative but still fully descriptive names for this entity." sort of thing?

zavod/zavod/extract/names/clean.py
pudo: +1

zavod/zavod/helpers/names.py
```diff
@@ -230,3 +248,113 @@ def split_comma_names(context: Context, text: str) -> List[str]:
...
             return [text]
+
+
+def is_name_irregular(entity: Entity, string: Optional[str]) -> bool:
+    """Determine whether a name string potentially needs cleaning."""
+    if not string:
+        return False
+
+    for spec in entity.dataset.names.specs_for_schema(entity.schema):
+        for char in spec.reject_chars_consolidated:
```
pudo: Pretty :) I think eventually we'll want to try and compile as many of the heuristics as we can into one regex, but that's mk 2

zavod/zavod/meta/names.py
```diff
@@ -0,0 +1,69 @@
...
+                    default_spec = instance.root[schema]
+                    merged_spec = default_spec.model_copy(update=spec)
+                    instance.root[schema] = merged_spec
+                else:
+                    instance.root[schema] = CleaningSpec.model_validate(spec)
+            return instance
+        raise TypeError(f"object must be a dict, got {type(obj)}")
+
+    def specs_for_schema(self, schema: Schema) -> List[CleaningSpec]:
+        """Returns the specs for all schemas that match the entity."""
```
pudo: Did you eventually decide implement this over the "most narrow" logic you were thinking about before? I would vote that if we have `Person` and `LegalEntity`, then `LegalEntity` gets ignored entirely. 

This could be done eg. by getting all the schema objects that match, and then picking the one with the shortest `schema.schemata` (inheritance) set.

pudo (commented):
It's in a really really good place. I wouldn't have any issues with landing it on a Friday, how do you feel?

zavod/zavod/meta/names.py
jbothma: Yeah I figured this is simplest and we can do most specific if we feel it's needed.

Happy to go with most specific.

PR #3038, opened 2025-11-07
UK NCA Press Releases

nvmbrasserie:
opensanctions/crawlers#412

### DIFF
--- a/datasets/gb/nca_press_releases/crawler.py
+++ b/datasets/gb/nca_press_releases/crawler.py
@@ -44,7 +44,7 @@ Specific fields:
 - entity_schema: {schema_field.description}
 - name: The name of the entity precisely as expressed in the text.
 - aliases: ONLY extract aliases that follow an explicit indication of an _alternative_ name, such as "also known as", "alias", "formerly", "aka", "fka". Otherwise the aliases field should just be an empty array.
-- address: The full address or location details as they appear in the article (do not split into components; capture the complete expression such as “123 Main Street, Birmingham”).
+- address: The full address or location details as they appear in the article (do not split into components; capture the complete expression such as “123 Main Street, Birmingham”). If no address is given for a specific entity, leave this field as an empty array.
 - country: Any countries the entity is indicated to reside, operate, or have been born or registered in. Leave empty if not explicitly stated.
 """
 


datasets/gb/nca_enforcements/crawler.py
```diff
@@ -0,0 +1,137 @@
...
+        context.emit(article)
+        context.emit(documentation)
+
+
+def crawl(context: Context):
+    page: int = 0
+    base_url: str = "https://www.nationalcrimeagency.gov.uk/news?start={offset}"
+
+    while True:
+        index_url = base_url.format(offset=page * 16)
```
leonhandreke: What if the default number of items on the page changes? Does it make sense to count `links` instead?

datasets/gb/nca_enforcements/gb_nca_enforcements.yml
```diff
@@ -0,0 +1,98 @@
+title: UK NCA Press Releases
```
jbothma: We don't really know what the NCA is generally, so it's worth expanding this in the title

```suggestion
title: UK National Crime Agency Press Releases
```

datasets/gb/nca_enforcements/gb_nca_enforcements.yml
```diff
@@ -0,0 +1,98 @@
...
+      - match:
+          - Cybercrime
+          - Cyber choices
+        value: crime.cyber
+      - match:
+          - Fraud
+          - False identity documents
+        value: crime.fraud
+      - match:
+          - Modern slavery
```
jbothma: I think we do forced.labor for modern slavery, see the br_slavery dataset.

datasets/gb/nca_enforcements/crawler.py
```diff
@@ -0,0 +1,137 @@
...
+            "//div[@class='blog news-page']/div[@class='row-fluid']//div[@class='page-header']//a/@href"
+        )
+        if links == [] or len(links) == 0:
+            context.log.info("No more links found, stopping crawl")
+            break
+        for link in links:
+            crawl_enforcement_action(context, link)
+        if page >= 200:  # Limit to 200 pages to avoid infinite loops
+            break
+        page += 1
```
jbothma: it's usually better to do the pagination more in line with the interface they provide. Whether it's an API or a web page, it often gives you hints about whether there are more pages to keep looping.

In this case, we can do something like 

```python
index_url = context.data_url
while index_url:
    doc = fetch(index_url)
    next_links = doc.xpath("a[aria-abel=\"Go to next > page\"]")
    assert len(next_links) <= 1
    if next_links:
        index_url = next_link[0]
    else
        index_url = None
    ...
```

That way

- if they change the page size, our code doesn't miss anything
- we can do away with the max pages assertion

datasets/gb/nca_enforcements/crawler.py
```diff
@@ -0,0 +1,137 @@
...
+"""
+
+
+def crawl_enforcement_action(context: Context, url: str) -> None:
+    article_doc = context.fetch_html(url, cache_days=7, absolute_links=True)
+    if article_doc is None:
+        return
+    article_name = article_doc.xpath(NAME_XPATH)[0]
+    assert article_name, "Article name not found"
+    article_content = article_doc.xpath(CONTENT_XPATH)
```
jbothma: I think it's better for the xpath to be defined in context. It makes it much easier to know what we're talking about when I'm reading the code. If it's all defined at the top, it looks nice and neat, but I forget where I was if I have to scroll up and down to see it. We're not using it in multiple places in the crawler, so I don't think we gain from defining them up top.

datasets/gb/nca_enforcements/crawler.py
```diff
@@ -0,0 +1,137 @@
...
+    assert len(article_content) == 1
+    article_element = article_content[0]
+    date = article_doc.xpath(DATE_XPATH)
+    # The date is absent in some articles
+    if date != []:
+        date = date[0]
```
jbothma: just a nitpick. Since the way we query for things with xpath() gives us lists, I prefer calling the variables their plural. Then we also don't have to use the same variable with two different types (list and string)
```suggestion
    dates = article_doc.xpath(DATE_XPATH)
    # The date is absent in some articles
    if dates != []:
        date = dates[0]
```

datasets/gb/nca_enforcements/crawler.py
```diff
@@ -0,0 +1,137 @@
...
+        entity.id = context.make_id(item.name, str(item.address), str(item.country))
+        entity.add("name", item.name, origin=review.origin)
+        if item.address != item.country:
+            entity.add("address", item.address, origin=review.origin)
+        entity.add("country", item.country, origin=review.origin)
+        for topic in topics:
+            entity.add("topics", topic)
+
+        # We use the date as a key to make sure notices about separate actions are separate sanction entities
+        sanction = h.make_sanction(context, entity, date)
```
jbothma: it'd be really nice if date is defined in the context where it's used, since that context is just these few lines

datasets/gb/nca_enforcements/crawler.py
```diff
@@ -0,0 +1,137 @@
...
+        date = date[0]
+    # Extract topics and look them up
+    raw_topics: List[str] = article_doc.xpath(TOPIC_XPATH)
+    topics: List[str] = []
+    for topic in raw_topics:
+        res = context.lookup("topics", topic)
+        if res is not None:
+            topics.append(res.value)
+        else:
+            context.log.warning("Topic lookup not found", topic=topic)
```
jbothma: you can now do
```suggestion
    for topic in raw_topics:
        res = context.lookup("topics", topic, warn_unmatched=True)
        if res is not None:
            topics.append(res.value)
```

datasets/gb/nca_enforcements/crawler.py
```diff
@@ -0,0 +1,137 @@
...
+        context.emit(documentation)
+
+
+def crawl(context: Context):
+    page: int = 0
+    base_url: str = "https://www.nationalcrimeagency.gov.uk/news?start={offset}"
+
+    while True:
+        index_url = base_url.format(offset=page * 16)
+        doc = context.fetch_html(index_url, absolute_links=True, cache_days=5)
```
jbothma: Best not to cache index pages https://zavod.opensanctions.org/best_practices/caching/#issue-with-pagination-and-cached-pages
```suggestion
        doc = context.fetch_html(index_url, absolute_links=True)
```

jbothma:
it'd be nice if we can get the LLM to give us the full address stated, rather than splitting parts. We don't want to blindly join them on the python side. Are there cases where we get multiple addresses/localities? Should it be a string instead of list?
image

datasets/gb/nca_enforcements/crawler.py
```diff
@@ -0,0 +1,132 @@
...
+def crawl(context: Context):
+    index_url = context.data_url
+    while index_url:
+        doc = context.fetch_html(index_url, absolute_links=True)
+        links = doc.xpath(
+            "//div[@class='blog news-page']/div[@class='row-fluid']//div[@class='page-header']//a/@href"
+        )
+        if links == [] or len(links) == 0:
+            context.log.info("No more links found, stopping crawl")
+            break
```
jbothma: in theory we don't need this any more. Do we still need this?

datasets/gb/nca_enforcements/crawler.py
```diff
@@ -0,0 +1,132 @@
...
+        entity.add("country", item.country, origin=review.origin)
+        for topic in topics:
+            entity.add("topics", topic)
+
+        dates = article_doc.xpath(
+            "//div[@itemprop='articleBody']/p[strong][last()]/strong/text()"
+        )
+        # The date is absent in some articles
+        if dates != []:
+            date = dates[0]
```
jbothma: we should probably check `enforcements.within_max_age` too

jbothma (commented):
Thanks for the changes!

Just a couple more things.

PR #3041, opened 2025-11-07
[au_dfat_sanctions] names/aliases fallback

nvmbrasserie:
Fallback: if no "name" property was found (e.g. only aliases exist), promote the first available entry to be the main name.


nvmbrasserie:
Closing this as the issue was resolved through direct communication with the publisher.

PR #3054, opened 2025-11-11
helpers: add xpath_{elements,strings} and use in datasets/de

leonhandreke:
- **helpers: add xpath_{elements,strings} and use in datasets/de**
- **pre-commit: datasets/de now passes mypy --strict**

Diffs are quite large since it ruff'd some files in `datasets/de`, sorry

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -86,7 +86,6 @@ repos:
           datasets/il|
           datasets/im|
           datasets/in|
-          datasets/iq|
           datasets/ir|
           datasets/is|
           datasets/jp|
--- a/datasets/iq/aml_list/crawler.py
+++ b/datasets/iq/aml_list/crawler.py
@@ -4,6 +4,13 @@ from typing import Optional, Set
 from normality import latinize_text
 from openpyxl import load_workbook
 from rigour.mime.types import XLSX
+<<<<<<< HEAD
+=======
+from typing import Dict, List, Optional
+
+from zavod import Context, helpers as h
+from zavod.shed import zyte_api
+>>>>>>> 3278f6768 ([iq] Make the typechecker happy)
 from zavod.shed.trans import (
     apply_translit_full_name,
 )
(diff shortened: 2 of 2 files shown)


zavod/zavod/helpers/html.py
```diff
@@ -129,3 +129,39 @@ def links_to_dict(el: Element) -> Dict[str | None, str | None]:
...
+) -> List[Element]:
+    """
+    Return a list of HtmlElement objects that match the given XPath expression.
+    """
+    result = el.xpath(xpath)
+    if not isinstance(result, list) or not all(
+        isinstance(r, HtmlElement) for r in result
+    ):
+        raise ValueError(
+            f"Expected list[HtmlElement] as result of xpath, got {type(result)}"
```
jbothma: I thinkfriendrich mentioned wanting to get rid of HtmlElement. Any reason why we're using that? Can Element work here?

leonhandreke: Seems to work just as well, thanks for the hint

datasets/de/bundesrat/crawler.py
```diff
@@ -112,5 +113,6 @@ def crawl_item(context: Context, item: HtmlElement) -> None:
 def crawl(context: Context) -> None:
     doc = context.fetch_html(context.data_url, cache_days=3, absolute_links=True)
     container = doc.find(".//div[@class='row']/ul[@class='members-list']")
```
jbothma, line 115: Could skip the `assert is not none` with

```suggestion
    container = h.xpath_elements(doc, ".//div[@class='row']/ul[@class='members-list']", expect_exactly=1)[0]
```

jbothma (approved):
cool!

zavod/zavod/helpers/html.py
```diff
@@ -129,3 +129,37 @@ def links_to_dict(el: Element) -> Dict[str | None, str | None]:
...
+    Return a list of strings that match the given XPath expression.
+    """
+    result = el.xpath(xpath)
+    if not isinstance(result, list) or not all(isinstance(r, str) for r in result):
+        raise ValueError(f"Expected list[str] as result of xpath, got {type(result)}")
+    if expect_exactly is not None and len(result) != expect_exactly:
+        raise ValueError(
+            f"Expected {expect_exactly} elements, got {len(result)} for xpath {xpath!r}"
+        )
+    return [cast(str, r) for r in result]
```
pudo, line 165: HTML strings often contain weird characters from HTML entities, like `\x0a`. Maybe we can filter using `normality.cleaning:remove_unsafe_chars`

PR #3058, opened 2025-11-12
Simplify name parser, make it more radical

pudo:


datasets/tw/shtc/crawler.py
```diff
@@ -99,49 +94,67 @@ def clean_names(
...
+    aliases: Set[str] = set()
+    for alias in aliases_str.split(";"):
+        alias = alias.strip()
+        if len(alias) == 0:
             continue
-        names.append(name)
+        for split in NAME_CHINESE:
+            splitted = alias.split(split, 1)
+            if len(splitted) > 1:
+                _, chinese_alias = splitted
```
jbothma, line 111: can we drop the bit before the chinese writing system prefix?

pudo: I think so: the structure seems to be `alias; chinese: alias` (ie. the number of occurrences of `; chinese:` in the file is same as `chinese:`)

PR #3066, opened 2025-11-13
Hungarian National Assembly

nvmbrasserie:
opensanctions/crawlers#555

### DIFF
--- a/datasets/hu/national_assembly/crawler.py
+++ b/datasets/hu/national_assembly/crawler.py
@@ -38,10 +38,6 @@ def crawl_row(context: Context, url: str, raw_name: str, party: str) -> None:
     entity = context.make("Person")
     entity.id = context.make_id(raw_name, party)
 
-    for honorific in ["Dr.", "Prof."]:
-        if raw_name.startswith(honorific):
-            raw_name = raw_name[len(honorific) :].strip()
-
     last_name, first_name = raw_name.split(",", 1)
 
     h.apply_name(entity, first_name=first_name, last_name=last_name)


datasets/hu/national_assembly/crawler.py
```diff
@@ -0,0 +1,100 @@
...
+        if occupancy is not None:
+            context.emit(occupancy)
+            context.emit(position)
+            context.emit(entity)
+
+
+def crawl(context: Context) -> None:
+    unblock = ".//table[@class=' table table-bordered']"
+    doc = zyte_api.fetch_html(context, context.data_url, unblock, cache_days=5)
+    table = doc.find(unblock)
```
jbothma: when its primary purpose is as the unblock validator, a better name would be `validator` - the noun. But since its primary purpose is selecting the table, and it happens to be a sane unblock validator, I think the best name is `table_xpath`
```suggestion
    table_xpath = ".//table[@class=' table table-bordered']"
    doc = zyte_api.fetch_html(context, context.data_url, unblock, cache_days=5)
    table = doc.find(unblock)
```

datasets/hu/national_assembly/crawler.py
```diff
@@ -0,0 +1,100 @@
...
+
+def crawl_row(context: Context, row: html.Element) -> None:
+    url_el = row.find(".//a[@href]")
+    if url_el is None:
+        context.log.warning("No URL found in row, skipping")
+        return
+    url = url_el.get("href")
+    assert url is not None, "No URL found in row"
+    unblock_pep = ".//div[@class='pair-content']"
+    pep_doc = zyte_api.fetch_html(context, url, unblock_pep, cache_days=5)
```
jbothma:
```suggestion
    validator = ".//div[@class='pair-content']"
    pep_doc = zyte_api.fetch_html(context, url, validator, cache_days=5)
```

datasets/hu/national_assembly/crawler.py
```diff
@@ -0,0 +1,100 @@
...
+    if "(" not in full_name:
+        context.log.warning(f"No party found in: {full_name}")
+        return full_name, None
+    name, party = full_name.rsplit("(", 1)
+    return name.strip(), party.rstrip(")").strip()
+
+
+def parse_name(context: Context, full_name: str) -> tuple[str, str]:
+    """Strip title prefix and return (last_name, first_name) from 'Last, First' format."""
+    name = full_name.strip().removeprefix("Dr. ").removeprefix("Prof. ")
```
jbothma: there's now a better way... you can make . an "irregular character" by adding it to Person's `reject_chars`
https://zavod.opensanctions.org/extract/names/#whats-a-dirty-name

@pudo do we want to use it for something like stripping titles? It isn't optimised for it yet but it could be.

datasets/hu/national_assembly/crawler.py
```diff
@@ -0,0 +1,100 @@
...
+        name="Member of the National Assembly of Hungary",
+        wikidata_id="Q17590876",
+        country="hu",
+        topics=["gov.legislative", "gov.national"],
+        lang="eng",
+    )
+
+    elections_table = h.xpath_elements(
+        pep_doc, '//table[.//th[@colspan="5" and text()="Elections"]]', expect_exactly=1
+    )[0]
```
jbothma, line 59: how sure are we that Eections is the table we want? My intuition would be that Elections is when they've stood for election but weren't necessarily elected, and `Parliamentary Group Membership` is when they've been an MP. The only way I can think of checking this is finding an instance of a current MP who we know didn't get elected for some term, and see what both tables look like.

jbothma: I think it'd be worthwhile getting the `State functions` table in addition to this. Some but not all of those seem to be executive functions so let's use the positions UI to set that topic rather than the crawler.

image

leonhandreke:
How do you feel about getting name and party from the root table where they're already separated? https://github.com/opensanctions/opensanctions/commit/7ff4b6018459fb672a8b419b7ac3a59bd30b05f7

datasets/hu/national_assembly/crawler.py
nvmbrasserie: I just implemented @leonhandreke's suggestion to get the names from the main page, so we don't need to clean them up.

datasets/hu/national_assembly/crawler.py
nvmbrasserie: I double checked and both tables are always identical. I think we're good with the `Elections` one.

datasets/hu/national_assembly/crawler.py
```diff
@@ -0,0 +1,123 @@
...
+
+def crawl_row(context: Context, url: str, raw_name: str, party: str) -> None:
+    unblock_pep = ".//div[@class='pair-content']"
+    pep_doc = zyte_api.fetch_html(context, url, unblock_pep, cache_days=5)
+    entity = context.make("Person")
+    entity.id = context.make_id(raw_name, party)
+
+    for honorific in ["Dr.", "Prof."]:
+        if raw_name.startswith(honorific):
+            raw_name = raw_name[len(honorific) :].strip()
```
jbothma: let's use `rigour.names.remove_person_prefixes`

datasets/hu/national_assembly/crawler.py
```diff
@@ -0,0 +1,123 @@
...
+
+    last_name, first_name = raw_name.split(",", 1)
+
+    h.apply_name(entity, first_name=first_name, last_name=last_name)
+    entity.add("political", party)
+    entity.add("sourceUrl", url)
+    entity.add("citizenship", "hu")
```
jbothma, line 46: is this a standard thing we're doing now? Is it something we can assume about national legislatures? make_occupancy does add the positions country to `country` already.

datasets/hu/national_assembly/crawler.py
```diff
@@ -0,0 +1,123 @@
...
+                country="hu",
+                topics=["gov.national"],
+            )
+            categorise_and_emit(
+                context,
+                entity=entity,
+                position=position,
+                start_date=str_row.pop("from"),
+                end_date=str_row.pop("to", None),
+                # is_pep is not set, since we rely on positions UI to set that topic rather than the crawler
```
jbothma:
```suggestion
                # is_pep is not set, since not all these positions are necessarily PEP positions
```

At least, is_pep isn't a topic, it's a flag on a position.

datasets/hu/national_assembly/crawler.py
nvmbrasserie: yes, we decided to standardise it recently
Screenshot 2025-12-11 at 17 08 27

Issue #2541, opened 2025-07-24
Better semantics for failed runs


leonhandreke:
Pulled out from the archived PR https://github.com/opensanctions/opensanctions/pull/2483 and https://github.com/opensanctions/opensanctions/issues/2364 and conversations. Core issue: right now, we don't really have the notion of a successful/unsuccessful run and consequently don't deal with them very well.

- The website just displays the latest run, and if that's empty, tough luck.
- The backfiller in zavod just iterates through the versions until it finds a `statements.pack` it can backfill. `index.json` will contain stats from the last `statistics.json` retrieved from the backfiller, but no resources, and the issues from the failed run.
- We don't currently have a great way to figure out which version of a dataset went into a collection export.
- `latest` can also be a failed run (I think?)

Room for improvement here. Also remember to write docs on this before considering this ticket complete ;)

jbothma:
customers who have already loaded data and are just looking for an update are ok, but it could be made clearer why no resources are listed in the latest index, if we stick to that. 
Not listing the previous crawl's resources in a failed run's index is probably a good thing - it should be very clear to which export a resource belongs to, and that that failed run's index did not produce that data.

customers who need to load data even if it's stale (new customers, existing customers newly provisioning a database) need to easily find the best data we have. 
perhaps we could include a "last_successful_export": "https://data.opensanctions.org/datasets/20250808/interpol_red_notices/index.json" in the metadata in those cases

jbothma:
maybe this is a good place for it - instead of [global state as a separate thing for change tracking](https://github.com/opensanctions/opensanctions/pull/2747), we discussed introducing a flag to Context to indicate that a crawler shouldn't publish, but we do want it to run as far as it can e.g. to cache HTTP or LLM requests

One thing about the current approach of raising an exception, is that that means a job that we told not to publish, gets rerun - maybe that's ok - it's just extra weird with the review crawlers because while the crawler is running, the "latest" version isn't complete.

jbothma:
It'd be nice if some or all disk state could be kept around for a while even if a run doesn't publish. Having the statements.pack for https://github.com/opensanctions/opensanctions/issues/2913 would have helped speed up debugging significantly because a) it takes a long time to generate, and b) what I generated the first time didn't actually reflect what was happening in prod

leonhandreke:
More awfulness:

```sh
$ gsutil cat gs://data.opensanctions.org/artifacts/opencorporates/20251112035734-hej/issues.log
[...]

PR #3079, opened 2025-11-18
WIP: Don't publish failed runs `index.json` to `/datasets`

leonhandreke:
This commit tries to establish some new language around publishing: "Archiving" means uploading to `/artifacts`, while "publishing" means pushing the file out to our customers at `/datasets`.

By no longer unlinking `statistics.json` before calling `write_dataset_index`, we will also be archiving the stats not from the last successful run, but from the current one (which may not have any stats at all if things failed very early on. Neither `index.json` nor `statistics.json` will end up in `/latest` for a failed run.

Unanswered questions:

- Does it make sense to archive (not publish!) all artifacts for forensics? Do we want to exclude some?
- Why were we unlinking RESOURCES_FILE and VERSIONS_FILE at the end of `publish_failure`? What purpose did that serve?

I don't think we have to tell our customers that we're no longer serving them bullshit `index.json`s that don't contain any resources but instead stats from the last successful run... do we?

See https://github.com/opensanctions/opensanctions/issues/2541

### DIFF
--- a/zavod/zavod/tests/test_publish.py
+++ b/zavod/zavod/tests/test_publish.py
@@ -11,7 +11,7 @@ from zavod.crawl import crawl_dataset
 from zavod.store import get_store
 from zavod.exporters import export_dataset
 from zavod.integration import get_dataset_linker
-from zavod.publish import publish_dataset, publish_failure
+from zavod.publish import publish_dataset, archive_failure
 from zavod.exc import RunFailedException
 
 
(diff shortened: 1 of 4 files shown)


zavod/zavod/tests/test_publish.py
```diff
@@ -71,25 +71,27 @@ def test_publish_dataset(testdataset1: Dataset):
...
+    artifact_path = artifacts_path / testdataset1.name / history.latest.id
 
     assert not latest_path.joinpath("statements.pack").exists()
     assert not latest_path.joinpath("issues.json").exists()
     assert artifact_path.joinpath("issues.json").exists()
     assert artifact_path.joinpath("index.json").exists()
-    assert latest_path.joinpath("index.json").exists()
+
+    # We don't want failed runs to end up in /datasets
+    assert len(list(latest_path.glob("*"))) == 0
```
jbothma, line 97: we also don't want to backfill from them https://github.com/opensanctions/opensanctions/pull/3159

Issue #2237, opened 2025-05-09
[pk_proscribed_persons] New JS-driven website


jbothma:
the ops guy I found on linkedin working for pakistan's counter terrorism agency connected me with a developer working for them, and he said he can look into it if we can provide the request on a document with a company letterhead.

But I think we can also just scrape their site and and leave their useless JSON link alone.

Right now the website isn't accessible - maybe because of geoblocking? Maybe temporarily because of the conflict with india?

So probably worth giving it a week to see what's happening.

jbothma:
today it decided to load. it looks like the UI is some auto-generated .net thing. The Export buttons make multiple requests.

The pagination buttons are the same, so scraping the html is not as simple as a browser load in zyte - we'd need a full browser, or to figure out the viewstate-esque nonsense.

@pudo shall we try the letterhead thing? In the past when I've phoned them I haven't really reached anyone helpful.

![Image](https://github.com/user-attachments/assets/2b68596b-83b4-4861-8132-7afc853115fd)

jbothma:
A browser-based visit to the site to download the file and then normal file crawling will do it. No luck asking them to improve the download button

leonhandreke:
We need a JS browser to click the export link, which then downloads a `blob:`. Haven't found a way to do that with Zyte. I think it's time to pronounce it dead.

leonhandreke:
Decision: Manually check in the data every 3 months.

PR #3102, opened 2025-11-18
Issues 2025-11-18

leonhandreke:
- **[us_finra_actions] remove custom table parsing and appease typechecker**
  

- **pre-commit mypy: add the individual datasets to exclude for us/**
  Work them down, one by one!
  

- **[in_med_exclusions] Make the typechecker happy**
  

- **[pk_proscribed_persons] check in source data and instructions to update manually**
  Fixes https://github.com/opensanctions/opensanctions/issues/2237
  

- **[pk_proscribed_persons] wanted -> sanctions, it's a UNSCR1373 list**
  

- **[ps] Make the typechecker happy**
  

- **[ps_ohchr_settlement] Update hash, no changes, 2025 report was already added**
  

- **[ch_parlament] Properly detect the end of the list with Zyte**

### DIFF
--- a/zavod/zavod/extract/zyte_api.py
+++ b/zavod/zavod/extract/zyte_api.py
@@ -158,6 +158,9 @@ class ZyteResult:
 
     response_text: str
 
+    # If the response is served from the cache, status_code is None
+    status_code: Optional[int]
+
     # The cache fingerprint
     cache_fingerprint: str
     from_cache: bool
@@ -232,7 +235,10 @@ def fetch(
                 "HTTP cache hit", url=zyte_request.url, fingerprint=fingerprint
             )
             return ZyteResult(
-                response_text=text, from_cache=True, cache_fingerprint=fingerprint
+                response_text=text,
+                status_code=None,
+                from_cache=True,
+                cache_fingerprint=fingerprint,
             )
 
     context.log.debug(f"Zyte API request: {zyte_request.url}", data=zyte_data)
@@ -255,6 +261,7 @@ def fetch(
         text = b64_text.decode(charset) if charset is not None else b64_text.decode()
 
     return ZyteResult(
+        status_code=api_response.json()["statusCode"],
         response_text=text,
         media_type=media_type,
         charset=charset,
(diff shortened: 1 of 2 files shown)


zavod/zavod/extract/zyte_api.py
```diff
@@ -255,6 +261,7 @@ def fetch(
         text = b64_text.decode(charset) if charset is not None else b64_text.decode()
 
     return ZyteResult(
+        status_code=api_response.status_code,
```
jbothma: does this work? I think what you're looking for is `api_response.json()["statusCode"]`

https://docs.zyte.com/zyte-api/usage/reference.html#operation/extract/response/200/statusCode

datasets/pk/proscribed_persons/pk_proscribed_persons.yml
```diff
@@ -5,8 +5,16 @@ entry_point: crawler.py
 prefix: pk-proscribed
 summary: List of Proscribed Persons in Pakistan under the Anti Terrorism Act, 1997
 coverage:
-  frequency: daily
+  # Both monthly and yearly would be a lie, it's roughly every 90 days.
+  frequency: unknown
```
jbothma: mmm so if this is `never` we'll show `frequency: not automated` on the website.

If it's `unknown` and `coverage.schedule` is unset, we'll show neither cron text, nor `frequency: unknown`

So I think `never` is the better value, since `frequency: quarterly` could show on the website but that's inconsistent with the other manually updated datasets.

zavod/zavod/extract/zyte_api.py
leonhandreke: Yes, thanks. I did briefly test locally and it didn't crash, weird...

datasets/pk/proscribed_persons/pk_proscribed_persons.yml
leonhandreke: What is the relationship of `coverage.schedule` and `deploy.schedule`? I've seen both.

I'll make it `never` for now.

PR #3103, opened 2025-11-19
Norway Members of the Storting

nvmbrasserie:
opensanctions/crawlers#557

### DIFF
--- a/datasets/no/storting/crawler.py
+++ b/datasets/no/storting/crawler.py
@@ -24,7 +24,8 @@ def parse_ms_date(ms_date: str | None) -> date | None:
       which in the -0700 timezone was 2015-10-19T17:40:15"
     - The offset doesn't modify the timestamp; it's metadata about local time
 
-    For dates (like dates of birth), we need the LOCAL date, so we:
+    We use naive (non-timezone-aware) dates everywhere in our data model and interpret based
+    on context. For dates of birth, that means local time. So to get local time, we:
     1. Parse the UTC timestamp
     2. Apply the offset to get the local datetime
     3. Extract the date from that local datetime


datasets/no/storting/no_storting.yml
```diff
@@ -0,0 +1,70 @@
...
+coverage:
+  frequency: monthly
+  start: 2025-11-19
+load_statements: true
+entry_point: crawler.py
+summary: >-
+  Current and recent members of the Norwegian Storting, the national legislature.
+description: |
+  This dataset provides information about the current and recent members of the Storting,
+  Norwegian Parliament. It includes details about each member, such as name, date and
```
leonhandreke:
```suggestion
  the Norwegian Parliament. It includes details about each member, such as name, date and
```

leonhandreke: Which date?

datasets/no/storting/crawler.py
```diff
@@ -0,0 +1,118 @@
...
+from zavod import Context, helpers as h
+from zavod.stateful.positions import categorise
+
+DATA_URL = "https://data.stortinget.no/eksport/representanter?stortingsperiodeid={term}&format=json"
+CUTOFF_YEAR = date.today().year - 20
+
+
+def parse_ms_date(ms_date: str) -> str | None:
+    """
+    Converts /Date(1763485375798+0100)/ or /Date(-106963200000+0200)/
```
leonhandreke: Is there any special meaning to the "-" prefix on the second one?

datasets/no/storting/crawler.py
```diff
@@ -0,0 +1,118 @@
...
+    ms = int(match.group("ms"))
+    tz = match.group("tz")
+    # Convert milliseconds to UTC datetime
+    dt = datetime.utcfromtimestamp(ms / 1000)
+    # Apply timezone offset if present
+    if tz:
+        sign = 1 if tz[0] == "+" else -1
+        hours_offset = int(tz[1:3])
+        minutes_offset = int(tz[3:5])
+        dt += timedelta(hours=hours_offset, minutes=minutes_offset) * sign
```
leonhandreke: I think this interpretation is wrong. See https://learn.microsoft.com/en-us/dotnet/standard/datetime/system-text-json-support and https://learn.microsoft.com/en-us/dotnet/api/system.datetimeoffset.tooffset?view=net-10.0#system-datetimeoffset-tooffset(system-timespan)

> An object that is equal to the original DateTimeOffset object (that is, their ToUniversalTime() methods return identical points in time) but whose Offset property is set to offset.

So really, we can ignore the timezone. This is merely a display hint, and I think in this context we can safely ignore it.

datasets/no/storting/crawler.py
```diff
@@ -0,0 +1,118 @@
+import re
+from datetime import date, datetime, timedelta
+
+from zavod import Context, helpers as h
+from zavod.stateful.positions import categorise
+
+DATA_URL = "https://data.stortinget.no/eksport/representanter?stortingsperiodeid={term}&format=json"
+CUTOFF_YEAR = date.today().year - 20
```
leonhandreke: Use `EXTENDED_AFTER_OFFICE` here, this is a policy thing and we should share it as much as possible throughout the codebase in case we ever need to update it.

datasets/no/storting/no_storting.yml
```diff
@@ -0,0 +1,70 @@
...
+  name: Stortinget
+  description: |
+    The Storting (Stortinget) is the supreme legislature of Norway, consisting of 169 representatives
+    elected every four years. It is a unicameral body, meaning it has only one chamber, and is
+    responsible for making laws, approving the budget, and overseeing the government.
+  country: "no"
+  url: https://www.stortinget.no/en/
+  official: true
+data:
+  url: https://data.stortinget.no/eksport/representanter?stortingsperiodeid=2025-2029&format=json
```
leonhandreke: (nit) Can we put the endpoint here and then append `stortingsperiodeid={term}` in the code? That would remove the duplication. I think we do this in quite a few other crawlers where `data.url` isn't something you can just access, it needs additional params/auth...

datasets/no/storting/crawler.py
```diff
@@ -0,0 +1,118 @@
...
+    # Apply timezone offset if present
+    if tz:
+        sign = 1 if tz[0] == "+" else -1
+        hours_offset = int(tz[1:3])
+        minutes_offset = int(tz[3:5])
+        dt += timedelta(hours=hours_offset, minutes=minutes_offset) * sign
+    return str(dt.date().isoformat())
+
+
+def translate_keys(member, context) -> dict:
```
leonhandreke: (nit) we often put `context: Context` first in our type signatures.

datasets/no/storting/crawler.py
```diff
@@ -0,0 +1,118 @@
...
+    h.apply_date(entity, "deathDate", parse_ms_date(item.pop("death_date", None)))
+    entity.add("political", item.pop("party").pop("name"))
+    entity.add("citizenship", "no")
+    entity.add("gender", context.lookup_value("gender", item.pop("gender")), lang="eng")
+    if not entity.get("gender")[0]:
+        context.log.warning(f"Unknown gender for {entity.id}")
+
+    position = h.make_position(
+        context,
+        name="Member of the Parliament of Norway",
```
leonhandreke, line 113: (nit) We use the OG names in all other crawlers - shouldn't we here? I think the Wikidata ID is to do the link properly, we can call it whatever we want.

datasets/no/storting/crawler.py
```diff
@@ -0,0 +1,118 @@
...
+from datetime import date, datetime, timedelta
+
+from zavod import Context, helpers as h
+from zavod.stateful.positions import categorise
+
+DATA_URL = "https://data.stortinget.no/eksport/representanter?stortingsperiodeid={term}&format=json"
+CUTOFF_YEAR = date.today().year - 20
+
+
+def parse_ms_date(ms_date: str) -> str | None:
```
leonhandreke: (nit) I think returning `datetime.date` or `datetime.date` here would be nice? Just less strings being thrown around everywhere :)

datasets/no/storting/no_storting.yml
nvmbrasserie: date and place of birth

datasets/no/storting/crawler.py
nvmbrasserie: IIUC, it stands for the time before the Unix epoch

datasets/no/storting/crawler.py
nvmbrasserie: So far, I've always used the name from the wikidata. My feeling is that it's better to keep it consistent and not adjust unless it's absolutely necessary

datasets/no/storting/crawler.py
leonhandreke: Ah, that makes sense, old people :)

PR #3112, opened 2025-11-19
Sweden Members of Riksdag

### LINKED ISSUE #558: Swedish Parliament
### Data URL

https://www.riksdagen.se/en/members-and-parties/members/

### Publisher

https://www.riksdagen.se/en/

se

### Type of data

### Coverage region

region:Europe

### Can you tell us more?

_No response_

- [x] I understand.

nvmbrasserie:
fixes opensanctions/crawlers#558

The Riksdag Open Data API endpoint we use as a `data_url` (see [docs](https://www.riksdagen.se/sv/dokument-och-lagar/riksdagens-oppna-data/)) returns only individuals who have served in the Swedish Parliament in some capacity since 2018. The dataset includes both full members and temporary substitutes ("`ersättare`") but does not always provide explicit parliamentary term boundaries for every person.

Alternatively, we could use the API endpoint covering all MPs since 1991 (see [constructor](https://data.riksdagen.se/personlista/)). However, that data often lacks complete dates for older parliamentary terms. Since we keep MPs for 20 years, using the 1991 endpoint would result in many older MPs having no term boundaries, leading to a large number of entries with status UNKNOWN.


datasets/se/riksdag/crawler.py
```diff
@@ -0,0 +1,110 @@
...
+                "Riksdagsledamot",  # MP
+                "Statsrådsersättare",  # MP substituting for a minister
+            }:
+                start = role.get("from")
+                end = role.get("tom")
+                return start, end
+    return None, None
+
+
+def extract_gender(context, gender, entity_id) -> Optional[str]:
```
leonhandreke: I think in a bunch of other crawlers we just use a `type.gender` lookup. WDYT?

datasets/se/riksdag/crawler.py
```diff
@@ -0,0 +1,110 @@
...
+        first_name = item.pop("first_name")
+        last_name = item.pop("last_name")
+
+        entity = context.make("Person")
+        entity.id = context.make_id(first_name, last_name, id)
+        h.apply_name(entity, first_name=first_name, last_name=last_name)
+        h.apply_date(entity, "birthDate", item.pop("birth_year"))
+        entity.add("political", item.pop("party"))
+        entity.add("citizenship", "se")
+        entity.add("sourceUrl", item.pop("person_url_xml"))
```
leonhandreke: What sort of URL is this? Does it point to an XML file? I think so far, we've often used `sourceUrl` as a "this is where you can look at it" URL.

datasets/se/riksdag/crawler.py
```diff
@@ -0,0 +1,110 @@
...
+    return translated
+
+
+def extract_terms(
+    item: Dict[str, Any],
+) -> Tuple[Optional[str], Optional[str]]:
+    """
+    Extract the start and end dates of the main parliamentary term.
+    """
+    for role in item.pop("person_mandate", {}).pop("mandate", []):
```
leonhandreke, line 29: If this is the list, what's usually in it? If there are/were multiple mandates, does it make sense to extract a list of terms?

leonhandreke (commented):
I think I don't fully get the point about the cutoff being 2018. IIUC, we're using the same endpoint that can be used to get all since 1990, but we're using it with some different params?

datasets/se/riksdag/crawler.py
nvmbrasserie: yeah, there is a xml file with more details there

datasets/se/riksdag/crawler.py
nvmbrasserie: It's a list of all positions including the governmental ones (that are not relevant to this crawler), so we exclude all but the ones related to the Parliament

leonhandreke: Thanks for the detailed comment, that's great!

PR #3118, opened 2025-11-20
Add more name irregularity checks

jbothma:


zavod/zavod/helpers/names.py
```diff
@@ -252,14 +253,29 @@ def split_comma_names(context: Context, text: str) -> List[str]:
...
         return False
 
     for spec in entity.dataset.names.specs_for_schema(entity.schema):
+        # Contains rejected characters
         for char in spec.reject_chars_consolidated:
             if char in string:
                 return True
 
+        # is nullword
+        if not spec.allow_nullwords and is_nullword(string, normalize=True):
```
pudo, line 268: all of these rigour helpers kind of want a normalized version of the string to be passed in. So I wonder if after doing squash_space, we also generate a normalized form once (basically just `rigour.names.tokenize:normalize_name`) and use it for these checks.

zavod/zavod/helpers/names.py
```diff
@@ -252,14 +253,29 @@ def split_comma_names(context: Context, text: str) -> List[str]:
...
+            return True
+
+        # min length
+        if len(string) < spec.min_chars:
+            return True
+
+        # requires space
+        if spec.require_space and " " not in string:
```
pudo, line 276: Love simple stuff like this :) Squash space already fixed all the edge cases.

pudo (approved):
Very cool!

FWIW, I just introduced a new helper in `rigour` called `rigour.names.check:is_generic_person_name` which just checks against a denylist of names that are very common in the data and meaningless (`abu bakr`). Can do that next round.

PR #3130, opened 2025-11-24
Finland Members of Parliament

nvmbrasserie:
opensanctions/crawlers#560

### DIFF
--- a/datasets/fi/eduskunta/fi_eduskunta.yml
+++ b/datasets/fi/eduskunta/fi_eduskunta.yml
@@ -28,6 +28,8 @@ data:
   url: https://avoindata.eduskunta.fi/api/v1/
   format: JSON
   lang: fin
+http:
+  user_agent: "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.2 Safari/605.1.15 (zavod; OpenSanctions.org)"
 dates:
   formats: ["%d.%m.%Y"]
 


datasets/fi/eduskunta/fi_eduskunta.yml
```diff
@@ -0,0 +1,151 @@
...
+summary: >-
+  Current members of the Eduskunta, the Finnish Parliament.
+description: |
+  This dataset contains information on the current members of the Finnish Parliament (Eduskunta).
+  It includes the names and political affiliations of all members, along with details such as their
+  parliamentary terms and dates of birth.
+tags:
+  - list.pep
+publisher:
+  name: Suomen eduskunta
```
leonhandreke, line 18: (nit) `name_en` "Parliament of Finland"? That's what they call themselves at https://www.eduskunta.fi/EN/Pages/default.aspx

datasets/fi/parliament/fi_parliament.yml
```diff
@@ -0,0 +1,151 @@
+title: Finland Members of Parliament
+prefix: fi-parliament
+url: https://www.eduskunta.fi/EN/Pages/default.aspx
```
leonhandreke: (nit) Maybe actually `https://avoindata.eduskunta.fi/#/fi/members` cause that's where the data is "from"?

datasets/fi/parliament/fi_parliament.yml
```diff
@@ -0,0 +1,151 @@
+title: Finland Members of Parliament
+prefix: fi-parliament
```
leonhandreke: (nit) I think often we use the native term as the prefix (se-riksdag, de-bundestag). "eduskunta" is a bit weird to us English speakers. I think this is fine, so feel free to just resolve. I just noticed and think it might be good to think about for two minutes and be consistent in the future.

datasets/fi/parliament/crawler.py
```diff
@@ -0,0 +1,121 @@
...
+    "as_member_of_parliament",
+    "phone",
+    "email",
+    "nickname",
+    "education",
+    "parliamentary_term_suspended",
+    "work_history",
+    "government_memberships",
+]
+PEP_URL = "https://avoindata.eduskunta.fi/api/v1/memberofparliament/{person_id}/fi"
```
leonhandreke: (nit) Let's use `data_url` as a prefix and then build the endpoints on top of it. More of a style thing since when this changes, we'll have to rework it anyway. But like this, we have a "base URL" and endpoints at that URL, and that's very API-y :)

datasets/fi/parliament/crawler.py
```diff
@@ -0,0 +1,121 @@
...
+
+        entity = context.make("Person")
+        entity.id = context.make_id(person_id, first_name, last_name, birth_year)
+        h.apply_name(entity, first_name=first_name, last_name=last_name)
+        entity.add("birthDate", birth_year)
+        entity.add("birthPlace", pep.pop("birth_place", None))
+        entity.add("gender", pep.pop("gender"), lang="eng")
+        entity.add("position", pep.pop("occupation"))
+        entity.add("address", pep.pop("current_municipality"))
+        entity.add("sourceUrl", pep_url)
```
leonhandreke: Same here, let's add a question to the issue to figure out whether to put this in sourceUrl

datasets/fi/parliament/crawler.py
```diff
@@ -0,0 +1,121 @@
...
+        last_name = pep.pop("last_name")
+        birth_year = pep.pop("birth_year")
+        person_id = pep.pop("person_id")
+
+        entity = context.make("Person")
+        entity.id = context.make_id(person_id, first_name, last_name, birth_year)
+        h.apply_name(entity, first_name=first_name, last_name=last_name)
+        entity.add("birthDate", birth_year)
+        entity.add("birthPlace", pep.pop("birth_place", None))
+        entity.add("gender", pep.pop("gender"), lang="eng")
```
leonhandreke: Why `lang=eng`? Does it make sense to pass lang here? Isn't gender a type that only allows a few English values?

datasets/fi/eduskunta/crawler.py
```diff
@@ -0,0 +1,120 @@
...
+            topics=["gov.legislative", "gov.national"],
+            lang="eng",
+        )
+
+        categorisation = categorise(context, position, is_pep=True)
+        if not categorisation.is_pep:
+            continue
+
+        parliamentary_terms = pep_data.pop("parliamentary_terms", {})
+        all_terms = get_parliamentary_terms(parliamentary_terms)
```
leonhandreke, line 108: So IIUC, that's all the current ones, but with past terms as well? If so, let's document.

leonhandreke (commented):
Looks great and clean! I pushed a little commit with some renames and type fixes - hope you agree, if not let's quickly talk tomorrow :)

Some little comments inline.

datasets/fi/parliament/crawler.py
nvmbrasserie: We did it before in similar cases, and I think adding `lang` to the few non-Finnish values in the data makes for cleaner output

PR #3131, opened 2025-11-24
[name cleaning] Prompt with one or more strings, incl schema

jbothma:
Schema helps it put fairly long single token names for persons into weakAlias without wrecking all the single company names

- Cleans/Reviews names in eu_fsf, be_fod, eu_travel_bans and the needing cleaning (but doesn't apply them yet)
- Replaces us_sam logic-based extraction reviews with LLM-based extraction/reviews but applies the logic-based extraction.
- Adds a JSONSourceValue class for structured inputs, and allows SourceValues to compare to other types (treating other types as non-matches, i.e. as if the source value changed, which it must have, if the type changed)

### DIFF
--- a/zavod/zavod/tests/test_tune.py
+++ b/zavod/zavod/tests/test_tune.py
@@ -18,7 +18,8 @@ from zavod.extract.names.dspy.clean import load_optimised_module
 from zavod.tune import cli
 
 example = {
-    "string": "John Doe",
+    "entity_schema": "Person",
+    "strings": ["John Doe"],
     "full_name": ["John Doe"],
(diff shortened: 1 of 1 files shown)


jbothma:
this branch at the current point:

```
2025-11-24 17:19:35 [info     ] Delta export complete          [us_sam_exclusions] added=0 dataset=us_sam_exclusions deleted=8 metric=delta_counts modified=13 version=20251124171137-hwg
```

i.e. up to 13 entities will revert to the logic-based un-moderated values. Seems to be mainly moves from alias to weakAlias.

PR #3135, opened 2025-11-24
Move cleaned name reviewing out of helpers for circular dep

jbothma:
To discuss, haven't updated tests:

We currenlty have a circular dependency:

```
zavod.helpers.names -> zavod.stateful.review
zavod.stateful.review -> zavod.helpers.html
zavod.helpers -> zavod.helpers.html
zavod.helpers -> zavod.helpers.names
```

This proposes

```
zavod.extract.names.clean -> zavod.stateful.review
zavod.stateful.review -> zavod.helpers.html
```

So `zavod.extract.names.clean` becomes higher level than helpers.

The only alternative I see is to move the dependencies in `zavod.helpers.html` to a lower level module.


pudo:
No! This is architectural chaos. Hard veto. Please let's remove the helper use from stateful.

pudo:
This is all about avoiding one call to xpath("string()"), no?

jbothma:
Cool - yeah it didn't quite feel right but I hadn't checked yet where the helper call came from. It came from here https://github.com/opensanctions/opensanctions/commit/0a75e51141651267e89032d47ac40f63bdfea086#diff-46e4db1b061609cf6f42ed3f9420ab75373c8c509bebc3ecbe472eda391a2eadR10-L147 where it made sense for the crawler to use the helper, but not for stateful.

PR #3139, opened 2025-11-25
Stop name review applying

jbothma:
```
2025-11-25 11:21:29 [info     ] Delta export complete          [us_sam_exclusions] added=0 dataset=us_sam_exclusions deleted=0 metric=delta_counts modified=0 version=20251125111733-dic
```

```
2025-11-25 11:43:42 [info     ] Delta export complete          [eu_fsf] added=0 dataset=eu_fsf deleted=0 metric=delta_counts modified=0 version=20251125113150-mss
```

### DIFF
--- a/datasets/us/sam_exclusions/crawler.py
+++ b/datasets/us/sam_exclusions/crawler.py
@@ -177,7 +177,7 @@ def crawl(context: Context) -> None:
             origin=origin,
         )
 
-        # h.review_names(context, entity, [name], "eng")
+        h.review_names(context, entity, [name])
 
         entity.add("firstName", row.pop("First", None), quiet=True, lang="eng")
         entity.add("middleName", row.pop("Middle", None), quiet=True, lang="eng")
--- a/zavod/zavod/shed/fsf.py
+++ b/zavod/zavod/shed/fsf.py
@@ -151,8 +151,8 @@ def parse_entry(context: Context, entry: Element) -> None:
             context.log.warning("Unknown language", lang=lang2)
             continue
 
-        # if not is_weak:
-        #    h.review_names(context, entity, [name.get("wholeName")])
+        if not is_weak:
+            h.review_names(context, entity, [name.get("wholeName")])
 
         h.apply_name(
             entity,


zavod/zavod/helpers/names.py
```diff
@@ -326,36 +341,15 @@ def review_names(
...
-    if not strings:
+    non_blank_strings = [s for s in strings if s]
+    if not non_blank_strings:
         return None
 
-    if settings.CI or not any(is_name_irregular(entity, s) for s in strings):
-        prop = "alias" if alias else "name"
-        for string in strings:
-            entity.add(prop, string, lang=lang)
+    if settings.CI or not any(is_name_irregular(entity, s) for s in non_blank_strings):
```
leonhandreke, line 349: (nit) not really part of this, but might be nice to do the `settings.CI` check last just so that as much code as possible runs in CI mode?

zavod/zavod/helpers/names.py
```diff
@@ -326,36 +341,15 @@ def review_names(
...
-    strings = [s for s in strings if s]
-
-    if not strings:
+    non_blank_strings = [s for s in strings if s]
+    if not non_blank_strings:
         return None
 
-    if settings.CI or not any(is_name_irregular(entity, s) for s in strings):
-        prop = "alias" if alias else "name"
-        for string in strings:
```
leonhandreke, line 339: So this was the problematic code segment, right?

zavod/zavod/helpers/names.py
```diff
@@ -326,36 +341,15 @@ def review_names(
...
-        context,
-        source_value=source_value,
-        original_extraction=names,
-        origin=LLM_MODEL_VERSION,
-    )
-    return review
+    return _review_names(context, entity, non_blank_strings)
 
 
 def apply_reviewed_names(
```
leonhandreke, line 355: So IIUC, usually we'd use `apply_reviewed_names`, but if we only want to create the reviews, we use `review_names?`

I think I sort of liked the design (if I'm getting this correctly) where the return value of `review_names` would be fed to an `apply_names` function - thinking ahead, there might be some policy decisions we want to put into the `apply` function arguments (like what happens with new, unreviewed names maybe?). But I think this works, we can always evolve it.

zavod/zavod/helpers/names.py
jbothma: it's the call to clean_names that we want to skip. But yeah... my wish is to come up with a notion of real-world-based mocks for a crawler to use in CI instead.

zavod/zavod/helpers/names.py
jbothma: yeah specifically `entity.add(prop, string, lang=lang)`

PR #3140, opened 2025-11-25
Tighten US SAM name requirements

jbothma:
Results in 2931 reviews

### DIFF
--- a/datasets/us/sam_exclusions/us_sam_exclusions.yml
+++ b/datasets/us/sam_exclusions/us_sam_exclusions.yml
@@ -46,7 +46,7 @@ names:
     # G&G INTERNACIONAL S.A.S. and as a LegalEntity entry it can cause lots of false
     # positives on person names.
     min_chars: 11
-    # Review even even company names without spaces since they can be very generic.
+    # Review all company names without spaces since they can be very generic in this dataset.
     require_space: true
   Person:
     min_chars: 10
--- a/zavod/zavod/meta/names.py
+++ b/zavod/zavod/meta/names.py
@@ -45,9 +45,9 @@ class NamesSpec(RootModel[Dict[str, CleaningSpec]]):
     """Name cleaning requirements by schema. All matching schema configurations will apply"""
 
     ###################
-    # NB when introducing more specific defaults, these could take prescedence
-    # over the defaults defined here. Datasets extending these defaults might
-    # need to be updated to extend the new default instead.
+    # Beware that when introducing defaults for more specific schemata, these could take
+    # precedence over extensions to the existing defaults in some datasets' metadata.
+    # Those datasets might need to be updated to extend the new default instead.
     root: Dict[str, CleaningSpec] = {
         "Person": CleaningSpec(
             reject_chars_baseline=";\\/()[]<>{}:",


datasets/us/sam_exclusions/us_sam_exclusions.yml
```diff
@@ -39,6 +39,18 @@ data:
 
 dates:
   formats: ["%m/%d/%Y"]
+names:
+  LegalEntity:
+    # Review names shorter than 11 chars, e.g. Sebastiano from G&G INTERNACIONAL S.A.S.
+    # We make Sebastiano an alias because it doesn't seem to be the primary official name of
```
leonhandreke, line 45: This comment made me think: Do we have any link from the review to the entity that it's for? Might be interesting to a) click through from reviews b) "fix up" a specific entity.

Out of scope here, just a thought.

datasets/us/sam_exclusions/us_sam_exclusions.yml
```diff
@@ -39,6 +39,18 @@ data:
...
+    # G&G INTERNACIONAL S.A.S. and as a LegalEntity entry it can cause lots of false
+    # positives on person names.
+    min_chars: 11
+    # Review even even company names without spaces since they can be very generic.
```
leonhandreke:
```suggestion
    # Review all company names without spaces since they can be very generic.
```

datasets/us/sam_exclusions/us_sam_exclusions.yml
```diff
@@ -39,6 +39,18 @@ data:
 
 dates:
   formats: ["%m/%d/%Y"]
+names:
+  LegalEntity:
+    # Review names shorter than 11 chars, e.g. Sebastiano from G&G INTERNACIONAL S.A.S.
+    # We make Sebastiano an alias because it doesn't seem to be the primary official name of
+    # G&G INTERNACIONAL S.A.S. and as a LegalEntity entry it can cause lots of false
+    # positives on person names.
+    min_chars: 11
```
leonhandreke, line 48: I guess both of these constants are also chosen as the SAM names generally fall in the "rather messy" category, right? (just for my understanding, no change suggestion here)

zavod/zavod/meta/names.py
```diff
@@ -44,6 +44,10 @@ def reject_chars_consolidated(self) -> Set[str]:
 class NamesSpec(RootModel[Dict[str, CleaningSpec]]):
     """Name cleaning requirements by schema. All matching schema configurations will apply"""
 
+    ###################
+    # NB when introducing more specific defaults, these could take prescedence
```
leonhandreke: I don't understand this comment. My understanding is that this model is used to in the dataset config. How do they have to be updated? Maybe I'm missing something really obvious here.

datasets/us/sam_exclusions/us_sam_exclusions.yml
jbothma: yeah! https://github.com/opensanctions/operations/issues/1930

datasets/us/sam_exclusions/us_sam_exclusions.yml
jbothma: 100%

zavod/zavod/meta/names.py
jbothma: Yeah this corresponds to e.g.
```yaml
names:
  LegalEntity:
    # Plenty of names should be split on comma, but not all.
    reject_chars: ","
```
in https://github.com/opensanctions/opensanctions/blob/main/datasets/ru/cbr_banks/ru_cbr_banks.yml#L41

If the crawler is emitting Organization or Company, as it stands, the LegalEntity config will match.

If we then want to introduce some defaults for `Organization`, since the most specific name spec is used, this LegalEntity config extension in the crawler is not going to apply any more, because a more specific schema is available (in the defaults).

so the idea with the comment is just to make it clear that introducing new defaults is a big deal. I think it will be very rare, but when we introduce them, we should consider whether crawler metadata need an update, e.g. adding the same extension for `Organization` in `ru_cbr`

jbothma: ahh I see why this was unclear. Updated the doc.

PR #3146, opened 2025-11-25
exporters: write result success/failure to `index.json`

leonhandreke:
See https://github.com/opensanctions/opensanctions/issues/2541

### DIFF
--- a/zavod/zavod/exporters/metadata.py
+++ b/zavod/zavod/exporters/metadata.py
@@ -20,7 +20,7 @@ from zavod.util import write_json
 log = get_logger(__name__)
 
 
-class DatasetVersionExitStatus(StrEnum):
+class DatasetVersionResult(StrEnum):
     SUCCESS = "success"
     FAILURE = "failure"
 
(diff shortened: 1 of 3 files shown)


zavod/zavod/exporters/metadata.py
```diff
@@ -96,6 +104,7 @@ def write_dataset_index(dataset: Dataset) -> None:
         meta["issue_levels"] = issues.by_level()
         meta["issue_count"] = sum(meta["issue_levels"].values())
     meta["last_export"] = settings.RUN_TIME_ISO
+    meta["exit_status"] = exit_status.value
```
pudo: Why not just `status`, `success` or `outcome` or `result`? `exit_status` to me refers to a very specific UNIX concept, which this isn't 100% overlapped with.

leonhandreke: Thanks, that's exactly why I sent it out. I thought `status` was a bit vague, since this file has things about the dataset more generally and the run. I like `result`.

pudo: Yeah I can also totally see this having more than 2 values some time down the road. For example `failed`, `success`, `unchanged`, `unavailable` etc. etc. etc.

Issue #2969, opened 2025-10-20
Improve graph annotator


pudo:
Here's two potential improvements for the graph annotator:

* Normally, `sanction.linked` is only propagated down an ownership chain. I think in the case that the `sanction` entity is a `Vessel`, we should also propagate across other links (so as to mark the firm owning vessel as risky). 

* We have a function called `from followthemoney.helpers import post_summary` which we're using in a very limited set of crawlers to inline position descriptions ("Minister of Finance (2021 - 2025)") into the main entity. I wonder if this should be handled by the graph annotator instead. Duplicate code that is doing roughly the same thing also exists in `nomenklatura.wikidata.qualified:post_summary`, which is used by the Wikidata enricher.

jbothma:
~I suspect we have a bug where the order that an ownership chain is encountered affects how far down we propagate the ownership.~ 

A.sanc -> B -> C
if we encounter A then B then C we're good.
if we encounter A then C then B we'll never tag C

~I think going down the ownership chain should be a graph traversal rather than just looking at the adjacent in view order.~

~We can cache and skip the traversed chains too. So we don't double up on chains for owners that are linked to two sanctioned entities.~

Edit:I forgot that this propagates from one run to the next

jbothma:
> This should be sanction.linked, too: https://www.opensanctions.org/entities/isin-CNE100000981/

This is done

> Normally, sanction.linked is only propagated down an ownership chain. I think in the case that the sanction entity is a Vessel, we should also propagate across other links (so as to mark the firm owning vessel as risky).

This is not done

> We have a function called from followthemoney.helpers import post_summary which we're using in a very limited set of crawlers to inline position descriptions ("Minister of Finance (2021 - 2025)") into the main entity. I wonder if this should be handled by the graph annotator instead. Duplicate code that is doing roughly the same thing also exists in nomenklatura.wikidata.qualified:post_summary, which is used by the Wikidata enricher.

I think this should be a different ticket. I think it should maybe be in ann_pep_positions, or does it go beyond peps?

> Ok so I think the next step here would be to be more precise in which type of risk topic does recursive propagation, vs. one step out: https://github.com/opensanctions/followthemoney/issues/247

So this is is the next step along with the _inline position descriptions_ ticket?

pudo:
I think the question of what should or should not be transitive is answered in more detail here: 

* #4496 
* #3008

PR #3147, opened 2025-11-25
sanction.linked topic for Sanction linked securities

jbothma:
This addresses the gap identified in https://github.com/opensanctions/opensanctions/issues/2969#issuecomment-3546904614 where Sanction-linked securities are not currently given the sanction.linked topic.

### DIFF
--- a/datasets/_analysis/ann_graph_topics/analyzer.py
+++ b/datasets/_analysis/ann_graph_topics/analyzer.py
@@ -22,11 +22,11 @@ def emit_patch(
 ) -> None:
     context.log.info(
         f"Adding topic: {topic}",
-        source=risk_source.caption,
-        source_id=risk_source.id,
-        linked=related_entity.caption,
-        linked_id=related_entity.id,
-        topics=existing_topics,
+        risk_source=risk_source.caption,
+        risk_source_id=risk_source.id,
+        related_entity=related_entity.caption,
+        related_entity_id=related_entity.id,
+        existing_topics=list(existing_topics),
     )
 
     patch = context.make(related_entity.schema)
@@ -67,7 +67,7 @@ def analyze_entity(context: Context, view: View, entity: Entity) -> None:
         if "sanction" in topics:
             if entity.schema.is_a("Security"):
                 continue
-            if adjacent.schema.is_a("Security"):
+            if prop.name == "securities" and adjacent.schema.is_a("Security"):
                 adj_topics = non_graph_topics(context, adjacent)
                 if adj_topics.intersection({"sanction", "sanction.linked"}):
                     continue


jbothma:
```
colordiff <(csvcut -c1-3 artifacts_ann_graph_topics_20251125184001-jzu_statements.pack | sort) <(csvcut -c1-3 ann_graph_topics-statements-main-branch.pack | sort) > main.diff
colordiff <(csvcut -c1-3 artifacts_ann_graph_topics_20251125184001-jzu_statements.pack | sort) <(csvcut -c1-3 data/datasets/ann_graph_topics/statements.pack | sort) > securities.diff
colordiff main.diff securities.diff |less -R
```

shows sanction-linked securities tagged sanction.linked incl https://www.opensanctions.org/entities/isin-CNE100000981/ and no other changes.

datasets/_analysis/ann_graph_topics/analyzer.py
```diff
@@ -6,57 +8,72 @@
...
         if "sanction" in topics:
             if entity.schema.is_a("Security"):
                 continue
+            if adjacent.schema.is_a("Security"):
+                adj_topics = non_graph_topics(context, adjacent)
+                if adj_topics.intersection({"sanction", "sanction.linked"}):
+                    continue
+                emit_patch(context, entity, adjacent, "sanction.linked", adj_topics)
+            if not asch.edge:
+                continue
```
leonhandreke, line 76: Just for my understanding, this is redundant because we're explicitly checking for schemas which are all edges in the next line, right? Or is this important for magic further down?

datasets/_analysis/ann_graph_topics/analyzer.py
```diff
@@ -6,57 +8,72 @@
...
+def emit_patch(
+    context: Context,
+    risk_source: Entity,
+    related_entity: Entity,
+    topic: str,
+    existing_topics: Set[str],
+) -> None:
+    context.log.info(
+        f"Adding topic: {topic}",
+        source=risk_source.caption,
```
leonhandreke: (nit) I think the params have better names than the fields in this log message, why not stick with them? I think they're well-named.

datasets/_analysis/ann_graph_topics/analyzer.py
```diff
@@ -6,57 +8,72 @@
...
+    patch.id = related_entity.id
+    patch.add("topics", topic)
+    context.emit(patch)
+
+
 def analyze_entity(context: Context, view: View, entity: Entity) -> None:
     topics = entity.get_type_values(registry.topic)
     for prop, adjacent in view.get_adjacent(entity):
         asch = adjacent.schema
-        if prop.reverse is None or not asch.edge:
```
leonhandreke, line 13: I'm trying to figure out what the `prop.reverse` check did and don't even understand what it does :(

datasets/_analysis/ann_graph_topics/analyzer.py
```diff
@@ -6,57 +8,72 @@
...
-                "Skipping entity with end date",
-                adjacent=adjacent.id,
-                entity=entity.id,
+                "Skipping entity with end date", adjacent=adjacent.id, entity=entity.id
             )
             continue
 
+        # Tag sanction.linked for sanction-linked entities
         if "sanction" in topics:
             if entity.schema.is_a("Security"):
```
leonhandreke, line 68: I *think* this check is there so that `sanction` doesn't propagate up a security's ownership chain.

datasets/_analysis/ann_graph_topics/analyzer.py
```diff
@@ -6,57 +8,72 @@
...
                 continue
+            if adjacent.schema.is_a("Security"):
```
leonhandreke: Are there any other links than `issuer` that we want to get here? Maybe worth a comment what the intention here is, for the next time I'm trying to figure out what's actually going on here :)

datasets/_analysis/ann_graph_topics/analyzer.py
jbothma: I agree it's not not needed for correctness - all the logic further down handles it.
I think it was at the top of the loop for performance, and I kept it, still for the minor performance it might gain.

datasets/_analysis/ann_graph_topics/analyzer.py
jbothma: nuanced :)
https://github.com/opensanctions/opensanctions/issues/2969#issuecomment-3581711648

datasets/_analysis/ann_graph_topics/analyzer.py
jbothma: very good point - I'll check for the prop name as well - I think that will be nice and clear

datasets/_analysis/ann_graph_topics/analyzer.py
jbothma: hmmm yeah - as far as I can tell, every property with type `entity` has a reverse. Maybe it was an optimisation.

datasets/_analysis/ann_graph_topics/analyzer.py
jbothma: they felt a bit long, but they're pretty-printed in the log explorer so it matters much less and they can be clearer - thanks

PR #3151, opened 2025-11-26
Fixes-26-11

nvmbrasserie:


datasets/ng/nigsac_sanctions/ng_nigsac_sanctions.yml
```diff
@@ -48,6 +48,11 @@ data:
   format: HTML
 dates:
   formats: ["%m/%d/%Y %H:%M:%S %p"]
+http:
+  additional_retry_statuses: [443]
```
leonhandreke: Why 443? Isn't this just the HTTPS port? And why POST, isn't it failing on the GET right now?

nvmbrasserie: you're right, it looks like it's up. shall we just rerun it?

PR #3157, opened 2025-11-27
Apply fsf reviews

jbothma:
verified that eu_fsf and be_fod_sanctions provide full name when they provide first/last, but can't verify eu_travel_bans while sanctionsmap.eu is down, and it doesn't seem like overkill to be sure we include a full name.

Reverts multiple-string input support at the API level but not the prompt/model level because nothing's using it right now and I'm not sure what to pass as original value for multiple strings (but probably not the whole json sent with the prompt) - `", ".join(strings)` technically works, but I'm not sure if serialising it like that is ok and it's not a decision we need to make today..

image

### DIFF
--- a/zavod/zavod/shed/fsf.py
+++ b/zavod/zavod/shed/fsf.py
@@ -167,14 +167,17 @@ def parse_entry(context: Context, entry: Element) -> None:
                 lang=lang,
             )
         else:
+            if not full_name and (first_name and last_name):
+                # Currently full_name always exists with first and last, but just make sure.
+                full_name = h.make_name(
+                    first_name=first_name,
+                    middle_name=middle_name,
+                    last_name=last_name,
+                )
             h.apply_reviewed_names(context, entity, full_name, lang=lang)
             entity.add("firstName", first_name, quiet=True, lang=lang)
             entity.add("middleName", middle_name, quiet=True, lang=lang)
             entity.add("lastName", last_name, quiet=True, lang=lang)
-        if (first_name and last_name) and not full_name:
-            # Currently the source data has a full name for every entry with first+last,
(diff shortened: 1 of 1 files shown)


zavod/zavod/helpers/names.py
```diff
@@ -307,13 +306,12 @@ def apply_names(
                 name,
                 lang=lang,
                 origin=review.origin,
-                original_value=review.source_value,
+                original_value=string,
```
jbothma, line 309: here we were setting json like this as the source value.

When fixing that, I wasn't sure what to do with an array of input strings. But I don't think we need to handle multiple strings today, while we do want to ship this change today.

image

zavod/zavod/shed/fsf.py
```diff
@@ -151,19 +151,34 @@ def parse_entry(context: Context, entry: Element) -> None:
...
+        else:
+            if not full_name and (first_name and last_name):
+                # Currently full_name always exists with first and last, but just make sure.
+                full_name = h.make_name(
+                    first_name=first_name,
+                    middle_name=middle_name,
+                    last_name=last_name,
+                )
+            h.apply_reviewed_names(context, entity, full_name, lang=lang)
+            entity.add("firstName", first_name, quiet=True, lang=lang)
```
leonhandreke, line 178: Am I reading this right that only `full_name` is worth reviewing? Because we're applying first/middle/last in any case.

jbothma: 100%

jbothma: The idea with the reviews is to take things they're putting into full name, and move them to more appropriate fields if they aren't good full names. i.e. things that cause lots and lots of false positives
image

PR #3172, opened 2025-12-01
New crawler for consolidated Belgian sanctions

jbothma:
Belgium sanctions file has moved so the old file isn't getting updated.

### DIFF
--- a/datasets/be/fod_sanctions/crawler.py
+++ b/datasets/be/fod_sanctions/crawler.py
@@ -2,7 +2,6 @@ from csv import DictReader
 from typing import Dict
 from lxml import etree
 from zipfile import ZipFile
-import re
 
 from rigour.mime.types import CSV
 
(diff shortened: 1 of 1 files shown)


datasets/be/fod_sanctions/crawler.py
```diff
@@ -1,13 +1,112 @@
...
+# "03 01  118013 (other-Other identification number) (Passport number,  national ID  number, other numbers of  identity  documents: 03 01  118013)",
+# "0363464 (passport-National passport)  (issued by Palestinian Authority)"
+def apply_identifier(context: Context, entity: Entity, id_number_line: str):
+    parts = id_number_line.split("(")
+    prop = None
+
+    if len(parts) == 1:
+        prop = "idNumber"
+    if len(parts) == 2:
+        # Intentionally don't set the prop here
```
jbothma, line 36: not sure what I meant here. Fix tomorrow.

PR #3173, opened 2025-12-01
[be_fod_sanctions] Fix splitting birthplace before making id

jbothma:
Split values in one place, after making id.

### DIFF
--- a/datasets/be/fod_sanctions/crawler.py
+++ b/datasets/be/fod_sanctions/crawler.py
@@ -1,5 +1,5 @@
 from csv import DictReader
-from typing import Dict
+from typing import Dict, List
 from lxml import etree
 from zipfile import ZipFile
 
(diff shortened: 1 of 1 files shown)


datasets/be/fod_sanctions/crawler.py
```diff
@@ -47,47 +43,45 @@ def apply_identifier(context: Context, entity: Entity, id_number_line: str):
...
-        "programId",
-        h.lookup_sanction_program_key(context, row.pop("Embargos")),
-    )
+    for listing_date in row.pop("Publication date"):
+        h.apply_date(sanction, "listingDate", listing_date)
+    for program_id in row.pop("Embargos"):
+        sanction.add(
+            "programId",
+            h.lookup_sanction_program_key(context, program_id),
+        )
```
jbothma, line 84: naming

PR #3178, opened 2025-12-02
Review entity link

jbothma:
- sometimes one review results in multiple entities being created after acceptance.
- sometimes one entity can exist with multiple pending reviews, and it can be useful to see more about that entity, e.g. other notions of its name

This
- adds 
  - `zavod.stateful.review.delete_review_links()` to reset links for the dataset at the start of a crawl
  - `zavod.stateful.review.Review.link_entity()` per entity during a crawl
- Fixes the review content getting pushed off the page
- Updates tsconfig automatically done by next.js, I think

Questions
- should `zavod.crawl.crawl_dataset` just do the delete call? I don't love the random call that has to be added to each crawler using the name cleaning helper. We could also just not garbage collect these links for now.

Thoughts
- multiple entities can end up linked to the same review, e.g. if two entities have the same name. Maybe this is a feature. Maybe it's an early optimisation.

image

### DIFF
--- a/zavod/migrations/2025-12-02-review-entity-table.sql
+++ b/zavod/migrations/2025-12-02-review-entity-table.sql
@@ -1,0 +1,11 @@
+CREATE TABLE review_entity (
+    dataset VARCHAR(255) NOT NULL,
+    review_key VARCHAR(255) NOT NULL,
+    entity_id VARCHAR(255) NOT NULL,
+    last_seen_version VARCHAR(255) NOT NULL
+);
+
+CREATE INDEX ix_review_entity_review_key ON review_entity (review_key);
+CREATE INDEX ix_review_entity_last_seen_version ON review_entity (last_seen_version);
+CREATE INDEX ix_review_entity_dataset ON review_entity (dataset);
+CREATE UNIQUE INDEX ix_review_entity_unique_review_key_entity_id_dataset ON review_entity (review_key, entity_id, dataset);


ui/app/layout.tsx
```diff
@@ -27,13 +27,12 @@ export default function RootLayout({
       <body
         className={`${geistSans.variable} ${geistMono.variable} antialiased`}
       >
-        <div className="container-fluid d-flex flex-column pl-4 pr-4">
+        <div className="container-fluid d-flex flex-column pl-4 pr-4" style={{ height: '100vh' }}>
```
jbothma: maybe it's time to be less hacky and put this into scss

pudo:
I'd be +1 on moving the reset into the context. The big failure mode here is incomplete runs, but I feel like that's survivable :)

zavod/zavod/stateful/model.py
```diff
@@ -79,6 +79,23 @@
...
 
+review_entity_table = Table(
+    "review_entity",
+    meta,
+    # If this was review.id, versioned updates of reviews would need to update these entries.
+    Column("review_key", Unicode(KEY_LEN), nullable=False, index=True),
+    Column("entity_id", Unicode(KEY_LEN), nullable=False, index=True),
+    Column("dataset", Unicode(KEY_LEN), nullable=False, index=True),
+    Index(
+        "ix_review_entity_unique_review_key_entity_id_dataset",
```
leonhandreke, line 92: I can't quite keep all of this in my head - why is that index part of this index?

zavod/zavod/stateful/model.py
```diff
@@ -79,6 +79,23 @@
 )
 
 
+review_entity_table = Table(
+    "review_entity",
+    meta,
+    # If this was review.id, versioned updates of reviews would need to update these entries.
```
leonhandreke, line 85: But since we reset all the links anyway on every run, this wouldn't matter that much, right? It's just for my understanding, I don't have a strong opinion on this.

datasets/us/sam_exclusions/crawler.py
```diff
@@ -51,6 +53,7 @@ def crawl_data_url(context: Context) -> str:
 
 
 def crawl(context: Context) -> None:
+    reset_entity_links(context)
```
leonhandreke: When does this transaction get committed, i.e. will there be a time while the crawler is running when the links will be missing in the UI?

I'm wondering whether there are cases where we don't "touch" all reviews in a run (except for the failed run case). Would it be an option to clear them per-review when we touch that review? Or maybe we could add a `last_seen_version` like this the reviews itself and then somehow expire them? I haven't thought too deeply about this, just trying to think aloud.

zavod/zavod/stateful/model.py
jbothma: in a given dataset, the key and entity id should be unique together

zavod/zavod/stateful/model.py
jbothma: the review ui also inserts versioned entries which bump the id

datasets/us/sam_exclusions/crawler.py
jbothma: > When does this transaction get committed

The transaction is committed after 10k entities are emitted or at the end of the run.

> will there be a time while the crawler is running when the links will be missing in the UI?

can happen for big datasets like us_sam_exclusions - same as with reviews.

> Would it be an option to clear them per-review when we touch that review? Or maybe we could add a last_seen_version like this the reviews itself and then somehow expire them?

Actually using last_seen_version could work I think - I like that

zavod/zavod/stateful/model.py
leonhandreke: Ah, that makes sense!

PR #3183, opened 2025-12-03
Belgium Members of the Chamber of Representatives

nvmbrasserie:
opensanctions/crawlers#562

### DIFF
--- a/datasets/be/chamber/be_chamber.yml
+++ b/datasets/be/chamber/be_chamber.yml
@@ -1,5 +1,5 @@
 title: Belgium Members of the Chamber of Representatives
-prefix: be-chamber
+prefix: be-kam
 url: https://www.dekamer.be/kvvcr/showpage.cfm?section=/depute&language=fr&cfm=cvlist54.cfm?legis=56&today=y
 coverage:
   frequency: monthly


datasets/be/chamber/crawler.py
```diff
@@ -0,0 +1,120 @@
...
+from normality import squash_spaces
+
+from zavod import Context, helpers as h
+from zavod.extract import zyte_api
+from zavod.stateful.positions import categorise, EXTENDED_AFTER_OFFICE_YEARS
+
+CUTOFF_YEAR = datetime.now().year - EXTENDED_AFTER_OFFICE_YEARS
+
+
+def get_latest_terms(context: Context, term: str) -> tuple[int | None, int | None]:
```
leonhandreke: (nit) why latest?

datasets/be/chamber/crawler.py
```diff
@@ -0,0 +1,120 @@
...
+    res = context.lookup("term_dates", term, warn_unmatched=True)
+    if res:
+        return int(res.start_date), int(res.end_date)
```
leonhandreke: (nit) let's be more rigorous with the types here. If we want to pass around `int` years, let's do that, but call it `_years` and make it an `int` in the YAML. If we want dates, then let's pass around `date` objects and also make the cutoff a date object.

datasets/be/chamber/crawler.py
```diff
@@ -0,0 +1,120 @@
...
+    else:
+        return None, None
+
+
+def crawl_persons(
+    context: Context, row: HtmlElement, start_date: int, end_date: int | None
+) -> None:
+    cells = h.xpath_elements(row, ".//td[@class='td1' or @class='td0']")
+    name = h.xpath_strings(cells[0], ".//b/text()", expect_exactly=1)[0].strip()
```
leonhandreke, line 47: (nit) I created `xpath_string` this morning to do exactly this :)

datasets/be/chamber/crawler.py
```diff
@@ -0,0 +1,120 @@
...
+    profile_url = h.xpath_strings(cells[0], ".//a/@href", expect_exactly=1)[0]
+
+    group_texts = h.xpath_strings(cells[1], ".//a/text()")
+    political_group = group_texts[0].strip() if group_texts else ""
+    # Normalize whitespace in the name (different terms may include inconsistent spacing)
+    name = squash_spaces(name)
+
+    entity = context.make("Person")
+    # Use the normalized name as the stable identifier (it's the only consistent cross-term key)
```
leonhandreke: Could we make even more explicit why we're normalizing before building the ID? I think I get it, but a more explicit comment wouldn't hurt :)

datasets/be/chamber/crawler.py
```diff
@@ -0,0 +1,120 @@
...
+        absolute_links=True,
+        cache_days=4,
+    )
+    # Extract legislature terms from menu
+    terms: list = []
+    seen_urls: set[str] = set()
+    for link in doc.findall('.//div[@class="menu"]//a'):
+        url = link.get("href")
+        text = h.element_text(link).strip()
+        assert url, url
```
leonhandreke: (nit) what are we asserting here? `is not None`?

datasets/be/chamber/crawler.py
```diff
@@ -0,0 +1,120 @@
...
+
+        # Only keep entries with explicit date format "XX (YYYY-YYYY)"
+        # The first link "Actuels" (current members) duplicates the most recent term,
+        # so we filter for entries with parentheses to get unique dated terms.
+        if text and "(" in text and ")" in text and url not in seen_urls:
```
leonhandreke: Why do we need `seen_urls`? Are they duplicated?

datasets/be/chamber/crawler.py
```diff
@@ -0,0 +1,120 @@
...
+            seen_urls.add(url)
+            terms.append({"url": url, "text": text})
```
leonhandreke: Maybe we could just make this a single loop instead of first building the terms?

But if we build the terms first: let's use a `NamedTuple` instead of relying on these keys, it's a bit nicer ;)

datasets/be/chamber/be_chamber.yml
```diff
@@ -0,0 +1,67 @@
...
+entry_point: crawler.py
+summary: >-
+  Current and recent members of the Belgian Chamber of Representatives, the lower chamber of Parliament.
+description: |
+  This dataset provides information about the current and recent members of the Belgian Chamber of Representatives,
+  including their names, political affiliations, parliamentary terms, and other relevant information.
+tags:
+  - list.pep
+publisher:
+  name: Chamber of Representatives
```
leonhandreke: Let's settle on one of the official languages here (or both?) and put this in `name_en`.

leonhandreke (commented):
Nice! Looks like they don't have much info, but this crawler gets it all out :)

datasets/be/chamber/crawler.py
```diff
@@ -0,0 +1,129 @@
...
+        return
+
+    occupancy = h.make_occupancy(
+        context,
+        person=entity,
+        position=position,
+        start_date=str(start_date),
+        end_date=str(end_date) if end_date else None,
+        categorisation=categorisation,
+    )
```
jbothma, line 80: the dates indicate the legislative term, not the term of office. In the past we've avoided using the term of office as the occupancy dates because it kinda misrepresents when someone was in office - better to say nothing than the wrong thing. 

Have you guys discussed being more relaxed on that in the office? 

Otherwise I think we should adjust the logic so that we don't put dates in the occupancy, we just give an explicit Status.CURRENT to everyone on the Actuels page and an occupancy with Status.ENDED to everyone on the earlier legislative term pages. That also means removing the logic to skip the Actuals page.

I don't see any data that matches the actual date they start/end their term of office other than the free text blurb, which we don't want to try and parse. But that's why we have the option of passing an explicit status based on logic in the crawler.
image

datasets/be/chamber/crawler.py
```diff
@@ -0,0 +1,129 @@
...
+    context: Context, row: HtmlElement, start_date: int, end_date: int | None
+) -> None:
+    cells = h.xpath_elements(row, ".//td[@class='td1' or @class='td0']")
+    name = h.xpath_strings(cells[0], ".//b/text()", expect_exactly=1)[0].strip()
+    profile_url = h.xpath_strings(cells[0], ".//a/@href", expect_exactly=1)[0]
+
+    group_texts = h.xpath_strings(cells[1], ".//a/text()")
+    political_group = group_texts[0].strip() if group_texts else ""
+    # Normalize whitespace in the name (different terms may include inconsistent spacing)
+    name = squash_spaces(name)
```
jbothma: I'm inclined to say let's call strip() on the value but not squash_spaces for now. Dedupe should find the matches but as a rule, we don't do normalisation on the input to make_id.

datasets/be/chamber/crawler.py
```diff
@@ -0,0 +1,129 @@
...
+
+    entity = context.make("Person")
+    # Use the normalized name as the stable identifier. We don't have consistent unique identifiers
+    # (like national IDs or member numbers) across all legislature terms, so the normalized name
+    # is our only reliable way to link the same individual across multiple terms and avoid duplicates
+    entity.id = context.make_id(name)
```
jbothma: it looks like we can fairly reliably pull out the date of birth, which really helps with dedupe (and distinguishing distinct people with the same name!)

Could you try splitting on `|` then best effort finding the value containing `Né à`, h.multisplitting on `Né à` and ` le `, then you've got a birth place and name.

Then we can use the birth date in the id

image

datasets/be/chamber/crawler.py
```diff
@@ -0,0 +1,129 @@
...
+from zavod.extract import zyte_api
+from zavod.stateful.positions import categorise, EXTENDED_AFTER_OFFICE_YEARS
+
+
+class LegislatureTerm(NamedTuple):
+    url: str
+    text: str
+
+
+CUTOFF_YEAR = datetime.now().year - EXTENDED_AFTER_OFFICE_YEARS
```
jbothma: perhaps a bit more future-proof would be

```suggestion
POSITION_TOPICS = ["gov.legislative", "gov.national"]
CUTOFF_YEAR = datetime.now().year - get_after_office(POSITION_TOPICS)
```

datasets/be/chamber/be_chamber.yml
```diff
@@ -0,0 +1,83 @@
+title: Belgium Members of the Chamber of Representatives
+prefix: be-chamber
```
pudo: maybe `be-kam`? Having these short is neat :)

datasets/be/chamber/be_chamber.yml
```diff
@@ -0,0 +1,83 @@
...
+      Position: 1
+    country_entities:
+      be: 1
+
+lookups:
+  term_years:
+    options:
+      - match: "56 (2024-2025)"
+        start_date: 2024
+        end_date: 2025
```
pudo, line 59: I weirdly like this lookup table

datasets/be/chamber/crawler.py
```diff
@@ -0,0 +1,134 @@
+from datetime import datetime
+from lxml.html import HtmlElement
+
+from zavod import Context, helpers as h
+from zavod.extract import zyte_api
+from zavod.stateful.positions import categorise, get_after_office, OccupancyStatus
+
+
+POSITION_TOPICS = ["gov.legislative", "gov.national"]
+CUTOFF_DATE = (datetime.now() - get_after_office(POSITION_TOPICS)).year
```
pudo, line 10: haha this is fucking clever. Maybe later we want to make this a helper in `positions` :)

nvmbrasserie: just filed an issue for it here: [#3252](https://github.com/opensanctions/opensanctions/issues/3252)

Issue #1435, opened 2024-10-23
Date comparison coherence


pudo:
Problem:
* Wikidata dates can be negative, or really small (0927-01-01). 
* We're comparing in: 
  * https://github.com/opensanctions/opensanctions/blob/main/datasets/_wikidata/peps/crawler.py#L60-L66 
  * https://github.com/opensanctions/opensanctions/blob/main/zavod/zavod/logic/pep.py#L143-L222 
  * https://github.com/opensanctions/opensanctions/blob/main/zavod/zavod/shed/wikidata/human.py#L38-L53 
  * https://github.com/opensanctions/opensanctions/blob/main/zavod/zavod/shed/wikidata/position.py#L135-L139 
 
What are the functions we should have? 

* `years_since(xxx)`

pudo:
I think the concern was a bit that the string comparisons may 

a. not work correctly on all possible date strings (e.g. 0198-00) - probably not true 
b. we're doing a lot of math on these strings that may want to be extracted into utility functions

It's neither urgent nor important, and if we cannot think of meaningful actions I guess we can close this ticket.

PR #3190, opened 2025-12-05
Typing unicorns: Make PEP cooldown periods timedeltas

leonhandreke:


pudo:
Well if you're going to open that sarcophacus, why not have a deeper look: https://github.com/opensanctions/opensanctions/issues/1435 :)

PR #3201, opened 2025-12-08
[be_fod_sanctions] Fix all entities as Person

### LINKED ISSUE #3195: [be_fod] Company emitted as Person
I imagine it's a data entry error, but we should override: 

Image

jbothma:
Fixes #3195

The issue was that the value splitting on `\n` was resulting in lists like `[""]` being passed to add_cast, which then adds the `Person` schema before the empty value is dropped by `add`.

This

- Uses the `type` column to determine schema - it turns out that's what `P` means
- changes the line splitting to h.multisplit which drops empty values
- Lets `audit_data` treat empty list as an empty value. The value argument to `add` includes iterable after all.

### DIFF
--- a/datasets/be/fod_sanctions/crawler.py
+++ b/datasets/be/fod_sanctions/crawler.py
@@ -75,9 +75,9 @@ def crawl_row(context: Context, entity_id: str | None, row: Dict[str, List[str]]
         h.apply_dates(entity, "birthDate", row.pop("Birth date"))
         entity.add("position", row.pop("Function"))
     else:
-        entity.add("name", row.pop("Firstname"))
-        # Organization with name
+        # There's an organization with the following Firstname
         # "صنایع شهید ستاری"
+        entity.add("name", row.pop("Firstname"))
         entity.add("notes", row.pop("Function"))
         row.pop("Gender")  # Rustecdrone is male, apparently.
 


datasets/be/fod_sanctions/crawler.py
```diff
@@ -37,43 +37,50 @@ def apply_identifier(context: Context, entity: Entity, id_number_line: str):
...
+        entity.add("firstName", row.pop("Firstname"))
+        entity.add("middleName", row.pop("Middlename"))
+        entity.add("gender", row.pop("Gender"))
+        entity.add("birthPlace", row.pop("Birth place"))
+        entity.add("country", row.pop("Birth country"))
+        h.apply_dates(entity, "birthDate", row.pop("Birth date"))
+        entity.add("position", row.pop("Function"))
+    else:
+        entity.add("name", row.pop("Firstname"))
+        # Organization with name
```
leonhandreke: What does this comment mean?

leonhandreke: Ah, probably that this one Organization has a Function as well?

jbothma: It helps if the comment is above the row it's about. But also clarified a bit - thanks.

PR #3203, opened 2025-12-08
[gb_fcdo_sanctions] parse csv

nvmbrasserie:


datasets/gb/fcdo_sanctions/crawler.py
```diff
@@ -233,3 +256,221 @@ def crawl(context: Context):
...
+            context.emit(entity, external=True)
+            context.emit(sanction, external=True)
+
+            context.audit_data(
+                row,
+                [
+                    "National Identifier additional information",
+                    "Non-latin script type",
+                    "Business registration number (s)",
+                    "Alias strength",
```
pudo: Can we use this to qualify things as `weakAlias` in some cases?

datasets/gb/fcdo_sanctions/crawler.py
```diff
@@ -233,3 +256,221 @@ def crawl(context: Context):
...
+
+            context.emit(entity, external=True)
+            context.emit(sanction, external=True)
+
+            context.audit_data(
+                row,
+                [
+                    "National Identifier additional information",
+                    "Non-latin script type",
+                    "Business registration number (s)",
```
pudo: Maybe goes into `registrationNumber`?

pudo (approved):
I like this and it looks well-implemented to me. Let's merge when you've had a look at these two and then we'll dedupe-run to see if we've missed anything :)

PR #3214, opened 2025-12-10
Kingdom of Jordan National Sanctions List

brrttwrks:
opensanctions/crawlers#250

### DIFF
--- a/datasets/jo/jo_sanctions/crawler.py
+++ b/datasets/jo/jo_sanctions/crawler.py
@@ -27,7 +27,7 @@ def xlsx_check(doc: HtmlElement) -> bool | str:
     for link in doc.iterlinks():
         ext = link[2].split(".")[-1]
         if "xlsx" in ext:
-            return link[2]
+            return str(link[2])
     return False
 
 
@@ -59,7 +59,7 @@ def crawl_person(context: Context, path: Path) -> None:
         person.add("idNumber", national_id)
         person.add("nationality", row.pop("nationality"))
         h.apply_date(person, "birthDate", birth_date)
-        person.add("sourceUrl", context.dataset.data.url)
+        person.add("sourceUrl", context.data.url)
 
         address_ent = h.make_address(
             context,
@@ -106,7 +106,7 @@ def crawl_legal_entities(context: Context, path: Path) -> None:
         context.audit_data(row, AUDIT_FIELDS)
 
 
-def crawl(context: Context):
+def crawl(context: Context) -> None:
     doc = context.fetch_html(context.data_url, cache_days=5)
     url = xlsx_check(doc)
     if url is False:


datasets/jo/jo_sanctions/crawler.py
```diff
@@ -0,0 +1,127 @@
...
+        h.apply_date(sanction, "startDate", included_date)
+        sanction.add("sourceUrl", context.dataset.data.url)
+        sanction.add("entity", person)
+
+        context.emit(person)
+        context.emit(sanction)
+
+        context.audit_data(row, AUDIT_FIELDS)
+    
+def crawl_legal_entity(context: Context, path: Path) -> None:
```
pudo: I would pluralize this: we often also have functions that crawl just one person or entity, that's what I first thought this was.

datasets/jo/jo_sanctions/crawler.py
```diff
@@ -0,0 +1,127 @@
...
+        skiprows=1,
+        header_lookup=context.get_lookup("columns"),
+    ):
+        
+        full_name_en = row.pop("full_name_en")
+        included_date = row.pop("included_date")
+        
+        legalent = context.make("LegalEntity")
+        legalent.id = context.make_id(full_name_en)
+        sanction = context.make("Sanction")
```
pudo: the sanction should be instantiated via a helper (`h.make_sanction`) which sets lots of default fields and generates an ID.

datasets/jo/jo_sanctions/crawler.py
```diff
@@ -0,0 +1,127 @@
...
+        person.id = context.make_id(
+            national_id,
+            fullname_ar, 
+            birth_date,
+        )
+        sanction = context.make("Sanction")
+        sanction.id = context.make_id("jo", fullname_ar, included_date)
+
+        person.add("name", fullname_ar, lang="ara")
+        person.add("name", fullname_en, lang="eng")
```
pudo, line 57: nice, `lang`!

datasets/jo/jo_sanctions/crawler.py
```diff
@@ -0,0 +1,127 @@
...
+        country = row.pop("country")
+
+        person = context.make("Person")
+        person.id = context.make_id(
+            national_id,
+            fullname_ar, 
+            birth_date,
+        )
+        sanction = context.make("Sanction")
+        sanction.id = context.make_id("jo", fullname_ar, included_date)
```
pudo: see above: helper

datasets/jo/jo_sanctions/jo_sanctions.yml
```diff
@@ -0,0 +1,283 @@
+name: jo_sanctions
```
pudo: Not needed, inferred from file name

datasets/jo/jo_sanctions/crawler.py
brrttwrks: Agree about the countries. I went a bit overboard at first, but now that I understand how it uses `rigor.territories.lookup` first and then falls back to crawler specific lookups, I've removed almost all of those and just kept the offending ones after debugging with `zavod --debug crawl`.

datasets/jo/jo_sanctions/crawler.py
brrttwrks: Ok. That helper makes it a bit easier and cleaner. Thanks!

datasets/jo/jo_sanctions/jo_sanctions.yml
```diff
@@ -0,0 +1,110 @@
+title: "Kingdom of Jordan National Sanctions List"
```
jbothma, line 1: we tend to name the directory after the bit after the country prefix, so you can rename the path to be `jo/sanctions/jo_sanctions.yml`

datasets/jo/jo_sanctions/crawler.py
```diff
@@ -0,0 +1,121 @@
...
+def crawl(context: Context):
+    doc = context.fetch_html(context.data_url, cache_days=5)
+    url = xlsx_check(doc)
+    if url is False:
+        msg = "XLSX file does not exist!"
+        context.log.error(msg)
+        raise MissingDataFile(msg)
+
+    path = context.fetch_resource("lists.xlsx", url)
+    context.export_resource(path, XLSX, title=context.SOURCE_TITLE)
```
jbothma, line 118: small nitpick: we tend to name the source file `source.{extension}` rather than `list`, unless there are more than one resource and we want to distinguish by name.

datasets/jo/jo_sanctions/crawler.py
```diff
@@ -0,0 +1,121 @@
...
+        h.apply_date(sanction, "startDate", included_date)
+
+        context.emit(legalent)
+        context.emit(sanction)
+
+        context.audit_data(row, AUDIT_FIELDS)
+
+
+def crawl(context: Context):
+    doc = context.fetch_html(context.data_url, cache_days=5)
```
jbothma, line 110: for sanctions it can be nice not to cache a page like, so if an update changes the link to the PDF, we get that new PDF on the next crawl.

datasets/jo/jo_sanctions/crawler.py
```diff
@@ -0,0 +1,121 @@
...
+        legalent.add("name", full_name_en, lang="eng")
+        legalent.add("classification", row.pop("classification"), lang="ara")
+
+        sanction = h.make_sanction(context, legalent)
+        h.apply_date(sanction, "startDate", included_date)
+
+        context.emit(legalent)
+        context.emit(sanction)
+
+        context.audit_data(row, AUDIT_FIELDS)
```
jbothma, line 106: how about
```suggestion
        context.audit_data(row, IGNORE_FIELDS)
```

datasets/jo/jo_sanctions/crawler.py
```diff
@@ -0,0 +1,121 @@
...
+    if url is False:
+        msg = "XLSX file does not exist!"
+        context.log.error(msg)
+        raise MissingDataFile(msg)
+
+    path = context.fetch_resource("lists.xlsx", url)
+    context.export_resource(path, XLSX, title=context.SOURCE_TITLE)
+
+    crawl_person(context, path)
+    crawl_legal_entities(context, path)
```
jbothma, line 121: It could be nice to open the excel file here, access the worksheet by name, and pass it in to each of these functions. Then the logic is a bit more tied to the worksheet name.

The other thing that's nice is to assert that these are the only sheets, e.g. by asserting there are only two sheets. or logging a warning, or whatever. Then we get told when they add another sheet so we don't miss any new data, e.g. if they start sanctioning ~vessels~ (ok jordan is landlocked but you get the idea).

datasets/jo/jo_sanctions/crawler.py
```diff
@@ -0,0 +1,121 @@
...
+from zavod import helpers as h
+from lxml.html import HtmlElement
+from openpyxl import load_workbook
+from rigour.mime.types import XLSX
+
+AUDIT_FIELDS = [
+    "row_no",
+    "mother_name",
+    "birth_place",
+    "first_and_surname",
```
jbothma, line 12: Looks like it's worth including this, splitting on `،`.  These probably belong in the `alias` prop
image

datasets/jo/jo_sanctions/crawler.py
```diff
@@ -0,0 +1,121 @@
...
+    "country",
+    "doc_type",
+    "doc_no",
+    "doc_issue_auth",
+    "doc_issue_date",
```
jbothma, line 17: these can be added with `h.make_identification` which returns an https://www.opensanctions.org/reference/#schema.Identification and adds the number to the appropriate prop.

You can use a lookup to determine whether it's a passport with something like `is_passport = context.lookup_value("identification_type", doc_type, False, warn_if_unmatched=True)`

datasets/jo/jo_sanctions/crawler.py
```diff
@@ -0,0 +1,121 @@
...
+from lxml.html import HtmlElement
+from openpyxl import load_workbook
+from rigour.mime.types import XLSX
+
+AUDIT_FIELDS = [
+    "row_no",
+    "mother_name",
+    "birth_place",
+    "first_and_surname",
+    "country",
```
jbothma, line 13: any time we're not using this?

jbothma: birthplace and description can also happily be added to the entities on the props by the same name

datasets/jo/jo_sanctions/jo_sanctions.yml
```diff
@@ -0,0 +1,110 @@
...
+url: https://sanctions.amlu.gov.jo/EN/List/National1_List
+load_statements: true
+coverage:
+  frequency: weekly
```
jbothma, line 5:
```suggestion
  frequency: daily
```
This can run daily

datasets/jo/jo_sanctions/jo_sanctions.yml
brrttwrks: done

datasets/jo/jo_sanctions/crawler.py
brrttwrks: done

datasets/jo/jo_sanctions/crawler.py
brrttwrks: Got it. Done.

datasets/jo/jo_sanctions/crawler.py
brrttwrks: Why is IGNORE_FIELDS vs AUDIT_FIELDS? I thought the idea was to be able to see when a new field was added in the logs?

datasets/jo/jo_sanctions/crawler.py
brrttwrks: Appreciate the assertion! I've done this.

datasets/jo/jo_sanctions/crawler.py
brrttwrks: Ok. Agreed. Done.

datasets/jo/jo_sanctions/crawler.py
brrttwrks: Ok. I understand now. Done.

PR #3245, opened 2025-12-17
[eu_travel_bans] add program mappings

nvmbrasserie:


zavod/zavod/shed/fsf.py
```diff
@@ -50,8 +50,20 @@ def parse_address(context: Context, el: Element) -> Optional[Entity]:
     )
 
 
-def parse_sanctions(context: Context, entity: Entity, entry: Element) -> None:
+def parse_sanctions(
+    context: Context,
+    entity: Entity,
+    entry: Element,
+    program_key_override: str | None = None,
```
pudo: can we just set the default value here to `programme`? I'd rename the variable to `program_attrib` (attributes in the XML sense), and then the implementation would just be `regulation.get(program_attrib)`.

zavod/zavod/shed/fsf.py
```diff
@@ -50,8 +50,20 @@ def parse_address(context: Context, el: Element) -> Optional[Entity]:
...
+def parse_sanctions(
+    context: Context,
+    entity: Entity,
+    entry: Element,
+    program_attrib: str = "programme",
+) -> None:
     regulations = entry.findall("./regulation")
+    """
+    Args:
+        program_key_override: Optional XML attribute name to use instead of 'programme'
```
pudo: Let's rename, then it's good to go!

PR #3246, opened 2025-12-17
[tr_fcib] fixes

brrttwrks:
I think the culprit for the mysterious anthropomorphization of companies was due to a mismatch header mapping and the use of row.pop("birth_date", "") with a string fallback, followed by a condition checking for that value to be True, which it always was due to the fallback.


brrttwrks:
> Haha, you're trying to sneak in Jordan via the side channel, hm?

shoot. still getting used to using git via a gui. i really prefer the cli, tbh. but i'll learn ;)

PR #3247, opened 2025-12-17
Issues-17-12

nvmbrasserie:


datasets/es/parliament/crawler.py
```diff
@@ -154,18 +152,14 @@ def crawl_senator(context: Context, senator_url: str) -> bool:
...
-    first_names = h.element_text(
-        h.xpath_elements(datos, "nombre", expect_exactly=1)[0]
-    ).split(" ", 1)
+    first_name = h.element_text(h.xpath_elements(datos, "nombre", expect_exactly=1)[0])
     last_name = datos.findtext("apellidos")
 
     person = context.make("Person")
-    # We ignore the type error here because we don't want to rekey
-    person.id = context.make_id(web_id, first_names, last_name)  # type: ignore
+    person.id = context.make_id(web_id, first_name, last_name)
```
nvmbrasserie: It will rekey senators so I'll run the dedupe once this PR is in

datasets/de/bundesrat/crawler.py
```diff
@@ -9,6 +9,13 @@
 from zavod import Context
 from zavod import helpers as h
 
+# NOTE: Bundesrat members are regional executives (Minister-Presidents, Senators,
+# etc.) who hold their Bundesrat seat by virtue of their regional office. We label
+# all as "Mitglied des Bundesrates" as suggested by the website, since this is the
+# federal legislative role relevant for PEP tracking. Status is set to UNKNOWN
+# because members alternate between deputy and full member roles in ways not clearly
+# documented on the source website.
```
jbothma: We now know how to easily determine whether they're a full or deputy member so I think it's an easy fix. The point isn't whether they're PEP, the point is that these are different positions and someone won't necessarily ever be a full member.

So we're currently being a bit inaccurate in calling them all full members. If we're being inaccurate, let the user know why (in the metadata, not the python code). But now that we know how to be accurate quite easily, let's be accurate.

datasets/es/parliament/crawler.py
jbothma: can we do the re-key with the external emit approach we've been discussing? i.e.

1. emit the entity with old key as before; emit the entity with new key as external
2. dedupe the dataset on its own
3. emit the entity with the new key (confirm locally with up to date resolver that entities with canonical IDs have kept them before shipping)

datasets/us/bis_denied/crawler.py
```diff
@@ -37,7 +37,7 @@ def parse_row(context: Context, row):
     sanction = h.make_sanction(context, entity, key=citation, program_key=PROGRAM_KEY)
     sanction.add("program", citation)
     h.apply_date(sanction, "startDate", row.pop("effective_date"))
-    h.apply_date(sanction, "endDate", row.pop("expiration_date"))
+    h.apply_date(sanction, "endDate", row.pop("expiratin_date"))
```
jbothma: ermahgerd

Just a suggestion to save us changing this back next week: Maybe we can try the proper spelling, and if that's unsuccessful, try the redneck spelling

PR #3282, opened 2025-12-30
Re-key statements in context.emit if we mutate their state

pudo:


zavod/zavod/context.py
```diff
@@ -566,17 +566,28 @@ def emit(
             self.flush()
         stamps = {} if self.dry_run else self.timestamps.get(entity.id)
         for stmt in entity.statements:
-            if stmt.id is None:
-                self.log.warn("Statement has no ID", stmt=stmt.to_dict())
-                continue
+            rekey = stmt.id is None
```
jbothma: nitpick: might be worth naming it something like `re_id` or `rehash` or something since rekey has such a significant different meaning in the system

pudo: Fair point!

PR #3283, opened 2025-12-30
Stop mutating nested values in logged event dicts

jbothma:
- Stop mutating nested values in logged event dicts. Structlog copies the root dict but we go deeper. The mutation affects e.g. debug-logged zyte request objects (redacted war sanctions api URL)
- Serialize all values except Context before redacting so that e.g. exception objects would also be redacted. Let `log_issue` pop context.
- use zyte for UA war sanctions API since our request from k8s is blocked upon first request

I tried all kinds of ideas to keep the redactor unaware of the context type, but needing access to a context instance after redaction makes that so tricky, so I think making it explicit is reasonable.

It doesn't sound like importing in the redactor is a massive performance concern but I haven't done big benchmarks. Shout if it's horrid.


zavod/zavod/runtime/issues.py
```diff
@@ -35,20 +34,8 @@ def write(self, event: Dict[str, Any]) -> None:
...
-                value = {
-                    "id": value.id,
-                    "caption": value.caption,
-                    "schema": value.schema.name,
-                }
-            if isinstance(value, set):
-                value = list(value)
-            data[key] = value
 
+        data = dict(event)  # copy
```
pudo: Why not `deepcopy`?

zavod/zavod/logs.py
```diff
@@ -201,9 +219,19 @@ def format_json(_: Any, __: str, ed: Event) -> Event:
...
+    """
+    Ensure that all types are JSON-serializable and redactable,
+    converting everything to string, list or dict.
 
+    Assumes it will be called recursively on list and dict values.
+    """
+    if isinstance(value, (str, dict, list)):
+        # The redactor will recurse into these
+        return value
     if isinstance(value, (_Element, HtmlElement)):
```
pudo: Are these two different things?

pudo (commented):
Honestly: I don't quite understand what's happening here. Is there a world where we want to make the log redactor less integrated with structlog? Can we have a single function that sees all log messages and dictifies AND redacts them?

zavod/zavod/runtime/issues.py
jbothma: This is just a comment added to an existing dict() call - the copy seems to prevent affecting downstream processors after popping fields in this function so no need for deepcopy.

jbothma:
The benefit of it being a distinct processor from log_issue is that it's clearer that all log output (write_issue, and stdout/err) will be redacted.

What's the downside of being a distinct structlog processor? It's not ideal, but I don't think it's entirely unusual for earlier processors to be aware of the needs of processors more than 1 step down the chain.

The state at the moment is that both write_issue and redaction want the data to be "dictified", except that write_issue wants a context instance, when one is available).

Of course if we're happy with unredacted logs in GPC logs, we can just to the dictification/redaction as part of write_issue.

zavod/zavod/logs.py
jbothma: yeah! Thanks

zavod/zavod/runtime/issues.py
jbothma: ah but deepcopy is the answer - I forgot it's its own thing handling more mutables than just dict.

jbothma: we could deepcopy in logs.py but we need to recurse to stringify anyway. In this context I think the dict() call was still needed for its downstream processors

zavod/zavod/logs.py
```diff
@@ -64,17 +84,21 @@ def redact_dict(self, dict_: Event) -> Event:
...
             if isinstance(value, dict):
                 value = self.redact_dict(value)
-            if isinstance(value, str):
+            elif isinstance(value, str):
                 value = self.redact_str(value)
-            if isinstance(value, list):
+            elif isinstance(value, list):
                 value = self.redact_list(value)
+            else:
+                value = self.redact_str(value)
```
jbothma, line 101: these are the key safety bits, ensuring we treat everything else as string and fail if it isn't one.

pudo: That makes a ton of sense.

zavod/zavod/logs.py
```diff
@@ -45,16 +44,37 @@
...
+                has_context_instance = True
+            # else who knows what it is - redact it as normal.
+
+        event_copy = issue_event_value(event_dict)
+        redacted_event = self.redact_dict(event_copy)
+
+        # Set it if and only if it existed before.
+        if has_context_instance:
+            redacted_event["context"] = context
+        return redacted_event
```
pudo, line 77: https://www.youtube.com/watch?v=suai1lU4x-g

zavod/zavod/logs.py
```diff
@@ -28,6 +26,7 @@
     "EU_JOURNAL_SEEN_PATH",
     "ZAVOD_DATA_PATH",
     "ZAVOD_OPENSANCTIONS_API_URL",
+    "GITHUB_JOB",
```
pudo, line 29: lol why :)

pudo (approved):
Honestly this looks great. If we fuck it up again with a leak, we'll just move to the wilderness.

zavod/zavod/logs.py
jbothma: image

🙃

Issue #2985, opened 2025-10-23
Another Arabic caption despite an English name


leonhandreke:
<img width="1328" height="1138" alt="Image" src="https://github.com/user-attachments/assets/a4dd69e3-f8f4-4a4e-8643-de2855b31f15" />

https://www.opensanctions.org/statements/Q47526/

pudo:
FWIW there's still this fix for nomenklatura that will feed into the Wikidata enricher, which is unreleased: https://github.com/opensanctions/nomenklatura/commit/7222f07003cccdfd96324e5b24ff8560cf5e5263 

I'm also working on a fix in rigour that should relate to the US SAM issue reported by a customer recently (I think the fact that whitespace was counted as non-latin text made the thing bias against multi-word names):
https://github.com/opensanctions/rigour/commit/f915e9eccddfa01c3e36ed62c9390b31fe116d01

pudo:
The fix obviously didn't work, argh.

Issue #3241, opened 2025-12-16
Re-key statement IDs


pudo:
At the moment we continue having a problem with code where two statements are emitted with the same value but different language codes:

```python
entity.add("Jeff", lang="eng")
entity.add("Jeff", lang="deu")
```

This behaviour is not well defined: both statements are written to the export CSV, but may overwrite each other eg. in the LevelDB store and in the SQL table. That's because they're considered to have the same checksum. My proposal therefore: let's add `lang` to the key part of the statement ID. 

Theoretically, we may also want to add `original_value` and `origin` - the last bits of the statement that may distinguish it from another. I can see a stronger case for `origin` than for `original_value`, the latter just doesn't seem ... important enough. But perhaps `lang` is good enough for now. 

### Migration 

The problem is that this will mess up the `first_seen` timestamps in `zavod`. So I think our process needs to be:

* Make a version of `followthemoney` with the new hash function
* Before adopting it, add code to `zavod.runtime.timestamps:TimeStampIndex.index` to compute the previous hash and index that as well.
* Deploy and wait a month or two
* Remove the old hash from the timestamper
* Carry on with life

pudo:
No fallout from this, phew.

PR #3288, opened 2025-12-31
Re-key statements and make them immutable

pudo:
See #3241 and https://github.com/opensanctions/followthemoney/pull/188


pudo:
Should also fix https://github.com/opensanctions/opensanctions/issues/2985

PR #3327, opened 2026-01-09
[gb_fcdo_sanctions] Only demote weak types to weakAlias if it's an alias

jbothma:
We treat blank "alias strength" as "weak", but it's always blank when name type isn't an alias, but rather a primary name. So we were sending primary names from the CSV to weakAlias incorrectly. The name consolidation them moved lots of matching primary names to weakAlias in collections including this dataset. (edit: the name demotion was limited to names shorter than 15 chars and single token names)

Expect deltas of up to 5720

```
Delta export complete          [gb_fcdo_sanctions] added=8 dataset=gb_fcdo_sanctions deleted=0 metric=delta_counts modified=5720 version=20260109134533-jmi
```


pudo:
Argh, not great. I worked on this, crap... Please lets get the fix rolled out ASAP.

PR #3337, opened 2026-01-14
[gb_fcdo_sanctions] Selectively treat blank alias strength as weak/ab…

jbothma:
…breviation

Previously we'd treat all blank alias strength rows as weak aliases. Some are indeed not great, but there are real names in there like "Boko Haram".

This changes the behaviour so that some obviously bad aliases move to weakAlias or abbreviation as needed, and the rest remain in `alias`.

```
Delta export complete          [gb_fcdo_sanctions] added=101 dataset=gb_fcdo_sanctions deleted=77 metric=delta_counts modified=638
```

The adds/deletes should be my out of date resolver.


pudo:
Oh that's a nice delta, much smaller than anticipated

PR #3349, opened 2026-01-19
US bis export

herrhellana:
fixes https://github.com/opensanctions/crawler-planning/issues/568

### DIFF
--- a/datasets/us/bis_export/us_bis_export.yml
+++ b/datasets/us/bis_export/us_bis_export.yml
@@ -46,18 +46,6 @@ assertions:
       LegalEntity: 2200
 
 lookups:
-  # exempt legitimate company names containing "&" from HTML/XSS false positives
-  type.name:
-    options:
-      - match: "M&M Wireless, Inc."
-      - match: "MIC P&I, LLC"
-      - match: "R&A International Trading, Inc"
-      - match: "R&A International Logistics"
-      - match: "B&H Foto & Electronics Corp."
-      - match: "P&P Computers"
-      - match: "ST&T"
-      - match: "G&W International Forwarders"
-      - match: "B&S Aircraft Alloys, Inc."
   comma_names:
     options:
       - match: 'Applied Materials, Inc. and Applied Materials Korea'


datasets/us/bis_export/crawler.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def crawl(context: Context) -> None:
+    # context.data_url with validation <== context.dataset.data.url without validation
+    doc = context.fetch_html(context.data_url, cache_days=1, absolute_links=True)
+    table = h.xpath_elements(doc, ".//table", expect_exactly=1)
+
+    for row in h.parse_html_table(table[0]):
+        str_row = h.cells_to_str(row)
+        case_id = row.get("case_id")
```
pudo: We could put the `case_id` into the sanctions object we generate below. For example, the `authorityId` prop which might be suitable.

datasets/us/bis_export/crawler.py
```diff
@@ -0,0 +1,35 @@
...
+        assert case_id is not None
+        url = h.xpath_elements(case_id, ".//a")[0].get("href")
+
+        for name in h.multi_split(
+            str_row.pop("case_name"), ";"
+        ):  # case-level names, might contain multiple entities
```
pudo: I would make this two lines to have it more readable: 

```
case_name = str_row.pop('case_name')
for name in h.multi_split(case_name, ';'):
```

datasets/us/bis_export/us_bis_export.yml
```diff
@@ -0,0 +1,41 @@
...
+coverage:
+  frequency: weekly
+  start: "2026-01-19"
+load_statements: true
+summary: >
+  List of violations of export control and antiboycott regulations.
+description: |
+  Export Enforcement protects and promotes U.S. national security by
+  investigating violations of export control and antiboycott regulations and
+  by partnering with industry and academia to facilitate compliance with those regulations.
```
pudo: Maybe our friend Claude can help us make this a bit more detailed? :)

pudo (commented):
Very nice work! My comments are entirely cosmetic.

datasets/us/bis_export/crawler.py
```diff
@@ -0,0 +1,35 @@
...
+                name, str_row.get("case_id")
+            )  # use raw name strings to generate IDs
+            entity.add("alias", [alias.strip().replace(",", "") for alias in aliases])
+            entity.add("name", entity.get("alias")[0])
+            entity.add("topics", "reg.warn")  # or export.control
+            entity.add("sourceUrl", url)
+
+            sanction = h.make_sanction(context, entity)
+            h.apply_date(
+                sanction, "listingDate", str_row.get("order_date")
```
nvmbrasserie:
```suggestion
                sanction, "listingDate", str_row.pop("order_date")
```

datasets/us/bis_export/crawler.py
```diff
@@ -0,0 +1,35 @@
...
+
+def crawl(context: Context) -> None:
+    # context.data_url with validation <== context.dataset.data.url without validation
+    doc = context.fetch_html(context.data_url, cache_days=1, absolute_links=True)
+    table = h.xpath_elements(doc, ".//table", expect_exactly=1)
+
+    for row in h.parse_html_table(table[0]):
+        str_row = h.cells_to_str(row)
+        case_id = row.get("case_id")
+        assert case_id is not None
```
nvmbrasserie:
```suggestion
        str_row = h.cells_to_str(row)
        case_id = row.pop("case_id")
        assert case_id is not None
```

datasets/us/bis_export/crawler.py
```diff
@@ -0,0 +1,35 @@
...
+            entity.add("topics", "reg.warn")  # or export.control
+            entity.add("sourceUrl", url)
+
+            sanction = h.make_sanction(context, entity)
+            h.apply_date(
+                sanction, "listingDate", str_row.get("order_date")
+            )  # sanction object schema date
+            context.emit(entity)
+            context.emit(sanction)
+            context.audit_data(str_row, ignore=["case_id", "order_date"])
```
nvmbrasserie:
```suggestion
            context.audit_data(str_row)
```

datasets/us/bis_export/crawler.py
```diff
@@ -0,0 +1,35 @@
...
+        url = h.xpath_elements(case_id, ".//a")[0].get("href")
+
+        for name in h.multi_split(
+            str_row.pop("case_name"), ";"
+        ):  # case-level names, might contain multiple entities
+            aliases = name.split("a/k/a")
+
+            entity = context.make("LegalEntity")
+            entity.id = context.make_id(
+                name, str_row.get("case_id")
```
nvmbrasserie:
```suggestion
                name, case_id
```

datasets/us/bis_export/crawler.py
```diff
@@ -0,0 +1,35 @@
...
+    for row in h.parse_html_table(table[0]):
+        str_row = h.cells_to_str(row)
+        case_id = row.get("case_id")
+        assert case_id is not None
+        url = h.xpath_elements(case_id, ".//a")[0].get("href")
+
+        for name in h.multi_split(
+            str_row.pop("case_name"), ";"
+        ):  # case-level names, might contain multiple entities
+            aliases = name.split("a/k/a")
```
nvmbrasserie: splitting on `aka` does the job for 90% of cases, but some are left unhandled:

```
- Ihsan Elashi Tetrabal Corporation Al Kayali Corporation Mynet.Net Corporation Infocom Corporation Synaptix.Net Maysoon Al Kayali Hazim Elashi Bayan Elashi Ghassan Elashi Basman Elashi Majida Salem and Fadwa Elafrangi
- Megatech Engineering & Services Pvt. Ltd.Ajay AhujaRavi ShettigarT.K. Mohan
- Xu Weibo, Hao Li Chen, Xiu Ling Cheng and Kwan Chun Chan
- Winter Aircraft/Ana Belen Diaz Sanchez
- Winter Aircraft/Iberair Lines
- Kabba & Amir Investments, Inc.d.b.a. International Freight Forwarders
```

One thing you can do is look up whether the string contains commas or slashes. Then, during the manual review, you can decide if it needs more splits.

datasets/us/bis_export/crawler.py
herrhellana: this will break at multi_split as i create an entity inside the split by ";" loop and need to assign the same date found in row for each entity found in row, hence .get()

datasets/us/bis_export/crawler.py
herrhellana: fixed issue with lookups

datasets/us/bis_export/us_bis_export.yml
```diff
@@ -0,0 +1,714 @@
...
+          - Mahan Air General Trading LLC
+          - Skyco (UK) Ltd # move country to?
+          - Equipco (UK) Ltd # move country to?
+          - Mehdi Bahrami
+      - match: 'Mahan Airways, Zarand Aviation, Gatewick LLC., PejMan Mahmood Kosarayanifard, Mahoud Amini, Kerman Aviation, Sirjanco Trading LLC, Ali Eslamian, Mahan Air General Trading LLC, Skyco (UK) LTD. and Equipco (UK) LTD. and Mehdi Bahrami'
+        entities:
+          - Mahan Airways
+          - Zarand Aviation
+          - Gatewick LLC
+          - ["Pejman Mahmood Kosarayanifard", "Kosarian Fard"]
```
nvmbrasserie: could you check the file to make sure the `entities` list only contains the entities from the `match`? It is crucial that the data is mirrored precisely from every `match`, without any entities being lost or added

herrhellana: yeah i can double check! note re: aliases: certain rows mentioned the same entities but didn't list all aliases for the entities. i still mapped those to rows with entities containing their complete alias lists

nvmbrasserie: yes, that's exactly the issue here. when cleaning up the data in lookups, we only want to map the values available to us without inferring any additional information. even if we are aware of additional aliases from elsewhere, they should not be listed unless they are explicitly stated in the specific `match` string

herrhellana: i see, then it's going to be a pain point bc it'll inflate the length of the .yml file twice (i.e. ill need to create a new lookup for every row with the same list of entities). is it sth we absolutely have to do? if not, i'd leave it as is

nvmbrasserie: yes, I know, but it's absolutely necessary. the idea is that we should never infer data from strings when there is no data on a specific entity. what you can do is remove the lookups for inferred entities completely, or just start from scratch. we don't want to leave any unused lookups in there either

herrhellana: aight understood, ill hop on it

datasets/us/bis_export/crawler.py
```diff
@@ -0,0 +1,97 @@
...
+def crawl_entity(
+    context: Context,
+    name: str | list[str],
+    url: str,
+    case_id_string: str,
+    order_date: str,
+) -> None:
+    # Lookup returns [primary_name, alias1, alias2, ...] or single string
+    primary_name = name[0] if isinstance(name, list) else name
+    aliases = name[1:] if isinstance(name, list) else []
```
jbothma: I think it'd be better to make the lookup results be dicts

```yaml
        entities:
          - name: Mahan Airways
          - name: Pejman Mahmood Kosarayanifard
            aliases: Kosarian Fard
```

or always lists with aliases as `names[1:]`

```yaml
        entities:
          - ["Mahan Airways"]
          - ["Pejman Mahmood Kosarayanifard", "Kosarian Fard"]
```

then you can avoid doing the isinstance dance. Simpler types is more understandble/less surprising code.

datasets/us/bis_export/us_bis_export.yml
jbothma: Thanks both.

Don't worry, you're far from the [Danger Zone](https://github.com/opensanctions/opensanctions/blob/main/datasets/fr/tresor/fr_tresor_gels_avoir.yml#L6062) :D

datasets/us/bis_export/crawler.py
```diff
@@ -0,0 +1,97 @@
...
+        case_id_string = str_row.pop("case_id")
+        order_date = str_row.pop("order_date")
+        assert case_id_string and order_date
+
+        cleaned_name = case_name
+        # Remove common suffixes (Inc., LLC, etc.) to check for delimiters
+        for suffix in SUFFIXES:
+            cleaned_name = cleaned_name.removesuffix(suffix)
+        # Check if the name contains delimiters indicating multiple entities (commas, semicolons, "and", etc.)
+        if any(char in cleaned_name for char in CUSTOM_STOPPHRASES):
```
jbothma:
```suggestion
        if any(phrase in cleaned_name for phrase in CUSTOM_STOPPHRASES):
```

datasets/us/bis_export/crawler.py
herrhellana: didn't know about casefold(), p neat!

datasets/us/bis_export/us_bis_export.yml
herrhellana: oh my god.... look at that, my new nightmare

jbothma: lol. LLMs or [multiline editing](https://code.visualstudio.com/docs/editing/codebasics#_multiple-selections-multicursor) are really good at this shit. And git diff is really good at checking that only the right things changed.

herrhellana: yeah! god bless opt+cmd+arrow

PR #3350, opened 2026-01-19
Introduce XSS scanning via regex

### LINKED ISSUE #1539: XSS detection on content
Our data is going into a lot of legacy downstream systems, so it may be nice if we do a bit of XSS validation to make sure we don't give them bonus javascript functionality by accident. 

* Precedent: https://www.theregister.com/2020/10/30/companies_house_xss_silliness/ 
* Suggested lib from audit: https://github.com/messense/nh3

pudo:
Fixes https://github.com/opensanctions/opensanctions/issues/1539 

This is a bit scary in that it uses some custom regex (from the machines), but the issue with using existing frameworks like `nh3` was that they are much more strict than we can tolerate: `&`, `"` and `<` are valid values in our output data, they just become a risk if they meet very specific conditions. `nh3.clean` and `nh3.is_html` don't quite do the right thing here.

### DIFF
--- a/zavod/zavod/runtime/lookups.py
+++ b/zavod/zavod/runtime/lookups.py
@@ -17,6 +17,16 @@ def get_type_lookup(dataset: Dataset, type_: PropertyType) -> Optional[Lookup]:
     return dataset.lookups.get(f"type.{type_.name}")
 
 
+def is_lookup_value(entity: "Entity", type_: PropertyType, value: str) -> bool:
+    """Check if a given value for a certain property type was obtained from
+    a lookup. This is used to skip validation for looked-up values."""
+    lookup = get_type_lookup(entity.dataset, type_)
+    if lookup is None:
+        return False
+    result = lookup.match(value)
+    return result is not None
+
+
 def type_lookup(
     dataset: Dataset, type_: PropertyType, value: Optional[str]
 ) -> List[str]:
--- a/zavod/zavod/runtime/lookups.py
+++ b/zavod/zavod/runtime/lookups.py
@@ -17,14 +17,17 @@ def get_type_lookup(dataset: Dataset, type_: PropertyType) -> Optional[Lookup]:
     return dataset.lookups.get(f"type.{type_.name}")
 
 
-def is_lookup_value(entity: "Entity", type_: PropertyType, value: str) -> bool:
-    """Check if a given value for a certain property type was obtained from
-    a lookup. This is used to skip validation for looked-up values."""
+def is_type_lookup_value(entity: "Entity", type_: PropertyType, value: str) -> bool:
+    """Check if a given value is the result of a type-based lookup.
+
+    Used to skip validation for looked-up values, because we've added them manually
+    we deem them to be safe.
+    """
     lookup = get_type_lookup(entity.dataset, type_)
     if lookup is None:
         return False
-    result = lookup.match(value)
-    return result is not None
+    # True if the value is the result of any of the lookup's options.
+    return any(option.result.value == value for option in lookup.options)
 
 
 def type_lookup(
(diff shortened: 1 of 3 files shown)


zavod/zavod/runtime/lookups.py
```diff
@@ -17,6 +17,16 @@ def get_type_lookup(dataset: Dataset, type_: PropertyType) -> Optional[Lookup]:
     return dataset.lookups.get(f"type.{type_.name}")
 
 
+def is_lookup_value(entity: "Entity", type_: PropertyType, value: str) -> bool:
+    """Check if a given value for a certain property type was obtained from
+    a lookup. This is used to skip validation for looked-up values."""
+    lookup = get_type_lookup(entity.dataset, type_)
+    if lookup is None:
+        return False
+    result = lookup.match(value)
```
leonhandreke: I know this was there before but... how does this work? Doesn't this check whether `value` yields a cleaned value, whereas what we want to check here is whether `value` is the result of a lookup, i.e. on the other side of the mapping? Hope this comment makes sense.

zavod/zavod/runtime/safety.py
```diff
@@ -0,0 +1,53 @@
...
+    """A very basic HTML entity and XSS check to prevent script injections."""
+    # This is very .... rustic. The reason for that design is that there's a lot of things
+    # here we do not want to do, which normal XSS filters would do, like stripping tags
+    # or escaping entities. Both of these would render our data useless for matching purposes.
+    # So we only want to identify the most egregious cases and log them for manual review.
+    has_xss = XSS_SUSPECT_PATTERN.search(value)
+    has_entities = HTML_ENTITY_PATTERN.search(value)
+    if not has_xss and not has_entities:
+        return value
+    if is_lookup_value(entity, prop.type, value):
```
leonhandreke: (nit) should this go a level above to skip some/all of the other validation? Or is this very specific to this check?

zavod/zavod/runtime/safety.py
```diff
@@ -0,0 +1,53 @@
...
+    r"<[^>]*>|"  # Tags
+    r"javascript:|data:|vbscript:|"  # URI schemes
+    r"on\w+\s*=|"  # Event handlers
+    r"&#|&[a-zA-Z]",  # Entity references
+    re.IGNORECASE,
+)
+
+
+def check_xss(entity: "Entity", prop: Property, value: str) -> str | None:
+    """A very basic HTML entity and XSS check to prevent script injections."""
```
leonhandreke: I definitely don't know enough about Javascript/HTML to gauge whether this guarantees the values safe to render un-escaped and in which context. So I guess this is more of a smell detector? This might be good to document because if someone will change it at some point, it might be good to know what the "contract" is that our string values adhere to.

zavod/zavod/runtime/lookups.py
pudo: So the objective here was to be able to say: 

```
type.name:
  options:
  - match: 1928
     value: 1928
```

To signify that an otherwise invalid value should be allowed to pass through unhindered. You might be entirely right to say that this doesn't really do that, and we should make the last line eg. do something like `return value in result.values`?

zavod/zavod/runtime/safety.py
pudo: Yeah I think our ambition should be to warn about possible script/xss content and escaped HTML entities in our data - not to actually guarantee that none of what comes out has any HTML formatting built in.

leonhandreke: I updated the docs to be more conservative so that future editors will know that it's more of a smoke detector than a strong guarantee about safety in some context that we've committed to downstream.

zavod/zavod/runtime/lookups.py
leonhandreke: I don't hate the mechanism, but I feel that using the values as the allow-list, rather than the match expressions, makes more sense (and is also what is conveyed by the function signature). I updated the function.

PR #3351, opened 2026-01-19
[ae_local_terrorist] more detailed data extraction

nvmbrasserie:


datasets/ae/local_terrorists/crawler.py
```diff
@@ -61,12 +62,36 @@ def parse_row(
...
+            case "issuer" | "passportNumber" | "is_passport":
+                identification[header.name] = value
             case _:
                 entity.add(header.name, value, lang=header.lang)
 
     if len(address):
+        # Build address entity with detailed breakdown of components (street, city, country).
+        # We use apply_address() instead of copy_address() because we want to emit
+        # a separate Address entity with the full structured data, not just copy
+        # the address string to the entity.
```
leonhandreke, line 80: Why? Are these `Address` entities actually useful for anything, do we want to move into this direction more generally?

datasets/ae/local_terrorists/ae_local_terrorists.yml
```diff
@@ -122,8 +136,8 @@ lookups:
         value: city
         lang: ara
       - match: النوع
-        # looks like: type of ban
-        value: provisions
+        # type of document (e.g. passport)
+        value: is_passport
```
leonhandreke: (nit) Make `document_type` or `identification_type` is a better name?

datasets/ae/local_terrorists/crawler.py
```diff
@@ -61,12 +62,36 @@ def parse_row(
...
         addr = h.make_address(context, **address)
-        h.copy_address(entity, addr)
+        h.apply_address(context, entity, addr)
+
+    if len(identification):
+        # The lookup returns string "True" for passport types or None for non-passport documents.
+        # We convert to boolean: any non-None/non-empty value becomes True.
+        is_passport = bool(
```
leonhandreke, line 88: Oh god nothing is safe from the `stringify` (I think this is totally fine given the constraints. Other option would be a dict in python, which I think would also be quite okay)

datasets/ae/local_terrorists/crawler.py
```diff
@@ -132,9 +156,14 @@ def parse_excel(context: Context, path: Path):
 
 def crawl(context: Context):
     doc = context.fetch_html(context.data_url, absolute_links=True)
-    section = doc.xpath(".//h5[text()='Local Terrorist List']")[0].getparent()
-    link = section.xpath(
-        ".//p[text()='Download Excel File']/ancestor::*[contains(@class,'download-file')]//a"
+    section = h.xpath_elements(
```
leonhandreke: I just added `xpath_element` to helper-ify the `expect_exactly=1)[0]` pattern

datasets/ae/local_terrorists/crawler.py
```diff
@@ -61,12 +62,36 @@ def parse_row(
...
+            )
+        )
+        ident = h.make_identification(
+            context,
+            entity,
+            number=identification.get("passportNumber"),
+            country=identification.get("issuer"),
+            passport=is_passport,
+        )
+        if ident is not None:
```
leonhandreke: When does it happen that ident is `None`?

nvmbrasserie: in some cases we don't have the number of the doc. if that's the case, `h.make_identification` will return None

PR #3362, opened 2026-01-22
Don't backfill issues log

leonhandreke:
I don't know why this was ever a thing, but this breaks the reasonable assumption that the issues log contains the issues from "this run". For collections, this currently means that `issues.log` is append-only and forever. Leaf datasets currently have their issues log cleared by `Context.begin(clear=True)`.

If I understand the mechanics of this correctly, this means that collections will now export and `issues.log` containing issues logged during the export stage, i.e. validation and assembly.

### DIFF
--- a/zavod/zavod/runtime/issues.py
+++ b/zavod/zavod/runtime/issues.py
@@ -28,7 +28,6 @@ class DatasetIssues(object):
     def __init__(self, dataset: Dataset) -> None:
         self.dataset = dataset
         self.fh: Optional[BinaryIO] = None
-        get_dataset_artifact(self.dataset.name, ISSUES_LOG)
 
     def write(self, event: Dict[str, Any]) -> None:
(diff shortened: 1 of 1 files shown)


leonhandreke:
Already done for resources: https://github.com/opensanctions/opensanctions/commit/ec6b9ce497ff77a775e862229a9bf4a4659d4154

For statistics, the whole thing is a bit more complicated, basically our website showing statistics even for failed datasets currently depends on backfilling `statistics.json`: https://github.com/opensanctions/opensanctions/blob/879234667d2dd93b7fd672b04252c4f052eeb865/zavod/zavod/exporters/metadata.py#L38

Will merge this tomorrow so that I can intervene if the next default export blows up.

pudo:
Thanks for doing this. It's hard to un-fuddle, but will leave us with a much clearer process.

PR #3381, opened 2026-01-26
Fixes-26-01

nvmbrasserie:


datasets/bg/jud_declarations/crawler.py
```diff
@@ -100,7 +101,7 @@ def crawl_row(context: Context, row: Dict[str, HtmlElement], index_url: str):
     # Link is in the same cell as the name
     name_link_elem = HtmlElement(row["name"]).find(".//a")
     declaration_url = name_link_elem.get("href")
-    if not declaration_url:
+    if not declaration_url or declaration_url in BROKEN_LINKS:
```
leonhandreke: Don't love this - should we wait a couple days? Alternatively, please file a bug (and assign to yourself maybe?) to take care of this.

datasets/id/dttot/crawler.py
```diff
@@ -82,10 +83,20 @@ def apply_date(entity: Entity, prop: str, value: str) -> None:
...
-    xls = context.fetch_resource("source.xls", last_link)
-    wb = xlrd.open_workbook(xls)  # This will break loudly once they switch to XLSX
+    # Check if the source file has been updated on the website
+    # If the hash changes and the file is not corrupt anymore,
+    # we should switch back to fetching directly from the URL
+    try:
+        html = context.fetch_html(context.data_url)
+        last_link = find_last_link(context, html)
+        h.assert_url_hash(context, last_link, "jnvkd")
+    except Exception as e:
```
leonhandreke: I'm not quite sure, is `assert` also an `Exception`? Anyway, I feel like there is little reason not to crash here, it's a pretty low-priority dataset.

nvmbrasserie: what would be the better way to implement this?

leonhandreke: Just don't catch the exception? That's what we do everywhere else. And in this case, it's not even catching an assert, cause it's just a warning.

(Anyway, I checked, it actually just warns with the default args... so we sorta failed on the name  :P )

PR #3402, opened 2026-01-28
Daily-issues-28-01

nvmbrasserie:


datasets/us/fcc_covered_list/us_fcc_covered_list.yml
```diff
@@ -31,6 +31,15 @@ description: |
   > outlined in sections 2(b)(2)(A), (B), or (C) of the Secure and Trusted Communications
   > Networks Act of 2019, 47 U.S.C. § 1601(b)(2)(A)-(C).
 
+  As of January 7, 2026, the list includes uncrewed aircraft systems (UAS) and UAS
+  critical components produced in a foreign country, with temporary exemptions until
```
leonhandreke: We should at least mention that we're not including it on the list.

> As of December 22, 2025, the source listing also includes a broad entry that covers all uncrewed aircraft systems (UAS) and UAS critical components produced in a foreign country (with some exceptions) as well as all communications and video surveillance equipment specified in Section 1709(a)(1) of the FY25 National Defense Authorization Act. These are not included in this dataset, as they amount to sector-level debarments rather than a list of entities.

This isn't completely true, the PDF linked includes two companies. But I don't think we should be in the business of making sense of these laws. @brrttwrks to confirm.

Image

nvmbrasserie: I'll move it here so that we can merge this PR: https://github.com/opensanctions/opensanctions/pull/3404

PR #3404, opened 2026-01-28
[us_fcc_covered_list] add policy outline to the metadata

nvmbrasserie:
@leonhandreke's suggestion is here: https://github.com/opensanctions/opensanctions/pull/3402#discussion_r2737405723


leonhandreke:
I suggest we add a sentence that we don't include it and why. Also note that the source lists a different start date for this policy. In my comment, I also editorialized the whole thing a bit, but no strong opinions on that.

PR #3412, opened 2026-01-29
br-pep tiny assertion fix

herrhellana:


datasets/br/pep/br_pep.yml
```diff
@@ -38,8 +38,8 @@ dates:
 assertions:
   min:
     schema_entities:
-      Person: 95000
-      Position: 20700
+      Person: 54000  # 67429 unique persons in df
+      Position: 800  # 1021 unique positions in df
```
jbothma: what's `df`?

Could you tell why the number has decreased? Is our parsing missing out some people? Have they purged a bunch from the files? i.e. did a previous CSV file include more people than now?

It doesn't look like we changed the code in a relevant way recently.

herrhellana: df stands for dataframe, a common convention in research (though it turns out not to be as common outside of it). the count you see is for unique persons in the source dataset. the number of rows in the source file is ~71k. so it seems like they just dropped the numbers. we'll look into dedupes with @nvmbrasserie tomorrow.

nvmbrasserie: I ran a check on unique tax numbers and found `65,501`. I also got `13365` unique positions. This file has only 70k rows compared to the previous one, which had 132k rows. I'll lower the assertions and dedupe it further after the first successful export

Issue #3359, opened 2026-01-21
Assertions for property fill rates


leonhandreke:
- Also by schema/property pair?

pudo:
Now rolled out for sanctions, let's do more as we go.

PR #3416, opened 2026-01-30
assertions: add property_fill_rate

leonhandreke:
I made this a separate validator because it works a bit differently to the other assertions, so I didn't want to lump it into the same class.

There is a bit of a caveat here: the property fill rate validator understands schema inheritance, whereas the statistics-based assertions do not.

See https://github.com/opensanctions/opensanctions/issues/3359

### DIFF
--- a/zavod/pyproject.toml
+++ b/zavod/pyproject.toml
@@ -39,6 +39,7 @@ dependencies = [
     "openai >= 1.33.0, < 3.0.0",
     "requests[security]",
     "requests_oauthlib",
+    "urllib3>=2.0",
     "sqlalchemy[mypy]",
     "structlog",
     "xlrd == 2.0.2",
--- a/zavod/zavod/validators/assertions.py
+++ b/zavod/zavod/validators/assertions.py
@@ -20,15 +20,18 @@ def check_value(
             raise ValueError(f"Unknown comparison: {comparison}")
 
 
+def is_assertion_fatal(assertion: Assertion) -> bool:
+    return assertion.comparison == Comparison.GTE
+
+
 def check_assertion(
     context: Context,
     stats: Dict[str, Any],
     assertion: Assertion,
-    log_error: bool = True,
 ) -> bool:
     """Returns true if the assertion is valid, false otherwise."""
 
-    log_fn = context.log.error if log_error else context.log.warning
+    log_fn = context.log.error if is_assertion_fatal(assertion) else context.log.warning
     results_valid: list[bool] = []
 
     if assertion.metric == Metric.ENTITY_COUNT:
@@ -113,6 +116,10 @@ def check_assertion(
                     )
                 results_valid.append(valid)
 
+    elif assertion.metric == Metric.PROPERTY_FILL_RATE:
+        # We handle that in a separate validator below because it's a special cookie
+        # that doesn't just want the Statistics
+        pass
     else:
         raise ValueError(f"Unknown metric: {assertion.metric}")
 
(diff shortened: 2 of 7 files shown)


zavod/zavod/validators/assertions.py
```diff
@@ -135,10 +142,69 @@ def finish(self) -> None:
...
+            if not valid and is_assertion_fatal(assertion):
                 self.abort = True
+
+
+class PropertyFillRateAssertionsValidator(BaseValidator):
+    """Warns if the fill rate of a property is below a threshold.
+
+    The reason this is implemented as a separate validator is that it doesn't just want the Statistics,
+    but the actual entities to compute the fill rate. Putting it in the AssertionsValidator would have
+    made it too complex, so instead I chose to spread the assertions functionality across multiple validators.
```
jbothma:
```suggestion
    The reason this is implemented as a separate validator is that Statistics only counts the specific schema, while fill rate considers all the subclasses of an asserted schema. Putting it in the AssertionsValidator would have
    made it too complex, so instead I chose to spread the assertions functionality across multiple validators.
```

fill rate is `entities with some value for the schema:prop / entities of this schema (or its subclasses)`, right?

And the reason statistics' schema count and entities-with-value-for-prop counts don't work for this is that they only count the specific schema while we want `{"LegalEntity": {"name": 0.5}}` to mean all legal entities, including persons and companies.

If this is the case, maybe we can explain a bit more specifically why statistics doesn't work for this.

Am I right in thinking the main benefit of this approach is that we don't have to specify the assertion for every schema that might we emit that might have this prop, while we can still be more specific if we want to have different expectations on Persons vs Companies. I think that's fair justification for more complexity, just making sure I'm on the same page.

leonhandreke: I think the main reason for me is that I didn't want to make `statistics.json` even more fat. Now that I say it, I'm no longer as sure, because I guess we could also have non-exported stats in the `Statistics` class. But I guess that just shifts the complexity (i.e. doing the counting) elsewhere? I feel this approach is quite straightforward actually because it doesn't take the detour through another counter class that needs to export just the right things to support the usecase at hand.

jbothma (approved):
It'd be great if this was documented on https://zavod.opensanctions.org/metadata/#data-assertions

zavod/pyproject.toml
```diff
@@ -39,6 +39,7 @@ dependencies = [
     "openai >= 1.33.0, < 3.0.0",
     "requests[security]",
     "requests_oauthlib",
+    "urllib3>=2.0",
```
jbothma, line 42: when I update dependencies locally, I use this:

```
pip install -U  --upgrade-strategy=eager -e 'zavod[dev]'
```

The eager causes indirect dependencies to also be upgraded to the latest meeting constraints, which is similar to a fresh CI build with no previously installed dependencies. Maybe doing that reduces the need to explicitly specify more dependencies in zavod - not sure if applicable in this case or if zavod directly needs this version though.

jbothma (approved):
Nice! Looks like this is now _exact schema match_ rather than _matching schema or ancestors_. I think _exact schema match_ is more intuitive in a context where we're counting entities and summarising by schema. Both for the assertion, and for presenting the summary.

Image

even though _matching schema or ancestors_ is technically correct.

zavod/pyproject.toml
leonhandreke: Somehow I managed to get the old one, not sure why, but since the constraints didn't seem to forbid it, I assume it was a valid outcome to the dependency resolution mechanism. zavod directly depends on it, so I think it's fine to specify it?

PR #3419, opened 2026-01-30
assertions: refactor to dumb down a bit

leonhandreke:
Refactor to dumb the whole thing down a bit. Disadvantage is we no longer get Dataset parse level validation for whether the format of the assertions is correct - we'll only know when the validator runs. But now assertions can have quite different config formats (some take just a value, some a value per country, some a value per schema/prop combo) and it won't be two layers of indirection anymore. That also made the error messages a bit better.

Also, make the assertions unit tests a bit mroe unit-testy. Previously, they relied on fat fixtures (`testdatasetN` that made it intransparent what was actually going on and hard to change these fixtures without breaking tests.

See https://github.com/opensanctions/opensanctions/issues/3359

### DIFF
--- a/zavod/zavod/meta/assertion.py
+++ b/zavod/zavod/meta/assertion.py
@@ -1,113 +1,66 @@
+from dataclasses import dataclass
 from enum import Enum
-from typing import Any, Dict, Generator, Optional
-
-from followthemoney import model
+from typing import Any, Dict, Generator
 
 
 class Metric(Enum):
     ENTITY_COUNT = "entity_count"
-    """Number of entities matching the filter in the dataset."""
-    COUNTRY_COUNT = "country_count"
+
+    SCHEMA_ENTITIES = "schema_entities"
+    """Number of entities with of a given schema in the dataset."""
+
+    COUNTRY_ENTITIES = "country_entities"
+    """Number of entities with a given country in the dataset."""
+
+    COUNTRY_COUNT = "countries"
     """Number of distinct countries occurring in the dataset."""
-    ENTITIES_WITH_PROP_COUNT = "entities_with_prop_count"
+
+    ENTITIES_WITH_PROP_COUNT = "entities_with_prop"
     """Number of entities with property values matching the filter in the dataset."""
+
+    def __str__(self) -> str:
+        return self.value
 
 
 class Comparison(Enum):
     GTE = "gte"
     LTE = "lte"
 
+    def __str__(self) -> str:
+        match self:
+            case Comparison.GTE:
+                return ">="
+            case Comparison.LTE:
+                return "<="
+            case _:
+                raise ValueError(f"Unknown comparison: {self}")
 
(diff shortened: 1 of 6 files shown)


zavod/zavod/tests/test_validate.py
```diff
@@ -36,6 +45,20 @@ def run_validator(clazz: Type[BaseValidator], dataset: Dataset):
     return validator, set(cap_logs)
 
 
+def e(ds: Dataset, schema: str, properties: dict[str, list[str]]) -> Entity:
```
jbothma: I think in other places a function e just returns an entity. I wonder if we can name this a bit more clearly. Maybe we could name it run_emit_e or add a docstring or something.

zavod/zavod/meta/assertion.py
```diff
@@ -1,118 +1,73 @@
...
 class Metric(Enum):
     ENTITY_COUNT = "entity_count"
-    """Number of entities matching the filter in the dataset."""
-    COUNTRY_COUNT = "country_count"
+
+    SCHEMA_ENTITIES = "schema_entities"
+    """Number of entities with of a given schema in the dataset."""
+
+    COUNTRY_ENTITIES = "country_entities"
+    """Number of entities with a given country in the dataset."""
```
jbothma, line 13: how did this ever work?!

https://www.pixelstech.net/images/fun/normal/6-stages-of-debugging.png

zavod/zavod/validators/assertions.py
```diff
@@ -18,64 +20,107 @@ def compare_threshold(value: int, comparison: Comparison, threshold: int) -> boo
...
+                log_fn(
+                    f"Assertion {assertion.metric} failed for {country}: {value} is not {assertion.comparison} threshold {threshold}"
+                )
+            results_valid.append(valid)
+
+    elif assertion.metric == Metric.ENTITIES_WITH_PROP_COUNT:
+        # stats["things"]["entities_with_prop"] is a list of dictionaries that look like
+        # [
+        #   { "schema": "Person", "property": "firstName", "count": 100 },
+        # ]
```
jbothma, line 99: very helpful

jbothma (approved):
Losing the upfront validation is a pity, and we might only find out about mistakes after a long slow dataset run.

Still, it looks like this will fail loudly if we get the structure wrong or make a typo on a name or prop. We can always add this back if the lack of validation really hurts.

zavod/zavod/meta/assertion.py
leonhandreke: Yeah I think it was all `entity_count` with a different `filter_attribute` or something like that? I thought it was quite hard to reason about (it probably made sense before it grew)

zavod/zavod/tests/test_validate.py
leonhandreke: Renamed to `emit_entity`

Issue #3348, opened 2026-01-19
validator: Person should only have one gender


pudo:
(please file under "we're building a government watchlist database, not a reflection of social reality")

Let's make a validator for Person entities that checks that they have only one distinct value for `gender`.

pudo:
See #3348: the world is complicated, let's go have cookies instead.

PR #3420, opened 2026-01-30
validators: Add one gender validator

leonhandreke:
https://github.com/opensanctions/opensanctions/issues/3348

### DIFF
--- a/zavod/zavod/tests/test_validate.py
+++ b/zavod/zavod/tests/test_validate.py
@@ -11,6 +11,7 @@ from zavod.meta.dataset import Dataset
 from zavod.store import get_store
 from zavod.validators import (
     DanglingReferencesValidator,
+    OneGenderValidator,
     SelfReferenceValidator,
     EmptyValidator,
 )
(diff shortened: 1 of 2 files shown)


pudo (approved):
Let's deploy it and then figure out how to remediate the resulting crawler issues? This doesn't need to go into issues week, it can be a maybe a bit spread out on the data team.

goose-life:
@leonhandreke wanna go ahead on this and see what happens?

leonhandreke:
Also, from my comment in the issue:

> There will always be cases such as  [https://opensanctions.org/entities/Q29564107]: 'Billie Eilish' -> ['female', 'other'] where it's not super clear or we have to make an editorial decision, and I wonder where that editorial decision should go, from a technical perspective.

pudo:
I computed the most recent version of them yesterday and sent them into a validation loop, it turns out that for the Wikidata cases in a surprising number of cases the multiple statements actually represent social reality. There were like 2 or 3 errors, but mostly just ... humans. There's a part of me that wants to abandon this validator.

leonhandreke:
I would also vote to abandon this for now. Maybe there is a world were we add it and also have a way to silence the warning where the multiple genders match social reality (I still love the Billie Eilish is in our database and nobody quite knows what's going on with the gender thing). But that seems like a lot of work right now for very little gain.

PR #3431, opened 2026-02-02
US Texas Medical Exclusion List

herrhellana:
opensanctions/crawlers#587

### DIFF
--- a/datasets/us/tx/med_exclusions/crawler.py
+++ b/datasets/us/tx/med_exclusions/crawler.py
@@ -42,12 +42,12 @@ def crawl_row(drow: Dict[str, Any], context: Context, wb: xlrd.book.Book) -> Non
         entity.add("description", occupation)
 
         if contains_split_phrase(company_name):
-            res = context.lookup("aliases", company_name, warn_unmatched=True)
-            if res is not None and res.aliases:
-                cleaned_company_name = res.aliases[0]
-                aliases = res.aliases[1:]
-                entity.add("name", cleaned_company_name)
-                entity.add("alias", aliases)
+            res = context.lookup("names", company_name, warn_unmatched=True)
+            if res is not None:
+                entity.add("name", res.names[0], original_value=company_name)
+                entity.add("alias", res.names[1:], original_value=company_name)
+            else:
+                entity.add("name", company_name)
         else:
             entity.add("name", company_name)
 
--- a/datasets/us/tx/med_exclusions/us_tx_med_exclusions.yml
+++ b/datasets/us/tx/med_exclusions/us_tx_med_exclusions.yml
@@ -60,68 +60,68 @@ lookups:
           - 1225533057
           - 118771347
         prop: registrationNumber
-  aliases:
+  names:
     options:
       - match: Pro Medic EMS, LLC dba Pro Medic EMS, LLC
-        aliases:
+        names:
           - Pro Medic EMS, LLC
       - match: Executive Healthcare, LLC d/b/a The Apothecary Sho
-        aliases:
+        names:
           - Executive Healthcare, LLC
           - The Apothecary Shop
       - match: ERX Consultants LLC, dba Zoetic Pharmacy
-        aliases:
+        names:
           - ERX Consultants LLC
           - Zoetic Pharmacy
       - match: Ukie Emergency Medical Services, Inc. d/b/a Ukie E
-        aliases:
+        names:
           - Ukie Emergency Medical Services, Inc
           - Ukie EMS
       - match: Maxim Mossman PLLC dba Amigos Dental
-        aliases:
+        names:
           - Maxim Mossman PLLC
           - Amigos Dental
       - match: Sechanz Enterprises Inc. d/b/a Hovic Pharmacy and
-        aliases:
+        names:
           - Sechanz Enterprises Inc
           - Hovic Pharmacy
       - match: MossPrime Dental PLLC dba North Arlington Smiles
-        aliases:
+        names:
           - MossPrime Dental PLLC
           - North Arlington Smiles
       - match: RightChoice Dental, PLLC dba The Dental House
-        aliases:
+        names:
           - RightChoice Dental, PLLC
           - The Dental House
       - match: Mossman Dental, PLLC dba The Dental  Park
-        aliases:
+        names:
           - Mossman Dental, PLLC
           - The Dental Park
       - match: Pharmplex LTD III dba Penitas Family Pharmacy
-        aliases:
+        names:
           - Pharmplex LTD III
           - Penitas Family Pharmacy
       - match: Western Dental dba West Texas Dental
-        aliases:
+        names:
           - Western Dental
           - West Texas Dental
       - match: M Ayman Ghraowi, MD PA dba South Texas Comprehensi
-        aliases:
+        names:
           - M Ayman Ghraowi, MD PA
           - South Texas Comprehensive Cancer Centers PLLC (STCCC)
       - match: Southwest Medical Consultants Inc. d/b/a Los Ninos
-        aliases:
+        names:
           - Southwest Medical Consultants Inc
           - Los Ninos Pediatric Rehab
       - match: Acosta Medical Supply, Inc. dba Affordab
-        aliases:
+        names:
           - Acosta Medical Supply, Inc
           - Affordable Medical Supply
       - match: Family Healthcare, Inc d/b/a Unique Medical & Reha
-        alises:
+        names:
           - Family Healthcare, Inc
           - Unique Medical & Reha  # could be Rehab or Rehabilitation, i was able to find them in the registry but without the alias: https://comptroller.texas.gov/taxes/franchise/account-status/search/18707535466
       - match: SCZ Investments, LLC d/b/a Rotunda Pharmacy
-        aliases:
+        names:
           - SCZ Investments, LLC
           - Rotunda Pharmacy


datasets/us/tx/us_tx_oig_exclusions.yml
```diff
@@ -0,0 +1,62 @@
+title: US Texas HHSC OIG Exclusions
+entry_point: crawler.py
+prefix: us-tx-med  # unsure, should it mirror folder name?
+coverage:
+  schedule: "@monthly"  # unsure if good pick
+  start: "2026-01-29"
+load_statements: true  # unsure what it does
+ci_test: false  # unsure what it does
```
jbothma: False is correct here, with a comment that zyte isn't available in CI

https://zavod.opensanctions.org/metadata/#:~:text=Continuous%20Integration-,ci_test,-%2D%20boolean%2C%20default%20true

datasets/us/tx/us_tx_oig_exclusions.yml
```diff
@@ -0,0 +1,62 @@
+title: US Texas HHSC OIG Exclusions
+entry_point: crawler.py
+prefix: us-tx-med  # unsure, should it mirror folder name?
+coverage:
+  schedule: "@monthly"  # unsure if good pick
```
jbothma: clicking a few medicaid datasets, it looks like we cron it to run a couple of days a month

```
coverage:
  start: "2024-11-15"
  schedule: "20 4 1,15 * *"
```

datasets/us/tx/us_tx_oig_exclusions.yml
```diff
@@ -0,0 +1,62 @@
+title: US Texas HHSC OIG Exclusions
```
jbothma:
```suggestion
title: US Texas Medicaid Provider Exclusions
```

I try to be careful with abbreviations. While it sometimes makes sense when really anyone looking at it would more easily understand the abbreviation than the full name (e.g. NASA), in this case I think we can work the terms from the intro para into a nice clear title.

For domain experts it can be nice to still keyword-stuff the page with the abbreviations, and I see you're already doing that nicely with summary - perfect.

datasets/us/tx/us_tx_oig_exclusions.yml
```diff
@@ -0,0 +1,62 @@
...
+entry_point: crawler.py
+prefix: us-tx-med  # unsure, should it mirror folder name?
```
jbothma: looks like we ended up with a pattern like `us-medmi` based on some other medicaid datasets. Context you might not have - we did most states around the same time as a big project so we could come up with some consistent patterns

datasets/us/tx/us_tx_oig_exclusions.yml
```diff
@@ -0,0 +1,62 @@
...
+coverage:
+  schedule: "@monthly"  # unsure if good pick
+  start: "2026-01-29"
+load_statements: true  # unsure what it does
```
jbothma: I see the docs don't really explain when this should be true. It populates the etlprod db with statements to enable the statements page on the website (and several other uses) https://www.opensanctions.org/statements/Q19878/

We set it to true on the leaf datasets (e.g. this one) and false on collections (e.g. `debarment.yml` to avoid duplication) and company registries (too big and not really necessary)

datasets/us/tx/crawler.py
```diff
@@ -0,0 +1,99 @@
...
+
+    context.audit_data(
+        drow,
+        ignore=["waiver", "eligibletoreapplydate"],
+    )
+
+
+def crawl(context: Context) -> None:
+    """
+    Website is geofenced (U.S.). Connection to a US server is required.
```
jbothma:
```suggestion
    Website is geofenced (U.S.). Connection from a US server is required.
```

datasets/us/tx/crawler.py
```diff
@@ -0,0 +1,99 @@
...
+        "__EVENTVALIDATION": eventvalidation,
+        "__EVENTTARGET": "dnn$ctr384$DownloadExclusionsFile$lb_DLoad_ExcFile_XLS",
+    }
+
+    path = context.fetch_resource(
+        "source.xls",
+        context.data_url,
+        method="POST",
+        data=form_params,
+    )  # really need to remember to delete files in datasets/us_tx_oig_exclusions in case of a bad first download...
```
jbothma: I know what you mean! This has bitten me so many times. 
I think we're planning on making crawl delete the data dir by default so this will be less of an issue.
Does it make sense to remove this comment since I think it could be a bit confusing without more context?

datasets/us/tx/crawler.py
```diff
@@ -0,0 +1,99 @@
...
+
+    context.export_resource(path, XLS, title=context.SOURCE_TITLE)
+
+    wb = xlrd.open_workbook(str(path))  # old .xls version not supported by openpyxl
+    sh = wb.sheet_by_index(0)
```
jbothma: I thought we had this in best practices but I don't see it.

It's a good idea to assert somehow that a) the sheet you're scraping means what you think it means, and b) they haven't added more sheets or you're scraping all of them. 

I'd suggest you a) select the sheet by name rather than index, unless they randomly change the name because humans and excel, and b) assert that there are exactly 1 sheets (or if there are more, that they're blank)

What we've seen is that occasionally a sheet gets added. We don't want to miss the new data. And occasionally the data on one sheet is banned people, and another sheet is bans that have been lifted. Incorrectly listing those as banned people is a legal liability so we need to be careful for both possibilities.

datasets/us/tx/crawler.py
```diff
@@ -0,0 +1,99 @@
...
+            last_name=last_name,
+            middle_name=drow.pop("midinitial"),
+        )
+        entity.add("position", occupation)
+
+    else:
+        entity = context.make("Company")
+        entity.id = context.make_id(
+            company_name
+        )  # note that npi/licenseNumber isnt listed for all companies though so not including in ID
```
jbothma: Generally, try to put the comment in the row above, rather than at the end only to get wrapped and become three lines.

In this case, there's nothing wrong with doing this - it handles empty strings and Nones just fine, and when an npi is available, we have a better chance at uniqueness for name conflicts. Sure it's unlikely, but it's easy to be safer.

```suggestion
        entity.id = context.make_id(company_name, npi_number)
```

jbothma: worth bearing in mind, adding information to keys later is a fair bit of effort because we need to gracefully manage re-keying everyone. So it's better to het the best unique but consistent IDs you can from the start.

datasets/us/tx/crawler.py
herrhellana: friedrich and i chatted about it, yeah it's a good idea to bring some nice flags in here. e.g. in case of my current Norwegian crawler, i download an enormously large zip file that i really don't want to redownload every time i debug stuff. 

and yes, ill remove the comment!

datasets/us/tx/crawler.py
herrhellana: hmm i'll try running it this way again, but this was my initial version, and it was breaking when calling a null `npi_number` inside `context.make_id()`

herrhellana: lol i reran it, and it didn't complain. ok, great

datasets/us/tx/crawler.py
```diff
@@ -0,0 +1,98 @@
...
+        "__EVENTVALIDATION": eventvalidation,
+        "__EVENTTARGET": "dnn$ctr384$DownloadExclusionsFile$lb_DLoad_ExcFile_XLS",
+    }
+
+    path = context.fetch_resource(
+        "source.xls",
+        context.data_url,
+        method="POST",
+        data=form_params,
+    )
```
jbothma: as discussed yesterday, this needs to go through zyte for geolocation, but zyte_api.fetch_resource doesn't support POST yet. Have a go at making the request work and let's check in after a timebox of about 3 hours

datasets/us/tx/us_tx_oig_exclusions.yml
```diff
@@ -0,0 +1,127 @@
+title: US Texas Medicaid Provider Exclusions
+entry_point: crawler.py
+prefix: us-medtx
+coverage:
+  schedule: "20 4 1,15 * *"
+  start: "2026-01-29"
+load_statements: true
+ci_test: false
```
jbothma:
```suggestion
ci_test: false  # zyte not available in CI
```

datasets/us/tx/crawler.py
```diff
@@ -0,0 +1,108 @@
...
+        ignore=["waiver", "eligibletoreapplydate"],
+    )
+
+
+def crawl(context: Context) -> None:
+    """
+    Website is geofenced (U.S.). Connection from a US server is required.
+    """
+    # fetch the XLS file via ASP.NET postback like in datasets/sk/public_officials
+    # doc = context.fetch_html(context.data_url)
```
jbothma:
```suggestion
```

datasets/us/tx/crawler.py
```diff
@@ -0,0 +1,108 @@
...
+    npi = drow.pop("npi")
+    license_number = drow.pop("licensenumber")
+    occupation = drow.pop("occupation")
+
+    if not last_name and not company_name:
+        return
+
+    if last_name:
+        entity = context.make("Person")
+        entity.id = context.make_id(last_name, license_number)
```
jbothma: sorry I didn't see this before - let's add first name too.

datasets/us/tx/us_tx_oig_exclusions.yml
```diff
@@ -0,0 +1,127 @@
...
+        prop: registrationNumber
+  aliases:
+    options:
+      - match: Pro Medic EMS, LLC dba Pro Medic EMS, LLC
+        aliases:
+          - Pro Medic EMS, LLC
+      - match: Executive Healthcare, LLC d/b/a The Apothecary Sho
+        aliases:
+          - Executive Healthcare, LLC
+          - The Apothecary Sho
```
jbothma: we can fix typos or data errors like this when we spot them and find sufficient evidence for the correct value fairly quicly

Image

```suggestion
      - match: Executive Healthcare, LLC d/b/a The Apothecary Sho
        aliases:
          - Executive Healthcare, LLC
          - The Apothecary Shop
```

datasets/us/tx/us_tx_oig_exclusions.yml
```diff
@@ -0,0 +1,127 @@
...
+      - match: ERX Consultants LLC, dba Zoetic Pharmacy
+        aliases:
+          - ERX Consultants LLC
+          - Zoetic Pharmacy
+      - match: Ukie Emergency Medical Services, Inc. d/b/a Ukie E
+        aliases:
+          - Ukie Emergency Medical Services, Inc
+          - Ukie E
```
jbothma:
```suggestion
          - Ukie EMS
```

https://npiregistry.cms.hhs.gov/provider-view/1568762821

datasets/us/tx/us_tx_oig_exclusions.yml
```diff
@@ -0,0 +1,127 @@
...
+          - Pharmplex LTD III
+          - Penitas Family Pharmacy
+      - match: Western Dental dba West Texas Dental
+        aliases:
+          - Western Dental
+          - West Texas Dental
+      - match: M Ayman Ghraowi, MD PA dba South Texas Comprehensi
+        aliases:
+          - M Ayman Ghraowi, MD PA
+          - South Texas Comprehensi
```
jbothma: Image

datasets/us/tx/us_tx_oig_exclusions.yml
```diff
@@ -0,0 +1,127 @@
...
+      - match: Southwest Medical Consultants Inc. d/b/a Los Ninos
+        aliases:
+          - Southwest Medical Consultants Inc
+          - Los Ninos
+      - match: Acosta Medical Supply, Inc. dba Affordab
+        aliases:
+          - Acosta Medical Supply, Inc
+          - Affordab
```
jbothma: etc

datasets/us/tx/crawler.py
herrhellana: i thought of it but we have some entities just with last_name and not first_name, so im not sure what to do with these; do we want to drop em?

datasets/us/tx/us_tx_oig_exclusions.yml
herrhellana: oh yep im trying to do this whenever i can, this is a human oversight (aka blind copy-paste), sorry! could be cool to have safeguards against it, too lol

datasets/us/tx/us_tx_oig_exclusions.yml
herrhellana: ok taking my words back, didn't read company names that carefully in this crawler. ill fix the rest!

datasets/us/tx/crawler.py
jbothma: remember it's ok with blanks (None, ""), so just do context.make_id(first_name, middle_name, last_name, license_number) for all persons and it will do the right thing (empty string becomes null, null doesn't become part of the key, happy happy)

datasets/us/tx/us_tx_oig_exclusions.yml
jbothma: no stress :) I spotted them cos I'm reviewing. If you don't already, it could be good to read your own review as though you're checking someone else's work.

herrhellana: some alias names are definitely trimmed but not that easy to look up; i can mark these in comments?

datasets/us/tx/crawler.py
herrhellana: ah you were talking about context.make_id(), gotcha. yes, makes a lot of sense!

datasets/us/tx/med_exclusions/us_tx_med_exclusions.yml
```diff
@@ -0,0 +1,127 @@
...
+          - Western Dental
+          - West Texas Dental
+      - match: M Ayman Ghraowi, MD PA dba South Texas Comprehensi
+        aliases:
+          - M Ayman Ghraowi, MD PA
+          - South Texas Comprehensive Cancer Centers PLLC (STCCC)
+      - match: Southwest Medical Consultants Inc. d/b/a Los Ninos
+        aliases:
+          - Southwest Medical Consultants Inc
+          - Los Ninos
```
jbothma:
```suggestion
          - Los Ninos Pediatric Rehab
```

datasets/us/tx/med_exclusions/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+    else:
+        entity = context.make("Company")
+        entity.id = context.make_id(company_name, npi)
+        entity.add("name", company_name)
+        entity.add("description", occupation)
+
+        if contains_split_phrase(company_name):
+            res = context.lookup("aliases", company_name, warn_unmatched=True)
+            aliases = res.aliases if res else []
+            entity.add("alias", aliases)
```
jbothma: ah - what we actually want here is

- if it contains a split phrase and we found a lookup
  - put the cleaned name in the name prop, passing the original name as original_value
  - put any aliases in the alias prop, passing the original name as original_value
- else
  -  put the original name in the name prop

jbothma: FYI I found this by running and exporting locally and scrolling through names.txt to look for anything weird at a glance - I found names with ` dba ` still in them.

datasets/us/tx/med_exclusions/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+
+    entity.add("idNumber", license_number)
+    entity.add("npiCode", npi)
+    entity.add("country", "us")
+    entity.add("topics", "debarment")
+
+    sanction = h.make_sanction(context, entity)
+    h.apply_date(sanction, "startDate", drow.pop("startdate"))
+    h.apply_date(sanction, "listingDate", drow.pop("adddate"))
+    h.apply_date(sanction, "endDate", drow.pop("reinstateddate"))
```
jbothma: sorry for just remembering this.

So we don't want to add the debarment topic to everyone - only for those whose debarment is still active. We do want to include the listed-but-not-debarred-any-more people with their Sanction entity explaining why they're here.

The way we normally do that is roughly

```
sanction = h.make_sanction()
h.apply_date(sanction, "endDate")
if h.is_active(sanction):
    entity.add("topic", "debarment")
```

PR #3439, opened 2026-02-03
Make `uv sync` just work by switching to dependency-groups and installing deps & docs by default

leonhandreke:
See for why this is better: https://peps.python.org/pep-0735/.

Claude:

> Migrate development dependencies from project.optional-dependencies to standardized dependency-groups table (PEP 735). This change properly separates published package extras from local-only development dependencies, ensuring dev tooling (pytest, ruff, etc.) is not included in package metadata when published to PyPI. The migration maintains compatibility with both uv (which has supported dependency-groups since its initial release) and pip 25.1+ (released April 2025), which added --group flag support.

I don't really care that it's better, I just want `uv sync` to do the right thing by default.

PSA: If you're using pip, instead of `pip install ".[dev]"`, you'll have to do `pip install --group dev .` now.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,7 @@ jobs:
         working-directory: zavod
         run: |
           python -m pip install --upgrade pip wheel
-          pip install --no-cache-dir -q -e ".[dev,dspy]"
+          pip install --no-cache-dir -q --group dev --group dspy -e "."
           pip freeze
       - name: Test crawlers with tests
         run: pytest datasets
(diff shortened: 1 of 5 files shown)


jbothma (approved):
Looks good to me.

I'm not sure whether we have any customers running this or if we want to alert external users of zavod. We should at least note internally that people will need to upgrade pip to 25.1 or up. The python installations I have came with 24 and 25.0 so I had to upgrade pip before I could update deps with this change.

Another thing to be aware of in case someone gets stuck - it looks like you now have to be in the project directory to install, e.g. I can't be in `opensanctions`, I have to be in `zavod` for pip to find the pyproject.toml.

PR #3451, opened 2026-02-04
Fixes-04-02

nvmbrasserie:


datasets/id/dttot/crawler.py
```diff
@@ -44,109 +26,54 @@ def clean_addresses(raw_addresses: str) -> list[str]:
...
+    doc = context.fetch_html(context.data_url)
+    xlsx_link = h.xpath_string(
+        doc,
+        ".//p[contains(., 'Daftar Terduga Teroris')]//a[contains(@href, '.xlsx')]/@href",
+    )
+    path = context.fetch_resource("source.xlsx", xlsx_link)
+    context.export_resource(path, XLSX, title=context.SOURCE_TITLE)
+    workbook: openpyxl.Workbook = openpyxl.load_workbook(path, read_only=True)
+    for row in h.parse_xlsx_sheet(
+        context, workbook["Export"], header_lookup=context.get_lookup("headers")
```
jbothma: could also assert num sheets is 1

datasets/id/dttot/crawler.py
```diff
@@ -44,109 +26,54 @@ def clean_addresses(raw_addresses: str) -> list[str]:
...
+
+    dob_raw = row.pop("birth_date")
+    birth_place = row.pop("birth_place")
+    if entity.schema.is_a("Person"):
+        entity.add("birthPlace", birth_place)
+        for dob in h.multi_split(dob_raw, [" atau ", "\n", "- ", " / "]):
+            if match := DATE_RANGE_RE.match(dob):
+                # Date range like "00/00/1970-1973" - add both years
+                for dob in list(match.groups()):
+                    h.apply_date(entity, "birthDate", dob)
```
jbothma:
```suggestion
                h.apply_dates(entity, "birthDate", list(match.groups()))
```

datasets/id/dttot/crawler.py
```diff
@@ -44,109 +26,54 @@ def clean_addresses(raw_addresses: str) -> list[str]:
...
-            h.apply_date(entity, prop, value)
-    else:
-        raise ValueError(f"Unexpected value type {type(value)}")
+def crawl_row(context: Context, row: Dict[str, str | None]):
+    item_id = row.pop("id")
+    res = context.lookup("type", row.pop("type"), warn_unmatched=True)
+    assert res and res.value
+    entity = context.make(res.value)
+    names = h.multi_split(row.pop("name"), ["alias", "ALIAS"])
+    entity.id = context.make_id(item_id, *names)
```
jbothma, line 35: looks like about 1/3 are getting re-keyed. 
I don't think we need to change this to use the raw string now, which would rekey most entities forcing us to do a graceful migration now, but maybe just try and do a dedupe on this ASAP after merge/rerun

```
Delta export complete          [id_dttot] added=375 dataset=id_dttot deleted=371 metric=delta_counts modified=694
```

datasets/id/dttot/crawler.py
nvmbrasserie: just saw it in your another review, completely forgot here. on it!

PR #3452, opened 2026-02-04
Norwegian Legal Entities (BRREG)

herrhellana:
[issue](https://github.com/opensanctions/crawler-planning/issues/592)

### DIFF
--- a/datasets/no/brreg/crawler.py
+++ b/datasets/no/brreg/crawler.py
@@ -99,14 +99,14 @@ def crawl_company(
     row: Dict[Any, Any],
     company_name: str,
     org_number: str,
-    company_juris: str,
 ) -> Entity:
-    entity = context.make("LegalEntity")
-    entity.id = context.make_id(company_name, org_number)
+    entity = context.make("Organization")
+    entity.id = context.make_id(org_number)
     entity.add("name", company_name)
     entity.add("alias", row.pop("foreign_register_name"))
-    entity.add("jurisdiction", company_juris)
+    entity.add("jurisdiction", row.pop("subject_to_legislation_country_code"))
     entity.add("legalForm", row.pop("org_form_description"))
+    entity.add("registrationNumber", org_number)
     entity.add("registrationNumber", row.pop("registration_number_home_country"))
     # about industry codes: https://www.brreg.no/bedrift/naeringskoder/
     industry_codes = {
@@ -144,13 +144,12 @@ def crawl_soe_peps(
     context: Context,
     org_number: str,
     company_name: str,
-    company_juris: str,
     entity: Entity,
-) -> int:
+) -> None:
     api_org_url = f"{BASE_API}/enheter/{org_number}/roller"
     rollegrupper = context.fetch_json(api_org_url)["rollegrupper"]
 
-    pos_in_company = 0
+    num_positions = 0
     for group in rollegrupper:
         group_type = group.get("type")
         group_code = group_type.get("kode")
@@ -185,10 +184,7 @@ def crawl_soe_peps(
             birth_date = person_data.pop("fodselsdato")
 
             pep = context.make("Person")
-            if birth_date:
-                pep.id = context.make_id(first_name, last_name, birth_date)
-            else:
-                pep.id = context.make_id(first_name, last_name, company_name)
+            pep.id = context.make_id(first_name, last_name, birth_date, company_name)
 
             h.apply_name(pep, first_name=first_name, last_name=last_name)
             pep.add("birthDate", birth_date)
@@ -198,8 +194,7 @@ def crawl_soe_peps(
                 context,
                 name=f"{role_desc}, {company_name}",
                 topics=["gov.soe"],
-                # adding company's country of jurisdiction, too
-                country=["no", company_juris],
+                country=["no"],
                 organization=entity,
             )
             categorisation = categorise(context, position, is_pep=True)
@@ -222,22 +217,18 @@ def crawl_soe_peps(
                 context.emit(position)
                 context.emit(occupancy)
 
-                pos_in_company += 1
+                num_positions += 1
 
-    if pos_in_company < 1 or pos_in_company > 40:
+    if num_positions < 1 or num_positions > 40:
         context.log.warning(
             "Expected between 1 and 40 positions per company.",
             company_name=company_name,
             org_number=org_number,
-            emitted_positions=pos_in_company,
+            emitted_positions=num_positions,
         )
-    return pos_in_company
 
 
 def crawl(context: Context) -> None:
-    total_positions = 0
-    n_companies = 0
-
     page = context.fetch_html(
         context.data_url, method="GET", absolute_links=True, cache_days=3
     )
@@ -256,8 +247,10 @@ def crawl(context: Context) -> None:
             for header in nok_header_row
         ]
 
+    with gzip.open(path, "rt") as f:
         # iterate over the rows with english field names
         dict_reader = csv.DictReader(f, fieldnames=eng_header_row)
+        next(dict_reader)  # skip header row
         for row in dict_reader:
             row = {k: v.strip() for k, v in row.items() if k is not None}
 
@@ -268,25 +261,16 @@ def crawl(context: Context) -> None:
 
             company_name = row.pop("name")
             org_number = row.pop("org_number")
-            company_juris = row.pop("subject_to_legislation_country_code")
-            n_companies += 1
 
             entity = crawl_company(
                 context,
                 row,
                 company_name,
                 org_number,
-                company_juris,
             )
-            total_positions += crawl_soe_peps(
+            crawl_soe_peps(
                 context,
                 org_number,
                 company_name,
-                company_juris,
                 entity,
             )
-
-        assert n_companies <= total_positions, (
-            f"Sanity check failed: extracted {total_positions} positions from {n_companies} SOEs, "
-            "expected at least one position per company"
-        )
--- a/datasets/no/brreg/no_brreg.yml
+++ b/datasets/no/brreg/no_brreg.yml
@@ -4,14 +4,14 @@ prefix: no-nbim
 coverage:
   frequency: monthly
   start: 2026-02-02
-load_statements: false # we'll add a second crawler that enriches against this data that will get added to the default collection (via enrichers)
+load_statements: true
 summary: >-
-  C-level execitives and board members of state-owned companies from the official registry
-  of Norwegian legal entities with organisation details, addresses,
-  industry codes, and registration statuses from Brønnøysund Register Centre.
+  Executives and board members of state-owned companies from the official registry
+  of Norwegian legal entities
 description: |
-  The dataset contains information about c-level executives from all registered state-owned legal entities in Norway,
+  The dataset contains information about executives from all registered state-owned legal entities in Norway,
   sourced from two registers managed by the Brønnøysund Register Centre:
+
       1. The Central Coordinating Register for Legal Entities (CCR / Enhetsregisteret):
        A central registry of basic data about legal persons and other entities in Norway.
        The CCR assigns unique nine-digit organisation numbers for identification of entities.
@@ -42,10 +42,14 @@ dates:
 assertions:
   min:
     schema_entities:
-      Person: 1600 # raw count: 2021 PEPs
+      Organization: 600 # As of 2026 we expect 703
+      Position: 1800 # Assuming generally at least 3 per company
+      Person: 1900 # As of 2026 we expect 2016. Generally around 3
   max:
     schema_entities:
-      Person: 4050
+      Organization: 1200
+      Position: 3600
+      Person: 4000
 
 ### dict for header: https://data.brreg.no/enhetsregisteret/api/dokumentasjon/en/index.html#section/Dictionary
 lookups:
@@ -81,10 +85,12 @@ lookups:
         value: has_registered_employee_count
       - match: antallAnsatte
         value: employee_count
+        # Enhetsregisteret is the Norwegian Register of Legal Entities
       - match: registreringsdatoAntallAnsatteEnhetsregisteret
-        value: registration_date_employee_count_entity_register # Enhetsregisteret is the Norwegian Register of Legal Entities
+        value: registration_date_employee_count_entity_register
+        # NAV Aa-registeret is the Norwegian Labour and Welfare Administration's Employment Register
       - match: registreringsdatoantallansatteNAVAaregisteret
-        value: registration_date_employee_count_nav_employment_register # NAV Aa-registeret is the Norwegian Labour and Welfare Administration's Employment Register
+        value: registration_date_employee_count_nav_employment_register
       #
       - match: hjemmeside
         value: website
@@ -187,8 +193,9 @@ lookups:
       - match: tvangsavvikletPgaManglendeSlettingDato
         value: forced_liquidation_missing_deletion_date
       #
+      # Organisation number of superior entity in public sector. String of 9 digits
       - match: overordnetEnhet
-        value: parent_entity_org_number # Organisation number of superior entity in public sector. String of 9 digits
+        value: parent_entity_org_number
       - match: maalform
         value: language_form
       - match: vedtektsdato
@@ -247,8 +254,9 @@ lookups:
         value: company_form_home_country_code
       - match: foretaksformIHjemlandet.beskrivelse
         value: company_form_home_country_description
+        # Bokmål is one of the two official written standards of the Norwegian language
       - match: foretaksformIHjemlandet.beskrivelseBokmaal
-        value: company_form_home_country_description_bokmaal # Bokmål is one of the two official written standards of the Norwegian language
+        value: company_form_home_country_description_bokmaal
 
   # all positions listed here: https://data.brreg.no/enhetsregisteret/api/roller/rollegruppetyper
   is_pep_role:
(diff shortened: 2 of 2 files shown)


datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,214 @@
...
+    path = context.fetch_resource("source.csv.gz", url)
+    context.export_resource(path, GZIP, title=context.SOURCE_TITLE)
+
+    with gzip.open(path, "rt") as f:
+        reader = csv.reader(f)  # single-pass iterator
+
+        headers: Dict[str, int] = {}
+        for col_idx, cell in enumerate(next(reader)):
+            val = cell.strip()
+            headers[context.lookup_value("headers", val, val)] = col_idx
```
jbothma: Maybe a simpler way to do this is something like

```python
# Read the norwegian header
with gzip.open(path, "rt") as f:
    reader = csv.reader(f) 
    nok_header_row = next(reader)

# translate to english headers
eng_header_row = []
for header in nok_header_row:
    eng_header_row.append(context.lookup(..., header, warn_if...=True))

# iterate over the rows with english field names
with gzip.open(path, "rt") as f:
    reader = csv.DictReader(f, headers=eng_header_row)
    for row in reader:
        # if we need the stripping:
        row = {k: v.strip() for k, v in row}
        crawl_row(context, row)
```

- that removes the need for row_to_dict and the column index based logic
- falling back to the norwegian header if we're missing a translation is a nice touch.
- warning if we don't have a translation isn't critical because other things will warn/fail but it might help identify the root cause

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,214 @@
...
+
+        # k = 0
+        for row in reader:
+            # k += 1
+            dict_row = row_to_dict(row, headers)
+            crawl_row(dict_row, context)
```
jbothma: totally a nitpick but we tend to put context as the first argument - it's almost a bit like "self" in significance

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,214 @@
...
+
+    oldest = min(dict_row[k] for k in present)
+    for k in present:
+        del dict_row[k]
+    return str(oldest)
+
+
+def crawl_row(dict_row: Dict[str, Any], context: Context) -> None:
+    name = dict_row.pop("name")  # what about foreign_register_name?
+    id = dict_row.pop("id")
```
jbothma: it could be nice to give this a slightly more descriptive name like "company_number" or "registration_number". This is like the national ID of a person or company.

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,214 @@
...
+
+    entity = context.make("LegalEntity")
+    entity.id = context.make_id(name, id)
+    entity.add("name", name)
+    entity.add(
+        "jurisdiction", "no"
+    )  # note there is also subject_to_legislation_country_code in input source, maybe use it too?
```
jbothma: yeah that feels like a really good fit for the jurisdiction prop.

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,214 @@
...
+    # --- not sure about these --- :
+    is_bankrupt = dict_row.pop("bankruptcy") == "true"
+    is_liquidated = dict_row.pop("under_forced_liquidation_or_dissolution") == "true"
+
+    if is_bankrupt and is_liquidated:
+        entity.add("status", "bankrupt and under forced liquidation or dissolution")
+    elif is_bankrupt:
+        entity.add("status", "bankrupt")
+    elif is_liquidated:
+        entity.add("status", "under forced liquidation or dissolution")
```
jbothma: we don't transform data into strings that aren't in the text. (ok, we're slowly phasing in some translation but that's still fully grounded in the source text and the original text is also added).

I do understand the temptation to do so but we really try to reflect only what's in the source, i.e. nothing "in our own words".

Maybe the best way to maintain these principles is to not translate these column names, and if their value is true, add the column name to the "status" property. Then status becomes e.g. "konkurs" (nok) and we can leave translation to the future.

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,214 @@
...
+
+    # registrationNumber
+    entity.add(
+        "registrationNumber", dict_row.pop("registration_number_home_country", None)
```
jbothma: id also goes in here

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,214 @@
...
+    )
+    industry_codes = {
+        k: v
+        for k, v in dict_row.items()
+        if k.startswith("industry_code") and k.endswith("code")
+    }  # about industry codes: https://www.brreg.no/bedrift/naeringskoder/
+    for k, v in industry_codes.items():
+        entity.add("registrationNumber", v)
```
jbothma: these rather go into LegalEntity:sector

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,214 @@
...
+
+    # registrationNumber
+    entity.add(
+        "registrationNumber", dict_row.pop("registration_number_home_country", None)
+    )
+    industry_codes = {
+        k: v
+        for k, v in dict_row.items()
+        if k.startswith("industry_code") and k.endswith("code")
+    }  # about industry codes: https://www.brreg.no/bedrift/naeringskoder/
```
jbothma: I'm sure you'll remember from texas, but documenting comments generally go on the line above unless they fit without wrapping.

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,214 @@
...
+    for k, v in industry_codes.items():
+        entity.add("registrationNumber", v)
+        dict_row.pop(k)
+
+    # contacts and addresses
+    email = context.lookup("email", dict_row.pop("email", None))
+    if email is not None and email.values:
+        entity.add("email", email.values)
+    else:
+        entity.add("email", email)
```
jbothma: a cleaner way of doing this is to just let the email validator complain about them, so remove all the logic except entity.add("email", row.pop("email"))

then rename your "email" lookup to "type.email" and let zavod to the work for you

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,214 @@
...
+    )
+    COLS_NOT_IN_USE = (
+        [
+            "org_form_code",
+            "support_unit_code",
+            "support_unit_code_description",
+            "institutional_sector_code",
+            "institutional_sector_description",
+            "last_submitted_annual_accounts",
+            "bankruptcy_date",
```
jbothma: another dissolution date

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,214 @@
...
+            "under_liquidation",
+            "language_form",
+            "articles_of_association_date",
```
jbothma: another registration date input

jbothma: actually probably not quite applicable to that prop

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,214 @@
...
+    Get the oldest date value from keys to check, delete the rest.
+    """
+    present = [k for k in keys_to_check if k in dict_row]
+    if not present:
+        return None
+
+    oldest = min(dict_row[k] for k in present)
+    for k in present:
+        del dict_row[k]
+    return str(oldest)
```
jbothma: if you do the row to dict with DictReader as I suggest, you can assume every field exists, then this logic might become simpler - just a matter of popping each field and taking the min.

datasets/no/brreg/no_brreg.yml
```diff
@@ -0,0 +1,286 @@
...
+          - 'valhallaoslovikinglag'
+          - '91812020'
+          - '97796058'
+          - 'Ubonthaimassasje'
+          - '91708719'
+          - '91581052'
+          - 'facebook'
+        value: null
+  country_code:
+      options:
```
jbothma: two spaces :)

datasets/no/brreg/no_brreg.yml
```diff
@@ -0,0 +1,286 @@
...
+      - match: cd
+        value: cg
```
jbothma: what's the context for remapping this?
It looks like both are valid, but different.
https://www.iso.org/obp/ui/#iso:code:3166:CG
https://www.iso.org/obp/ui/#iso:code:3166:CD

Image

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,214 @@
...
+    entity.add("website", website)
+    entity.add("phone", dict_row.pop("phone", None))
+    entity.add("phone", dict_row.pop("phone_mobile", None))
+    h.apply_date(entity, "incorporationDate", get_oldest_date(dict_row, INC_DATES))
+    h.apply_date(entity, "dissolutionDate", get_oldest_date(dict_row, DISSOLV_DATES))
+
+    country_code_business = (
+        dict_row.pop("business_address_country_code", None) or ""
+    ).lower() or None
+    country_code_business = context.lookup_value("country_code", country_code_business)
```
jbothma: another one where type.country would let zavod do the lookup for you upon `add()`

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,214 @@
...
+            "is_in_corporate_group",
+            "foreign_register_name",  # maybe we need it? as an alias maybe? it is tied to foreign address
+            "parent_entity",  # do we need to add it?
+        ]
+        + [
+            k
+            for k in dict_row
+            if k.startswith("industry_code") and k.endswith("description")
+        ]
+        + [k for k in dict_row if any(substr in k for substr in COLS_NOT_IN_USE_SUBSTR)]
```
jbothma: maybe these can be assigned to descriptive variable names and then the three lists can be concatenated. That just takes a little bit of cognitive load off the reader.

jbothma:
Remember to grab the directors from the API

> Using the company ID from the open data JSON file (above), you can then use the public API to acquire the board of directors:
>
> https://data.brreg.no/enhetsregisteret/api/dokumentasjon/no/index.html#tag/Roller

datasets/no/brreg/no_brreg.yml
```diff
@@ -0,0 +1,286 @@
+title: Main Entities from The Central Coordinating Register for Legal Entities and The Register of Business Enterprises
```
jbothma: I think this could be a bit more concise and start with the Norway prefix. Perhaps


```suggestion
title: Norway Register of Business Enterprises Main Entities
```

datasets/no/brreg/no_brreg.yml
```diff
@@ -0,0 +1,286 @@
...
+entry_point: crawler.py
+prefix: no-nbim
+coverage:
+  frequency: weekly  # how often shall we crawl it?
```
jbothma: typically monthly for company registers

datasets/no/brreg/no_brreg.yml
```diff
@@ -0,0 +1,286 @@
...
+  start: 2026-02-02
+load_statements: true
```
jbothma:
```suggestion
load_statements: false
```

for company registers. We'll add a second crawler ext_no_brreg which enriches against this data. Its data will get added to the `default` collection (via `enrichers`) and its load_statements will be true.

Would you mind reflecting the comments on load_statements in this and the texas PRs in the metadata docs page?

datasets/no/brreg/no_brreg.yml
```diff
@@ -0,0 +1,286 @@
...
+summary: >-
+  Complete official registry of Norwegian legal entities with organisation details, addresses,
+  industry codes, employee data, and registration statuses from Brønnøysund Register Centre.
```
jbothma: no employee data.
not entirely complete since we're not including parties, non profits, and potentially not state owned enterprises.

datasets/no/brreg/no_brreg.yml
```diff
@@ -0,0 +1,286 @@
...
+description: |
+  The dataset contains information about all registered main entities in Norway,
+  sourced from two primary registers managed by the Brønnøysund Register Centre:
```
jbothma: do we source from two registers? At the moment we're sourcing from one register, right? Do they source from two?

When we add the director search via that API mentioned in the ticket, if that's a separate register, documenting it as two registers might make sense after all.

datasets/no/brreg/no_brreg.yml
```diff
@@ -0,0 +1,286 @@
...
+       The CCR assigns unique nine-digit organisation numbers for identification of entities.
+       It serves as a "master register" coordinating data with other Norwegian registers.
+
+      2. The Register of Business Enterprises (Foretaksregisteret):
+       Contains detailed information about businesses registered in Norway, including
+       company structure, management, ownership, and legal obligations.
+
+url: https://www.brreg.no/en/use-of-data-from-the-bronnoysund-register-centre/datasets-and-api/data-about-organisations/
+publisher:
+  name: Brønnøysund Register Centre
```
jbothma, line 28: it's fun to put the native language name in `name` and the English version (this) in `name_en`

datasets/no/brreg/no_brreg.yml
```diff
@@ -0,0 +1,286 @@
...
+  acronym: BRREG
+  description: |
+    Brønnøysund Register Centre is a Norwegian government agency that is responsible for the management of numerous
+    public registers for Norway, and governmental systems for digital exchange of information.
+  country: "no"
+  url: https://www.brreg.no/en/
+  official: true
+data:
+  url: https://www.brreg.no/en/use-of-data-from-the-bronnoysund-register-centre/datasets-and-api/data-about-organisations/
+  format: ZIP
```
jbothma: this isn't technically zip, it's gzip. But I think referring to the underlying format we're using (CSV) is more interesting here

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,214 @@
...
+    name = dict_row.pop("name")  # what about foreign_register_name?
+    id = dict_row.pop("id")
+
+    entity = context.make("LegalEntity")
+    entity.id = context.make_id(name, id)
+    entity.add("name", name)
+    entity.add(
+        "jurisdiction", "no"
+    )  # note there is also subject_to_legislation_country_code in input source, maybe use it too?
+    entity.add("description", dict_row.pop("org_form_description", None))
```
jbothma: legal_form prop

jbothma: Also, what we often do, and what's quite cool for the data, is to do a lookup from from these to Schema

```
jq .[].organisasjonsform.beskrivelse enheter_alle.json | sort | uniq -c | sort -rn
450724 "Enkeltpersonforetak"
419828 "Aksjeselskap"
127176 "Forening/lag/innretning"
32444 "Eierseksjonssameie"
27536 "Utenlandsk enhet"
25692 "Norskregistrert utenlandsk foretak"
10842 "Ansvarlig selskap med delt ansvar"
10130 "Borettslag"
7670 "Konkursbo"
```

e.g. this one's a Person

```json
  {
  "organisasjonsnummer" : "811544922",
  "navn" : "KRISTINA THOMASSEN",
  "organisasjonsform" : {
    "kode" : "ENK",
    "beskrivelse" : "Enkeltpersonforetak",
    "links" : [ ]
  },
```

technically it's more correct to do the mapping on the org form code, but I imagine the descriptions will change so rarely, so it won't be a maintenance burden, and it'll be LOT more convenient to just read the norwegian labels in the mapping.

Bonus points for comments with the english translation.

anything you're unsure of, just map to LegalEntity.

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,214 @@
...
+    entity.add("summary", dict_row.pop("statutory_purpose", None))
```
jbothma: I reckon we can leave this out for now and not inflate the data with this

datasets/no/brreg/no_brreg.yml
herrhellana: what would be the eng name, just dropping the norwegian special chars?

datasets/no/brreg/no_brreg.yml
herrhellana: gzip.csv maybe?

datasets/no/brreg/no_brreg.yml
herrhellana: changed to `false` and included a comment about it in the .yml file. didnt get the texas PR bit -- are you talking about `ci_test`?

datasets/no/brreg/no_brreg.yml
herrhellana: the dataset name is "Main entities from The Central Coordinating Register for Legal Entities and The Register of Business Enterprises" so i assumed two registers...

datasets/no/brreg/crawler.py
herrhellana: wym, which ID?

datasets/no/brreg/crawler.py
jbothma:
```
450724 "Enkeltpersonforetak" -> LegalEntity # because sometimes it's a company name
419828 "Aksjeselskap" -> Company
127176 "Forening/lag/innretning" -> Organization
32444 "Eierseksjonssameie" -> LegalEntity
27536 "Utenlandsk enhet"
25692 "Norskregistrert utenlandsk foretak"
10842 "Ansvarlig selskap med delt ansvar"
10130 "Borettslag"
7670 "Konkursbo"
```

https://github.com/opensanctions/opensanctions/commit/537b1b5760bb1bed91eb6444e032b0d736c20c30

context.lookup("schema", company_form, "LegalEntity", warn_if_not_matched=True)

herrhellana: among state-owned entities im crawling, there are

```
 685 Aksjeselskap -> Joint stock company
   7 Ansvarlig selskap med delt ansvar -> General partnership
   5 Statsforetak -> State enterprise
   4 Allmennaksjeselskap -> Public limited company
   2 Stiftelse -> Foundation
   1 Kommandittselskap -> Limited partnership
   1 Ansvarlig selskap med solidarisk ansvar -> Joint liability partnership
```
   
is it safe to emit all these as LegalEntity? Or emit Aksjeselskap and Allmennaksjeselskap as Company, and the rest is LegalEntity? Or we do not care that much about it here bc n=705?

datasets/no/brreg/crawler.py
```diff
@@ -28,109 +29,152 @@
...
         values.append(v)
     return min(values) if values else None
 
 
-def crawl_row(context: Context, dict_row: Dict[Any, Any]) -> None:
-    # === get state-owned enterpreses details === #
-    company_name = dict_row.pop("name")
-    org_number = dict_row.pop("org_number")
-    company_juris = dict_row.pop("subject_to_legislation_country_code", None)
+def crawl_address(context: Context, row: Dict[str, Any], entity: Entity) -> None:
```
herrhellana: yes, makes sense for readability, thank you @nvmbrasserie

datasets/no/brreg/crawler.py
```diff
@@ -196,81 +169,75 @@ def crawl_pep(
...
-                categorisation=categorisation,
-                status=OccupancyStatus.UNKNOWN,
-            )
-
-            if occupancy is not None:
-                context.emit(pep)
-                context.emit(entity)
-                context.emit(position)
-                context.emit(occupancy)
+        if role_code not in ROLE_CODES:
```
herrhellana: fair enough

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,263 @@
...
+DISSOLV_DATES = [
+    "bankruptcy_date",
+    "under_liquidation_date",
+    "forced_dissolution_missing_ceo_date",
+    "forced_dissolution_missing_auditor_date",
+    "forced_dissolution_missing_accounts_date",
+    "forced_dissolution_deficient_board_date",
+    "forced_liquidation_missing_deletion_date",
+]
+# should we put this there ^: under_foreign_insolvency_proceedings_date ?
```
jbothma: nah - should be fine to ignore

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,263 @@
...
+        person_data = role.get("person")
+        if person_data is None:
+            continue
+        names = person_data.get("navn")
+        first_name = names.pop("fornavn")
+        last_name = names.pop("etternavn")
+
+        pep = context.make("Person")
+        pep.id = context.make_id(first_name, last_name)
+        pep.add("birthDate", person_data.pop("fodselsdato"))
```
jbothma: definitely worth including birth date in make_id to give more uniqeness. If you didn't have birth date, I'd say put company name in there, even if it's really the same person. It's better to have a false duplicate than a false merge

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,263 @@
...
+        pep.add("firstName", first_name)
+        pep.add("lastName", last_name)
```
jbothma: we normally add firstName and lastName using h.apply_name(...)
That will construct a full name for us, add that, and also add these parts.

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,263 @@
...
+            place=row.pop(f"{prefix}_municipality"),
+            country_code=country.lower() if country else None,
+        )
+        h.copy_address(entity, address)
+
+    context.audit_data(row, IGNORE)
+    return entity
+
+
+def crawl_pep(
```
jbothma:
```suggestion
def crawl_soe_peps(
```

considering this is intended to crawl all of them

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,263 @@
...
+            name=f"{role_name}, {company_name}",
+            topics=["gov.soe"],
+            # adding company's country of jurisdiction, too
+            country=["no", company_juris],
+            organization=entity,
+        )
+        categorisation = categorise(context, position, is_pep=True)
+
+        if not categorisation.is_pep:
+            return
```
jbothma:
```suggestion
            continue
```

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,263 @@
...
+    api_org_url = f"{BASE_API}/enheter/{org_number}/roller"
+    rollegrupper = context.fetch_json(api_org_url)["rollegrupper"]
+
+    # flatten to iterate over "rollegrupper" that contains a list of roles
+    roles = chain.from_iterable(group.get("roller") for group in rollegrupper)
+    for role in roles:
+        role_data = role.get("type")
+        role_code = role_data.get("kode")
+
+        if role_code not in ROLE_CODES:
```
jbothma: it looks like we're only finding a bunch of `Daglig leder` and one `Forretningsfører` so something's going wrong, but I'm not sure what.

```
grep Position data/datasets/no_brreg/entities.ftm.json | jq  '.properties.name[0]'|sort
```

It does look like more roles are available

```
curl https://data.brreg.no/enhetsregisteret/api/enheter/985198292/roller|jq .
```

Image

datasets/no/brreg/crawler.py
```diff
@@ -0,0 +1,263 @@
...
+    org_number: str,
+    company_juris: str,
+) -> Entity:
+    entity = context.make("LegalEntity")
+    entity.id = context.make_id(company_name, org_number)
+    entity.add("name", company_name)
+    entity.add("alias", row.pop("foreign_register_name"))
+    entity.add("jurisdiction", company_juris)
+    entity.add("legalForm", row.pop("org_form_description"))
+    entity.add("registrationNumber", row.pop("registration_number_home_country"))
```
jbothma, line 110: for some reason I'm not seeing registrationNumber for Avinor even though we have one PEP from there so far.

jbothma (commented):
Nice!

I think there's just one weirdness stoppinng us getting all the PEPs then we're good to go

datasets/no/brreg/crawler.py
herrhellana: ah right, sorry forgot about it

datasets/no/brreg/crawler.py
herrhellana: ok great one! first of all, i was filtering by kode types in rollegrupper > roller > kode, but we also have kode types in rollegrupper itself. and looking at your curl statement, i realized there are a bunch of roles that are not in ROLE_CODES but part of rollegrupper that satisfies our filter criterion (rollergrupper's kode is in ROLE_CODES). 
 
so i rewrote the code to filter by rollegrupper's kode instead and catch all the roles inside it even if the role's kode is not in ROLE_CODES (let me know if this needs to be changed, this is true only for STYR rollergrupper). E.g. (first is rollergrupper's kode, second is role's kode inside the rollergrupper):
```
STYR LEDE
STYR MEDL
STYR MEDL
DAGL DAGL
STYR LEDE
STYR MEDL
STYR MEDL
STYR MEDL
```

then i also realized that STYR was misspelled (lowercase instead of uppercase), ie we weren't scraping it... fixed it!

herrhellana: i also checked the inverse, i.e. when rollergrupper's kode is not in ROLE_CODES (renamed to GROUP_CODES) but it has roles that are in  _CODES -- it seems like we don't have situations like that, so probably safe to just filter by rollergrupper's kode type? at least this is the logic i've implemented, but i can also scrape data for a case like that if this changes in the future

datasets/no/brreg/crawler.py
herrhellana: it looks like registreringsnummerIHjemlandet (registrationNumber) is null for it? https://data.brreg.no/enhetsregisteret/api/enheter/985198292

datasets/no/brreg/crawler.py
jbothma: Great! Will review in a bit. The important lesson here is to think about every dimension, e.g. number of companies * number of positions per company, and satisfy yourself that you're really extracting 

- something from each company
- all the things in scope for any given company

then add assertions that help verify that. Let's add a data assertion for the min/max Position entities expected.

jbothma: It can be helpful to add comments beside the assertion for a thumb-suck/heuristic for validating the numbers, e.g. roughly n companies x roughly m positions per company = 200, then you add your min and max assertion and see if it fits

herrhellana: right, assertion there is a great idea instead of setting up an extra guardrail (scrapping loop) that returns nothing as of now. i'll add in a sec

herrhellana: ok im not 100% sure about what heuristics to use here for validation, i added the most basic and boring case of total n of positions not exceeding n of companies

datasets/no/brreg/crawler.py
jbothma: I think just Organization is fine

datasets/no/brreg/crawler.py
jbothma: I've moved the type mapping to a lookup against the metadata - we like pulling things like that out of the coee and into metadata because it makes it neat and more accessible to less technical people.

datasets/no/brreg/crawler.py
jbothma: org number is the company registration number in norway.

datasets/no/brreg/no_brreg.yml
jbothma: now that it's not an enricher, we want it to load, acgually

datasets/no/brreg/crawler.py
jbothma: I also added some assertions

PR #3471, opened 2026-02-05
[il_mod_crypto] split 'accountId' and 'publicKey' of the crypto wallet

nvmbrasserie:
If approved: we'll need to adjust `il_mod_crypto` google sheet accordingly


pudo (approved):
I don't really know the structure of the source data well enough to make sense of this, but shall we merge it?

PR #3478, opened 2026-02-06
Fixes-06-02-26

nvmbrasserie:


pudo (commented):
Thank you for digging into the crypto thing. Here's a deal pitch: you add a few test cases on the crypto wallets patterns to the zavod tests and your weekend starts after? Let's celebrate dry feb!

datasets/jo/sanctions/crawler.py
```diff
@@ -52,7 +52,7 @@ def crawl_person(context: Context, worksheet: Worksheet) -> None:
         person.add("nationality", row.pop("nationality"))
         h.apply_date(person, "birthDate", birth_date)
         person.add("birthPlace", row.pop("birth_place"))
-        person.add("description", row.pop("description"))
+        person.add("notes", row.pop("description"))
```
jbothma, line 55: do we prefer notes over description?

jbothma (approved):

PR #3485, opened 2026-02-09
[me_acp_peps] Address mappings, also fixes #3356

### LINKED ISSUE #3356: [me_acp_peps] No country/nationality
It looks like this crawler doesn't emit a country/citizenship for PEPs. Let's at least set `country`.

pudo:
Fixes https://github.com/opensanctions/opensanctions/pull/3356

### DIFF
--- a/datasets/me/acp_peps/crawler.py
+++ b/datasets/me/acp_peps/crawler.py
@@ -221,7 +221,7 @@ def crawl_person(context: Context, person_data) -> bool:
         person = context.make("Person")
         # TODO(Leon Handreke): I think we should be passing *function_labels here,
         # but that would mean a re-key.
-        person.id = context.make_id(full_name, *sorted(function_labels))
+        person.id = context.make_id(full_name, sorted(function_labels))  # pyright: ignore[reportArgumentType]
         h.apply_name(person, full_name)
 
     person.add("citizenship", "me")
--- a/datasets/me/acp_peps/crawler.py
+++ b/datasets/me/acp_peps/crawler.py
@@ -2,7 +2,7 @@ import io
 import re
 from csv import DictReader
 from pathlib import Path
-from typing import Dict, List, Optional, Tuple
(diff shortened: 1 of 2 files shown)


nvmbrasserie (approved):

datasets/me/acp_peps/crawler.py
```diff
@@ -219,9 +221,11 @@ def crawl_person(context: Context, person_data) -> bool:
         person = context.make("Person")
         # TODO(Leon Handreke): I think we should be passing *function_labels here,
         # but that would mean a re-key.
-        person.id = context.make_id(full_name, sorted(function_labels))
+        person.id = context.make_id(full_name, *sorted(function_labels))
```
leonhandreke: See comment above - this leads to a re-key, maybe? Or am I missing something? But let's remove the comment above in any case ;) (I can do that, but let's make sure we want it first)

pudo: Argh I was way too hasty here. I think this currently hashes on the string repr of the list, so it would force a re-key. We're looking at 15,000 people, so it's not nothing. Maybe we should do the resolver trick here for the first time?

leonhandreke: https://github.com/opensanctions/opensanctions/issues/3494

jbothma: Is the resolver trick doing `resolver.decide(old_key, new_key, Judgement.POSITIVE)` within the crawler?

brrttwrks: Is this is a one-off, in-crawler trick to replace the manual deduping we've been doing until now for "graceful rekeying"?

leonhandreke: This was a thought that @pudo had a while ago. It's a bit hacky but yes, in cases where we know both the old and new IDs in the crawler, we could let the crawler write to the resolver in the manner suggested by @jbothma above

PR #3491, opened 2026-02-09
Sweden State-Owned Enterprises Leadership

nvmbrasserie:
opensanctions/crawlers#570

We might expect an update in summer 2026. The only question is, what is the best way to catch the next report?

### DIFF
--- a/datasets/se/soe/extract_csv_from_report_pdf_2024.py
+++ b/datasets/se/soe/extract_csv_from_report_pdf_2024.py
@@ -1,4 +1,6 @@
-# pyright: basic
+# This file does not run in prod, it's a one-time script to extract data from a PDF report,
+# so our requirements are relaxed.
+# mypy: ignore-errors
 
 import csv
 import sys


datasets/se/soe/crawler.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+# import pdfplumber
+# from normality import squash_spaces
+# from pathlib import Path
+
+# # Define position markers in order of appearance
+# POSITION_MARKERS = ["Ordf:", "Vd:", "Led:", "Arb rep:", "Rev:"]
+# POSITION_NAMES = {
+#     "Ordf:": "Chair",
```
nvmbrasserie: I'm not sure if I should remove all of it

leonhandreke: Maybe we can put this in a separate file in the same directory (`extract_csv_from_report_pdf_2024.py` or something?) and add a tiny bit of documentation how we expect it to be used the next time? I also wonder if we can save the PDF somewhere so it's possible to figure out what the thing actually did at some point? And mention if we did manual cleaning, etc.... 

I don't know, this is all messy... :(

datasets/se/soe/crawler.py
```diff
@@ -0,0 +1,210 @@
...
+            position = h.make_position(
+                context,
+                name=f"{position_name}, {company_name}",
+                topics=["gov.soe"],
+                country="se",
+                organization=company,
+            )
+            categorisation = categorise(context, position, is_pep=True)
+
+            occupancy = h.make_occupancy(
```
leonhandreke, line 72: As mentioned by @pudo , maybe `Directorship` is better? But is a board membership also a directorship? I guess so?

datasets/se/soe/crawler.py
leonhandreke: Can we make this "Board Chair"? Or is this standard business vocab?

datasets/se/soe/crawler.py
```diff
@@ -0,0 +1,210 @@
...
+#     text = text.replace(".", ",")
+#     # Remove parenthetical content (e.g., company names in parentheses)
+#     while "(" in text and ")" in text:
+#         start = text.find("(")
+#         end = text.find(")", start)
+#         if end > start:
+#             text = text[:start] + text[end + 1 :]
+#         else:
+#             break
+#     # Split on commas
```
leonhandreke: (nit) could we just do a `h.multi_split` here for `och`, `\n`...?

Or maybe just add a comment now that it's done that this may be a future improvement.

datasets/se/soe/crawler.py
jbothma: the point of this dataset was to serve a PEP need of a customer or a jurisdiction, so if we deviate from the normal PEP data modelling, let's have a broader conversation. It affects other nordics crawlers in development (at least norway)

datasets/se/soe/crawler.py
```diff
@@ -0,0 +1,210 @@
+import csv
+
+from zavod import Context, helpers as h
+from zavod.stateful.positions import categorise, OccupancyStatus
+
+
+SOE_REPORT_URL = "https://www.regeringen.se/contentassets/a2be3c80b3384f3eadc64530f6a2ff23/verksamhetsberattelse--for-bolag-med-statligt-agande-2024.pdf"
+
+
+def crawl(context: Context) -> None:
```
jbothma, line 32: To discover updates, you could fetch https://www.regeringen.se/regeringens-politik/bolag-med-statligt-agande/ and assert that there is exactly one link whose label contains `[Verksamhetsberättelse för bolag med statligt ägande` and that its href is `https://www.regeringen.se/rapporter/2025/06/verksamhetsberattelse-for-bolag-med-statligt-agande-2024/`

datasets/se/soe/crawler.py
jbothma: I agree - I think everything that went into turning the PDF into the CSV should go into `datasets/se/soe/extract_csv_from_report_pdf_2024.py`

I've uploaded the PDF to https://storage.googleapis.com/internal-data.opensanctions.org/se_soe/verksamhetsberattelse--for-bolag-med-statligt-agande-2024.pdf
The script can just take a pdf path as a command line argument. It doesn't have to fetch from internal data bucket.

jbothma: I think Chairperson is the more fleshed out form that's closest to the original

datasets/se/soe/se_soe.yml
```diff
@@ -0,0 +1,114 @@
+title: Sweden State-Owned Enterprises Leadership
+prefix: se-soe
+url: https://www.regeringen.se/rapporter/2025/06/verksamhetsberattelse-for-bolag-med-statligt-agande-2024/
+coverage:
+  frequency: weekly
```
jbothma: maybe we should give this a `never` and deploy.schedule @monthly like 

```
coverage:
  frequency: never
  start: "2024-05-02"
  end: "2025-05-23"
manual_check:
  last_checked: "2026-02-01"
  interval: 30
  message: Please check the website for any updates to the sanctions list. Navigate to "Full List" near the bottom of the home page. Each of the four sections should have a "Last updated" date for the list.
deploy:
  schedule: "@monthly"
```

We could even add an annual manual check, can't we?

datasets/se/soe/crawler.py
jbothma: To make the lookup work, you can do

```
from zavod.meta import load_dataset_from_path,

dataset = load_dataset_from_path(...)
context = Context(dataset)
```

datasets/se/soe/se_soe.yml
```diff
@@ -0,0 +1,114 @@
...
+      - match: 'Apotek Produktion & Laboratorier AB (APL) utveckla'
+        name: Apotek Produktion & Laboratorier AB
+      - match: 'Apoteket AB erbjuder läkemedel, apoteksprodukter o'
+        name: Apoteket AB
+      - match: 'Arlandabanan Infrastructure AB (AIAB) äger och för'
+        name: Arlandabanan Infrastructure AB
+      - match: 'Green Cargo AB är Sveriges största aktör inom håll'
+        name: Green Cargo AB
+      - match: 'AB Göta kanalbolag driver och förvaltar Göta kanal'
+        name: AB Göta
```
jbothma:
```suggestion
        name: AB Göta kanalbolag
```

datasets/se/soe/crawler.py
```diff
@@ -0,0 +1,210 @@
...
+#     text = raw_text
+#     if "Arvode" in text:
+#         text = text.split("Arvode")[0]
+
+#     # Find positions of each marker in the text
+#     marker_positions = []
+#     for marker in POSITION_MARKERS:
+#         if marker in text:
+#             idx = text.find(marker)
+#             marker_positions.append((idx, marker))
```
jbothma: Are there cases where the marker wasn't found in the text? Are they all truly absent or are any inconsistencies we need to deal with?

datasets/se/soe/crawler.py
```diff
@@ -0,0 +1,210 @@
...
+#     return results
+
+
+# def clean_names(section_text: str) -> list[str]:
+#     """Extract individual names from a section of text."""
+#     # Replace common separators with commas
+#     text = section_text
+#     text = text.replace(" och ", ",")
+#     text = text.replace("\n", ",")
+#     text = text.replace(".", ",")
```
jbothma: How does this do the right thing? My expectation would be that we'd skip`Lars-Johan` since it doesn't have a space and output `Patrik Hofbauer Jarnheimer` because of how the wrapping worked out.

Image

jbothma: I wonder if it isn't better to replace \n with empty string, and use a regular expression to split on `Statens ägarandel: \d+ %` and `Styrelse och revisorer valda för 2025/26`

then instead of dropping values containing % or \d, we warn on them because it suggests something's up. Also on `:`

nvmbrasserie: The issue with the extraction code is that it was written specifically for that one-time exercise. I validated the output and corrected a few entries that had not been extracted correctly. We don't know what next year's report will look like or whether we'll ever use this code again, so I don't think we should spend any more time on the extraction.

That said, I'll move it to a different file and add a detailed comment explaining what it's doing and how it could be improved, so that if/when we return to it, we know how to make the extraction process more efficient.

jbothma: That's fine.

But I think it's super important if you've done manual corrections, to note that fact at the top of the script or somewhere where the person running this next time will find, so that they know if they want to replicate the results of the previous run, they'll have to do manual corrections too.

datasets/se/soe/crawler.py
leonhandreke: I made it a script now. However, I get this:

```
❯ python datasets/se/soe/extract_csv_from_report_pdf_2024.py verksamhetsberattelse--for-bolag-med-statligt-agande-2024.pdf
2026-02-17 17:10:21 [warning  ] No matching lookup found.      lookup=company_names value='Luossavaara\xadKiirunavaara Aktiebolag (LKAB) är mode'
  File "/Users/leon/Development/opensanctions/datasets/se/soe/extract_csv_from_report_pdf_2024.py", line 193, in <module>
    main()
  File "/Users/leon/Development/opensanctions/zavod/.venv/lib/python3.12/site-packages/click/core.py", line 1442, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/leon/Development/opensanctions/zavod/.venv/lib/python3.12/site-packages/click/core.py", line 1363, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/leon/Development/opensanctions/zavod/.venv/lib/python3.12/site-packages/click/core.py", line 1226, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/leon/Development/opensanctions/zavod/.venv/lib/python3.12/site-packages/click/core.py", line 794, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/leon/Development/opensanctions/datasets/se/soe/extract_csv_from_report_pdf_2024.py", line 189, in main
    csv_from_pdf(context, pdf_path)
  File "/Users/leon/Development/opensanctions/datasets/se/soe/extract_csv_from_report_pdf_2024.py", line 61, in csv_from_pdf
    entity_name = get_entity_name(context, page, width)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/leon/Development/opensanctions/datasets/se/soe/extract_csv_from_report_pdf_2024.py", line 178, in get_entity_name
    assert entity_name
           ^^^^^^^^^^^
AssertionError
```

So something with the special character is off?

I'm happy to merge like this, the next person will have to adapt it a bit anyway, but just thought I'd mention.

PR #3503, opened 2026-02-11
cli: rename --clear to --rebuild-store

leonhandreke:
That's one less thing named `--clear` where it's not quite clear what is actually being cleared. In the past, `clear` was responsible for:

- rebuilding the store
- deleting the data directory on crawls
- appending a new version to the version history (lol)
- deleting a few, but not all, resources in `Context.begin`

So yeah, it's a grab bag. Enough!

### DIFF
--- a/zavod/zavod/cli.py
+++ b/zavod/zavod/cli.py
@@ -78,8 +78,8 @@ def crawl(dataset_path: Path, dry_run: bool = False, clear: bool = False) -> Non
 
 @cli.command("validate", help="Check the integrity of a dataset")
 @click.argument("dataset_path", type=InPath)
-@click.option("-c", "--clear", is_flag=True, default=False)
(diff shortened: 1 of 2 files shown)


zavod/zavod/cli.py
```diff
@@ -78,16 +78,16 @@ def crawl(dataset_path: Path, dry_run: bool = False, clear: bool = False) -> Non
 
 @cli.command("validate", help="Check the integrity of a dataset")
 @click.argument("dataset_path", type=InPath)
-@click.option("-c", "--clear", is_flag=True, default=False)
-def validate(dataset_path: Path, clear: bool = False) -> None:
+@click.option("--rebuild-store", is_flag=True, default=False)
```
pudo: Can a fellow hacker get a shortcut, perhaps? `-r`?

pudo (approved):
Very nice clarification overall.

PR #3511, opened 2026-02-12
Czech Republic PEPs

nvmbrasserie:
opensanctions/crawlers#601

A couple of things:

- For now, I’ve only added judges, senators, and deputies. We might want to expand this later to cover more people. The question is whether we want to include all ~39k declarable positions, or focus only on the more senior ones.

- I’m also not sure how to name judges in the crawler, since their position ID (`omv_id`) doesn’t distinguish between the different types of judges (`soudce Obvodní soud, soudce Okresní soud, soudce Krajský soud, soudce Vrchní soud, soudce Městský soud`, etc.).

This is still a pre-final version, just looking for an opinion on these two points for now, keeping it big picture

### DIFF
--- a/datasets/cz/pep_declarations/cz_pep_declarations.yml
+++ b/datasets/cz/pep_declarations/cz_pep_declarations.yml
@@ -38,7 +38,7 @@ dates:
 assertions:
   min:
     schema_entities:
-      Person: 3200
-      Position: 120
+      Person: 500
+      Position: 5
     country_entities:
       cz: 1


datasets/cz/pep/crawler.py
```diff
@@ -0,0 +1,119 @@
...
+
+        # We mark them all as PEPs by the source definition
+        categorisation = categorise(context, position, is_pep=True)
+
+        end_date = wp.get("end")  # alternative key: writtenDateOfEnd
+        topics = position.get("topics")
+        if end_date is not None and topics:
+            cutoff = datetime.now() - get_after_office(topics)
+            if datetime.strptime(end_date, "%Y-%m-%d") < cutoff:
+                continue
```
jbothma: we've added a declaration date prop to occupancy. Without the other dates, I think we can assume that the person holds the position on that date - at least for the purpose of determining whether they are a PEP on the basis of that declaration, if not for setting the status of the position.

But maybe that could be a separate ticket because it looks like this dataset contains start/end dates?

Image

Or is that data actually often missing? can missing end dates not be assumed to mean current?

datasets/cz/pep/crawler.py
```diff
@@ -0,0 +1,119 @@
...
+                context,
+                name=f"{role_name}, {org_name}" if org_name else role_name,
+                country="cz",
+                lang="ces",
+            )
+
+        # We mark them all as PEPs by the source definition
+        categorisation = categorise(context, position, is_pep=True)
+
+        end_date = wp.get("end")  # alternative key: writtenDateOfEnd
```
jbothma: maybe max() of the two?

datasets/cz/pep/cz_pep.yml
```diff
@@ -0,0 +1,133 @@
...
+          name: Member of the Senate of the Parliament of the Czech Republic
+          qid: Q18941264
+          topics: ["gov.legislative", "gov.national"]
+      - match: f2986792-0539-44d8-ae4a-30faba776b36
+        items:
+          name: Minister of the Czech Republic
+          qid: null
+          topics:
+            - gov.national
+            - gov.executive
```
jbothma: any reason not to let this follow the `{function}, {role}` pattern?

I think in za_pmg_legislators we might have an example of generating the position names, and the rewriting a handful to be prettier, e.g. replace("minister, ministry of", "minister of")

datasets/cz/pep/crawler.py
nvmbrasserie: it's safe to assume that no end_date means current

jbothma: are you re-implementing what occupancy_status() does for make_occupancy()?
https://github.com/opensanctions/opensanctions/blob/main/zavod/zavod/stateful/positions.py#L181

Any reason why we can't just let make_occupancy decide whether they're a PEP or not based on the end date and topics?

datasets/cz/pep_declarations/crawler.py
```diff
@@ -0,0 +1,89 @@
...
+            person=entity,
+            position=position,
+            start_date=wp.get("start"),
+            end_date=end_date,
+            categorisation=categorisation,
+        )
+        if occupancy is not None:
+            context.emit(
+                entity,
+                origin=f"https://cro.justice.cz/verejnost/api/funkcionari/{person_id}",
```
jbothma, line 83: I know we've talked about using a filename or the LLM name as origin, and I haven't documented or announced it anywhere.

Is using a deep/API URL for origin a standard practice now or a new idea?

datasets/cz/pep/cz_pep.yml
```diff
@@ -0,0 +1,43 @@
+title: Czech Republic PEPs
+prefix: cz-pep
```
jbothma: we seem to have a pattern of generally calling PEP datasets based on declarations of interests `cc_declarations` or `cc_pep_declarations`.
The best title seems to touch on both the PEP and declaration nature:
`Lithuania PEPs from the register of private interests`

shall we make this cz_pep_declarations and use a similar title?

datasets/cz/pep/cz_pep.yml
```diff
@@ -0,0 +1,43 @@
...
+entry_point: crawler.py
+summary: >-
+  Czech public officials registered in the Central Register of Declarations
+  (Centrální registr oznámení), covering parliament, judiciary, and local
+  government.
+description: |
+  The Central Register of Declarations (CRO) publishes asset and interest
+  declarations filed by around 39,000 public officials under the Czech Conflict
+  of Interest Act (No. 159/2006 Coll.), spanning all branches and levels of
+  government, from MPs and senators to mayors, judges, and state institution managers.
```
jbothma: maybe let's leave mayors out of the description/summary if we're not committing to including them all on release.

datasets/cz/pep/crawler.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+        entity.add("position", position_name)
+
+        categorisation = categorise(context, position, is_pep=is_pep)
+        if not categorisation.is_pep:
+            continue
+
+        end_date = wp.get("end") or wp.get("writtenDateOfEnd")
+        if wp.get("end") and wp.get("writtenDateOfEnd"):
+            end_date = max(wp.get("end"), wp.get("writtenDateOfEnd"))
```
jbothma: claude told me this works, and it looks quite tidy. Maybe we can just ad a comment to say we're picking the latest

```
# Pick the latest available
max(filter(None, [wp.pop("end"), wp.pop("writtenDateOfEnd")), default=None)
```

```python
>>> max(filter(None, ["a", "b"]))
'b'
>>> max(filter(None, ["a", None]))
'a'
>>> max(filter(None, [None, "b"]))
'b'
>>> max(filter(None, [None, None]))
  File "<python-input-3>", line 1, in <module>
    max(filter(None, [None, None]))
    ~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: max() iterable argument is empty
>>> max(filter(None, [None, None]), default=None)
>>>
```

jbothma: Just a suggestion, but I think it's easier to understand with the comment, than three lines of gets.

datasets/cz/pep/crawler.py
```diff
@@ -0,0 +1,89 @@
...
+        if is_deputy or is_senator or is_judge:
+            is_pep = True
+        else:
+            is_pep = None
+
+        role_name = wp_data.get("name")
+        org_name = wp.get("organization")
+        # Construct a clean position_name:
+        # e.g. poslanec, Kancelář Poslanecké sněmovny Parlamentu
+        position_name = f"{role_name}, {org_name}" if org_name else role_name
```
jbothma:
```suggestion
        position_name = f"{role_name.capitalize()}, {org_name}" if org_name else role_name
```

because

```
poslanec, Kancelář Poslanecké sněmovny Parlamentu České republiky

senátor, Senát Parlamentu ČR
```

It sounds like "".title() wouldn't make sense for Czech (and I'm personally not a fan of titlecasing generally)

datasets/cz/pep_declarations/crawler.py
```diff
@@ -0,0 +1,89 @@
...
+    for wp in item.get("workingPositions", []):
+        wp_data = wp.get("workingPosition")
+        # is_deputy and is_senator are position-level flags
+        is_deputy = wp_data.get("deputy")
+        is_senator = wp_data.get("senator")
+        # Only the main institutions are marked as PEP by default
+        if is_deputy or is_senator or is_judge:
+            is_pep = True
+        else:
+            is_pep = None
```
jbothma, line 49: in the last engineering meeting not to do any more position-specific bulk categorisation in the crawler itself. Let's leave is_pep as None and categorise the positions in the UI once it's run through.

datasets/cz/pep/cz_pep.yml
jbothma: It could be nice to add a paragraph to then say this dataset identifies PEPs based on the positions they currently or recently held according to their declarations.

jbothma (commented):
Great! just a few items and a question

datasets/cz/pep_declarations/crawler.py
nvmbrasserie: I think that was also mentioned and it's somehow similar to the `filename`.  I saw the filename usage in `us_al_med_exclusions` and thought that it's in line with that. It's not entirely new but is not widely implemented yet either

datasets/cz/pep_declarations/crawler.py
nvmbrasserie: I think since we have the flags it'd be nice to have some basic coverage. It's very explicit and we have the same `is_pep` or `not is_pep` logic in quite a few places. My understanding is that what we discussed was more about removing `CUTOFF_DATE` logic from everywhere

datasets/cz/pep_declarations/crawler.py
jbothma: I asked the oracle (slack, not an LLM), and it has spoken

> if sourceUrl vomits JSON or XML at civilians, then maybe we want to make it origin?

Thanks for this.

datasets/cz/pep_declarations/crawler.py
jbothma: We decided that if PEP status is effectively indicated in the source as booleans in the data, it's ok to use that.

jbothma: The data includes municipal court judges. They're not covered by our policy. Shall we play it safe and exclude judge from this and categorise the rest using the position categorisation UI?

datasets/cz/pep_declarations/crawler.py
```diff
@@ -0,0 +1,92 @@
...
+        entity.add("position", position_name)
+
+        categorisation = categorise(context, position, is_pep=is_pep)
+        if not categorisation.is_pep:
+            continue
+
+        # Pick the latest available
+        end_date = max(
+            filter(None, [wp.get("end"), wp.get("writtenDateOfEnd")]), default=None
+        )
```
jbothma: Readability nit :)

```suggestion
        # Pick the latest available
        end_dates = [wp.get("end"), wp.get("writtenDateOfEnd")]
        end_date = max(filter(None, end_dates), default=None)
```

datasets/cz/pep_declarations/crawler.py
```diff
@@ -0,0 +1,92 @@
...
+
+    entity = context.make("Person")
+    entity.id = context.make_id(person_id)
+    h.apply_name(
+        entity,
+        first_name=first_name,
+        last_name=last_name,
+    )
+    entity.add("title", item.get("titleBefore"))
+    entity.add("title", item.get("titleAfter"))
```
jbothma: @pudo this data has things like PhD or LL.M in titleAfter. the ftm docs says nameSuffix is for stuff like OBE so it looks like a fit. We have things like Jr or III in there at the moment. 

Would nameSuffix be a reasonable prop for this?

datasets/cz/pep_declarations/crawler.py
```diff
@@ -0,0 +1,92 @@
...
+            )
+            context.emit(position)
+            context.emit(occupancy)
+
+
+def crawl(context: Context) -> None:
+    page = 0
+    while True:
+        url = f"{context.data_url}?sort=created&order=DESC&page={page}&pageSize=100"
+        data = context.fetch_json(url, cache_days=7)
```
jbothma: I think we ought to disable caching for this endpoint.

In theory if we sorted **ascending** by creation date **and**  they didn't delete people, we could safely cache. But we can't assume they don't delete people, which means all the results could shift across pages, so we can intermittently miss people making it a pain to debug.

datasets/cz/pep_declarations/crawler.py
```diff
@@ -0,0 +1,92 @@
...
+        if data is None:
+            context.log.error("Empty response", url=url)
+            break
```
jbothma: how about an assertion or valueError here? When would this be empty when we'd want to actually publish the result?

datasets/cz/pep_declarations/crawler.py
```diff
@@ -0,0 +1,92 @@
...
+        # Pick the latest available
+        end_date = max(
+            filter(None, [wp.get("end"), wp.get("writtenDateOfEnd")]), default=None
+        )
+
+        occupancy = h.make_occupancy(
+            context,
+            person=entity,
+            position=position,
+            start_date=wp.get("start"),
```
jbothma: Sorry for commenting on this so late, but let's pop then it fails loudly if fields go missing. Then we could even audit() like usual.

```
>>> {}.get("foo")
>>> {}.pop("foo")
  File "<python-input-1>", line 1, in <module>
    {}.pop("foo")
    ~~~~~~^^^^^^^
KeyError: 'foo'
```

PR #3512, opened 2026-02-12
Heuristic name categorisation and crawler suggestions

jbothma:
Draft - some TODOs:

- [ ] tw probably wants more love
- [x] us_sam is a good demo of what experimental heuristics/suggestions in a crawler might look like before they get promoted to zavod. Do we want to add these suggestions to zavod right away or let them percolate a little while we compare the same idea on another crawler or two?
- [ ] more tests are needed for all this logic of when something's irregular, applying suggestions correctly, etc. 
- [x] ~~maybe the original == suggested logic for whether a review should be created can be dropped now that it's shown that is_irregular probably has to be an optional argument for crawler custom checks anyway~~ Actually I think we want both checks - if a review is created if either is_irregular=True is passed, or a suggested differing from original, I think it simplifies the work for the call site a bit.
- [x] I don't think `h.apply_names` needs to be exposed
- I want to look a bit more closely still at the llm use case
  - [x] the simplest thing right now is just to add an `assert not (suggested is not None and enable_llm_cleaning)` because right now we just stuff all the names for all the props into a "list of strings" for the LLM, and mixing crawler/heuristic suggestions and llm isn't well defined or tested and out of scope.

This introduces and changes a few methods:

- `h.apply_reviewed_name_string` is a simple case of a single name that maybe needs cleaning
- `h.apply_reviewed_names` is the powerhouse
  - `apply_reviewed_name_string` calls it with that single name string.
  - a crawler can do its own categorisation. It does that on the result of `h. check_names_regularity` so that regularity's heuristics don't override the crawler's
  - it can insist that a Names suggestion is irregular (even if it hasn't recategorised anything)
- `h. review_names` interface updates to be consistent with `apply_reviewed_names` - it is still a way to prepare reviews without applying anything - even the original - for crawlers that already have some logic in place that we're not ready to remove yet

### DIFF
--- a/zavod/zavod/helpers/names.py
+++ b/zavod/zavod/helpers/names.py
@@ -326,7 +326,7 @@ def check_names_regularity(entity: Entity, names: Names) -> Tuple[bool, Names]:
     from "name" to "alias" or "weakAlias").
     """
     is_irregular = False
-    updated_suggested_data = defaultdict(list)
+    updated_suggested_data: Dict[str, List[str | None]] = defaultdict(list)
     for key, strings in names.nonempty_item_lists():
         for string in strings:
             regularity = check_name_regularity(entity, string)


datasets/us/sam_exclusions/crawler.py
```diff
@@ -204,7 +206,51 @@ def crawl(context: Context) -> None:
 
         # The low quality names tend to come from OFAC so check those.
         if agency == "TREAS-OFAC":
-            h.review_names(context, entity, name)
+            original = Names(name=name)
```
pudo: I really like what's happening here as a pattern, the composability is cool. In this specific case, I think we could even extend it beyond the `TREAS-OFAC` scope.

Do we want a helper to make `Names` objects, like `h.names()`? At the minimum, I would expose the class as part of the helper root module (or do we want `from zavod.extract import Names` as a pattern?).

datasets/ru/cbr_banks/crawler.py
```diff
@@ -88,20 +88,24 @@ def crawl_details(context: Context, bic: str, entity: Entity, short_name: str |
...
+        )
     else:
         ssv_date = co_data.findtext("SSV_Date")
         reg_date = co_data.findtext("MainDateReg")
 
         en_names = co_data.findtext("encname")
-        h.review_names(context, entity, en_names, enable_llm_cleaning=True)
-        h.review_names(
+        h.apply_reviewed_name_string(
+            context, entity, en_names, enable_llm_cleaning=True
```
pudo: total nitpick, but: is `llm_cleaning=True` enough? I feel like bool arguments sort of imply the `enable_` bit :)

datasets/tw/shtc/tw_shtc.yml
```diff
@@ -33,6 +33,11 @@ tags:
   - list.regulatory
   - list.risk
 
+names:
+  LegalEntity:
+    reject_chars: "()（）"
+    single_token_min_length: 6
```
pudo, line 40: so cool. can't wait to roll this out more

datasets/ru/cbr_banks/crawler.py
jbothma: yes!

PR #3557, opened 2026-02-16
[us_dod_chinese_milcorps] clean up the code; add weakAlias, sourceUrl and previousName

nvmbrasserie:


nvmbrasserie:
It will trigger a rerun, but we still have the latest update in the data that was rolled back. Should I move it somewhere else until further notice?

PR #3558, opened 2026-02-16
daily Issues 02-16-2026

herrhellana:


datasets/us/az/med_exclusions/us_az_med_exclusions.yml
```diff
@@ -39,7 +39,7 @@ dates:
 assertions:
   min:
     schema_entities:
-      LegalEntity: 225
+      LegalEntity: 170  # raw count 218
```
nvmbrasserie: it can be 200 in that case. it's good to keep this list in a bit tighter assertions

datasets/us/de/med_exclusions/us_de_med_exclusions.yml
```diff
@@ -161,12 +161,16 @@ lookups:
         prop: sector
       - match: "10\n93379521"
         value: 1093379521
-      - match: "OS024306/ \nOT021627"
+      - match:
+          - "OS024306/ OT021627"
+          - "OS024306/ \nOT021627"
```
nvmbrasserie: if you run `zavod --debug crawl --clear`, chances are that the previous lookup is unused and it's safe to remove it.

herrhellana: we dont want to keep it in case they roll back older versions?

datasets/us/ga/med_exclusions/crawler.py
```diff
@@ -19,6 +18,10 @@ def crawl_item(row: Dict[str, str], context: Context):
         entity = context.make("Company")
         entity.id = context.make_id(business_name, row.get("npi"))
         entity.add("name", business_name)
+
+        # the latest file update includes a middle name for a business entity, rm it
+        if row.pop("middle_name") is not None:
+            return
```
nvmbrasserie: this one is more complex than simply removing the middle name. we'll need to update the function to cover cases where there is both a person and a business entity in the same row. let's park this one for now

datasets/us/az/med_exclusions/us_az_med_exclusions.yml
nvmbrasserie: the same goes for all `*_med_exlusions`. their formats allow us to be certain about the number of expected entities

datasets/us/de/med_exclusions/us_de_med_exclusions.yml
nvmbrasserie: nope, we want to keep lookups 1:1 with the current state of affairs

datasets/us/ga/med_exclusions/crawler.py
herrhellana: aight sounds good

PR #3559, opened 2026-02-16
included sort for splitters in h.multi_split()

herrhellana:
hoping that nobody ever relied on a sequential order of splitters....

### DIFF
--- a/zavod/zavod/helpers/text.py
+++ b/zavod/zavod/helpers/text.py
@@ -2,6 +2,9 @@ import re
 from banal import is_listish, ensure_list
 from typing import Optional, List, Union, Iterable
 from normality import squash_spaces
+from zavod.logs import get_logger
+
+log = get_logger(__name__)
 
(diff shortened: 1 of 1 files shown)


pudo (approved):
This will definitely break something. I want to see what it is. cc @jbothma

pudo:
Alternative: make it regex based and add boundary?

jbothma:
Prefixes sort earlier than longer strings

```
>>> sorted(["also known as", "also"])
['also', 'also known as']
```

https://github.com/opensanctions/opensanctions/blob/a4fae68bf10b445a52e83da41ed6079158909f05/datasets/_global/adb_sanctions/crawler.py#L13

One option @pudo  suggested I think is that it can warn if there's a risk, i.e. if any earlier values are included in later values. Friedrich mentioned a flag - I'm not sure how that'd work, but without a flag, the crawler can just order splits such that i) comes after ii) and won't match prematurely.

PR #3566, opened 2026-02-17
Issues 02-17-2026

herrhellana:


datasets/_global/un_ga_protocol/un_ga_protocol.yml
```diff
@@ -65,13 +65,24 @@ lookups:
...
           - 09-Sep-220
           - Feb-23
           - Nov-21
           - Aug-19
+          - May-21
         value: null
+      - match: 09-Jan-0721
+        value:
+          - 09-Jan-07
+          - 21-Nov-24
```
nvmbrasserie: we shouldn't be inferring any data based on the contextual knowledge. there might be a case when `09-Jan-0721` is being picked up somewhere else in the file but the second date will be different from `21-Nov-24`. So, even if we want to extend it here, it should be just `2007-01-09` (mappings should be in the ISO format)

datasets/_global/un_ga_protocol/un_ga_protocol.yml
```diff
@@ -65,13 +65,20 @@ lookups:
           - 09-Sep-2207
           - 09-Sep-2201
           - 09-Sep-2228
```
nvmbrasserie: shall we remap all of them in that case? `Aug-19` will be just `2019-08`

PR #3582, opened 2026-02-18
Move programs and issuers from Directus CMS to YAML files

### LINKED ISSUE #3234: Move programs editor to review UI
I know this isn't strictly necessary, but I still think it'd leave us with a much more explainable data architecture in the long run: We should vibe up a WYSIWYG editor for sanctions programs into zavod UI. This way, there is no consistency issues between directus and the review system, and our data pipeline has no dependencies on the website CMS.

We'll need: 
* `issuers` for the authorities
* `programs` expanded with the other metadata properties
* Update the index exporter to use the table instead of the directus CMS

leonhandreke:
## Summary

Closes #3234.

- Programs and issuers are now stored as individual YAML files under `programs/` and `programs/issuers/`, replacing the Directus CMS as the source of truth.
- Each program file contains all metadata: title, key, URL, summary, dataset, issuer reference, aliases, target territories, and measures.
- Issuer files are referenced by their filename stem (e.g., `issuer: au_dfat`).
- `zavod.stateful.programs` reads the YAML files at startup using Pydantic models (`Program`, `Issuer`) and validates all territory codes against `rigour.territories`.
- Removed `contrib/load_programs_from_cms` which loaded programs from the CMS into the database.
- A follow-up PR will update the index exporter to use the YAML-backed getter instead of fetching from Directus.

## Test plan

- [ ] Run the existing unit tests (`test_sanctions.py`)
- [ ] Spot-check a few program YAML files against the Directus CMS


zavod/zavod/stateful/programs.py
```diff
@@ -1,34 +1,121 @@
 import functools
-from dataclasses import dataclass
+import yaml
+from pathlib import Path
 from typing import Optional
+from pydantic import BaseModel, Field
 
-from zavod import Context
-from zavod.stateful.model import program_table
+PROGRAMS_RESOURCE_PATH = Path(__file__).parent.parent.parent.parent / "programs"
```
pudo: Not in love with this. Can it live in `settings` and perhaps be an env var that defaults to this?

zavod/zavod/stateful/programs.py
```diff
@@ -1,34 +1,121 @@
...
+    )
+
+
+class Program(BaseModel):
+    """A sanctions regime."""
+
+    id: int  # from Directus, drop after migration
+    key: str = Field(
+        description="Hyphenated reference key (e.g., 'AU-AFGHANISTAN', 'AE-UNSC1373', 'US-AFGH')"
+    )
```
pudo, line 47: I wonder if we should autofill/override this from the file name stem to ensure the two always match?

zavod/zavod/stateful/programs.py
```diff
@@ -1,34 +1,121 @@
...
+        description="Alternative names or references (e.g., 'Resolution 1373', 'EO 13818')",
+    )
+    target_territories: list[str] = Field(
+        default_factory=list,
+        description="Territory codes targeted by this program (e.g., 'af', 'ru', 'by')",
+    )
+    measures: list[str] = Field(
+        default_factory=list,
+        description="Types of sanctions imposed (e.g., 'Asset freeze', 'Travel ban', 'Arms embargo')",
+    )
```
pudo: Do we want to introduce a controlled vocab of these next? They should also apply to `Sanction:provisions` in the pipeline itself.

zavod/zavod/stateful/programs.py
```diff
@@ -1,34 +1,121 @@
...
-            url=row.url,
-        )
-        for row in context.conn.execute(stmt).fetchall()
-    ]
-    return {p.key: p for p in programs}
-
-
-def get_program_by_key(context: Context, program_key: str) -> Optional[Program]:
-    return get_all_programs_by_key(context).get(program_key, None)
+def get_all_programs_by_key() -> dict[str, Program]:
```
pudo, line 91: little @cache and the memory profile does it's magic!

pudo (changes requested):
Impressive work! I would kind of propose we force a relationship between the file name and program key so that we don't run into a discoverability nightmare down the road. Could also imagine a folder structure a la:

```
meta/
  programs/
  issuers/
```

zavod/zavod/stateful/programs.py
leonhandreke: See lines 78-79, the diff makes it a bit hard to find :)

pudo: Argh, sorry!

zavod/zavod/stateful/programs.py
leonhandreke: Introduced a `Measure` type as a `Literal` with the 9 values currently in use across all programs:

```python
Measure = Literal[
    "Arms embargo",
    "Arms export restrictions",
    "Asset freeze",
    "Export control",
    "Financial block",
    "Investment ban",
    "Prohibition to satisfy claims",
    "Restrictions on goods",
    "Travel ban",
]
```

Pydantic now validates program YAML files against this vocab at load time. Applying these to `Sanction:provisions` in the pipeline is a natural next step — we could reuse the same `Measure` type there.

zavod/zavod/stateful/programs.py
leonhandreke: Done — the key is now derived from the filename stem, with an assertion for it to be the same as `key`. Files are also renamed to use the hyphenated key directly (e.g. `US-IRAQ.yml`), so no underscore-to-hyphen conversion needed.

zavod/zavod/stateful/programs.py
leonhandreke: Moved to `settings.META_RESOURCE_PATH`, defaults to `meta` and overridable via `ZAVOD_META_RESOURCE_PATH`.

pudo (approved):
lol this one is big but somehow feels non-threatening.

PR #3585, opened 2026-02-18
[tr_parliament] add birth date/place to IDs (externally)

herrhellana:
emit external entities using new id scheme incl birth date/place 

This doesn't modify existing entities. It only emits external entities with matching names to gracefully move over to the new IDs in https://github.com/opensanctions/opensanctions/pull/3686

issue [#596](https://github.com/opensanctions/crawler-planning/issues/596)

### DIFF
--- a/datasets/tr/parliament/crawler.py
+++ b/datasets/tr/parliament/crawler.py
@@ -44,7 +44,9 @@ def crawl_birth_year_place(context: Context, url: str) -> tuple[str | None, str 
         actions=UNBLOCK_ACTIONS,
         javascript=True,
         absolute_links=True,
-        cache_days=1,
+        # Cache disabled because their bot blocking includes regularly-changing UUIDs in the URL
+        # making caching fairly futile.
+        cache_days=None,
     )
 
     # Pick the first element under the span in the profile free text section
@@ -87,8 +89,6 @@ def crawl_item(context: Context, item: etree):
     entity.add("political", party)
 
     birth_year, birth_place = crawl_birth_year_place(context, deputy_url)
-    entity.add("birthDate", birth_year)
-    entity.add("birthPlace", birth_place)
 
     position = h.make_position(
         context, "Member of the Grand National Assembly", country="tr"
@@ -116,8 +116,7 @@ def crawl_item(context: Context, item: etree):
         context.emit(entity)
         context.emit(position)
         context.emit(occupancy)
-        if birth_year is not None and birth_place is not None:
-            context.emit(entity_temp, external=True)
+        context.emit(entity_temp, external=True)
 
 
 def crawl(context: Context):
--- a/datasets/tr/parliament/tr_parliament.yml
+++ b/datasets/tr/parliament/tr_parliament.yml
@@ -39,6 +39,11 @@ assertions:
     country_entities:
       tr: 590
     countries: 1
+    property_fill_rate:
+      Person:
+        # The birth date extraction is prone to failure because it's a regex on free text.
+        # If it breaks, we risk re-keying the dataset unintentionally.
+        birthDate: 0.7
   max:
     schema_entities:
       Person: 605


datasets/tr/parliament/crawler.py
```diff
@@ -51,6 +72,9 @@ def crawl_item(context: Context, item: etree):
     entity.add("sourceUrl", deputy_url)
     entity.add("political", party)
 
+    birth_year = crawl_birth_year(context, deputy_url)
```
herrhellana: actually maybe made the id on the birth_year + name, not party + name? or all three?

jbothma: At the moment there are two people named `Mustafa DEMİR`. Luckily they're in different parties so we don't have a dupe yet. But it might be worth the rekey.

This is how we should do it:

1. create a temporary additional person entity for each person
2. make their id based on name, birth date, birth place
3. add the same name, the birth place and date, and party to the new entity (don't add/change anything on the old entity
4. emit these with external=True alongside the original entities

we'll merge the dataset, and dedupe the internal (current) entities with the external (additional) ones.
The existing entities and the new entities will end up with the same canonical IDs

Then we make another PR, we give it the disruptive tag and only merge it after it's been announced in the tech bulletin:

1. Make sure both `Mustafa DEMİR`'s have different canonical IDs
1. replace the original IDs with the new IDs based on birth place and name
2. remove the code for creating/emitting the temporary entities
3. add birth date and place to the persons as normal

when we release this, the original entities should see an update rather than be replaced.

datasets/tr/parliament/crawler.py
```diff
@@ -35,6 +36,26 @@
...
+        javascript=True,
+        absolute_links=True,
+    )
+
+    birth_string = h.xpath_strings(
+        doc,
+        '//*[@id="milletvekili-detay-holder-desktop"]//div[contains(@class, "profile-ozgecmis-div")]/span//*[normalize-space()][1]/text()',
+    )
+    year_match = re.search(r"\b\d+\b", birth_string[0])
+    year = int(year_match.group()) if year_match else None
```
jbothma: Since this picking the first `p` tag in a longish list of personal details (not a very reliable way to be sure we're finding the correct thing), let's tighten the regex up a bit:

- require four digits for the year
- match the alphabetic chars with optional / and space between as birth place - that's handy too
- match alphabetic chars with optional comma after
- require this to match the entire string in that element

That makes it much less likely that we're matching the wrong thing if they change things around.

Then let's use the new [property fill rate assertion](https://github.com/opensanctions/opensanctions/pull/3416/changes#diff-8371e998a7ad2da001d6d302d80cc8492d7f4f23b2416ea6c6416dae94dc7c50R167) Leon just added to assert that we're finding, let's say, 80% of them? it'd be interesting to see what % we actually find - start with 100% and pull back if needed.

This might only work once we're adding birth date to the internal entities - not sure if this will pick up for external entities.

datasets/tr/parliament/crawler.py
```diff
@@ -35,6 +36,26 @@
...
+        '//*[@id="milletvekili-detay-holder-desktop"]',
+        actions=UNBLOCK_ACTIONS,
+        javascript=True,
+        absolute_links=True,
+    )
+
+    birth_string = h.xpath_strings(
+        doc,
+        '//*[@id="milletvekili-detay-holder-desktop"]//div[contains(@class, "profile-ozgecmis-div")]/span//*[normalize-space()][1]/text()',
+    )
```
jbothma: defining the xpath variable should improve the wrapping on this and make it more readable.
Does it work to specify p tag and remove the normalise?

```suggestion
    xpath = '//*[@id="milletvekili-detay-holder-desktop"]//div[contains(@class, "profile-ozgecmis-div")]/span//p[1]/text()'
    birth_string = h.xpath_strings(
        doc,
        xpath
    )
```

herrhellana: not quite, DOM alternates between using p tags and div tags there across politicians pages, so i had to use sth that is more flexible. but it looks like it's safe to remove normalize-space()

datasets/tr/parliament/crawler.py
herrhellana: added these! didn't add the property fill rate assertion because leon's branch hasn't been merged yet.

datasets/tr/parliament/crawler.py
```diff
@@ -51,6 +72,10 @@ def crawl_item(context: Context, item: etree):
     entity.add("sourceUrl", deputy_url)
     entity.add("political", party)
 
+    birth_year, birth_place = crawl_birth_year_place(context, deputy_url)
+    entity.add("birthDate", birth_year)
+    entity.add("birthPlace", birth_place)
```
jbothma: let's only add this to `entity` in the next pull request where we replace the old ID with the new ID.

PR #3595, opened 2026-02-19
Export statements before consolidation

### LINKED ISSUE #3593: statements.csv sometimes misses statements which are in crawler statements.pack files (consolidation)
Value consolidation in the export process causes the statement exporter's output to differ from the union of crawler statements.pack's

jbothma:
Fixes #3593

Consolidation removes statements. Exporting statements.csv based on consolidated entities means that export doesn't match the statements database which is the union of the relevant source datasets' statements.pack outputs.

This fixes that by adding a feed of pre-consolidation entities for exporters like this who need it.

### DIFF
--- a/zavod/zavod/exporters/__init__.py
+++ b/zavod/zavod/exporters/__init__.py
@@ -69,11 +69,16 @@ def export_data(context: Context, view: View) -> None:
         exporter.setup()
 
     for idx, entity in enumerate(view.entities()):
-        # Use it once we figure memory explosion out
-        entity = consolidate_entity(view.store.linker, entity)
-        fragment = ViewFragment(view, entity)
         if idx > 0 and idx % 10000 == 0:
             log.info("Exported %s entities..." % idx, scope=context.dataset.name)
+
+        # feed_unconsolidated must be called before consolidate_entity, because
+        # consolidate_entity mutates the entity in place.
+        for exporter in exporters:
+            exporter.feed_unconsolidated(entity)
+
+        entity = consolidate_entity(view.store.linker, entity)
+        fragment = ViewFragment(view, entity)
         for exporter in exporters:
             exporter.feed(entity, fragment)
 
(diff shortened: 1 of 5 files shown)


zavod/zavod/exporters/__init__.py
```diff
@@ -69,11 +69,16 @@ def export_data(context: Context, view: View) -> None:
         exporter.setup()
 
     for idx, entity in enumerate(view.entities()):
-        # Use it once we figure memory explosion out
-        entity = consolidate_entity(view.store.linker, entity)
```
jbothma, line 73: this signature makes it look like consolidate_entity is functional but it's not. It could be nice to either make it functional, or remove the return type to make it more obvious that it mutates entity (if mutation is significantly more performant than copying)

jbothma: I'd leave that to a different PR

zavod/zavod/tests/exporters/util.py
```diff
@@ -16,7 +17,10 @@ def harnessed_export(exporter_class, dataset, linker=None) -> None:
     exporter = exporter_class(context)
     exporter.setup()
     for entity in view.entities():
+        exporter.feed_unconsolidated(entity)
         fragment = ViewFragment(view, entity)
+        entity = consolidate_entity(view.store.linker, entity)
+        entity = fragment.get_entity(entity.id)
```
jbothma: export has gotten more complex - it'd be nice if harness now called export_data to follow all the same logic without remembering to keep it up to date.

I'll also leave this to a different PR

zavod/zavod/exporters/__init__.py
pudo: Oh nice point. I feel like cloning the entity once would not be the end of the world performance-wise, but I like keeping it apart.

pudo (approved):
That looks really good, I'm happy for this to go in!

PR #3623, opened 2026-02-23
Don't publish failed runs `index.json` to `/datasets`

leonhandreke:
See https://github.com/opensanctions/opensanctions/issues/2541. Trying to make the PRs as small as possible to make it easy for my meager brain to reason about.

The website already reads the latest archived run to display metadata and issues, so this will continue working.


pudo (approved):
Yeah that looks good and is a real nice step forward in terms of untangling this mess. We're still uploading the versions file for discovery - looks good!

leonhandreke:
This one is scary to me, so thanks for the encouragement :)

PR #3637, opened 2026-02-25
[ca-sema-sanctions-at] canadian sanctions announcement table

herrhellana:
issue: https://github.com/opensanctions/crawler-planning/issues/580

### DIFF
--- a/datasets/ca/sema_sanctions_news/ca_sema_sanctions_news.yml
+++ b/datasets/ca/sema_sanctions_news/ca_sema_sanctions_news.yml
@@ -61,6 +61,14 @@ assertions:
       Vessel: 1600
 
 lookups:
+  sanction.program:
+    options:
+      - match:
+          - "Russia"
+          - "Belarus"
+          - "Ukraine"
+          - "Moldova"
+        value: CA-SEMA
   schema_type:
     options:
       - match:
--- a/datasets/ca/sema_sanctions_news/crawler.py
+++ b/datasets/ca/sema_sanctions_news/crawler.py
@@ -8,9 +8,6 @@ from zavod import Context, helpers as h
 from zavod.stateful.review import assert_all_accepted
 
 
-PROGRAM_KEY = "CA-SEMA"
-
-
 def split_names(name: str) -> tuple[bool, h.Names]:
     """
     Returns:
@@ -114,10 +111,11 @@ def crawl_entity_notice(context: Context, row: Dict[str, _Element]) -> None:
         )
 
         entity.add("topics", "sanction")
+        program_key = h.lookup_sanction_program_key(context, country)
         sanction = h.make_sanction(
             context,
             entity,
-            program_key=PROGRAM_KEY,
+            program_key=program_key,
         )
         sanction.add("program", country)
         sanction.add("reason", reason)
@@ -141,7 +139,7 @@ def crawl_vessel(context: Context, row: Dict[str, _Element]) -> None:
     sanction = h.make_sanction(
         context,
         vessel,
-        program_key=PROGRAM_KEY,
+        program_key="CA-SEMA",
     )
     context.emit(vessel)
     context.emit(sanction)


datasets/ca/sema_sanctions_at/ca_sema_sanctions_at.yml
```diff
@@ -0,0 +1,47 @@
...
+  format: HTML
+dates:
+  formats: ["%Y", "%B %d, %Y", "%B %d %Y"]
+
+assertions:
+  min:
+    schema_entities:
+      Person: 1500  # 1927 raw count
+      LegalEntity: 600  # 770 raw count
+      Vessel: 80 # 109 raw count
```
nvmbrasserie: It's important that we check the number of entities crawler emits against the source coverage. In this case, we'd expect 619 vessels, as per the source coverage: 

`Total number of ships listed as of February 24, 2026: 619`

datasets/ca/sema_sanctions_at/crawler.py
```diff
@@ -0,0 +1,99 @@
...
+        context,
+        vessel,
+        program_key=PROGRAM_KEY,
+    )
+    context.emit(vessel)
+    context.emit(sanction)
+
+
+def crawl(context: Context) -> None:
+    doc = context.fetch_html(context.data_url, cache_days=1)
```
jbothma: let's drop the cache days otherwise we don't benefit from crawling multiple times per day

datasets/ca/sema_sanctions_at/crawler.py
```diff
@@ -0,0 +1,99 @@
...
+    if "(russian:" in name.lower():
+        name, name_ru = name.split("(Russian:", 1)
+        name_ru = name_ru.strip().rstrip(")")
+    if "(also known as" in name.lower():
+        name, aka = name.split("(also known as", 1)
+        aliases = [a.strip().rstrip(")") for a in aka.split(",")]
+    return name.strip(), dob, name_ru, aliases
+
+
+def crawl_entity(context: Context, row: Dict[str, _Element]) -> None:
```
jbothma: nitpick: I'd call this 

```suggestion
def crawl_entity_notice(context: Context, row: Dict[str, _Element]) -> None:
```
just because each row in the table isn't an entity, it's a notice containing a list of entities

datasets/ca/sema_sanctions_at/crawler.py
```diff
@@ -0,0 +1,99 @@
...
+    aliases = []
+    if "(born " in name.lower():
+        name, dob = name.split("(born", 1)
+        dob = dob.strip()
+    if "(russian:" in name.lower():
+        name, name_ru = name.split("(Russian:", 1)
+        name_ru = name_ru.strip().rstrip(")")
+    if "(also known as" in name.lower():
+        name, aka = name.split("(also known as", 1)
+        aliases = [a.strip().rstrip(")") for a in aka.split(",")]
```
jbothma: there are a small handful of cases where they wrote `bornon`. I found that by scrolling through `sort data/datasets/.../names.txt | less`.

Instead of adding more parsing logic, let's add a a condition `if h.is_name_irregular` - the ( should trigger it, then we can use lookups for further cleaning as needed.

datasets/ca/sema_sanctions_at/crawler.py
```diff
@@ -0,0 +1,99 @@
...
+    people = h.xpath_strings(
+        specific_prohibition, ".//div[contains(., 'List of individuals')]//li/text()"
+    )
+    entities = h.xpath_strings(
+        specific_prohibition, ".//div[contains(., 'List of entities')]//li/text()"
+    )
+    for entity_name in chain(people, entities):
+        entity = context.make(schema)
+        entity.id = context.make_id(entity_name)
+        name, dob, name_ru, aliases = parse_name_dob(entity_name)
```
jbothma: I think returning a four-tuple is a good argument for doing 

```
for person in people:
    crawl_person(...)
```

I know some of the things are the same between . But it'd also be nice to be able to add the date of birth to make_id for people

datasets/ca/sema_sanctions_at/crawler.py
jbothma: there also person names that start with numbers but shouldn't and company names that start with numbers and should.

and company aliases separated by ` and `

datasets/ca/sema_sanctions_at/crawler.py
jbothma: alternatively maybe we can make a dataclass to return instead of a 4-tuple

datasets/ca/sema_sanctions_at/ca_sema_sanctions_at.yml
```diff
@@ -0,0 +1,75 @@
...
+prefix: ca-sema-at
+coverage:
+  schedule: "0 */3 * * *" # every 3 hours
+  frequency: daily
+  start: 2026-02-24
+load_statements: true
+summary: >-
+  Sanctions against individuals, entities, and ships imposed by Canada in response to Russia’s
+  unprovoked and unjustified invasion of Ukraine under the Special Economic Measures Act (SEMA).
+  The current announcement table is more regularly maintained.
```
jbothma: Let's keep the summary short and crisp, and add a blank line and paragraph to the description like "The announcements these data are sourced from are updated more quickly than the full Canadian Consolidated Autonomous Sanctions List" or something

```suggestion
```

jbothma: On the website, the summary goes here 

Image

datasets/ca/sema_sanctions_at/ca_sema_sanctions_at.yml
```diff
@@ -0,0 +1,75 @@
+title: Canadian Consolidated Autonomous Sanctions List on the Russian invasion of Ukraine
+entry_point: crawler.py
+prefix: ca-sema-at
```
jbothma: Maybe ua since the dataset has a focus on the ukraine invasion?

```suggestion
prefix: ca-sema-ua
```

datasets/ca/sema_sanctions_at/crawler.py
```diff
@@ -1,57 +1,46 @@
...
-        name = name.strip()
-
+    entities = h.xpath_strings(
+        specific_prohibition, ".//div[contains(., 'List of entities')]//li/text()"
+    )
+    for name in chain(people, entities):
         entity = context.make(schema)
         entity.id = context.make_id(name)
 
-        # --- clean the name + extract DOB if present
```
herrhellana: why you dont like my dashes.... @nvmbrasserie

datasets/ca/sema_sanctions_at/crawler.py
herrhellana: i used this chunk of code previously to send it into review, could you please check the latest pushed up version and tell if this is the correct use of the new helpers functions @jbothma :

```
# send the rest of the irregular aliases into review
            original = h.Names(name=alias)
            is_irregular, suggested = h.check_names_regularity(entity, original)

            h.review_names(
                context,
                entity,
                original=original,
                suggested=suggested,
                is_irregular=is_irregular,
            )
```

datasets/ca/sema_sanctions_announcements/ca_sema_sanctions_announcements.yml
```diff
@@ -0,0 +1,112 @@
+title: Canadian Consolidated Autonomous Sanctions List on the Russian invasion of Ukraine
+entry_point: crawler.py
+prefix: ca-sema-announcements
```
jbothma:
```suggestion
prefix: ca-sema-ann
```

datasets/ca/sema_sanctions_at/ca_sema_sanctions_at.yml
jbothma: Sorry, I was unclear.

I think

> The data sourced from this announcement table is updated more quickly than the full Canadian Consolidated Autonomous Sanctions List.

should be in `description`

datasets/ca/sema_sanctions_announcements/crawler.py
```diff
@@ -0,0 +1,141 @@
...
+def parse_name_dob(name: str) -> PersonBio:
+    name_ru = None
+    dob = None
+    aliases = []
+
+    born_patterns = ["(born ", "(bornon "]
+    for born in born_patterns:
+        if born in name.lower():
+            name, dob = name.split(born, 1)
+            dob = dob.strip(")")
```
jbothma: @pudo if we move this date split out of this function, shall we have a go at using `h.apply_reviewed_names` on this set of suggested splits + a check_names_regularity to catch the things these splits don't catch?

The date splits don't seem to overlap with the other splits. There are around 350 (Russian's, 191 (also's.

datasets/ca/sema_sanctions_announcements/ca_sema_sanctions_announcements.yml
herrhellana: or `ca-sema-news` then, as friedrich proposed?

datasets/ca/sema_sanctions_news/ca_sema_sanctions_news.yml
```diff
@@ -0,0 +1,124 @@
...
+      - match: 'onOctober 27, 1963'
+        value: 1963-10-27
+      - match: 'on April19, 1952'
+        value: 1952-04-19
+  types.name:
+    options:
+      - match: 'A7 and A7 Limited Liability Company'
+        values:
+          - 'A7'
+          - 'A7 Limited Liability Company'
```
jbothma: let's remove these - I forgot to.

I moved the regex to detect these and included it in is_irregular so we'll review them and clean them in the reviews.

It's nice to have either of lookups or reviews for names, but ideally not have both.

datasets/ca/sema_sanctions_news/crawler.py
```diff
@@ -0,0 +1,140 @@
...
+    oranizations = h.xpath_strings(
+        specific_prohibition, ".//div[contains(., 'List of entities')]//li/text()"
+    )
+    assert sum([bool(persons), bool(oranizations)]) == 1, (
+        "Expected exactly one of persons or organizations",
+        len(persons),
+        len(oranizations),
+        entity_type,
+        listing_date,
+    )
```
jbothma, line 81: this is raising - let's see what we're missing and fix

datasets/ca/sema_sanctions_news/crawler.py
```diff
@@ -0,0 +1,140 @@
...
+            is_irregular=crawler_is_irregular or helper_is_irregular,
+        )
+
+        entity.add("topics", "sanction")
+        sanction = h.make_sanction(
+            context,
+            entity,
+            program_key=PROGRAM_KEY,
+        )
+        sanction.add("program", country)
```
jbothma, line 120: mmmm are you sure? @nvmbrasserie is this right for canada?

herrhellana: adding CA-SEMA? or are you talking about adding a country? i just looked at how we do it at the main canadian sanctions list via a lookup -- maybe i do the same here using `h.lookup_sanction_program_key()` and warn on unmatched countries? but basically we have only russia, belarus, and ukraine (disinfo agents) targeted there. 

all regulations are just CA-SEMA (the Special Economic Measures Act), at least what i get from the source page description.

nvmbrasserie: it's a good point. yes, let's do it here in the same way as we do in `ca_dfatd_sema_sanctions` (via `h.lookup_sanction_program_key()`)

datasets/ca/sema_sanctions_news/ca_sema_sanctions_news.yml
```diff
@@ -48,17 +48,18 @@ names:
...
-      Person: 1500
-      LegalEntity: 600
-      Vessel: 80
-  max:
-    schema_entities:
-      Person: 4000
-      LegalEntity: 1500
-      Vessel: 1600
+# TODO once we review the names:
+# assertions:
```
nvmbrasserie: the number of entities won't be affected by the names reviews. it might only affect the number of statements, so we can add assertions back in

datasets/ca/sema_sanctions_news/crawler.py
herrhellana: added! the program key for the vessels remains hardcoded though: the website doesn't mention a country name next to the ship, just all active sanctions + country name before the vessels table. we decided not to crawl this section

jbothma: sorry - I was getting confused between the `programId` and `program` props - I see in ca_dfatd, `program` is also `country`

PR #3640, opened 2026-02-25
[us-fincen-enforcements] US FINCEN Enforcement Actions

herrhellana:
issue: https://github.com/opensanctions/crawler-planning/issues/371

### DIFF
--- a/datasets/us/fincen_enforcement/us_fincen_enforcement.yml
+++ b/datasets/us/fincen_enforcement/us_fincen_enforcement.yml
@@ -3,7 +3,7 @@ prefix: us-fincenenf
 entry_point: crawler.py
 resolve: true
 load_statements: true
-ci_test: true
+ci_test: false # CI fails assertions without reviewed entity splits
 coverage:
   frequency: weekly
   start: 2026-02-25


datasets/us/fincen-enforcement/us_fincen_enforcement.yml
```diff
@@ -0,0 +1,184 @@
...
+        entities:
+          - ["Gibraltar Private Bank and Trust Company"]
+      - match: "Sigue Corporation and Sigue, LLC"
+        entities:
+          - ["Sigue Corporation and Sigue, LLC"]
+      - match: "Deprez’s Quality Jewelry and Loans, Inc."
+        entities:
+          - ["Deprez’s Quality Jewelry and Loans, Inc."]
+
+assertions:
```
jbothma: it's nice to organise the metadata such that everything including assertions come before lookups, just because lookups is usually the looooooooongest section by far.

datasets/us/fincen-enforcement/us_fincen_enforcement.yml
```diff
@@ -0,0 +1,184 @@
+title: "US Enforcement Actions List"
```
jbothma:
```suggestion
title: "US FinCEN Enforcement Actions"
```

datasets/us/fincen-enforcement/us_fincen_enforcement.yml
```diff
@@ -0,0 +1,184 @@
...
+prefix: us-fincen
```
jbothma:
```suggestion
prefix: us-fincenenf
```

Let's differentiate from us_fincen_special_measures

datasets/us/fincen-enforcement/us_fincen_enforcement.yml
```diff
@@ -0,0 +1,184 @@
...
+entry_point: crawler.py
+resolve: false
```
jbothma: We only set resolve to false for enrichment target datasets

```suggestion
```

datasets/us/fincen-enforcement/us_fincen_enforcement.yml
```diff
@@ -0,0 +1,184 @@
...
+load_statements: false
```
jbothma: this will be normal dataset representing risk directly, rather than an enrichment target, so let's load
```suggestion
load_statements: true
```

datasets/us/fincen-enforcement/us_fincen_enforcement.yml
```diff
@@ -0,0 +1,184 @@
...
+ci_test: false
```
jbothma: any reason not to run in CI? we prefer running in CI by default, and adding a comment explaining why we don't run in CI, e.g. if it's super slow, or if we're using zyte (becuase we don't want to put zyte credentials into CI)

datasets/us/fincen-enforcement/crawler.py
```diff
@@ -0,0 +1,53 @@
...
+
+    assert case_name is not None
+    if "in the matter of" in case_name.lower():
+        case_name = case_name.split("In the Matter of")[1].strip()
+
+    entity = context.make("LegalEntity")
+    entity.id = context.make_id(case_name, str_row.pop("matter_number"))
+
+    # custom lookups to split entities and aliases
+    if contains_split_phrase(case_name) or " and " in case_name:
```
jbothma: Let's use h.is_name_irregular instead of contains_split_phrase. it doesn't support " and " yet so the outcome would be the same at the moment but it will make it easy for us to find as a candidate for entity splitting in future

datasets/us/fincen-enforcement/crawler.py
```diff
@@ -0,0 +1,53 @@
...
+        res = context.lookup("comma_names", case_name, warn_unmatched=True)
+        if res and res.entities:
+            for entity_name in res.entities:
+                primary_name = entity_name[0]
+                entity.add("name", primary_name)
+                aliases = entity_name[1:] if len(entity_name) > 1 else []
```
jbothma: his could be simplifed to this, right? the list slice syntax tolerates being out of bounds

```suggestion
                aliases = entity_name[1:]
```

datasets/us/fincen-enforcement/crawler.py
```diff
@@ -0,0 +1,53 @@
...
+                for alias in aliases:
+                    entity.add("alias", alias)
+        else:
+            print(case_name)
```
jbothma: stray print :)

datasets/us/fincen-enforcement/crawler.py
```diff
@@ -0,0 +1,53 @@
...
+            entity.add("name", case_name)
+
+    entity.add("sourceUrl", url)
+    entity.add("sector", str_row.pop("financial_institution"))
```
jbothma: let's add topic `reg.action` and country `us`

datasets/us/fincen-enforcement/crawler.py
```diff
@@ -0,0 +1,53 @@
...
+
+    sanction = h.make_sanction(context, entity)
+    h.apply_date(sanction, "listingDate", str_row.pop("date_sort_ascending"))
```
jbothma: @pudo fincen enforcement actions data goes back to 1999 - should we apply within_max_age just to limit age a bit? maybe to 10 years already?

datasets/us/fincen-enforcement/crawler.py
```diff
@@ -0,0 +1,53 @@
...
+                primary_name = entity_name[0]
+                entity.add("name", primary_name)
+                aliases = entity_name[1:] if len(entity_name) > 1 else []
+                for alias in aliases:
+                    entity.add("alias", alias)
+        else:
+            print(case_name)
+            entity.add("name", case_name)
+
+    entity.add("sourceUrl", url)
```
jbothma: @pudo should we also do articles/documents for consistency or is sourceUrl ok?

datasets/us/fincen-enforcement/crawler.py
```diff
@@ -0,0 +1,53 @@
...
+    assert url_el is not None
+    url = h.xpath_strings(url_el, ".//a/@href")
+
+    # process row
+    str_row = h.cells_to_str(row)
+    case_name = str_row.pop("enforcement_action")
+
+    assert case_name is not None
+    if "in the matter of" in case_name.lower():
+        case_name = case_name.split("In the Matter of")[1].strip()
```
jbothma: if you wanna be a touch more conservative you can do `else lookup_value(clean_name_prefix, name, name, warn=True)`

datasets/us/fincen-enforcement/crawler.py
pudo: Yeah 10 years sounds great!

datasets/us/fincen-enforcement/crawler.py
pudo: Hm I suggested skipping `Document` to @herrhellana - although doing articles/documentation might make some sense. But I think it's also OK to just ship with `sourceUrl`.

datasets/us/fincen-enforcement/us_fincen_enforcement.yml
herrhellana: golden words

datasets/us/fincen-enforcement/crawler.py
herrhellana: woopppsssssss

datasets/us/fincen-enforcement/crawler.py
herrhellana: mm not 100% sure what you mean here, like what kind of a lookup are we talking about?

pudo:
@jbothma What do you think about this one as a candidate for using review over YAML? I feel like basically it has a signficant percentage of payload data in the metadata, which feels like a threshold criteria.

datasets/us/fincen_enforcement/crawler.py
```diff
@@ -0,0 +1,67 @@
...
+    entity.add("sector", str_row.pop("financial_institution"))
+    entity.add("topics", "reg.action")
+    entity.add("country", "us")
+
+    sanction = h.make_sanction(context, entity)
+    sanction_date = str_row.pop("date_sort_ascending")
+    if sanction_date and not enforcements.within_max_age(
+        context, sanction_date, MAX_AGE_DAYS
+    ):
+        return
```
jbothma, line 45: it could be nice to do this at the top, especially since it might prevent unnecessary name splitting/cleaning

datasets/us/fincen_enforcement/crawler.py
```diff
@@ -0,0 +1,107 @@
...
+    assert case_name is not None
+    case_name = case_name.replace("In the Matter of", "")
+    assert "in the matter of" not in case_name.lower()
+
+    entity = context.make("LegalEntity")
+    matter_number = str_row.pop("matter_number")
+    entity.id = context.make_id(case_name, matter_number)
+
+    # send to cleaning if a string contains split entities and aliases
+    if contains_split_phrase(case_name) or " and " in case_name:
```
jbothma: remember the point is that the case name can contain multiple entities, so we want to iterate over respondents and make an entity for each

Image

what you could do is instantiate Respondents([Respondent(name=case_name])) in the else clause and then make the entities after

jbothma: double-check the names in your output to see if there are more cases we need to catch - let's just throw anything weird to review 

Image

datasets/us/fincen_enforcement/crawler.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def crawl(context: Context) -> None:
+    doc = context.fetch_html(context.data_url, cache_days=1, absolute_links=True)
+    table = h.xpath_element(doc, ".//table")
+
+    for row in h.parse_html_table(table):
+        crawl_row(context, row)
+
+    # assert_all_accepted(context, raise_on_unaccepted=False)
```
jbothma: any reason to leave this commented out?

datasets/us/fincen_enforcement/crawler.py
herrhellana: i added an extra irregularity check on names + iterating over multiple entities now...

PR #3647, opened 2026-02-26
HATVP French declarations crawler

herrhellana:
issue: https://github.com/opensanctions/crawler-planning/issues/611

### DIFF
--- a/datasets/fr/hatvp_declarations/fr_hatvp_declarations.yml
+++ b/datasets/fr/hatvp_declarations/fr_hatvp_declarations.yml
@@ -39,7 +39,7 @@ data:
 tags:
   - list.pep
 
-# Wait until we resolve the position taxonomy in the UI and set is_pep=True
+# TODO: Wait until we resolve the position taxonomy in the UI and set is_pep=True
 # assertions:
 #   min:
 #     schema_entities:


datasets/fr/hatvp-declarations/crawler.py
```diff
@@ -0,0 +1,81 @@
...
+            occupancy = h.make_occupancy(
+                context,
+                person,
+                position,
+                categorisation=categorisation,
+            )
+            if occupancy is None:
+                return
+            h.apply_date(occupancy, "declarationDate", row.pop("date_depot"))
+            # type_mandat -> where to?
```
herrhellana: i'd love to preserve it, but i'm not sure which field it could go to. 

from Slack:
in my french data, i have both position names and mandate types

- Maire de Bobigny (position name) of "commune" mandate type (member of a municipal council)
- Vice-Président de la Métropole Rouen Normandie (position name) of "epci" mandate type (intercommunal public body)
- Sénateur de la Seine-Saint-Denis -- "senator" mandate type

herrhellana: [JD]: what we sometimes do is a mapping from level to position topics, e.g.
senator -> [gov.national, gov.legislative]
ideally as a lookup

datasets/fr/hatvp-declarations/crawler.py
```diff
@@ -0,0 +1,89 @@
...
+from zavod import Context, helpers as h
+from zavod.stateful.positions import categorise
+from rigour.mime.types import CSV
+
+
+def crawl(context: Context) -> None:
+    doc = context.fetch_html(context.data_url)
+
+    url_el = h.xpath_element(
+        doc,
```
nvmbrasserie: you can use `xpath_string` here with `'@href'`. then there is no need to assert of the length (`xpath_string` will warn unless it  exactly one result)

datasets/fr/hatvp-declarations/crawler.py
```diff
@@ -0,0 +1,89 @@
...
+                person,
+                position,
+                categorisation=categorisation,
+            )
+            if occupancy is None:
+                return
+            h.apply_date(occupancy, "declarationDate", row.pop("date_depot"))
+
+            url_declaration_pdf = "http://www.hatvp.fr/livraison/dossiers/" + row.pop(
+                "nom_fichier"
```
nvmbrasserie: `from urllib.parse import urljoin`
`url_declaration_pdf = urljoin("http://www.hatvp.fr/livraison/dossiers/", row.pop("nom_fichier"))`

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
```diff
@@ -0,0 +1,95 @@
...
+  prevent conflicts of interest.
+publisher:
+  name: Haute Autorité pour la Transparence de la vie Publique
+  name_en: High Authority for Transparency in Public Life
+  acronym: HATVP
+  description: |
+    La Haute Autorité pour la Transparence de la vie Publique (HATVP) is an independent administrative authority, responsible for ascertaining and
+    preventing potential conflicts of interest among French public servants and officials.
+    HATVP promotes the integrity and exemplarity of public officials,
+    monitors the ethics of civil servants, oversees lobbying and disseminates a cuture of integrity. Its independence is guaranteed
```
nvmbrasserie:
```suggestion
    monitors the ethics of civil servants, oversees lobbying and disseminates a culture of integrity. Its independence is guaranteed
```

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
```diff
@@ -0,0 +1,95 @@
...
+    by the functioning of its executive board and its financial and administrative autonomy.
+  country: fr
+  url: https://www.hatvp.fr/
```
nvmbrasserie:
```suggestion
  url: https://www.hatvp.fr/en/high-authority/institution/
```

datasets/fr/hatvp-declarations/crawler.py
```diff
@@ -0,0 +1,89 @@
...
+        for row in dict_reader:
+            title = row.pop("civilite")
+            first_name = row.pop("prenom")
+            last_name = row.pop("nom")
+            role = row.pop("qualite")
+
+            person = context.make("Person")
+            person.id = context.make_id(first_name, last_name, role)
+            h.apply_name(
+                person, prefix=title, first_name=first_name, last_name=last_name
```
nvmbrasserie: to construct cleaner names, it's better to move the title out of `h.apply_name`

datasets/fr/hatvp-declarations/crawler.py
```diff
@@ -0,0 +1,89 @@
...
+            )
+            person.add(
+                "idNumber", row.pop("id_origine")
+            )  # populated for deputies and senators only
+
```
nvmbrasserie:
```suggestion
person.add("title", row.pop("civilite")
```

datasets/fr/hatvp-declarations/crawler.py
```diff
@@ -0,0 +1,89 @@
...
+            url = "https://www.hatvp.fr" + row.pop("url_dossier")
+            person.add("sourceUrl", url)  # url to person's dossier
+
+            person.add("position", role)
+            person.add("topics", "role.pep")
```
nvmbrasserie: `  person.add("topics", "role.pep")` is done by `Occupancy`, so can be omitted here

datasets/fr/hatvp-declarations/crawler.py
```diff
@@ -0,0 +1,89 @@
...
+            person.add("country", "fr")
+
+            mandate_type = row.pop("type_mandat")
+            res = context.lookup("positions", mandate_type)
```
nvmbrasserie:
```suggestion
            res = context.lookup("positions", mandate_type, warn_unmatched=True)
```

datasets/fr/hatvp-declarations/crawler.py
```diff
@@ -0,0 +1,89 @@
...
+
+            position = h.make_position(
+                context,
+                name=role,
+                topics=topics,
+                country="fr",
+            )
+            position.add(
+                "subnationalArea", row.pop("departement")
+            )  # int for French departement number
```
nvmbrasserie: (nit) it makes for cleaner code if you add comments above the row: 

```suggestion
            # int for French departement number
            position.add("subnationalArea", row.pop("departement") ) 
```

datasets/fr/hatvp-declarations/crawler.py
```diff
@@ -0,0 +1,89 @@
...
+            h.apply_date(occupancy, "declarationDate", row.pop("date_depot"))
+
+            url_declaration_pdf = "http://www.hatvp.fr/livraison/dossiers/" + row.pop(
+                "nom_fichier"
+            )
+            occupancy.add("sourceUrl", url_declaration_pdf)  # url to declaration's PDF
+
+            context.emit(person)
+            context.emit(position)
+            context.emit(occupancy)
```
nvmbrasserie: (nit) the convention is: 
```suggestion
if Occupancy is not None:
            context.emit(person)
            context.emit(position)
            context.emit(occupancy)
```

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
```diff
@@ -0,0 +1,95 @@
...
+  official: true
+data:
+  url: https://www.hatvp.fr/open-data/
+  format: CSV
+tags:
+  - list.pep
+  - target.fr
```
nvmbrasserie: we use target for `sanctions` datasets usually. `list.pep` would be sufficient here

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
```diff
@@ -0,0 +1,94 @@
...
+          - gov.national
+          - gov.head
+      - match: senateur
+        topics:
+          - gov.national
+          - gov.legislative
+      - match: depute
+        topics:
+          - gov.national
+          - gov.legislative
```
jbothma: let's comment these out with a comment on which other official sources supply them

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
```diff
@@ -0,0 +1,94 @@
+title: France HATVP Declarations and Assessments
+prefix: fr-hatvp
+url: https://www.hatvp.fr/open-data/
+coverage:
+  frequency: daily  # unsure
```
jbothma: weekly is fine for for this

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
```diff
@@ -0,0 +1,94 @@
...
+entry_point: crawler.py
+summary: >-
+  Declarations of interests and assets, and associated assessments,
+  published by the French High Authority for Transparency in Public Life (HATVP).
+description: |
+  This dataset lists the declarations and assessments published by the
+  Haute Autorité pour la Transparence de la vie Publique (HATVP). French
+  public officials are required to declare their interests and assets to
+  the HATVP, which reviews and publishes them to promote transparency and
+  prevent conflicts of interest.
```
jbothma: Let's add a couple of paragraphs like

> this dataset provides the national cabinet,  regional council presidents and members, departmental council presidents, executives of large public bodies.
>
> Members of the national assembly and senate, and mayors, are not included from this dataset since other official sources are already covered by other datasets.

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
```diff
@@ -0,0 +1,94 @@
...
+          - gov.soe
+
+
+assertions:
+  min:
+    schema_entities:
+      Person: 6300  # raw count 7958
+  max:
+    schema_entities:
+      LegalEntity: 16000
```
jbothma: we can probably make assertions on the number of positions too

datasets/fr/hatvp-declarations/crawler.py
```diff
@@ -0,0 +1,86 @@
...
+            # int for French departement number
+            position.add("subnationalArea", row.pop("departement"))
+            categorisation = categorise(context, position, is_pep=True)
+
+            occupancy = h.make_occupancy(
+                context,
+                person,
+                position,
+                categorisation=categorisation,
+            )
```
jbothma: how old do these declarations get?

I think we can't really tell whether they currently occupy this position, so we need to add the arguments `no_end_implies_current=False` and `status=OccupancyStatus.UNKNOWN`

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
```diff
@@ -0,0 +1,94 @@
...
+          - gov.state
+          - gov.legislative
+      - match: departement
+        topics:
+          - gov.state
+          - gov.legislative
+      - match: commune
+        topics:
+          - gov.muni
+          - gov.executive
```
jbothma: this is mostly but not exclusively mayors.

We want to exclude mayors.

Do we already have a dataset providing `Conseillère déléguée à la Ville de Paris`? But I think it's safe enough to just drop this entire category.

```
grep -i commune data/datasets/fr_hatvp_declarations/source.csv | grep -i -v mair | csvlens -i  -d\;
```

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
```diff
@@ -0,0 +1,94 @@
...
+tags:
+  - list.pep
+
+lookups:
+  positions:
+    options:
+      - match: europe
+        topics:
+          - gov.igo
+          - gov.legislative
```
jbothma: if these are govered by https://www.opensanctions.org/datasets/eu_meps/ lets skip this entire category. Could you check if there are any positions here that aren't MEPs?

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
```diff
@@ -0,0 +1,94 @@
...
+          - gov.executive
+      # overseas and special-status territories:
+      - match: ctsp
+        topics:
+          - gov.state
+          - gov.legislative
+      # HATVP leadership:
+      - match: autre
+        topics:
+          - gov.soe
```
jbothma: I wuuldn't say this is a state-owned enterprise. It's some kind of public body but not a company.

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
herrhellana: only deputes and senators? maybe presidents too then?

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
herrhellana: yes i know, i thought of excluding mayors via a string match logic in .py but left it as is since we didnt come to any conclusion last week re: leaving vs excluding

herrhellana: this category also lists a bunch of mayor's assistants or sth along this line, i doubt we cover it elsewhere?

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
herrhellana: yes.... i made a blunt guess here and couldnt find a good enough field from the metadata dict to supply

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
herrhellana: yep i checked last week, MEPs only!

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
jbothma: Do we have another official source for the president?

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
jbothma: position topics are very much best effort. Rather missing than wrong. In part because we use e.g. the national/subnational level topic to determine how long people are marked as PEP after they leave office. Getting that wrong is more bad than just embarrassing.

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
```diff
@@ -0,0 +1,94 @@
+title: France HATVP Declarations and Assessments
```
jbothma: looks like these files are now in the - and the _ directory. Remember to commit the delete.

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
jbothma: should be resolved by delegating the decision to categorise()

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
herrhellana: not sth exclusive for France, thought we crawled it via Wikidata, but then it'd be enrichment. keep the presidents?

datasets/fr/hatvp_declarations/fr_hatvp_declarations.yml
```diff
@@ -0,0 +1,101 @@
...
+  published by the French High Authority for Transparency in Public Life (HATVP).
+description: |
+  This dataset lists the declarations and assessments published by the
+  Haute Autorité pour la Transparence de la vie Publique (HATVP). French
+  public officials are required to declare their interests and assets to
+  the HATVP, which reviews and publishes them to promote transparency and
+  prevent conflicts of interest.
+
+  This dataset provides the national cabinet, regional council presidents and members,
+  departmental council presidents, executives of large public bodies.
```
jbothma:
```suggestion
  departmental council presidents, and executives of large public bodies.
```

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
jbothma: should be resolved by delegating the decision to categorise()

datasets/fr/hatvp-declarations/fr_hatvp_declarations.yml
jbothma: should be resolved by delegating the decision to categorise()

datasets/fr/hatvp-declarations/crawler.py
jbothma: currently this would result in all of the occupancies having status "current", but we can't make such a strong assertion about their tenure, so we need to mark them as UNKNOWN

datasets/fr/hatvp_declarations/fr_hatvp_declarations.yml
```diff
@@ -0,0 +1,101 @@
+title: France HATVP Declarations and Assessments
```
jbothma:
```suggestion
title: France HATVP Declarations of interests and Assessments
```

datasets/fr/hatvp-declarations/crawler.py
herrhellana: the oldest `date_publication == 2019-07-23`. Set `no_end_implies_current=False` and `status=OccupancyStatus.UNKNOWN`

datasets/fr/hatvp_declarations/crawler.py
```diff
@@ -0,0 +1,100 @@
...
+        person, first_name=row.pop("first_name"), last_name=row.pop("last_name")
+    )
+    person.add("title", row.pop("title"))
+
+    url = urljoin("https://www.hatvp.fr", row.pop("dossier_url"))
+    person.add("sourceUrl", url)  # url to person's dossier
+    role = row.pop("role")
+    assert role is not None
+    person.add("position", role)
+    person.add("citizenship", "fr")
```
jbothma, line 23: are we 100% sure all the positions we might add using this require citizenship? Otherwise we need to flip this to country.

herrhellana: every position in the dataset requires French citizenship because at the municipal level, only mayors and adjoints are included, which require citizenship. We have some conseillers with special delegations that still require citizenship (eg conseillers départementaux and conseillers régionaux); EU citizens can be ordinary conseillers municipaux, but neither those nor conseillers without a delegation are subject to HATVP declarations. all other positions in the dataset (parliamentary, governmental, regional, departmental, EPCI, territorial) require French nationality.

(the only exception is MEPs, but we're dropping them here)

herrhellana: Law on who is subject to HATVP declarations: https://www.legifrance.gouv.fr/loda/article_lc/LEGIARTI000052092895

Note that the dataset also includes HATVP board members. Those are typically elected by the Conseil d'État, the Cour de cassation, and the Cour des comptes, the government, and the presidents of the National Assembly and Senate. Those all are French nationals in practice, https://www.hatvp.fr/la-haute-autorite/linstitution/organisation/

PR #3670, opened 2026-03-01
Kenya domestic list

### LINKED ISSUE #599: Kenya Domestic Terrorism List
### Data URL

https://www.frc.go.ke/?page_id=193

### Publisher

https://www.frc.go.ke

KE

### Type of data

Sanctions (Governments barring specific engagement e.g. financial transaction with these entities)

### Coverage region

region:Africa

### Can you tell us more?

Kenyan domestic terrorism list. XLSX file: https://www.frc.go.ke/wp-content/uploads/2026/02/Domestic-List_Kenya.xlsx

- [x] I understand.

pudo:
Fixes https://github.com/opensanctions/crawler-planning/issues/599 

Test case for #3669

### DIFF
--- a/datasets/ke/frc_sanctions/crawler.py
+++ b/datasets/ke/frc_sanctions/crawler.py
@@ -1,0 +1,87 @@
+from typing import Dict
+
+from normality import stringify
+from openpyxl import load_workbook
+from rigour.mime.types import XLSX
+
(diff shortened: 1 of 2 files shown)


pudo:
I think the underlying thing here is that they're lower-casing the text? I could definitely make it more explicit

PR #3687, opened 2026-03-02
Move `{black_sea,tokyo}_mou_psc` common code to shed

nvmbrasserie:
`tokyo_mou_psc` and `black_sea_mou_psc` are two crawlers for Port State Control inspection databases published by the Tokyo MoU and Black Sea MoU respectively. The two sites have nearly identical structure, so the crawlers had almost identical code.

This PR extracts the shared logic into `zavod/zavod/shed/bs_tokyo_mou_psc.py`, leaving only the login/CAPTCHA handling (which differs between the two) in each crawler.


leonhandreke (commented):
Added some commits to rename things and build URLs from `context.data_url` - hope you agree, if not, let me know :) Doing this not because I want to be prescriptive but to save some roundtrips.

One question: why do we do the `total_pages` dance in the crawler and not in the `shed` module?

Issue #3311, opened 2026-01-07
heuristic-based name categorisation


jbothma:
When llm-based extraction is disabled in `h.apply_reviewed_names`, strings are placed in `CleanNames.full_name` by default.

Some of the `is_name_regular` checks can propose a property for name strings. Specifically `us_sam_exclusions` has a lot of single token names that could default to weak_alias.

let's try that.

pudo:
I really liked the idea of having a background function, let's say `check_name_regularity` that returns an object (eg. `Regularity`) with (derived) properties like `is_irregular` and `suggested_prop`, and then making `is_name_regular` a caller of that. 

Some heuristic ideas (!): 
* Person entity single word name -> `weakAlias` (perhaps after running `remove_person_name_prefix` to clean off `The`, `El`)
* Organization name shorter than 8 letters, all uppercase -> `abbreviation`

Not sure what the full set here should be?

jbothma:
The person single word prefix suggestion would cover the vast majority of fsf family edits done by hand right now.

jbothma:
1. is_modern_alphabet  
2. statements diff cli
3. LangText
4. Person single token suggestion
5. eu_fsf multi names together

Issue #3601, opened 2026-02-20
tw_shtc still has lots of ; and a.k.a in 'name', could benefit from standard suggestions


jbothma:
Apply standard heuristics

Issue #3312, opened 2026-01-07
cleaning multiple name inputs (e.g. tw_shtc)


jbothma:
We'd like to standardise more of the tw_shtc name cleaning/categorisation.

the names logic for taiwan is a bit more fancy than the rest, since we get a primary name and an aliases field, and then do the swap between name and longest alias if needed (80 cases).

it looks like that makes quite a nice improvement to the names:

<img width="566" height="282" alt="Image" src="https://github.com/user-attachments/assets/db929f98-c4d3-4877-80c6-09b77e6c0f47" />

the name cleaner currently assumes the provided string is either provided as a primary name, or an alias (apply_name alias argument)

How important is it that this logic gets removed from the tw crawler?

How much does it overlap with needs from other sources, e.g. ofac?

> we really do need to consider language metadata (any solution that doesn't correctly handle it ends up being a stopgap)

- some sources give us multiple name fields for an entity, e.g. fsf family
- some give one or two strings, each containing any number of names e.g. tw_shtc, ru_cbr_banks
- splitting 
- some sources indicate a language for a name, e.g. eu_fsf (tw_shtc has a prefix in the string denoting a mandarin name)

This comes out of https://github.com/opensanctions/opensanctions/issues/2656#issuecomment-3702512696 where some initial options were proposed which don't yet consider language or other crawlers' needs

jbothma:
In https://github.com/opensanctions/opensanctions/pull/4324 we ensured reviews are created for the current automated cleaning.

In https://github.com/opensanctions/opensanctions/pull/4325 we removed the current automated cleaning.

80 names out of 10,000 suggests automating name swapping isn't that important. We already have heuristics to demote abbreviations, and we've set the minimum name length to 8 characters. Human review can probably easily fix the rest as they come.

PR #3690, opened 2026-03-02
 Move name suggestion heuristics to zavod

jbothma:
- Moves initial suggesting heuristics from us_sam_exclusions to zavod
- Nests schema-specific name regularity settings under `names.schema_rules`

[See example change](https://github.com/opensanctions/opensanctions/pull/3690/changes#diff-70ad474990a784002c1c58fa41522b202b7595baddd6f5a5962771f8da5a7da8L43)

Initially these names are really verbose - see comment in _check_heuristics() - I think it's better to start off with these ugly feature flag names and add more distinct rules with similarly ugly names if we need to while we figure out which heuristics are reused, and later settle on prettier names for the heuristics that survive. I think we should avoid the temptation to add lots of conditions to a smaller number of heuristics.

This PR lets us start trying out the suggesting heuristics in more crawlers to see how they want to evolve. e.g. can we get rid of most of the custom logic in tw_shtc with this plus tw_shtc's alias swapping?

Wraps up https://github.com/opensanctions/opensanctions/issues/3311

### DIFF
--- a/zavod/zavod/meta/names.py
+++ b/zavod/zavod/meta/names.py
@@ -70,19 +70,19 @@ class NamesSpec(BaseModel):
     schema_rules: Dict[str, CleaningSpec] = dict(_DEFAULT_SCHEMA_RULES)
     """Name cleaning requirements by schema. All matching schema configurations will apply."""
 
-    suggest_person_single_token: bool = False
+    suggest_weak_alias_person_single_token: bool = False
     """
     If True, single-token Person names (after stripping name prefixes such as "Mr.")
     are suggested as weakAlias rather than name.
     """
 
-    suggest_uppercase_org_single_token_shorter_than: Optional[int] = None
+    suggest_abbreviation_uppercase_org_single_token_shorter_than: Optional[int] = None
     """
     If set, Organization names that are all-uppercase, contain no spaces, and are
     shorter than this threshold are suggested as abbreviation rather than name.
     """
 
-    suggest_non_person_single_token_shorter_than: Optional[int] = None
+    suggest_abbreviation_non_person_single_token_shorter_than: Optional[int] = None
     """
     If set, LegalEntity-but-not-Person names (i.e. companies, organisations, vessels, etc.)
     that are all-uppercase, contain no spaces, and are shorter than this threshold are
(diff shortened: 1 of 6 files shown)


zavod/zavod/meta/names.py
```diff
@@ -45,49 +44,77 @@ def reject_chars_consolidated(self) -> Set[str]:
...
+    schema_rules: Dict[str, CleaningSpec] = dict(_DEFAULT_SCHEMA_RULES)
+    """Name cleaning requirements by schema. All matching schema configurations will apply."""
+
+    suggest_person_single_token: bool = False
+    """
+    If True, single-token Person names (after stripping name prefixes such as "Mr.")
+    are suggested as weakAlias rather than name.
+    """
+
+    suggest_uppercase_org_single_token_shorter_than: Optional[int] = None
```
leonhandreke: (nit) very minor nit, but I feel like adding the consequence to the name, like `suggest_abbreviation_uppercase_org_single_token_shorter_than` would make these a lot more speaking when reading the YAML.

datasets/tw/shtc/tw_shtc.yml
```diff
@@ -34,10 +34,11 @@ tags:
...
 
 names:
-  LegalEntity:
-    reject_chars: "()（）"
-    allow_chars: ";" # We're splitting on semicolons so don't bother requiring review by default
-    single_token_min_length: 6
+  schema_rules:
+    LegalEntity:
+      reject_chars: "()（）"
+      allow_chars: ";" # We're splitting on semicolons so don't bother requiring review by default
```
leonhandreke, line 40: I don't quite understand - we're allowing `;` to pass, and the splitting happens after the review step?

zavod/zavod/meta/names.py
jbothma: I'm already so awkward about these long names, but I do agree it's an improvement.

leonhandreke: I like long names :)

datasets/tw/shtc/tw_shtc.yml
jbothma: hm - it might have been needed when I tried to add check_names_regularity before the crawler-custom stuff but that isn't there now.

What I found at the time was that even the current stuff seams to leave a lot of ; in there so I made https://github.com/opensanctions/opensanctions/issues/3601

But we might already deal with it in https://github.com/opensanctions/opensanctions/issues/3312

Since it isn't a change here, I'm gonna rely on figuring it out in those upcoming tasks. Good spot though.

PR #3699, opened 2026-03-03
Apply length and single token checks only to modern scripts

jbothma:
https://github.com/opensanctions/opensanctions/issues/3311#issuecomment-3913033763

### DIFF
--- a/zavod/zavod/meta/names.py
+++ b/zavod/zavod/meta/names.py
@@ -29,11 +29,13 @@ class CleaningSpec(BaseModel):
     Remember that characters defined for other matching schema specs will still apply.
     """
     min_length: int = 2
-    """Minimum length for names. Applies only to "modern" alphabets."""
+    """Minimum length for names. Does not apply to "dense" writing systems like Han for Chinese."""
     single_token_min_length: int = 2
-    """Minimum length for names with no spaces, i.e. a single token. Applies only to "modern" alphabets."""
+    """Minimum length for names with no spaces, i.e. a single token.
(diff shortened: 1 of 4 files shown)


zavod/zavod/meta/names.py
```diff
@@ -4,13 +4,15 @@
...
 from followthemoney.schema import Schema
-from pydantic import BaseModel, RootModel
+from pydantic import BaseModel, ConfigDict, RootModel
 
 
 log = getLogger(__name__)
 
 
 class CleaningSpec(BaseModel):
+    model_config = ConfigDict(extra="forbid")
```
leonhandreke, line 14: What does this do?

jbothma: This causes a validation error if unexpected keys are in the metadata. 
e.g. I didn't realise the code expected `min_chars` but the crawler was using `min_length`.

jbothma: Reckon a comment is needed to explain?

leonhandreke: Just added one :).

jbothma:
ok, I think what we want is `is_modern_alphabet` as a rough proxy for "does it make sense to look for spaces" and `is_cjk` as a rough proxy for "length checks of the order of magnitude of latin script make sense"

jbothma:
> What if we made a new is_logographic helper in rigour

Will do

> I thought Arabic also had spaces in names?

You're right. It turns out Google Translate is transliterating the arabic names in eu_fsf sensibly (with spaces) doesn't mean they're correct.

PR #3703, opened 2026-03-03
AE DFSA debarment list

herrhellana:
issue: https://github.com/opensanctions/crawler-planning/issues/336

### DIFF
--- a/datasets/ae/dfsa_debarred/ae_dfsa_debarred.yml
+++ b/datasets/ae/dfsa_debarred/ae_dfsa_debarred.yml
@@ -49,7 +49,7 @@ dates:
 assertions:
   min:
     schema_entities:
-      Person: 20  # 30 raw count
+      Person: 20  # 27 raw count
   max:
     schema_entities:
       Person: 60
--- a/datasets/ae/dfsa_debarred/crawler.py
+++ b/datasets/ae/dfsa_debarred/crawler.py
@@ -1,12 +1,7 @@
 from lxml.etree import _Element
 
 from zavod import Context, helpers as h
-from zavod.extract.zyte_api import fetch_html
-
-
-ACTIONS = [
-    {"action": "scrollBottom"},
-]
+from zavod.extract.zyte_api import fetch_html, UnblockFailedException
 
 
 def crawl_person(context: Context, row: _Element) -> None:
@@ -43,16 +38,31 @@ def crawl_person(context: Context, row: _Element) -> None:
 
 
 def crawl(context: Context) -> None:
-    xpath_table = ".//div[@class='table-content prohibited--grid']"
+    xpath_el = ".//a[@class='table-row']"
+    page_num = 1
 
-    # using Zyte API to bypass the 403 error
-    doc = fetch_html(
-        context,
-        context.data_url,
-        xpath_table,
-        actions=ACTIONS,
-    )
-    table = h.xpath_element(doc, xpath_table)
+    # using Zyte to bypass the 403 status response
+    # paginate over the AJAX endpoint until an empty reponse
 
-    for row in h.xpath_elements(table, ".//a"):
-        crawl_person(context, row)
+    # xpath_el is used as a Zyte unblocking validator
+    # it raises UnblockFailedException on empty pages, which is used as the pagination stop condition
+    while True:
+        try:
+            doc = fetch_html(
+                context,
+                f"{context.data_url}?page={page_num}&status=&keywords=&isAjax=true",
+                xpath_el,
+            )
+        except UnblockFailedException:
+            # if UnblockFailedException past the fist page, stop paginating
+            # (we've likely hit an empty page)
+            if page_num > 1:
+                break
+            # loudly raise an error if we fail at page 1
+            raise
+
+        assert doc is not None
+        for row in h.xpath_elements(doc, xpath_el):
+            crawl_person(context, row)
+
+        page_num += 1


datasets/ae/dfsa_debarred/ae_dfsa_debarred.yml
```diff
@@ -0,0 +1,50 @@
+title: United Arab Emirates Prohibited/Restricted Individuals by the DFSA
+prefix: ae-dfsa-debarred
+entry_point: crawler.py
+coverage:
+  frequency: daily  # is it good?
+  start: 2026-03-03
+load_statements: true
```
herrhellana, line 7: maybe set `ci_test` to `True` for this crawler?

datasets/ae/dfsa_debarred/ae_dfsa_debarred.yml
```diff
@@ -0,0 +1,44 @@
+title: United Arab Emirates Prohibited/Restricted Individuals by the DFSA
```
jbothma:
```suggestion
title: United Arab Emirates DFSA Prohibited/Restricted Individuals
```

datasets/ae/dfsa_debarred/ae_dfsa_debarred.yml
```diff
@@ -0,0 +1,44 @@
...
+prefix: ae-dfsa-debarred
```
jbothma:
```suggestion
prefix: ae-dfsa
```

I can't think of any other list that would come from the DFSA. This is at least a solid default

datasets/ae/dfsa_debarred/crawler.py
```diff
@@ -0,0 +1,42 @@
...
+    sanction.add("sourceUrl", url)
+    sanction.add("status", status)
+    sanction.add("country", "ae")
+
+    context.emit(person)
+    context.emit(sanction)
+
+
+def crawl(context: Context) -> None:
+    doc = fetch_html(
```
jbothma: it's nice to note why zyte is needed. I haven't documented this anywhere yet. Looks like there's a cloudflare javascript challenge?

datasets/ae/dfsa_debarred/crawler.py
```diff
@@ -0,0 +1,42 @@
...
+        context,
+        context.data_url,
+        ".//div[@class='table-content prohibited--grid']",
+        actions=ACTIONS,
+    )
+    table = h.xpath_element(doc, ".//div[@class='table-content prohibited--grid']")
```
jbothma: stylistic: when using the same xpath for the unblock check, it can be quite tidy to assign the xpath to a variable e.g. `table_xpath`. Then it's also a bit more obvious what you're looking for in the unblock check

datasets/ae/dfsa_debarred/crawler.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def crawl_person(context: Context, row: _Element) -> None:
+    url = row.get("href")  # link to case's PDF
+    # not all rows have three values to unpack, this is a workaround to fetch text
+    date, name, status = [el.tail for el in h.xpath_elements(row, ".//p/span")]
+
+    person = context.make("Person")
+    person.id = context.make_id(name, url)
+    person.add("name", name)
```
jbothma, line 19: let's skip the redacted people 

Image

datasets/ae/dfsa_debarred/crawler.py
```diff
@@ -0,0 +1,42 @@
...
+    person.add("topics", "debarment")
+
+    sanction = h.make_sanction(context, person)
+    h.apply_date(sanction, "listingDate", date)
+    sanction.add("sourceUrl", url)
+    sanction.add("status", status)
```
jbothma: let's continue emitting the people where status is past, but

- only assign the topic when status does not contain "past"
- split the date from the status and if it contains past, add it to endDate on a Sanction entity
- move status to the Sanction entity

datasets/ae/dfsa_debarred/ae_dfsa_debarred.yml
```diff
@@ -0,0 +1,44 @@
...
+entry_point: crawler.py
+coverage:
+  frequency: daily  # is it good?
+  start: 2026-03-03
+load_statements: true
+summary: >-
+  The DFSA maintains a list of individuals and entities that are prohibited or restricted from participating in
```
jbothma:
```suggestion
  Individuals and entities that are prohibited or restricted from participating in...
```

datasets/ae/dfsa_debarred/ae_dfsa_debarred.yml
```diff
@@ -0,0 +1,44 @@
...
+  financial services within the Dubai International Financial Centre (DIFC).
+description: |
+  > Under the Rule 9.8.3 of the Markets Rules module of the DFSA Rulebook, Reporting Entities are
+  required to disclose to the Market information through a DFSA approved Regulatory Announcement Service.
```
jbothma: See if you can expand this a smidgen. Feel free to ask claude for help. One thing that could be interesting is examples of causes, e.g. "for example, providing clients with materially inaccurate information regarding the value of the their assets" or whatever floats your boat looking at a couple of reports.

It's also not wrong to repeat the key bit of the summary: "individuals and entities listed here are are prohibited or restricted from participating in..."

datasets/ae/dfsa_debarred/ae_dfsa_debarred.yml
```diff
@@ -0,0 +1,44 @@
+title: United Arab Emirates Prohibited/Restricted Individuals by the DFSA
+prefix: ae-dfsa-debarred
+entry_point: crawler.py
+coverage:
+  frequency: daily  # is it good?
```
jbothma: I guess weekly is fine? what does the rest of the debarments collection look like?

datasets/ae/dfsa_debarred/ae_dfsa_debarred.yml
jbothma: won't work if we're using zyte. Better set it to false.

herrhellana: ah true, i forgot about zyte.

datasets/ae/dfsa_debarred/crawler.py
herrhellana: it's a dynamic page load here, so without zyte (the ScrollDown action), we're fetching the first 10 entries only

datasets/ae/dfsa_debarred/crawler.py
herrhellana: this makes sense, thank you! i'll put it in.

datasets/ae/dfsa_debarred/crawler.py
herrhellana: oh, sorry, why zyte -- i think it was a 403 error, i'll double check and leave a comment about it!

datasets/ae/dfsa_debarred/ae_dfsa_debarred.yml
herrhellana: daily mostly in other collections, so i went with this freq. But maybe we want to adjust it to DFSA's data? Hard to tell how often the website is updated.

datasets/ae/dfsa_debarred/crawler.py
herrhellana: done, not sure what to do with this guy though: https://365343652932-web-server-storage.s3.eu-west-2.amazonaws.com/files/9316/7575/5147/20230130_Dalma_FMT_Sanctions_Susp_Order.pdf

(no endDate listed in the string, but the timeframe, 2 years, is mentioned in the PDF file)

datasets/ae/dfsa_debarred/crawler.py
jbothma: aaa I missed the scroll action. Clever, but there's a risk that the page hasn't loaded all the pages - currently it does three requests. this might mean we intermittently miss the last few people depending on timing.

Let's rather paginate over the endpoint giving the results directly by incrementing page number until the response is empty https://www.dfsa.ae/public-register/prohibited-individuals?page=1&status=&keywords=&isAjax=true. Maybe zyte isn't even needed then.

datasets/ae/dfsa_debarred/crawler.py
jbothma: I think you can do 
```
if "past" in status.lower() in status.lower():
        date_end = status.split("(")[1]
        date_ennd = date_end.strip(")")
        date_end = date_end.replace("expired", "")
        h.apply_date(sanction, "endDate", date_end)
```

datasets/ae/dfsa_debarred/crawler.py
herrhellana: Ok sounds good! Zyte is still needed unfortunately bc of the 403

datasets/ae/dfsa_debarred/crawler.py
herrhellana: hm... it's not that much different from my current implementation (any reason this ^ is preferred?) and will fail for the [guy](https://365343652932-web-server-storage.s3.eu-west-2.amazonaws.com/files/9316/7575/5147/20230130_Dalma_FMT_Sanctions_Susp_Order.pdf) in question (whose endDate is missing from the string but is present in the PDF doc). 

But I thought of another alternative (pushed up) where I just don't append the debarment status to the guy in question.

jbothma: basically the differences with my suggestion from above:

- do one cleaning step per line
- expired didn't exist in Zachary's status, so I think it would have been an exception
- my suggestion would log a warning on the law reference not validating as a date, which can be squashed with a type.date lookup to null.

```
        try:
            date_end = status.split("expired ")[1].strip(")")
            h.apply_date(sanction, "endDate", date_end)
        except IndexError:
            pass
```

A good mantra is "exceptions should be used for exceptional circumstances". Generally it's best not to use exceptions for control flow. There are exceptions to that (lol) - e.g. int(variable_which_is_occasionally_not_an_int) but that should be very rarely needed in crawler code.

Specifically here, I think `replace("expired", "")` is more explicit about what we're doing than `try split("expired")` therefore more easily understandable.

herrhellana: aha ok understood! makes sense with the lookups. sometimes i forget we can use it in instances as such, and i still want the code to handle everything...

datasets/ae/dfsa_debarred/crawler.py
```diff
@@ -0,0 +1,68 @@
...
+                f"{context.data_url}?page={page_num}&status=&keywords=&isAjax=true",
+                xpath_el,
+            )
+        except UnblockFailedException:
+            # if UnblockFailedException past the fist page, stop paginating
+            # (we've likely hit an empty page)
+            if page_num > 1:
+                break
+            # loudly raise an error if we fail at page 1
+            raise
```
jbothma, line 62: Can I be really annoying and ask you to try zyte_api.fetch_text + lxml.html.fromstring without any custom retry stuff - just tell it you're expecting text/html and charset=UTF-8

If it turns out we need the xpath-based unblock verification we'll find out quickly, but that feels safer and less weird than bailing on an UnblockFailedException

datasets/ae/dfsa_debarred/crawler.py
jbothma: yeah it ends up being a super nice pattern - whenever anything unexpected happens, we get a warning. And the weird stuff is pulled out from the depth of the code "to the forefront"

datasets/ae/dfsa_debarred/crawler.py
```diff
@@ -0,0 +1,68 @@
...
+    sanction.add("status", status)
+
+    # add endDate if the status is expired & don't add debarment topic
+    if "past" in status.lower():
+        date_end = status.split("(")[1]
+        date_end = date_end.strip(")")
+        date_end = date_end.replace("expired ", "")
+        h.apply_date(sanction, "endDate", date_end)
+    else:
+        person.add("topics", "debarment")
```
jbothma, line 34: sorry - late thought - let's warn if status is anything other than Ongoing in this clause

datasets/ae/dfsa_debarred/crawler.py
```diff
@@ -0,0 +1,68 @@
...
+    # skip anonymized entries
+    if name.lower() == "an individual":
+        return
+
+    person = context.make("Person")
+    person.id = context.make_id(name, url)
+    person.add("name", name)
+
+    sanction = h.make_sanction(context, person)
+    h.apply_date(sanction, "listingDate", date)
```
jbothma, line 22: "date of restriction" feels more like `startDate`

datasets/ae/dfsa_debarred/ae_dfsa_debarred.yml
jbothma: In that case I think daily is cool.

datasets/ae/dfsa_debarred/crawler.py
herrhellana: hm i think makes sense! when do we want to use `listingDate` ?

datasets/ae/dfsa_debarred/crawler.py
herrhellana: aight lemme try

datasets/ae/dfsa_debarred/crawler.py
jbothma: listingDate often makes sense when there's e.g. a publication date. Try a search in the codebase to see what else comes up. startDate refers to the precise dates during which the restriction applied, but a notice of a restriction could be published before or after the effective date.

PR #3752, opened 2026-03-09
[programs] revised taxonomy for sanction provisions

nvmbrasserie:


zavod/zavod/stateful/programs.py
```diff
@@ -6,14 +6,51 @@
...
 
 Measure = Literal[
+    # Suspension of foreign aid, development funding, or multilateral
+    # lending to a country.
+    "Aid suspension",
+    # Blanket prohibition on arms/military equipment to a country or regime.
+    # Country-wide scope only. If targeted at specific entities, use
+    # "Export control" instead.
     "Arms embargo",
-    "Arms export restrictions",
```
nvmbrasserie, line 10: merging the two as suggested by Claude: 

```
"Arms embargo" vs. "Arms export restrictions" — these two are closely related and the boundary is fuzzy. An arms embargo is typically a comprehensive, country-wide prohibition (often UN-mandated), while arms export restrictions can be more targeted or partial. 
```

zavod/zavod/stateful/programs.py
```diff
@@ -6,14 +6,51 @@
...
+    # Includes US-style "blocking" — treat these as equivalent.
     "Asset freeze",
+    # Exclusion from government procurement, contracts, or programs
+    # (Medicaid/Medicare exclusions, World Bank debarment, SAM.gov).
+    "Debarment",
+    # Restrictions on export of dual-use goods, technology, military items,
+    # or luxury goods to specific destinations or end-users. Covers both
+    # outright bans and licensing requirements.
     "Export control",
-    "Financial block",
```
nvmbrasserie, line 13: `"Financial block"` became `"Financial Restrictions"` for broader coverage

zavod/zavod/stateful/programs.py
```diff
@@ -6,14 +6,51 @@
...
+    # existing assets.
     "Investment ban",
+    # Prohibition on providing professional services (legal, accounting, IT,
+    # consulting, engineering, advertising, trust formation) to sanctioned
+    # countries or persons. No physical goods involved.
+    "Services ban",
+    # Bars sanctioned parties from claiming compensation for the effects of
+    # sanctions via litigation or arbitration.
     "Prohibition to satisfy claims",
-    "Restrictions on goods",
```
nvmbrasserie, line 16: `"Restrictions on goods"` was removed in favour of: `Import restrictions` and `Export Control`

zavod/zavod/stateful/programs.py
```diff
@@ -6,14 +6,51 @@
...
+    # Suspension of foreign aid, development funding, or multilateral
+    # lending to a country.
+    "Aid suspension",
+    # Blanket prohibition on arms/military equipment to a country or regime.
+    # Country-wide scope only. If targeted at specific entities, use
+    # "Export control" instead.
     "Arms embargo",
-    "Arms export restrictions",
+    # Freezing of funds and economic resources of a designated person/entity.
+    # Includes US-style "blocking" — treat these as equivalent.
```
leonhandreke: What's "US-style blocking"?

zavod/zavod/stateful/programs.py
```diff
@@ -6,14 +6,51 @@
...
     "Asset freeze",
+    # Exclusion from government procurement, contracts, or programs
+    # (Medicaid/Medicare exclusions, World Bank debarment, SAM.gov).
+    "Debarment",
+    # Restrictions on export of dual-use goods, technology, military items,
+    # or luxury goods to specific destinations or end-users. Covers both
+    # outright bans and licensing requirements.
     "Export control",
-    "Financial block",
+    # Systemic financial measures: correspondent banking bans, SWIFT cutoffs,
```
leonhandreke: Does that mean they only apply to whole countries?

nvmbrasserie: that's a good catch, I'll update the description

nvmbrasserie:
I skimmed through a few papers, but the academic definition isn’t quite what we’re looking for here. I think the current taxonomy represents a middle ground between the actual measures implemented by the EU and the US, although there isn’t a single clear source that explicitly grounds it in those frameworks. That said, I’m happy for it to be on the website, and I think it strikes a good balance between how the world works in practice and the level of detail we need from a practical perspective

PR #3756, opened 2026-03-09
Portugal parliament: current and historical members

herrhellana:
Issue: https://github.com/opensanctions/crawler-planning/issues/621

### DIFF
--- a/datasets/pt/parliament/pt_parliament.yml
+++ b/datasets/pt/parliament/pt_parliament.yml
@@ -38,7 +38,8 @@ dates:
 assertions:
   min:
     schema_entities:
-      Person: 1800  # raw count 2291 (current + historical, incl. temporary MPs)
+      # count of Efetivo statuses (not necessarily unique MPs) from source: 2738
+      Person: 1800  # raw count for Efetivo from stats: 2257 (current + historical, incl. temporary MPs)
       Position: 1
   max:
     schema_entities:
@@ -49,17 +50,17 @@ lookups:
     # sioDes indicates whether a period in the data represents active membership or non-membership.
     options:
       - match:
-          - "Efetivo" # Effective (used in current and historical periods)
+          - "Efetivo"  # Effective (used in current and historical periods)
           - "Efetivo Definitivo"
           - "Efetivo Temporário"
-          - "Suplente" # Alternate
         value: "true"
       - match:
-          - "Renunciou" # Resigned
-          - "Suspenso(Não Eleito)" # Suspended (Not Elected)
-          - "Suspenso(Efet Def)" # Suspended (Effective Definitive)
-          - "Suspenso(Eleito)" # Suspended (Elected)
-          - "Desistência" # Withdrawal
-          - "Impedido" # Disqualified
-          - "Falecido/a" # Deceased
+          - "Suplente"  # Substitutes (elected on the party list but are not occupying a seat until they're called up)
+          - "Renunciou"  # Resigned
+          - "Suspenso(Não Eleito)"  # Suspended (Not Elected)
+          - "Suspenso(Efet Def)"  # Suspended (Effective Definitive)
+          - "Suspenso(Eleito)"  # Suspended (Elected)
+          - "Desistência"  # Withdrawal
+          - "Impedido"  # Disqualified
+          - "Falecido/a"  # Deceased
         value: "false"


datasets/pt/parliament/crawler.py
```diff
@@ -0,0 +1,131 @@
...
+        )
+        position.add("subnationalArea", member.pop("DepCPDes"))  # constituency name
+        position.add("subnationalArea", member.pop("DepCPId"))  # constituency ID
+        categorisation = categorise(context, position, is_pep=True)
+
+        if not categorisation.is_pep:
+            continue
+
+        # --- fetch non-leadership seat occupancy in the plenary ---
+        for seat_occupancy in member.pop("DepSituacao") or []:
```
herrhellana, line 54: i looked a bit more into the occupancy status in the source data, and i need to change how i code `sioDtInicio` and `sioDtFim` -- the dates are related not to the MP seat but to the status of MP seat occupancy change if any. 

E.g. this is `seat_occupancy` for [Francisco Meneses de Lima](https://pt.wikipedia.org/wiki/Francisco_Meneses_de_Lima):

```
(Pdb) seat_occupancy
{'sioDes': 'Renunciou', 'sioDtFim': None, 'sioDtInicio': '2025-06-03'}
```

which means that Francisco resigned from his mandate on March 6th, 2025 (not reflected on wiki). I was able to find a parliamentary [hearing](https://debates.parlamento.pt/catalogo/r3/dar/01/17/01/001/2025-06-03/7?pgs=4&org=PLC&plcdf=true) about it:

> Electoral District of the Autonomous Region of the Azores—Francisco Gabriel Meneses de Lima resigned from his mandate and was replaced by Ana Maria Lourenço Martins, in light of the resignation submitted by the unsuccessful candidate from the Chega party.

but then I don't quite get why e.g. [Francisco Miguel Vital Gomes do Vale César](https://pt.wikipedia.org/wiki/Francisco_C%C3%A9sar) has his status record looking like this:
```
(Pdb) seat_occupancy
{'sioDes': 'Efetivo', 'sioDtFim': None, 'sioDtInicio': '2025-06-03'}
```
(note that this is the only observation we have in the list `member.get("DepSituacao")` for this person)


So my thinking: 
- I could hardcode legislature dates (based on wiki data) for MP seats (start-end), and then edit it based on `seat_occupancy` to capture any temporary or resigned MPs, and do nothing if `seat_occupancy['sioDes'] == 'Efetivo'`. Thoughts?

datasets/pt/crawler.py
```diff
@@ -0,0 +1,131 @@
...
+        "//div[@id='ctl00_ctl51_g_48ce9bb1_53ac_4c68_b897_c5870f269772_ctl00_pnlPastas']//a/text()",
+    )
+
+    for name, u in zip(parliament_names, parliament_urls):
+        # skip 'Constituinte' and parliaments older than the 10th (ran from 10 March 2005 to 14 October 2009)
+        if name == "Constituinte":
+            continue
+        roman = name.split()[0]
+        if roman_to_int(roman) < 10:
+            continue
```
jbothma: I think it's fine to just say `if name == "IX Legislatura": break`. (2009 is still in scope for after office, isn't it?)

This is just to save the expense of crawling 10 parliaments we won't bother using, right? So it they rename them in a way we don't emit any/enough, data assertions save us. And if they rename it in a way that we crawl too many, we're just wasting a couple of minutes of downloading and excluding all the unnecessary data. Then we don't need a function to parse roman numerals, cool as it may be.

datasets/pt/parliament/crawler.py
```diff
@@ -0,0 +1,131 @@
...
+        doc_landing,
+        "//a[@title='Recursos' and contains(@href,'DAComposicaoOrgaos')]/@href",
+    )
+
+    ### --- iterate over parliaments ---
+    doc_parliament_list = context.fetch_html(url_parliament_list, absolute_links=True)
+    parliament_urls = h.xpath_strings(
+        doc_parliament_list,
+        "//div[@id='ctl00_ctl51_g_48ce9bb1_53ac_4c68_b897_c5870f269772_ctl00_pnlPastas']//a/@href",
+    )
```
jbothma, line 129: a nice alternative might be


```suggestion
    parliament_urls = h.xpath_strings(doc_parliament_list, "//div[@class='archive-item']//a/@href")
```

jbothma: same for names.

datasets/pt/crawler.py
```diff
@@ -0,0 +1,131 @@
...
+        name = member.pop("DepNomeCompleto")
+
+        person = context.make("Person")
+        person.id = context.make_id(name, id)
+        person.add("name", name)
+        # unique MP identifier
+        person.add("idNumber", id)
+        # shorter parliamentary name
+        person.add("alias", member.pop("DepNomeParlamentar"))
+        person.add("country", "pt")
```
jbothma: we're using citizenship for PEPs now unless we find that people don't have to be a citizen of the country to be a member of parliament. In that case, please add a comment and source link.

datasets/pt/crawler.py
```diff
@@ -0,0 +1,131 @@
...
+
+        for party_history in member.pop("DepGP") or []:
+            person.add("political", party_history.pop("gpSigla"))  # party abbreviation
```
jbothma: I almost feel like we should pick the latest here, rather than all.

datasets/pt/crawler.py
```diff
@@ -0,0 +1,131 @@
...
+        ### each member also has:
+        # gpId -- party group ID
+        # gpDtFim and gpDtInicio -- dates of party affiliation
+
+        position = h.make_position(
+            context,
+            name="Member of the Portuguese Parliament",
```
jbothma:
```suggestion
            name="Member of the Assembly of the Portuguese Republic",
            wikidataId="Q19953703",
```

datasets/pt/crawler.py
```diff
@@ -0,0 +1,131 @@
...
+            topics=["gov.national", "gov.legislative"],
+            country=["pt"],
+        )
+        position.add("subnationalArea", member.pop("DepCPDes"))  # constituency name
+        position.add("subnationalArea", member.pop("DepCPId"))  # constituency ID
```
jbothma: Remember subnationalArea is jurisdiction

```suggestion
        position.add("constituency", member.pop("DepCPDes"))  # constituency name
        position.add("constituency", member.pop("DepCPId"))  # constituency ID
```

datasets/pt/parliament/crawler.py
jbothma: so it looks like the start date for Renunciou is when they resigned, and the end date is the end of the parliamentary term.

image

So if you make a lookup from sioDes to a boolean that says whether it's an active period, you can say Efetivo is an active period, Renunciou is not an active period, and just skip over anything we haven't categorised yet or isn't an active period.

we're left with Eduardo having an occupancy from 2024-03-26 to 2024-07-31. We have no need to represent the renunciou period.

datasets/pt/crawler.py
```diff
@@ -0,0 +1,131 @@
...
+                context.emit(person)
+                context.emit(occupancy)
+                context.emit(position)
+
+        # --- fetch leadership roles in the plenary ---
+        for leadership_role in member.pop("DepCargo") or []:
+            occupancy = h.make_occupancy(
+                context,
+                person,
+                position,
```
jbothma: how about making a dedicated position for this: `f"{role['carDes'}, Assembleia da República"`. If we check all of these, do the names all look correct and appropriate?

datasets/pt/crawler.py
```diff
@@ -0,0 +1,131 @@
...
+                start_date=leadership_role.pop("carDtInicio"),
+                end_date=leadership_role.pop("carDtFim"),
+            )
+            if occupancy is not None:
+                occupancy.add("recordId", leadership_role.pop("carId"))
```
jbothma: this isn't really useful to end-users - let's drop it instead of doing something special

datasets/pt/crawler.py
```diff
@@ -0,0 +1,131 @@
...
+                occupancy.add("description", leadership_role.pop("carDes"))
```
jbothma: see comment above - dedicated position

datasets/pt/pt_parliament.yml
```diff
@@ -0,0 +1,45 @@
+title: Portugese Members of the Parliament
```
jbothma: We still prefer making the country prefix in the title `Portugal` rather than `Portuguese` for consistency.

```suggestion
title: Portugal Members of the Assembly
```

datasets/pt/crawler.py
```diff
@@ -0,0 +1,131 @@
...
+                context,
+                person,
+                position,
+                start_date=seat_occupancy.pop("sioDtInicio"),
+                end_date=seat_occupancy.pop("sioDtFim"),
+            )
+
+            # MP seat occupancy status, e.g. Effective, Retired, Suspended, Disqulified, etc:
+            seatstatus = seat_occupancy.pop("sioDes")
+            position.add("notes", seatstatus)
```
jbothma: let's put this in Occupancy:summary. Position is shared by everyone so you'd just have a jumble of values on each person. But occupancy belongs to the individual.

datasets/pt/crawler.py
herrhellana: i was thinking about how much party history we want to preserve. the political scientist in me does not want to get rid of previous affiliations, but then the question is whether `political` is the right prop to store all historic party data in... @pudo wyt?

datasets/pt/crawler.py
herrhellana: this is so much easier lol and makes a lot of sense. thanks!

datasets/pt/crawler.py
herrhellana: yes, those are all Vice-Secretário, Secretário, Vice-Presidente, Presidente, and that's it. Made a separate position for it!

datasets/pt/crawler.py
jbothma: yeah... it's tricky. The historical data is useful. I've switched this to add all affiliations with no end date.

PR #3761, opened 2026-03-09
Ireland Oireachtas

herrhellana:
Issue: https://github.com/opensanctions/crawler-planning/issues/620

### DIFF
--- a/datasets/ie/parliament/crawler.py
+++ b/datasets/ie/parliament/crawler.py
@@ -3,7 +3,16 @@ from typing import Dict, Any
 from zavod import Context, helpers as h
 from zavod.stateful.positions import categorise
 
-HOUSE_TITLES = {"seanad": "Senator", "dail": "Teachta Dála"}
+POSITIONS = {
+    "seanad": {
+        "title": "Senator of Ireland",
+        "wikidata_id": "Q18043391",
+    },
+    "dail": {
+        "title": "Teachta Dála",
+        "wikidata_id": "Q654291",
+    },
+}
 
 
 def crawl_member(context: Context, member: Dict[str, Any]) -> None:
@@ -24,10 +33,9 @@ def crawl_member(context: Context, member: Dict[str, Any]) -> None:
         for party in membership["parties"]:
             person.add("political", party["party"]["showAs"])
 
-        title = HOUSE_TITLES.get(membership["house"]["houseCode"].lower())
-        # Dáil and Seanad are modelled as distinct positions in Wikidata
-        wikidata_id = "Q654291" if title == "Teachta Dála" else "Q18043391"
-        assert title is not None
+        house_code = membership["house"]["houseCode"].lower()
+        title = POSITIONS[house_code]["title"]
+        wikidata_id = POSITIONS[house_code]["wikidata_id"]
 
         position = h.make_position(
             context,


datasets/ie/parliament/crawler.py
```diff
@@ -0,0 +1,152 @@
...
+    while skip < total:
+        batch = context.fetch_json(f"{context.data_url}?limit={limit}&skip={skip}")[
+            "results"
+        ]
+        all_members.extend(batch)
+        skip += limit
+
+    for member in all_members:
+        member = member["member"]
+        crawl_member(context, member)
```
jbothma: could we do crawl_member directly on the response data instead of building them all up? It reduces the amount of RAM needed, and means we start seeing results from a partial crawl which could make debugging easier.

datasets/ie/parliament/crawler.py
```diff
@@ -0,0 +1,152 @@
...
+
+    # the API has a server-side cap of 1k results per request
+    # we need a skip parameter to paginate records
+    meta = context.fetch_json(context.data_url)
+    total = meta["head"]["counts"]["memberCount"]
+
+    while skip < total:
+        batch = context.fetch_json(f"{context.data_url}?limit={limit}&skip={skip}")[
+            "results"
+        ]
```
jbothma: keep an eye on awkward wrapping. The fix - factoring a long value out to its own variable - often makes code more readable


```suggestion
        url = f"{context.data_url}?limit={limit}&skip={skip}"
        batch = context.fetch_json(url)["results"]
```

datasets/ie/parliament/crawler.py
```diff
@@ -0,0 +1,152 @@
...
+HOUSE_TITLES = {"seanad": "Senator", "dail": "Teachtaí Dála"}
+
+
+def crawl_member(context: Context, member: Dict[str, Any]) -> None:
+    first_name = member.pop("firstName")
+    last_name = member.pop("lastName")
+    id = member.pop("memberCode")
+
+    person = context.make("Person")
+    person.id = context.make_id(first_name, last_name, id)
```
jbothma: shit, `memberCode` is as if it was made by us! I think it's entirely safe to use just `member_code` as the input to make_id.
```json
"memberCode": "Henry-J-J-Abbott.D.1987-03-10"
```

datasets/ie/parliament/crawler.py
```diff
@@ -0,0 +1,152 @@
...
+    h.apply_name(person, first_name=first_name, last_name=last_name)
```
jbothma: when they provide a full name, it's nice to also supply that to apply_name. That helps in cases when they've contructed it differently from how we would.

datasets/ie/parliament/crawler.py
```diff
@@ -0,0 +1,152 @@
...
+    person.add("country", "ie")
```
jbothma: citizenship unless that's untrue for Ireland, in which case a comment would be great.

datasets/ie/parliament/crawler.py
```diff
@@ -0,0 +1,152 @@
...
+            topics=["gov.national", "gov.legislative"],
+            country=["ie"],
+        )
+        title = HOUSE_TITLES.get(membership["house"]["houseCode"].lower())
+        if title:
+            position.add("name", title)
+        if title == "Teachtaí Dála":
+            # some TDs can represent several constituencies per term
+            for represents in membership["represents"]:
+                position.add("subnationalArea", represents["represent"]["showAs"])
```
jbothma: constituency is occupancy:constituency - position is shared by all the MPs.

datasets/ie/parliament/crawler.py
```diff
@@ -0,0 +1,152 @@
...
+
+        position = h.make_position(
+            context,
+            name="Member of the Irish Parliament",
+            topics=["gov.national", "gov.legislative"],
+            country=["ie"],
+        )
+        title = HOUSE_TITLES.get(membership["house"]["houseCode"].lower())
+        if title:
+            position.add("name", title)
```
jbothma: let's make positions specific to the house https://www.wikidata.org/wiki/Q654291, https://www.wikidata.org/wiki/Q18043391 incl wikidata id

datasets/ie/parliament/crawler.py
```diff
@@ -0,0 +1,152 @@
...
+        if title == "Teachtaí Dála":
+            # some TDs can represent several constituencies per term
+            for represents in membership["represents"]:
+                position.add("subnationalArea", represents["represent"]["showAs"])
+        # ^ this also features things like "Administrative Panel" or "Labour Panel" for Senators
+        # not sure what to do with these, so skipping for now
```
jbothma: oh this is actually interesting! That does fit for me in represents - if I wanted to understand funny influence of a senator, this saves me time.

datasets/ie/parliament/crawler.py
```diff
@@ -0,0 +1,152 @@
...
+            position_other_name = office["office"]["officeName"]["showAs"]
+
+            if position_other_name is not None:
+                position_other = h.make_position(
+                    context,
+                    name=position_other_name,
+                    topics=["gov.national", "gov.legislative"],
+                    country=["ie"],
+                )
+                categorisation = categorise(context, position_other, is_pep=True)
```
jbothma: ministerial positions are generally executive.
Are they all executive positions? Otherwise we should drop the topics and let those be categorised in the UI.
Are they all PEP positions? Othwerwise leave is_pep=None and let them be categorised in the UI.

datasets/ie/parliament/crawler.py
```diff
@@ -0,0 +1,152 @@
...
+                    start_date=office["office"]["dateRange"]["start"],
+                    end_date=office["office"]["dateRange"]["end"],
+                )
+                if occupancy_other is not None:
+                    context.emit(position_other)
+                    context.emit(occupancy_other)
+
+        # --- some MPs have committee positions listed: ---
+        if membership.get("committees"):
+            for committee in membership["committees"]:
```
jbothma: include based on the outcome of https://opensanctions-corp.slack.com/archives/C07C5G9KZED/p1773757301017919

jbothma: Decision: skip committee membership when the house member data is this good.

datasets/ie/parliament/crawler.py
herrhellana: hm ok i rewrote the code so it writes `represents["represent"]["showAs"]` as constituency for both TDs and Senators now :-) Just gonna be vocational panels (or uni or head of gov nomination) for senators.

datasets/ie/parliament/crawler.py
herrhellana: woops yep sorry this was an overlook. this list features exec or administrative roles mostly:
- cabinet ministers
- head of gov and their deputy
- attorney general
- administrative & civil service (eg Director of Agriculture or Secretary for Trade and Commerce)

And three legislative roles:
- Ceann Comhairle (speaker of Dáil Éireann, lower house)
- Cathaoirleach and Leas-Cathaoirleach (chairman and their deputy, upper house)

herrhellana: actually we can be more granular that just labeling everything as exec (i looked more closely into our data dict); implemented a new topic add logic for these positions in the code

Issue #3603, opened 2026-02-20
Migrate eu_fsf to multi-names input


jbothma:
depends on LangText

PR #3770, opened 2026-03-11
Add support for LangText in name cleaning helpers

### LINKED ISSUE #3602: Add support for LangText
so that we can provide input with multiple strings in different languages in a single review

It's useful to see as much context as possible when reviewing name cleaning. some sources provide multiple strings of one or more names, potentially in different properties. We can now review them together.

Some sources indicate a language for a name string. We'd like to support indicating the language in the input and cleaned/reviewed output of name cleaning so that multiple accepted names can be applied with their different indicated languages, if accepted.

This can be somewhat inspired by wikidata's LangText.

jbothma:
Fixes #3602 and #3603

- Adds `zavod.extract.names.clean.LangText` and adds that as a possible value in `zavod.extract.names.clean.Names` fields.
- Also adds `zavod.extract.names.clean.Names.add(prop_name, value, lang=lang)` so that crawlers can easily build up a Names instance in a loop. 
- Demonstrates use in eu_fsf and eu_sanctions_map.
- For rollout I'd suggest making the affected crawlers raise on unaccepted reviews until they're all reviewed.
- [x] make sure we can still optimise/compare (verify performance) of the DSPy program (prompt)
- [ ] https://github.com/opensanctions/opensanctions/pull/3637/changes#diff-5e106a2073abe41db40e9810626ca5027b2445a336963e2f72b56b902d062973R78

For a future task, Names.add feels nice. Maybe if it also accepts iterables, it could filter out Nones (allowed in a list of strings for crawler convenience) so the name helper type handling could be simpler all over.

Aside: With Names.add, the Names structure is beginning to feel almost like a strongly-typed entity proxy. That feels like it could be an interesting avenue to explore.

image

image

### DIFF
--- a/zavod/zavod/extract/names/clean.py
+++ b/zavod/zavod/extract/names/clean.py
@@ -109,6 +109,11 @@ class Names(BaseModel):
         Add a value to a property. If set as a single value, the values are added to a list.
         Value is wrapped in LangText if lang is provided.
 
+        Note: Names with LangText language values and llm_cleaning=True are not supported together.
+        If names share the same language, pass ``lang`` to ``apply_reviewed_names`` /
+        ``apply_reviewed_name_string`` instead. If names have different languages, use a
+        separate call per language.
+
         Args:
             prop: The property name to add the value to.
             value: The name value to add.
--- a/zavod/zavod/helpers/names.py
+++ b/zavod/zavod/helpers/names.py
@@ -542,6 +542,15 @@ def _review_names(
     return review
 
 
+def _original_has_lang(original: Names) -> bool:
+    """Check if original names contain any LangText values with a language set."""
+    for _prop, values in original.as_langtexts():
+        for value in values:
+            if value.lang is not None:
+                return True
+    return False
+
+
 def review_names(
     context: Context,
     entity: Entity,
@@ -591,6 +600,14 @@ def review_names(
         assert suggested is None, (
             "Suggested names can't be supplied if LLM cleaning is enabled"
         )
+        if _original_has_lang(original):
+            # LLM cleaning returns plain strings, so per-value language will be dropped.
+            # Use a separate review_names, apply_reviewed_names or apply_reviewed_name_string call
+            # with the lang argument for each language instead.
+            context.log.warning(
+                "Names with LangText language values and llm_cleaning=True are not supported together.",
+                original=original,
+            )
 
     # heuristic-based review unless suggestion was supplied
     if suggested is None:
--- a/datasets/be/fod_sanctions/crawler.py
+++ b/datasets/be/fod_sanctions/crawler.py
@@ -1,5 +1,5 @@
 from csv import DictReader
-from typing import Dict, List, cast
+from typing import Dict, List
 
 from rigour.mime.types import CSV
 
@@ -51,7 +51,9 @@ def crawl_row(context: Context, entity_id: str | None, row: Dict[str, List[str]]
 
     whole_names = row.pop("Wholename")
     assert whole_names, row
-    original = h.Names(name=cast(list[str | None], whole_names))
+    original = h.Names()
+    for whole_name in whole_names:
+        original.add("name", whole_name)
     h.apply_reviewed_names(context, entity, original=original)
 
     for id_number_line in row.pop("Number"):
--- a/datasets/ru/cbr_banks/crawler.py
+++ b/datasets/ru/cbr_banks/crawler.py
@@ -96,15 +96,15 @@ def crawl_details(context: Context, bic: str, entity: Entity, short_name: str | 
         reg_date = co_data.findtext("MainDateReg")
 
         en_names = co_data.findtext("encname")
-        names = h.Names(
-            name=[
-                en_names,
-                co_data.findtext("OrgName"),
-                co_data.findtext("OrgFullName"),
-                co_data.findtext("csname"),
-            ]
-        )
+        names = h.Names()
+        names.add("name", co_data.findtext("OrgName"))
+        names.add("name", co_data.findtext("OrgFullName"))
+        names.add("name", co_data.findtext("csname"))
         h.review_names(context, entity, original=names, llm_cleaning=True)
+        # en_names should apply with lang="eng"
+        h.review_names(
+            context, entity, original=h.Names(name=en_names), llm_cleaning=True
+        )
 
         phones = co_data.findtext("phones")
         lic_withd_num = co_data.findtext("licwithdnum")


zavod/zavod/tests/helpers/names/test_derive_originals.py
```diff
@@ -152,3 +152,32 @@ def test_derive_original_values_single_original_multiple_props():
         "Brandon": "John Brandon Doe",
         "JBD": "John Brandon Doe",
     }
+
+
+def test_derive_original_values_with_langtext():
+    """LangText values should work the same as str values."""
+    # LangText in original, str in extracted
+    original = Names(name=LangText(text="John/Jon Doe", lang="eng"))
+    extracted = Names(name="John Doe", alias="Jon Doe")
```
jbothma, line 162: Do we care if e.g. `LangText("Jean Smith;John Smith", lang="fra")` and `LangText("Jon Smith;John Smith", lang="eng")` are available and we pick `"Jean Smith;John Smith"` for `John Smith`?

I can't think of a realistic example where this would happen. The fact that we're not doing fuzzy matching but only `contains exactly` I think makes this fairly reasonable.

zavod/zavod/extract/names/clean.py
```diff
@@ -14,26 +14,37 @@
...
-        return len(text) == 0
-    return False
+    def __hash__(self) -> int:
+        return hash((self.text, self.lang))
 
 
-class Names(BaseModel):
-    """Names categorised and cleaned of non-name characters."""
+NamesValue = str | LangText | None
+NamesValues = NamesValue | Sequence[NamesValue]
```
pudo: Ok, I wonder if we can bring down the permutations on this. What if we got rid of `NamesValue` and try to use this for `NamesValues`:

```
NamesValues = None | str | Sequence[str, LangText]
```

This basically leaves us with easy mode (name: null, name: Jeff) and hard mode (name: ["Jeff", {name: "Joff", lang: fra}]).

zavod/zavod/extract/names/clean.py
```diff
@@ -66,23 +77,50 @@ def is_empty(self) -> bool:
             return False
         return True
 
-    def nonempty_item_lists(self) -> Generator[Tuple[str, List[str]], None, None]:
+    def nonempty_item_lists(
+        self,
+    ) -> Generator[Tuple[str, List[str | LangText]], None, None]:
```
pudo: Should this always return List[LangText]? Basically, this function is the opposite of simplify (below) and it wants to return the most complex output form that consumer code will want to handle.

zavod/zavod/helpers/names.py
```diff
@@ -339,9 +342,10 @@ def _check_schema_name_specs(string: str, spec: CleaningSpec) -> Optional[Regula
     return None
 
 
-def check_name_regularity(entity: Entity, string: Optional[str]) -> Regularity:
+def check_name_regularity(entity: Entity, names_value: NamesValue) -> Regularity:
```
pudo: Why does the regularity checker need to know about `LangText`? To me, its whole purpose is to look at the name string. Are we ever deriving value from the `lang` metadata in the context of the regularity checker? I'm not saying it isn't theoretically conceivable that one day it could do that, but we can still cross that bridge then.

zavod/zavod/helpers/names.py
```diff
@@ -379,47 +383,54 @@ def check_names_regularity(entity: Entity, names: Names) -> Tuple[bool, Names]:
...
     """
     is_irregular = False
-    updated_suggested_data: Dict[str, List[str | None]] = defaultdict(list)
-    for key, strings in names.nonempty_item_lists():
-        for string in strings:
-            regularity = check_name_regularity(entity, string)
+    updated_suggested_data: Dict[str, List[NamesValue]] = defaultdict(list)
+    for key, names_values in names.nonempty_item_lists():
+        for name_val in names_values:
+            regularity = check_name_regularity(entity, name_val)
```
pudo: If this always returned `LangText`, this would be a great moment to hand in `val.text`.

zavod/zavod/helpers/names.py
```diff
@@ -442,29 +453,42 @@ def apply_names(
...
 
     for prop, name_values in names.nonempty_item_lists():
         for name in name_values:
+            name_lang: Optional[str]
+            if isinstance(name, LangText):
+                name_string = name.text
+                name_lang = name.lang
+            else:
+                name_string = name
+                name_lang = lang
```
pudo: Can go if `nonempty_item_lists` returns langtext

datasets/eu/fsf/crawler.py
```diff
@@ -12,4 +12,4 @@ def crawl(context: Context) -> None:
     for entry in doc_.findall(".//sanctionEntity"):
         parse_entry(context, entry)
 
-    assert_all_accepted(context, raise_on_unaccepted=False)
+    assert_all_accepted(context, raise_on_unaccepted=True)
```
pudo: This is a policy thing I would hard disagree with: we ALWAYS need to publish FSF. Always. I assume this is just for debug purposes to block publication for a little while?

zavod/zavod/helpers/names.py
jbothma: yeah that makes sense. will do

zavod/zavod/extract/names/clean.py
jbothma: This wasn't enough to make gpt-4o happy but it's a good change anyway.

> I know it's duplication but what if the training for the DSPy ran on an output model without LangText? We can still use 5.4 - so we get the gains without burning them on the complexity?

making `avod.extract.names.clean.clean_names()` return `SimpleNames extends Names` which is basically `list[str]` makes gpt4o happy though - I think that's a nice fix.

pudo: Ah, I meant this irrespective of Mr. Altman's concerns, I just feel like we've created too many modes for this data to happily exist there. Having only `None | str | Sequence[str, LangText]` basically translates into easy, medium and hard mode :-P

datasets/eu/fsf/crawler.py
jbothma: this should definitely have had a task attached to not forget to review and revert very soon. But yes this change is actually entirely compatible so new reviews can be created but not applied while we review them and apply the accepted reviews.

zavod/zavod/helpers/names.py
jbothma: nice simplification. Thanks!

zavod/zavod/extract/names/clean.py
jbothma: that's nice - I renamed it as_langtexts() and introduced LangNames returned by check_names_regularity so that callers know they're always dealing with lists of LangTexts.

zavod/zavod/extract/names/clean.py
jbothma: It feels a bit exuberant to have three Names types, but it feels like that actually works really well for three use cases:

- SimpleNames for LLM - `Sequence[str]`
- LangNames for crawlers - `Sequence[LangText]`
- Names for humans - `None | str | Sequence[str, LangText]`

pudo:
"Optimise DSPy" just means having zavod-tune re-run the prompt optimizer, right? Or do we want to do another round of DSPy work in this PR?

datasets/eu/fsf/eu_fsf.yml
```diff
@@ -72,6 +72,11 @@ data:
   url: https://webgate.ec.europa.eu/fsd/fsf/public/files/xmlFullSanctionsList_1_1/content?token=dG9rZW4tMjAxNw
   format: XML
 
+names:
+  suggest_weak_alias_person_single_token: true
+  suggest_abbreviation_uppercase_org_single_token_shorter_than: 8
+  suggest_abbreviation_non_person_single_token_shorter_than: 5
```
pudo, line 78: You freaks with your long var names :-P

zavod/zavod/helpers/names.py
```diff
@@ -442,29 +449,32 @@ def apply_names(
...
 
-    for prop, name_values in names.nonempty_item_lists():
+    for prop, name_values in names.as_langtexts():
         for name in name_values:
             entity.add(
                 prop,
-                name,
-                lang=lang,
+                name.text,
+                lang=name.lang if name.lang is not None else lang,
```
pudo: name.lang or lang?

datasets/eu/fsf/eu_fsf.yml
leonhandreke: someone called me? :P

zavod/zavod/helpers/names.py
jbothma: wait I think I can write it more complicatedly

zavod/zavod/extract/names/clean.py
```diff
@@ -165,6 +252,6 @@ def clean_names(context: Context, raw_names: SourceNames) -> Names:
         context=context,
         prompt=prompt,
         string=input_string,
-        response_type=Names,
+        response_type=SimpleNames,
         model=LLM_MODEL_VERSION,
     )
```
jbothma, line 262: by turning it all to strings, we lose language we get from the source

> I might even add a warning so it shouts at us if we try.

https://github.com/opensanctions/opensanctions/pull/3770#issuecomment-4068994639

Lemme add this warning tomorrow

```diff
diff --git a/datasets/ru/cbr_banks/crawler.py b/datasets/ru/cbr_banks/crawler.py
index 8163e953c..23922e93c 100644
--- a/datasets/ru/cbr_banks/crawler.py
+++ b/datasets/ru/cbr_banks/crawler.py
@@ -96,14 +96,11 @@ def crawl_details(context: Context, bic: str, entity: Entity, short_name: str |
         reg_date = co_data.findtext("MainDateReg")

         en_names = co_data.findtext("encname")
-        names = h.Names(
-            name=[
-                en_names,
-                co_data.findtext("OrgName"),
-                co_data.findtext("OrgFullName"),
-                co_data.findtext("csname"),
-            ]
-        )
+        names = h.Names()
+        names.add("name", en_names, lang="eng")
+        names.add("name", co_data.findtext("OrgName"))
+        names.add("name", co_data.findtext("OrgFullName"))
+        names.add("name", co_data.findtext("csname"))
         h.review_names(context, entity, original=names, llm_cleaning=True)
```

Image

PR #3772, opened 2026-03-11
[us_ga_med_exclusions] clean up AKAs in middle_name

nvmbrasserie:


datasets/us/ga/med_exclusions/crawler.py
```diff
@@ -32,7 +53,11 @@ def crawl_item(row: Dict[str, str | None], context: Context):
         )
         if business_name:
             entity.add("alias", business_name)
-
+        if aliases:
+            for alias in aliases:
+                # One word after AKA means it's a different spelling of the firstName
+                prop = "firstName" if " " not in alias else "alias"
```
leonhandreke: Now we have this business logic scattered in two different places. Maybe it would be better to let `split_middle_name` return a `namedtuple` of `name, alias, firstName` or something like that?

PR #3794, opened 2026-03-13
Sort splitters by length

nvmbrasserie:


leonhandreke:
Closing this in favor of https://github.com/opensanctions/opensanctions/issues/3810. Basically, we're only going to manually sort them if the results are different, to keep the issues board clean but the number of changes low. At some point, we just enable `sorted` in `multi_split`

PR #3799, opened 2026-03-14
Set origin on more non-source data prop values

pudo:
When filling out an RfP the other day I realized we could make some really great and easy gains on adding `origin` to a few more categories of property values. Are there other instances of this that I missed here?

### DIFF
--- a/zavod/zavod/helpers/names.py
+++ b/zavod/zavod/helpers/names.py
@@ -196,10 +196,10 @@ def apply_name(
     if is_weak:
         name_prop = "weakAlias"
 
-    # Flag up full names created by
+    # Provenance for full names created from parts
+    full_origin = origin
     if full is None or len(full) == 0:
-        if origin is None:
-            origin = ORIGIN_INFERRED
+        full_origin = ORIGIN_INFERRED
     full = make_name(
         full=full,
         name1=name1,
@@ -219,7 +219,7 @@ def apply_name(
         suffix=suffix,
     )
     if full is not None and len(full):
-        entity.add(name_prop, full, quiet=quiet, lang=lang, origin=origin)
+        entity.add(name_prop, full, quiet=quiet, lang=lang, origin=full_origin)
 
 
 def split_comma_names(context: Context, text: str) -> List[str]:


zavod/zavod/helpers/addresses.py
```diff
@@ -233,10 +239,26 @@ def copy_address(entity: Entity, address: Optional[Entity]) -> None:
...
-                entity.add("country", country)
+                for stmt in address.get_statements("country"):
+                    entity.add(
+                        "country",
+                        stmt.value,
+                        lang=stmt.lang,
+                        origin=stmt.origin,
+                        original_value=stmt.original_value,
+                        cleaned=True,
+                    )
```
jbothma: It took me a moment to figure out that we're accessing statements here to access origin and original_value. It might be worth a comment just to some someone the journey

zavod/zavod/runtime/cleaning.py
```diff
@@ -80,6 +82,8 @@ def value_clean(
 
     for prop_, item in prop_lookup(entity, prop, value):
         clean: Optional[str] = item
+        if origin is None and item != value:
+            origin = ORIGIN_LOOKUP
```
jbothma, line 86: love this

zavod/zavod/helpers/addresses.py
pudo: I was almost thinking about a utility method like `entity.adopt_stmt`. How does that feel?

zavod/zavod/helpers/names.py
```diff
@@ -194,6 +195,11 @@ def apply_name(
         name_prop = "alias"
     if is_weak:
         name_prop = "weakAlias"
+
+    # Flag up full names created by
+    if full is None or len(full) == 0:
+        if origin is None:
```
jbothma: do we really only want ORIGIN_INFERRED if origin is None? That's different from the address behaviour which always uses ORIGIN_INFERRED if full was inferred.

zavod/zavod/helpers/addresses.py
jbothma: that's very neat

zavod/zavod/helpers/names.py
pudo: Oh that's fair let's make it coherent. Can you make the patch?

pudo:
One thing I'm wondering: if something gets sent to the review system currently and it comes back unmodified, do we still set the origin on it?

jbothma:
> One thing I'm wondering: if something gets sent to the review system currently and it comes back unmodified, do we still set the origin on it?

with the

```python
if not review.accepted:
    return

entity.add("name", item.name, origin=review.origin)
```

[pattern](https://zavod.opensanctions.org/data_reviews/), origin would only be that passed to review_extraction if it was accepted.

`apply_reviewed_names` applies the review's origin (`analyst`) if irregular and accepted

https://github.com/opensanctions/opensanctions/blob/pudo/origin-fills/zavod/zavod/helpers/names.py#L652

PR #3807, opened 2026-03-16
AE DFSA debarred list small fixes

herrhellana:
issue: https://github.com/opensanctions/crawler-planning/issues/336
related to older PR: https://github.com/opensanctions/opensanctions/pull/3703

### DIFF
--- a/datasets/ae/dfsa_debarred/crawler.py
+++ b/datasets/ae/dfsa_debarred/crawler.py
@@ -1,4 +1,4 @@
-from lxml.etree import _Element, ParserError
+from lxml.etree import _Element
 from lxml.html import document_fromstring
 
 from zavod import Context, helpers as h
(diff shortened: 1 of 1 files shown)


datasets/ae/dfsa_debarred/crawler.py
```diff
@@ -30,39 +31,37 @@ def crawl_person(context: Context, row: _Element) -> None:
         date_end = date_end.strip(")")
         date_end = date_end.replace("expired ", "")
         h.apply_date(sanction, "endDate", date_end)
-    else:
+    if "ongoing" in status.lower():
         person.add("topics", "debarment")
+    else:
+        context.log.warning(f"Unexpected case status: {status}")
```
jbothma:
```suggestion
        context.log.warning(f"Unexpected case status", status=status)
```

datasets/ae/dfsa_debarred/crawler.py
```diff
@@ -30,39 +31,37 @@ def crawl_person(context: Context, row: _Element) -> None:
...
-            # if UnblockFailedException past the fist page, stop paginating
-            # (we've likely hit an empty page)
+            doc = document_fromstring(html_text[3])
+
+            for row in h.xpath_elements(doc, ".//a[@class='table-row']"):
+                crawl_person(context, row)
+        except ParserError:
+            # break on empty page if past page 1
             if page_num > 1:
                 break
```
jbothma: now that it's just text in a variable, we can just 
```
# Pagination gives empty response after last page
if html_text == "":
    break 
```
instead of a try...catch.

One of the downsides of exceptions is that they're easily more broad than we intend. So it's nice to think about the condition we really want to implement and see if there's a nice normal way to implement that without resorting to exceptions.

Issue #3768, opened 2026-03-10
[ar_parliament] Switch to new JSON source


leonhandreke:
Something happened last December, switching the crawler over to a new JSON source: https://github.com/opensanctions/opensanctions/pull/3240. But only half of it happened, the statements are still emitted as `external`. We still get our data from the website. Or do we, are `external` that are deduped pulled in? I don't really know for sure.

The data from the HTML also contains DoB (important), emails and URLs (not important). Here is what should happen:

- Decide if we want to keep the HTML parsing around.
- Switch the crawler over to the new-new JSON source (https://datos.hcdn.gob.ar/dataset/legisladores). Super nice API!
- Clean the whole thing up, right now we have two positions emitted... what happening, help!

Adding to triage for decisions for what to do here. Feel free to reach out to me with questions, I just dug into this.

PR #3813, opened 2026-03-16
[ar_parliament] person.id rekey due to change in the source data

herrhellana:
source metadata change -> rename keys + make_id change (no deputy_id in source anymore)

https://github.com/opensanctions/opensanctions/issues/3768

### DIFF
--- a/datasets/ar/parliament/crawler.py
+++ b/datasets/ar/parliament/crawler.py
@@ -85,7 +85,7 @@ def crawl_personal_page(context: Context, url):
 
 def crawl(context: Context):
     # TODO: lower cache after dedupe
-    # crawl_csv(context)
+    crawl_csv(context)
     table_xpath = ".//table[@id='tablaDiputados']"
     doc = zyte_api.fetch_html(
         context,


datasets/ar/parliament/crawler.py
```diff
@@ -29,16 +29,21 @@ def crawl_json(context: Context) -> None:
...
+        first_name = entry.pop("NOMBRE")
+        party = entry.pop("BLOQUE")
+
+        mandate_dates = entry.pop("MANDATO")
+        mandate_dates_end = mandate_dates.split("-")[1]
 
         person = context.make("Person")
-        person.id = context.make_id(first_name, last_name, deputy_id)
+        # creating id on party instead of deputy_id bc source data does not have the id field anymore
+        person.id = context.make_id(first_name, last_name, party)
```
jbothma: shall we add distrito too? Just gives us a little bit more uniqueness

jbothma: let's remove party from id actually - it looks like they really mean block or coalition, and not party. And I think that's more variable than mere party which makes the IDs less stable. I think district is more stable and unique so it's a good trade.

let's rename party to coalition but I think it's still cool to put in the political prop.

datasets/ar/parliament/crawler.py
```diff
@@ -48,25 +53,24 @@ def crawl_json(context: Context) -> None:
...
 
         occupancy = h.make_occupancy(
             context,
             person,
             position,
             False,  # every tenure should have an end date (even if it is in the future)
-            start_date=entry.pop("inicio"),
-            end_date=entry.pop("fin"),
+            start_date=entry.pop("FECHA_DE_INICIO"),
+            end_date=mandate_dates_end,
```
jbothma: this will go nicely in the upcoming (parliamentary) termEnd prop but let's not use it as a date for the individual's term.

datasets/ar/parliament/crawler.py
```diff
@@ -87,7 +85,7 @@ def crawl_personal_page(context: Context, url):
 
 def crawl(context: Context):
     # TODO: lower cache after dedupe
-    crawl_json(context)
+    # crawl_csv(context)
```
nvmbrasserie:
```suggestion
    crawl_csv(context)
```

PR #3824, opened 2026-03-17
issues 03-17-2026

herrhellana:


datasets/de/abgeordnetenwatch/de_abgeordnetenwatch.yml
```diff
@@ -37,8 +37,8 @@ ci_test: false
 assertions:
   min:
     schema_entities:
-      Person: 2500
-      Position: 15
+      Person: 2200
```
leonhandreke, line 40: Do we have an idea what happened here? The commit message doesn't contain any information.

herrhellana: API has 2515 mandates, we're excluding MEPs (96 in Germany) and emitting 2418 persons. so looks good

leonhandreke:
Added some comments and pushed a commit. All the rest, LGTM, let's go!

PR #3826, opened 2026-03-17
cleaning: implement `silence_warnings: [address-too-short]` syntax

leonhandreke:
- **cleaning: implement `silence_warnings: [address-too-short]` syntax**
- **cleaning: also let xss-html-smell use `get_type_lookup_silence_warnings`**

### DIFF
--- a/contrib/issues_agent/prompt.md
+++ b/contrib/issues_agent/prompt.md
@@ -62,14 +62,23 @@ For merged countries `France / Syria`, make multiple `values` (`France`, `Syria`
 
 Try to remove or substitute the HTML content, leaving any text content in place: `<p>Hello</p>` -> `Hello`.
 
+If the content was erroneously marked as HTML, silence the warning:
+
+```
+  type.text:
+    options:
+      - match: "Also known under the pseudonym <tramp>"
+        silence_warnings: [xss-html-smell]
+```
+
 ### `Property for address looks too short for an address: Zug`
 
-If the string looks like a valid place name or address, introduce a lookup with the same return value:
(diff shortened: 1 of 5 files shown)


jbothma:
```
lookups:
  options:
    - match: abc
      ignore_reason: xss
```

jbothma:
We left this at requiring a bit more time to think about this. Just noting that down.

The following might be the validation/cleaning rewrite we said we don't want now. I'm just throwing this here because today we said "what if we had a better data linter" and it's fun to see what happens when we put some of these checks side by side and whether it starts looking like something we want. 

It'd be cool to have linting rules like

- for all datasets
  - xss check
  - field max size check
  - small addresses check
  - ...
- if the dataset has the tag `list.pep`
  - every Person should have either `role.pep` or `role.rca` 
  - every `role.pep` should be connected to an Occupancy 
  - every `role.pep` should have a `citizenship` value (except exceptions?)
- if the dataset has the tag `list.sanction`
  - ... `program` ...
  - if `endDate` is < crawl time it should not have the `sanction` topic
    - this doesn't work for a dataset with more than one sanction for some person

maybe this isn't leading down a nice path because some of these make sense to check before emit, and some after crawl, before publish. There are also important things it's not catching, like crawlers missing a second tab of data in a spreadsheet, or potentially paginating infinitely. It might help with our standardisation projects, but custom code for them doesn't feel like it hurts that much.

leonhandreke:
Nothing has moved here for a while and while I still think this change is a net positive, even without a rewrite of our field cleaning logic, it's not very important to me. Closing this for the moment until the need to revisit this arises again.

PR #3599, opened 2026-02-20
Add statements diff tool

jbothma:
Fixes https://github.com/opensanctions/operations/issues/879

Much vibes. Somewhat checked output and diff logic tests but not my usual vigour.

Supports downloading prod statements to a local copy for quicker repeat diffs when tweaking a crawler.

Supports copying local statements out of data dir to compare iterations on a crawler.

Initially this is doing everything in memory and implements the pager using Textual. That's nice to enable wrapping/truncation toggling similar to csvlens, but it makes it slow or useless for really big datasets.

Let's try it on smaller datasets and see if we like the idea and what we need, and figure out scaling after that.

```
$ zavod-stmt --help
Usage: zavod-stmt [OPTIONS] COMMAND [ARGS]...

  Utilities for working with statements.pack files.

  For example:

  - Compare a local statements.pack against the latest production version

  - Fetch a production statements.pack for quick local comparison

  - Copy your last local run's .pack to compare with subsequent runs

  - Compare two local statements.pack files against each other

Options:
  --help  Show this message and exit.

Commands:
  cp     Copy statements.pack for a dataset to...
  diff   Diff two statement sets.
  fetch  Download the latest production statements pack for a dataset.
```

```
$ zavod-stmt diff --help
Usage: zavod-stmt diff [OPTIONS] LEFT_PATH [RIGHT_PATH]

  Diff two statement sets. Arguments can be .yml dataset paths or .pack files.

  With one argument (must be a .yml), diffs local statements against the
  latest production version, treating production statements as left side and
  local statements as right side.

  With two arguments (each can be a dataset .yml or a statements.pack file),
  diffs the left statement set against the right.

  Examples:
      zavod-stmt diff datasets/tw/shtc/tw_shtc.yml
      zavod-stmt diff ../data/tw_shtc-20240101.pack datasets/tw/shtc/tw_shtc.yml
      zavod-stmt diff ../data/tw_shtc-20231201-archive.pack ../data/tw_shtc-20240101.pack
```

```
$ zavod-stmt fetch --help
Usage: zavod-stmt fetch [OPTIONS] DATASET_PATH DEST_DIR

  Download the latest production statements pack for a dataset.

  Saves to <dest_dir>/<dataset_name>-<version_id>-archive.pack.

  Example:
      zavod-stmt fetch datasets/tw/shtc/tw_shtc.yml ../data
```

```
$ zavod-stmt cp --help
Usage: zavod-stmt cp [OPTIONS] DATASET_PATH DEST_DIR

[...]

PR #3847, opened 2026-03-19
cli: add `archive url` command to resolve dataset resource URLs

leonhandreke:
So, for context: there's been ongoing discussion (#3599, https://github.com/opensanctions/operations/issues/879) about a statement diff tool. The design question was whether to build URL construction into the diff tool itself, or to let zavod own that knowledge — since zavod already knows the archive structure, the `ARCHIVE_SITE`, the `ARTIFACTS` path layout, and the version history.

This PR takes the latter approach: zavod can now resolve the public URL for a dataset resource, and any external tool (diff, fetch, whatever) can just take a URL. For now, `wget` does the job fine.

- Adds `zavod archive url <resource_type> <dataset.yml> --latest`
- Resolves the actual latest version from the dataset's `versions.json` (via `get_latest()`) rather than hardcoding `/latest/`
- Currently supports `statements` (`statements.pack`); easy to extend

```
$ zavod archive url statements --latest datasets/us/ofac/us_ofac_sdn.yml
https://data.opensanctions.org/artifacts/us_ofac_sdn/20250319120000/statements.pack
```

A `fetch` subcommand would be a natural next step, but keeping it out for now — the URL is enough to unblock the diff workflow.


zavod/zavod/cli/archive.py
```diff
@@ -0,0 +1,44 @@
...
+    is_flag=True,
+    default=False,
+    help="Resolve the latest version from versions.json",
+)
+def url(resource_type: str, dataset_path: Path, latest: bool = False) -> None:
+    dataset = load_dataset(dataset_path)
+    filename = RESOURCE_TYPES[resource_type]
+
+    if not latest:
+        raise click.ClickException("At least one version specifier must be given")
```
jbothma: this is kinda weird. How about just 


```suggestion
    if not latest:
        # No support for finding other versions yet
        raise click.ClickException("--latest is required.")
```

contrib/dedupe/pep_by_name.py
```diff
@@ -10,7 +10,7 @@
 from zavod.meta import Dataset
 from zavod.integration import get_resolver
 from zavod.store import get_view, clear_store
-from zavod.cli import _load_datasets
+from zavod.cli import load_datasets
```
jbothma: I wonder if it's maybe better to leave it as _load_datasets so that it's a bit clearer that these guys are violating the interface visibility and here be dragons.Maybe the bug is that they're using it, rather than that it's marked private. I see the underlying load_dataset sets some context thing. At least maybe deciding that this can be public can be kept out of scope of this change.

jbothma (commented):
It doesn't seem to work at the moment:

```
zavod archive url statements --latest datasets/us/finra_actions/us_finra_actions.yml
https://data.opensanctions.org/artifacts/us_finra_actions/20260330123934-iyl/statements.pack

wget https://data.opensanctions.org/artifacts/us_finra_actions/20260330123934-iyl/statements.pack
--2026-03-30 13:45:40--  https://data.opensanctions.org/artifacts/us_finra_actions/20260330123934-iyl/statements.pack
Resolving data.opensanctions.org (data.opensanctions.org)... 143.244.56.50
Connecting to data.opensanctions.org (data.opensanctions.org)|143.244.56.50|:443... connected.
HTTP request sent, awaiting response... 404 Not Found
2026-03-30 13:45:40 ERROR 404: Not Found.
```

same for us_sam.

contrib/dedupe/pep_by_name.py
leonhandreke: Agreed.

zavod/zavod/cli/archive.py
```diff
@@ -3,15 +3,24 @@
...
 
 from zavod import settings
-from zavod.archive import ARTIFACTS, STATEMENTS_FILE
+from zavod.archive import ARTIFACTS, STATEMENTS_FILE, iter_dataset_versions
 from zavod.cli import cli, DatasetInPath, _load_dataset
-from zavod.runtime.versions import get_latest
+from zavod.meta.dataset import Dataset
 
 RESOURCE_TYPES = {
     "statements": STATEMENTS_FILE,
```
jbothma: maybe this could just be `statements.pack` and not `statements: statements.pack` similarly to how dataset metadata exports refers to exports by file name, but I don't feel strongly about it. It's easy to change if we want it later.

jbothma (approved):

zavod/zavod/cli/archive.py
leonhandreke: Great idea.

PR #3850, opened 2026-03-19
[us_bis_export] lookup and split on ' and '

nvmbrasserie:


datasets/us/bis_export/crawler.py
```diff
@@ -47,7 +47,9 @@ def crawl_entity(
     entity.id = context.make_id(primary_name, case_id_string)
     entity.add("name", primary_name)
     for alias in aliases:
-        entity.add("alias", alias)
+        # Folowing tw_shtc logic, we move very short aliases (e.g., acronyms) to weakAlias
+        prop = "weakAlias" if len(alias) < 8 else "alias"
+        entity.add(prop, alias)
```
nvmbrasserie: is it a safe assumption to make? Mostly acronyms here (8 in total), but also a few very short names (e.g. `Shahi, Jim Chu, Ma Li`)

Issue #3805, opened 2026-03-16
PEP alignment plan


pudo:
We need to go virtually all PEP crawlers and look for the following: 

* Applying the new `earliest_term_start` helper
* Opportunities to fill `constituency` and `politicalGroup` on `Occupancy`
* Opportunities to fill `periodStart`, `periodEnd`, `electionDate` on `Position`
* Opportunities to pull in a narrative biography,  which is increasingly useful for PEP matching with LLMs. Next FtM release will introduce `Person:biography`. 
* Crawlers which don't yet set the citizenship prop and rely on `h.make_occupancy` to auto-fill `country` (will be a warning via #3860).
* Removing any instances where `h.make_occupancy` takes `birth_date` or `death_date` rather than checking the `person` object (will be a warning via #3860).

Because we need to avoid big splashy deltas, I suggest we do this one PR per crawler, no exceptions.

pudo:
Next FtM release will introduce `Person:biography`.

herrhellana:
do we also want to remove party names from "political" prop of Person if we're moving it to Occupancy:politicalGroup?

jbothma:
Sometimes we get a list of party affiliations with associated date. Do we want to add all party names to `Person:political` or just the latest? In the portuguese parliament I suggested only latest, but let's decide before we roll out the behaviour further.

> do we also want to remove party names from "political" prop of Person if we're moving it to Occupancy:politicalGroup?

just noting here, as discussed elsewhere, politicalGroup would be more like a caucus or other affiliation in a parliament, different from the party which remains on the person.

herrhellana:
re: a list of party affiliations, @pudo agreed with just adding the latest party when we discussed the portuguese crawler, but, to keep the logic consistent, if we scrape bios for PEPs, why not to save the entire party history, too? we could also think if we want to keep the current party in one prop, and the list of historic ones in another prop, new or even appended to bio?

jbothma:
>  but, to keep the logic consistent, if we scrape bios for PEPs, why not to save the entire party history, too? we could also think if we want to keep the current party in one prop, and the list of historic ones in another prop, new or even appended to bio?

Previously you said for political science, party history is interesting - do I remember correctly?

For pol sci questions, I imagine you need the time dimension too - just having a bag of parties this person has belonged to at some point isn't particularly useful, is it?

[...]

Issue #3321, opened 2026-01-08
Positions/occupancy helper/interface simplications


pudo:
Picking up from https://github.com/opensanctions/opensanctions/issues/2969#issuecomment-3718425091 - I want to look at the positions API, some comments:

### Big picture

* Would there ever be value to a function like `h.finalize_pep(context, person, position, occupancy)` where we can do logic stuff before emitting? 

### Small changes

* Definitely: remove `birthDate` and `deathDate` from `h.make_occupancy` - those must be read from the person, otherwise it's confusing.
  * Same in `occupancy_status` - read from entity. 
* Prefer to remove `propagate_country`, too. We want to move PEP country tags to `citizenship` anyway to align with the Wikidata enricher, and it just feels like we're being clever in a complex way. 
* Does `h.make_occupancy(current_time=)` ever actually need to be overridden? It's not used in the codebase as of today. 
* Should `country` be forced as a single string for `h.make_position`? I feel like un-countried or multi-countried positions are both ... bugs? 
* What is the use case for passing `PositionCategorisation` into `occupancy_status` when the position also goes in, and carries the `topics` prop?

pudo:
This also seems satisfied by #4445, #4446, #4447.

PR #3860, opened 2026-03-20
Warn for deprecated make_occupancy args

pudo:
Refs #3805 and #3321


pudo:
The period stuff definitely needs a better think than what I've just given it, but I wanted to sketch out the thing for hand-over. Would you have a look, @jbothma ?

jbothma (approved):
I think this is good to go - I fixed my comment on the three categories which should make it actually make sense, I hope.

zavod/zavod/stateful/positions.py
```diff
@@ -199,24 +223,35 @@ def occupancy_status(
...
+    # is still in office. An MP could leave a term early
+    if period_end is not None:
+        if period_end < current_iso:  # period_end is in the past
+            if period_end < h.backdate(current_time, after_office):
+                # period_end is beyond after-office threshold
                 return None
+            else:
+                return OccupancyStatus.ENDED
+        else:  # period_end is in the future
+            return OccupancyStatus.UNKNOWN
```
pudo: Should this reflect `no_end_implies_current`?

pudo: We can also just not return here quite yet.

zavod/zavod/tests/stateful/test_positions.py
```diff
@@ -36,14 +56,13 @@ def status(
...
-            datetime(2021, 1, 1),
-            start,
-            end,
-            birth,
-            death,
+            person=person,
+            position=pos,
+            occupancy=make_occ(start=start, end=end, period_end=period_end),
+            no_end_implies_current=implies,
+            current_time=datetime(2021, 1, 1),
```
pudo, line 63: I would really like to get rid of the `current_time` argument to `make_occupancy`. How much work would it be to (ask Claude to) make the tests generate start and end date dynamically instead of overriding runtime?

pudo (commented):
I think we should avoid being implausibly uptight about marking CURRENT - my sense right now is that we miss a lot of potential PEPs because we're aggressively conservative with this. If we have no end date but a period end etc. etc. - I think this generally does that?

zavod/zavod/tests/stateful/test_positions.py
jbothma: I started removing it but surprising tests broke so I decided to do that in a separate branch.

Is calculating test dates dynamically better than mocking settings using monkeypatch?

```
  def test_foo(monkeypatch):
      fake_time = datetime(2025, 1, 1, tzinfo=timezone.utc)
      monkeypatch.setattr(settings, "RUN_TIME", fake_time)
      monkeypatch.setattr(settings, "RUN_TIME_ISO", fake_time.isoformat(...))
      monkeypatch.setattr(settings, "RUN_DATE", fake_time.date().isoformat())
```

pudo: Oh no I'm here for a good monkeypatch in the tests! :)

PR #3889, opened 2026-03-26
[us_finra_actions] split concatenated names

nvmbrasserie:


datasets/us/finra_actions/crawler.py
```diff
@@ -13,22 +13,29 @@
...
+        # text content. This avoids relying on positional span[2] and also prevents
+        # the outer wrapper span from concatenating all descendant text into a single
+        # corrupted string.
+        dirty_name = h.xpath_string(
+            div_row, "./span[normalize-space(text())]/text()"
+        ).strip()
+        if not dirty_name:
+            context.log.warning("No name span found, page structure may have changed")
+            continue
+        name = re.sub(r",1$", "", dirty_name)
```
leonhandreke: I know this was there before, but let's make this a datapatch while we're at it?

datasets/us/finra_actions/crawler.py
```diff
@@ -38,6 +45,17 @@ def crawl_item(context: Context, row: Dict[str, _Element]) -> None:
     for name in names:
         entity = context.make("LegalEntity")
         entity.id = context.make_slug(name)
+
+        original = h.Names(name=name)
```
leonhandreke: I'm not super deep into the whole name cleaning, and this whole things looks sane enough. But I'm wondering whether in the philosophy of the name cleaning framework, we should instead be passing in the unsplit originals, and let the thing figure out the rest, or pass in the split names as a `suggested`? @jbothma will have better opinions.

datasets/us/finra_actions/crawler.py
```diff
@@ -38,6 +45,17 @@ def crawl_item(context: Context, row: Dict[str, _Element]) -> None:
...
+        is_irregular, suggested = h.check_names_regularity(entity, original)
```
leonhandreke: Let's maybe add a comment here mentioning these concatenated names and that/why the cleaning framework will catch them. It's not obvious from reading the code.

datasets/us/finra_actions/crawler.py
nvmbrasserie: It's just one name, and the code becomes much chunkier when we try to catch it. I think it's unnecessary there and it's cleaner to just use a regex in that case

leonhandreke: Can't we just patch it in `type.name` without catching it at all in the code?

datasets/us/finra_actions/crawler.py
nvmbrasserie: concatenated names were originating from here: `for dirty_name_el in name_els.findall(".//span"):`

name cleaning is tackling cases like: 
```

- Justly Markets LLC f/k/a DBOT ATS LLC
- Osaic Wealth Inc. formerly known as Royal Alliance Associates Inc.

```

jbothma: ohhh we were capturing this outer span (and maybe also the inner spans?!

Image

So we would have had an entity for `kellogg-partners-institutional-services-llc-mark-anthony-henriquez`, `kellogg-partners-institutional-services-llc` and `mark-anthony-henriquez`. Now we're deleting `kellogg-partners-institutional-services-llc-mark-anthony-henriquez` and we're left with the correct IDs.

```
grep DEL data/datasets/us_finra_actions/entities.delta.json | jq .entity.id|cut -c11-|grep -|wc -l
     558
```

these plus their sanctions adds up to roughly the number of deletions we see. No additions expected.

datasets/us/finra_actions/crawler.py
jbothma: on the one hand I want to say the name cleaning framework shouldn't be used at all because the original string here contains multiple entities, and yeah, now we have this question of what should be passed in `original` because that ends up in `original_value` in statements data.

On the other hand I can't come up with examples where h.split_comma_names hasn't split the entities, so we're just left with aliases and previousNames per entity, which the names framework does nicely.

I think the options we have are:

1. use the names framework as implemented here - I think providing the original original value would make the reviews confusing
2. add some more checks like contains_split_phrase to h.split_comma_names before returning so that its lookup group can be used to fix the 284 "irregular" names
3. add a custom lookup for this crawler

I'm leaning to (1) with llm cleaning enabled. BTW the llm cleaning is better than a couple of weeks ago now we're using gpt-5.4. I don't think it's a disaster that the original_value will not be the full original string. It will still be a decent deterministic version of the original.

jbothma: leaning to (1) because it introduces the least amount of new logic, and there's a migration path from here if we discover cases where entities aren't split, or when we introduce a better entity splitter.

PR #3902, opened 2026-03-30
[ir_uani_business_registry] fix crawler

brrttwrks:
Fix crawler to ensure comprehensive pagination and consistent record acquisition across crawls. Sourced from issue https://github.com/opensanctions/crawler-planning/issues/647

### DIFF
--- a/datasets/ir/uani_business_registry/crawler.py
+++ b/datasets/ir/uani_business_registry/crawler.py
@@ -34,29 +34,27 @@ def parse_facts_list(container: HtmlElement) -> Dict[str, List[HtmlElement]]:
     return data
 
 
-def is_500_page(doc: etree._Element) -> bool:
-    return (
-        "The website encountered an unexpected error. Try again later."
-        in doc.text_content()
-    )
-
-
-def crawl_subpage(context: Context, url: str, entity: Entity, entity_id: str | None):
+def crawl_subpage(context: Context, url: str, entity: Entity, entity_id: str):
     context.log.info(f"Starting to crawl company page: {url}")
-    validator_xpath = (
-        './/div[@class="c-full-node__info"] | '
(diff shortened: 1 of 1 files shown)


datasets/ir/uani_business_registry/crawler.py
```diff
@@ -131,12 +130,27 @@ def crawl_subpage(context: Context, url: str, entity: Entity, entity_id: str):
...
 def crawl(context: Context):
-    pages_processed = 0
+    end_page = get_end_page(context)
 
-    while True:
+    # paginate
+    for page_num in range(0, end_page + 1):
         # Construct the URL for the current page
-        url = f"https://www.unitedagainstnucleariran.com/iran-business-registry?page={pages_processed}"
+        url = f"https://www.unitedagainstnucleariran.com/iran-business-registry?page={page_num}"
```
leonhandreke: (nit) we can also construct this from `context.data_url` maybe :)

jbothma:
I'm not sure I understand what the cause was and that we're fixing it

- we used to stop paginating over index pages if we couldn't find the table
  - we now require the table to be present and paginate using pagination controls
  - I'll remove `if end_page is None:` because if they add data during the crawl, we're more likely to capture everything available
  - I take it this is what we believe was the cause? It's the most likely given the next item
- we skip detail pages with an error message. However I [can't find instances of this in the logs](https://console.cloud.google.com/logs/query;query=%0ASEARCH%2528%22Broken%20link%20detected%22%2529;cursorTimestamp=2026-03-28T21:07:03.708823106Z;duration=P30D?project=opensanctions-ops).
  - I'll remove the skip and just assert that the error message isn't in the content.
- This PR removes caching from detail pages, but I don't see how caching on detail pages would be causing this, and it can help bypass intermittent issues when the site is re-crawled. So I think it's worth keeping it.

PR #3925, opened 2026-04-02
Fixes 02 04

nvmbrasserie:
in `[jp_shugiin]` the structure of the source changed and we can no longer generate the same ids. we'll get the rekey on roughy 450 PEPs

### DIFF
--- a/datasets/ca/sema_sanctions_news/crawler.py
+++ b/datasets/ca/sema_sanctions_news/crawler.py
@@ -138,6 +138,8 @@ def crawl_vessel(context: Context, row: Dict[str, _Element]) -> None:
     vessel.add("type", str_row.pop("ship_type"))
     h.apply_date(vessel, "buildDate", str_row.pop("ship_build_date"))
     vessel.add("topics", "sanction")
+    # The program applies to all vessels. Since there is no country column in the
+    # vessel table, we cannot map from the original value.
     sanction = h.make_sanction(
         context,
         vessel,


datasets/ca/sema_sanctions_news/crawler.py
```diff
@@ -7,6 +7,8 @@
 from zavod import Context, helpers as h
 from zavod.stateful.review import assert_all_accepted
 
+PROGRAM_KEY = "CA-SEMA"
```
leonhandreke, line 10: We'r enot using this anywhere here?

leonhandreke: Sorry, I was being dumb

leonhandreke, line 10: Could we mention why in one sanction we're doing the lookup while in the other we're using the const?

datasets/jp/shugiin/crawler.py
```diff
@@ -64,7 +38,9 @@ def crawl(context: Context) -> None:
 
     doc = context.fetch_html(context.data_url)
     urls = [context.data_url]
-    nav_links = doc.findall('.//div[@id="LnaviArea"]//table//a')
+    nav_links = h.xpath_elements(
+        doc, './/div[@id="LnaviArea"]//table//a', expect_exactly=17
```
jbothma, line 42: this is based on the surnames of the members? So it might well break upon an election or or resignation. I wonder if expect_exactly is more strict than we need to be, and data assertions in the metadata are sufficient.

Are you tightening up because of a bug you found? If so, a comment for this unusual strictness could be useful.

datasets/us/ne/med_exclusions/crawler.py
```diff
@@ -4,7 +4,11 @@
 from zavod import Context, helpers as h
 from zavod.extract.zyte_api import fetch_resource
 
-PAGE_SETTINGS = {"join_y_tolerance": 2}
+PAGE_SETTINGS = {
+    "join_y_tolerance": 2,
+    "vertical_strategy": "explicit",
+    "explicit_vertical_lines": [37, 85, 170, 290, 325, 374, 531, 602, 655, 720],
+}
```
jbothma, line 11: at first I was worried this would be brittle but not brittle enough, as in changing column widths might result in bad data in future.

But I think this could work ok because we're fairly likely to get key errors if the headings have moved enough. Since that's what we're relying on for safety, a comment to that effect could be nice for the next person.

PR #3929, opened 2026-04-02
China Shenzhen Stock Exchange Trading Restriction

herrhellana:
issue: https://github.com/opensanctions/crawler-planning/issues/547


datasets/cn/szse_trading/cn_szse_trading.yml
```diff
@@ -0,0 +1,47 @@
+title: China Shenzhen Stock Exchange Trading Restrictions
```
jbothma, line 1: Missing `tags` field. Other enforcement/regulatory action datasets use tags like `list.enforcement`, `sector.financial`, and an issuer tag. For this dataset, something like:

```yaml
tags:
  - list.enforcement
  - sector.financial
```

See `us_cftc_enforcement_actions.yml` or `us_fed_enforcements.yml` for examples.

datasets/cn/szse_trading/crawler.py
```diff
@@ -0,0 +1,74 @@
...
+        entity.add("name", name)
+        # snippet as context; redacted IDs in body aren't useful as identifiers
+        entity.add("notes", doc_content)
+        entity.add("country", "cn")
+        entity.add("topics", "reg.action")
+
+        sanction = h.make_sanction(context, entity)
+        h.apply_date(sanction, "listingDate", date)
+        sanction.set("sourceUrl", url)
+
```
jbothma, line 44: Let's do the Document and Article thing too https://zavod.opensanctions.org/extract/enforcements/#create-article-and-documentation-entities - it links the related entities in our data

datasets/cn/szse_trading/crawler.py
```diff
@@ -0,0 +1,74 @@
...
+            "scope": "0",
+        }
+        result = context.fetch_json(
+            context.data_url,
+            data=data,
+            method="POST",
+            headers=HEADERS,
+            cache_days=1,
+        )
+        for item in result.get("data"):
```
jbothma, line 69: This ceil division pattern works but isn't immediately obvious. A comment or more readable alternative would help the next reader.

If you really like the ceiling approach,

```python
import math
total_pages = math.ceil(result["totalSize"] / 20)
```

although we have a lot of crawlers that do a `while` and figure out the max after the first request

datasets/cn/szse_trading/crawler.py
```diff
@@ -0,0 +1,74 @@
...
+    "Referer": "https://www.szse.cn/disclosure/supervision/transaction/restrict/index.html",
+    "X-Request-Type": "ajax",
+    "X-Requested-With": "XMLHttpRequest",
+}
+
+
+def parse_item(context: Context, item: dict) -> None:
+    # epoch millis from API; convert to ISO date for h.apply_date
+    date = (
+        datetime.fromtimestamp(item.pop("docpubtime") / 1000, tz=CST).date().isoformat()
```
jbothma, line 17: dedication!
How sure are we this is China Standard Time and not UTC? best practice with stored data is UTC, to then present to users in whatever is appropriate. 
If we do want China Standard Time, any reason not to use this?

```
from zoneinfo import ZoneInfo
ZoneInfo("Asia/Shanghai")
```

datasets/cn/szse_trading/crawler.py
```diff
@@ -0,0 +1,74 @@
...
+    # title format: 限制交易决定书（name1、name2）— split on first/last parens
+    doc_title = item.pop("doctitle")
+    name_part = doc_title.split("（", 1)[1].rsplit("）", 1)[0]
+    # multi-entity notices delimit names with 、
+    names = h.multi_split(name_part, ["、", "，", ","])
+
+    for name in names:
+        entity = context.make("LegalEntity")
+        # doc_id is per-notice, not per-entity; name disambiguates within a notice
+        entity.id = context.make_id(name, doc_id)
```
jbothma, line 33: what do you think of this?

```suggestion
        entity.id = context.make_id(name, doc_title)
```

I think it's good to have something in addition to name here, but the document title feels more stable than an id which might change if they change their content management system.

datasets/cn/szse_trading/crawler.py
```diff
@@ -0,0 +1,74 @@
...
+            "currentPage": str(page),
+            "pageSize": "20",
+            "scope": "0",
+        }
+        result = context.fetch_json(
+            context.data_url,
+            data=data,
+            method="POST",
+            headers=HEADERS,
+            cache_days=1,
```
jbothma, line 67:
```suggestion
```

https://zavod.opensanctions.org/best_practices/caching/#issue-with-pagination-and-cached-pages

jbothma (commented):
Have we tried throwing the document content at an LLM to see how well it extracts things like the restriction start and end dates?

> In accordance with Article 6.8 of the Shenzhen Stock Exchange Trading Rules , the Exchange has decided to impose trading restrictions on the securities accounts under the name of Xiang Yongzhi from May 23, 2024 to June 6 , 2024 , that is , to restrict the relevant accounts from buying and selling all stocks listed and traded on the Exchange during the aforementioned period .

That's quite a short restriction, if google translate is accurate!

pudo:
What's the deal here? I've now sent it through Claude to do a run-through, but I'm curious what is needed to get this merged or discarded?

herrhellana:
We parked this crawler together with other Shenzhen ones (`blocked:full-index`) because we couldn't reliable parse all entity names (which was solved for this crawler), and this particular dataset was too small to offer much data -- see Ivan's comment above.

pudo:
Parked until when? If it is not going to be done, let's close it.

PR #3943, opened 2026-04-08
[at] Make the typechecker happy

leonhandreke:


pudo:
Superseded: both AT crawlers are already type-annotated and migrated to the typed `h.xpath_*`/`h.element_text` helpers on main, and `datasets/at` is out of the mypy exclude. Closing.

PR #3956, opened 2026-04-09
Germany's banned organizations involved in foreign-related extremism

herrhellana:
issue: https://github.com/opensanctions/crawler-planning/issues/389

### DIFF
--- a/datasets/de/bfv_extremism/crawler.py
+++ b/datasets/de/bfv_extremism/crawler.py
@@ -1,0 +1,100 @@
+import re
+from pydantic import BaseModel
+from lxml.etree import _Element
+
+from zavod import Context, helpers as h
+from rigour.names.split_phrases import contains_split_phrase
+from zavod.stateful.review import (
+    TextSourceValue,
+    review_extraction,
+    assert_all_accepted,
+)
+
+
+class Organization(BaseModel):
+    name: str
+    aliases: list[str] = []
+    abbreviations: list[str] = []
+    previous_names: list[str] = []
+
+
+class Organizations(BaseModel):
+    organizations: list[Organization]
+
+
+def crawl_row(context: Context, row: dict[str, _Element]) -> None:
+    str_row = h.cells_to_str(row)
+    name = str_row.pop("organisation")
+    assert name is not None
+
+    if contains_split_phrase(name):
+        source_value = TextSourceValue(
+            key_parts=name,
+            label="Organization Name",
+            text=name,
+            url=context.data_url,
+        )
+        # basic split and let reviewers do the rest
+        name, *aliases = h.multi_split(name, ["auch agierend unter"])
+        organization = [Organization(name=name, aliases=aliases)]
+        original_extraction = Organizations(organizations=organization)
+
+        review = review_extraction(
+            context,
+            source_value=source_value,
+            original_extraction=original_extraction,
+            origin="heuristic",
+        )
+
+        if not review.accepted:
+            return
+        extracted_data = review.extracted_data
+    else:
+        extracted_data = Organizations(organizations=[Organization(name=name)])
+
+    for item in extracted_data.organizations:
+        entity = context.make("Organization")
+        entity.id = context.make_id(item.name)
+
+        entity.add("name", item.name)
+        entity.add("alias", item.aliases)
+        entity.add("previousName", item.previous_names)
+        entity.add("abbreviation", item.abbreviations)
+        entity.add("topics", "sanction")
+        entity.add("country", "de")
+
+        ban_info = str_row.pop("verbot_verbotsgrunde_auszug")
+        assert ban_info is not None
+        sanction = h.make_sanction(context, entity)
+
+        if "Vollzug des Verbots" in ban_info:
+            ban_info = ban_info.split("Vollzug des Verbots: ")[1]
+            date, reason = re.split(r"Verbotsgr[uü]nd\w*:\s*", ban_info, maxsplit=1)
+            h.apply_date(sanction, "listingDate", date.strip())
+            sanction.add("reason", reason.strip())
+        elif "Inkrafttreten des Kennzeichenverbots" in ban_info:
+            ban_info = ban_info.split("Inkrafttreten des Kennzeichenverbots: ")[1]
+            pattern = r"\d{1,2}\.\s+(?:Januar|Februar|März|April|Mai|Juni|Juli|August|September|Oktober|November|Dezember)\s+(?:19|20)\d{2}"
+            match = re.search(pattern, ban_info)
+            assert match is not None
+            date = match.group().strip()
+            reason = ban_info[match.end() :].strip()
+            h.apply_date(sanction, "listingDate", date.strip())
+            sanction.add("reason", reason.strip())
+        else:
+            context.log.warning(f"Unexpected ban info format: {ban_info}")
+
+        context.emit(entity)
+        context.emit(sanction)
+
+    context.audit_data(str_row, ignore=["verbotene_symbole_und_kennzeichen"])
+
+
+def crawl(context: Context) -> None:
+    doc = context.fetch_html(context.data_url)
+    table = h.xpath_element(doc, ".//table")
+
+    for row in h.parse_html_table(table):
+        crawl_row(context, row)
+
+    assert_all_accepted(context, raise_on_unaccepted=False)
--- a/datasets/de/bfv_extremism/crawler.py
+++ b/datasets/de/bfv_extremism/crawler.py
@@ -1,9 +1,7 @@
-import re
 from pydantic import BaseModel
 from lxml.etree import _Element
 
 from zavod import Context, helpers as h
-from rigour.names.split_phrases import contains_split_phrase
 from zavod.stateful.review import (
     TextSourceValue,
     review_extraction,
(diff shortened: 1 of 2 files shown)


herrhellana:
Note that the first org name contains aliases, so i set up a review for it instead of using lookups as part of transitioning away from the latter.

datasets/de/bfv_extremism/crawler.py
```diff
@@ -0,0 +1,100 @@
...
+            ban_info = ban_info.split("Inkrafttreten des Kennzeichenverbots: ")[1]
+            pattern = r"\d{1,2}\.\s+(?:Januar|Februar|März|April|Mai|Juni|Juli|August|September|Oktober|November|Dezember)\s+(?:19|20)\d{2}"
+            match = re.search(pattern, ban_info)
+            assert match is not None
+            date = match.group().strip()
+            reason = ban_info[match.end() :].strip()
+            h.apply_date(sanction, "listingDate", date.strip())
+            sanction.add("reason", reason.strip())
+        else:
+            context.log.warning(f"Unexpected ban info format: {ban_info}")
```
jbothma, line 85: Nice work getting it working, but I think we should er on the side of caution because this is going to be very brittle and just create work for us, for little value. Let's always imagine we have a quota of 0-1 regex per crawler. We don't want to use an LLM for data extraction on sanctions data, and this really is free text.

I think this crawler can be functionally complete by just splitting the names in the first column and sticking the rest into a notes field in the entity. I don't even think the start date is important enough for a regex. I think a regex would be justified if it gave us more identifying information.

jbothma: given that, and the fact that the names are also a bit of a mess, what do you think of just manually extracting everything? Just send every row to review, perhaps with names added to the name field of a single organization to get the reviewer started. So the source value is the entire row. 

- organizations
  - names
  - abbreviations
  - aliases
- start_date
- reason

If the review isn't accepted yet, just take the entire name column's value and emit an entity with that and the Verbot column as notes.

Out of interest, when the source value changes, the review system will invalidate the review, i.e. it becomes unaccepted and a reviewer has to re-review it. That way we detect changes that we have to incorporate in the extraction.

herrhellana: yeah that sounds reasonable and seems to be a better approach, i'll rewrite it now

datasets/de/bfv_extremism/crawler.py
```diff
@@ -0,0 +1,83 @@
...
+
+    review = review_extraction(
+        context,
+        source_value=source_value,
+        original_extraction=original_extraction,
+        origin="heuristic",
+    )
+
+    if not review.accepted:
+        return
```
jbothma: if review isn't accepted we can make the entities from original_extraction and still emit some entity with a technically usable name

datasets/de/bfv_extremism/crawler.py
```diff
@@ -0,0 +1,83 @@
...
+    assert name is not None
+    assert ban_info is not None
+
+    # send every row for review
+    source_value = TextSourceValue(
+        key_parts=name,
+        label="Organization and Ban Info",
+        text=f"{name}\n\n{ban_info}",
+        url=context.data_url,
+    )
```
jbothma, line 38: out of interest, did you try HtmlSourceValue?

I tried it quick but as expected, a table row element <tr> without the table just looks pretty shit, and wrapping it in a table would be quite weird for a crawler. So I think your approach is fine. I think we should always be a bit careful with "constructing" source values, like we do here, but I think it's ok if it's less weird than the alternatives

https://github.com/opensanctions/opensanctions/compare/de-bfv...de-bfv-html-source?

Image

datasets/de/bfv_extremism/crawler.py
jbothma: that should fix your current CI error because we'd then be emitting entities in CI.

datasets/de/bfv_extremism/de_bfv_extremism.yml
```diff
@@ -0,0 +1,63 @@
+title: Germany's banned organizations involved in foreign-related extremism
+prefix: de-bfv
+coverage:
+  frequency: monthly
```
jbothma: this is probably updated extremely rarely, but maybe we can do weekly, just to improve the chance that we include someone before we get emailed about why they're missing.

datasets/de/bfv_extremism/de_bfv_extremism.yml
```diff
@@ -0,0 +1,63 @@
+title: Germany's banned organizations involved in foreign-related extremism
```
jbothma:
```suggestion
title: Germany Banned organizations and symbols in foreign-related extremism
```

I was about to link to the documentation that says we prefix dataset names with the country name simply as `Country` not `Country's` so that it's consistent and clean and is grouping country datasets nicely on the dataset list ordered alphabetically. But it doesn't seem to be written there! Sorry! Would you mind adding the bit about country name to the docs?

https://zavod.opensanctions.org/metadata/#:~:text=Basics-,title,-%2D%20As%20close%20as

datasets/de/bfv_extremism/de_bfv_extremism.yml
```diff
@@ -0,0 +1,63 @@
...
+prefix: de-bfv
+coverage:
+  frequency: monthly
+  start: 2026-04-08
+load_statements: true
+entry_point: crawler.py
+summary: >-
+  Organizations from the field of foreign-related extremism banned at the federal level
+  by the Federal Ministry of the Interior (BMI) under Section 3 of the German Associations Act (Vereinsgesetz).
```
jbothma: I'm not sure about "field of foreign-related extremism" even though I see it in the google translate for the page. I guess it's a "field" but like, I doubt it has a SIC code.

My first thought is also that a reference to legislation is maybe a bit heavy for the summary that'll be the blurb in the google search result. I wonder if we can word this to be a bit more accessible. Maybe claude has useful words.

Image

jbothma (commented):
Thanks for the change - this is generally looking great. I think some easy changes and it's merged.

datasets/de/bfv_extremism/crawler.py
herrhellana: ok sounds good! i wasn't sure if i should emit none for now or create a temporary substitute.

datasets/de/bfv_extremism/crawler.py
herrhellana: interesting! i didnt; how did you make the changes visible in the UI if the crawler isnt merged?

datasets/de/bfv_extremism/de_bfv_extremism.yml
herrhellana: sounds good :-)

herrhellana: p.s.: i didnt include symbols there because we're not scraping those

datasets/de/bfv_extremism/de_bfv_extremism.yml
herrhellana: perhaps: 

"Extremist organizations with ties to political conflicts abroad (auslandsbezogener Extremismus) that have been banned across Germany by the Federal Ministry of the Interior (BMI)."

jbothma: that's excellent

datasets/de/bfv_extremism/de_bfv_extremism.yml
jbothma: Thanks - you're very right.

datasets/de/bfv_extremism/crawler.py
jbothma: It's easier to find when we link to it :) https://github.com/opensanctions/opensanctions/pull/4125/changes#diff-8eafc59cba50198fb2e4e5b852c75feae53aac1798f65c965c212edd0c0016efR24

datasets/de/bfv_extremism/crawler.py
```diff
@@ -0,0 +1,89 @@
...
+        temp_name = original_extraction.organizations[0].name
+        entity = context.make("Organization")
+        entity.id = context.make_id(temp_name)
+        entity.add("name", temp_name)
+        entity.add("country", "de")
+        context.emit(entity)
+        return
+    extracted_data = review.extracted_data
+
+    for item in extracted_data.organizations:
```
jbothma, line 56: Two ways this can improve:

- we're duplicating code and there's a nice way to avoid that
- we're assuming a single organization when the code above it may change without realising there's this assumption further down. (The principle here is to think how someone is going to misread the code in future - we're writing code for tomorrow's maintainer)

```suggestion
    if not review.accepted:
        extracted_date = original_extraction
    else:
        extracted_data = review.extracted_data

    for item in extracted_data.organizations:
```

jbothma (approved):
Thanks for this!
I think you'd basically be able to accept the proposed change and merge unless I missed something.

datasets/de/bfv_extremism/crawler.py
herrhellana: ok i'm falling in love with these new docs and github migration... i think i also just need to sit down and reread everything as my attention now will go to more specific vs broad topics

PR #3971, opened 2026-04-10
Skill name framework migration: Step 1

### LINKED ISSUE #3951: Migrate name-cleaning hacks to standardized name framework
Several crawlers still use ad-hoc name cleaning logic — `if/in name` checks, bracket stripping, manual splitting on delimiters, etc. — instead of the name cleaning framework. These are fragile, inconsistent across datasets, and hard to maintain.

Goal is to incrementally migrate these to the name framework, prioritizing crawlers where the hacks are most brittle or where we've seen data quality issues. Not a single big-bang rewrite — just chipping away at it as we touch each crawler.

The migration strategy per dataset:

1. Switch the dataset to write-only mode (`h.review_names`) — pump fixes into the review system without applying output yet
2. Leave the existing hack in place while reviews accumulate
3. Once all reviews are applied, switch to `h.apply_reviewed_names` and remove the hack

### LINKED ISSUE #3970: names framework claude skill
The datasets repo has a single large `claude.md` that inlines `zavod` documentation. We're restructuring it into task-specific skill files. This skill covers one specific migration task: replacing ad-hoc name cleaning logic in crawlers with the name framework.

What the skill should cover:
The problem it addresses: many crawlers clean names manually before calling h.apply_name — delimiter splits, bracket stripping, regex substitutions, conditional checks on name content. These are fragile and inconsistent.
The one-stage migration strategy per crawler:

Add h.review_names in write-only mode alongside the existing hack (do not remove the hack).

nvmbrasserie:


.claude/skills/name-framework-migration/SKILL.md
```diff
@@ -0,0 +1,79 @@
...
+if len(parts) == 2:
+    h.apply_name(entity, first_name=parts[1].strip(), last_name=parts[0].strip())
+else:
+    h.apply_name(entity, full=name)
+h.review_names(context, entity, original=h.Names(name=raw_name), default_accepted=False)
+```
+
+## Do not
+
+- Do not remove the existing hack
```
leonhandreke, line 75: Maybe "Do not modify or remove the existing name cleaning logic"?

.claude/skills/name-framework-migration/SKILL.md
```diff
@@ -0,0 +1,79 @@
...
+
+```python
+raw_name = name_raw  # capture before any mutation
+name = name_raw.replace("(Acting)", "").strip()
+parts = name.split(",", 1)
+if len(parts) == 2:
+    h.apply_name(entity, first_name=parts[1].strip(), last_name=parts[0].strip())
+else:
+    h.apply_name(entity, full=name)
+h.review_names(context, entity, original=h.Names(name=raw_name), default_accepted=False)
```
leonhandreke, line 70: Didn't we have a mechanism to build suggested names - do we maybe want to keep this around for very simple cases to reduce the review load?

leonhandreke (commented):
It was good reading this, if only to keep up to date with what's happening in Claude-land. I don't feel I have enough context on the whole name cleaning thing to review this well, but I added some comments anyway that you can take with a grain of salt :)

pudo:
I like this a lot. Some open questions:
* Do we want to frame the skill as the migration or the eventual `name-review` skill? Once the initial upgrades are done we want this to just be one of the things the thing knows to do, like `zyte-unblocking` or `pep-positions` etc.
* Maybe we should add a section to the root `CLAUDE.md` that says: here's the docs, whenever you develop a skill, consider upstreaming the best practices into the relevant `zavod/docs/<xxx>` file. I also want to expose a docs style guide somewhere at some point, but that need not be this PR.

datasets/us/ga/med_exclusions/crawler.py
```diff
@@ -59,6 +60,13 @@ def crawl_item(row: Dict[str, str | None], context: Context):
         h.apply_name(
             entity, first_name=first_name, middle_name=middle_name, last_name=last_name
         )
+        if raw_middle_name is not None:
+            h.review_names(
+                context,
+                entity,
+                original=h.Names(name=raw_middle_name),
+                default_accepted=False,
+            )
```
jbothma, line 69: what sort of irregularities are you catching here? we don't support name parts yet - only complete names. So there isn't really a path for it to apply the cleaned name part after review.

Maybe we should codify that if the skill is doing this a lot.

datasets/us/ddtc_debarred/crawler.py
```diff
@@ -51,10 +51,12 @@ def crawl_debarment(
         schema = "LegalEntity"
 
     entity = context.make(schema)
+    raw_name = row[name_field]
     name, aliases = split_names(row.pop(name_field))
     entity.id = context.make_slug(name, date_of_birth, strict=False)
     entity.add("name", name)
     entity.add("alias", aliases)
+    h.review_names(context, entity, original=h.Names(name=raw_name), default_accepted=False)
```
jbothma, line 59: it's a bit noisy if it's setting default_accepted-False all the time. It only needs to set it if it's changing from the default.

.claude/skills/name-framework-migration/SKILL.md
```diff
@@ -0,0 +1,79 @@
...
+if len(name_split) > 1:
+    entity.add("alias", name_split[1:])
+```
+
+## Migration steps
+
+1. Capture the raw name value **before** any split/clean/pop logic runs.
+2. Verify the exact `review_names` signature and `Names` type from `zavod/zavod/helpers/names.py` — do not guess.
+3. Add `h.review_names(context, entity, original=h.Names(name=<raw>), default_accepted=False)` immediately after the existing `entity.add`/`h.apply_name` calls.
+4. Guard with `if <raw> is not None:` if the value can be `None`.
```
jbothma, line 47: actually the design of Names is to allow `Names(name=str_or_none)` and `Names().add("name", str_or_none)` so crawlers don't have to do a ton of guarding.

jbothma:
> * Do we want to frame the skill as the migration or the eventual `name-review` skill? Once the initial upgrades are done we want this to just be one of the things the thing knows to do, like `zyte-unblocking` or `pep-positions` etc.

Now that I've finally documented the migration a bit more properly (https://zavod.opensanctions.org/extract/names/#migrating-to-the-name-cleaning-helpers) and (https://zavod.opensanctions.org/extract/names/#example-usage), I think we should either have two skills - one for step one and one for step 3, or one skill with an explicit instruction whether we're doing step 1 or step 3. On the one hand, there might be some duplication if we do two skills. On the other hand, perhaps it can cope with reading the common documentation from `zavod/docs/extract/names.md` and each skill just has more examples?

.claude/skills/name-framework-migration/SKILL.md
```diff
@@ -0,0 +1,79 @@
...
+if len(parts) == 2:
+    h.apply_name(entity, first_name=parts[1].strip(), last_name=parts[0].strip())
+else:
+    h.apply_name(entity, full=name)
+```
+
+### After
+
+```python
+raw_name = name_raw  # capture before any mutation
```
jbothma, line 63: raw_name = name_raw? :)

.claude/skills/name-framework-migration/SKILL.md
jbothma: reviews will only be created for "irregular" names. So basically things that are really short, single tokens, contains rejected chars like ( or /, and the list of rejected chars can be customised to e.g. include , when appropriate.

Do we want to enable llm_cleaning in the examples?

Should opting in/out of llm cleaning be an argument we supply to the skill? basically the rule of thumb is no llm for sanctions crawlers. But I'm unsure about letting the crawler decide that.

.claude/skills/name-framework-migration/examples/migrations.md
```diff
@@ -0,0 +1,131 @@
...
+entity.id = context.make_slug(name, date_of_birth, strict=False)
+entity.add("name", name)
+entity.add("alias", aliases)
+```
+
+```python
+# AFTER
+entity = context.make(schema)
+raw_name = row[name_field]                          # capture before pop
+name, aliases = split_names(row.pop(name_field))
```
jbothma, line 25: bit of a nit, but before the bots spray our code, this feels a bit funny. Wouldn't the normal thing (for our codebase) be

```python
raw_name = row.pop(name_field)
name, aliases = split_names(raw_name)
...
h.review_names(context, entity, original=h.Names(name=raw_name))
```

.claude/skills/name-framework-migration/examples/migrations.md
```diff
@@ -0,0 +1,131 @@
...
+entity.id = context.make_slug(name, date_of_birth, strict=False)
+entity.add("name", name)
+entity.add("alias", aliases)
+h.review_names(context, entity, original=h.Names(name=raw_name), default_accepted=False)
+```
+
+---
+
+## Pattern 2: Optional value, None guard required
```
jbothma, line 34: none guard can be skipped to trim crawler code. names helpers don't support name parts

.claude/skills/name-framework-migration/examples/migrations.md
```diff
@@ -0,0 +1,131 @@
...
+    organization.id = context.make_slug(name)
+    organization.add("topics", "sanction")
+    organization.add("name", name)
+    organization.add("alias", aliases)
+    h.review_names(
+        context,
+        organization,
+        original=h.Names(name=raw_name),
+        default_accepted=False,
+    )
```
jbothma, line 120: my feeling is that for sanctions crawlers where we're not using an LLM for cleaning, we can still do best effort cleaning in the crawler as long as it's very simple. I see no reason why `, and` and `,` can't be added to ALIAS_SPLITS. Then this is a single h.multi_split call. 

Then those get added to a `suggestions = Names()` and passed to apply_reviewed_names (when reviews are done).

Heuristics still gives us some automated categorisation and we catch any other irregularity like punctuation.

And the h.multi_split doesn't get removed in step 3 of the migration. https://zavod.opensanctions.org/extract/names/#migrating-to-the-name-cleaning-helpers

See an example final version in https://zavod.opensanctions.org/extract/names/#sanctions-crawler-with-some-custom-splitting

@pudo does that track with your thinking about this  (for the sanctions use case in particular)?

We could also introduce a split_on_split_phrases or something with a better name, which trims , and ; on the ends.

datasets/us/ga/med_exclusions/crawler.py
nvmbrasserie:
```
 middle_name = row.pop("middle_name")
 raw_middle_name: str | None = middle_name
...
if middle_name and "aka" in middle_name.lower():
        split = split_middle_name(context, middle_name)
        middle_name, aliases, first_names = split
```

sometimes there're aliases in the `middle_name` column. here we're trying to catch those

.claude/skills/name-framework-migration/examples/migrations.md
```diff
@@ -0,0 +1,151 @@
+# Real migration examples
+
+Three patterns from production crawlers. Each shows a distinct variation of trigger pattern and raw-capture approach.
+
+---
+
+## Pattern 1: dataset-specific delimiter splits name into primary + aliases
+
+**Crawler:** `datasets/pk/proscribed_persons/crawler.py`  
+**Trigger:** `split("@")` — a dataset-specific delimiter embedded in the raw name string.
```
jbothma:
```suggestion
**Trigger:** `split("@")` — a dataset-specific delimiter embedded in the raw name string denoting zero or more aliases.  
```

.claude/skills/name-framework-migration/examples/migrations.md
```diff
@@ -0,0 +1,151 @@
+# Real migration examples
+
+Three patterns from production crawlers. Each shows a distinct variation of trigger pattern and raw-capture approach.
+
+---
+
+## Pattern 1: dataset-specific delimiter splits name into primary + aliases
```
jbothma:
```suggestion
## Pattern 1: dataset-specific delimiter splits name string into primary name + aliases
```

.claude/skills/name-framework-migration-first-step/examples/migrations.md
```diff
@@ -0,0 +1,151 @@
+# Real migration examples
+
+Three patterns from production crawlers. Each shows a distinct variation of trigger pattern and raw-capture approach.
+
+---
+
+## Pattern 1: dataset-specific delimiter splits name into primary + aliases
+
+**Crawler:** `datasets/pk/proscribed_persons/crawler.py`
```
jbothma, line 9: does it help to name the specific crawler? I worry that it's not adding anything and maybe will cause the skill-running LLM confusion down the line.

.claude/skills/name-framework-migration/SKILL.md
```diff
@@ -0,0 +1,82 @@
+---
+name: name-framework-migration
```
jbothma: Let's explicitly reference `zavod/docs/extract/names.md#migrating-to-the-name-cleaning-helpers` Step 1.

Let's call it something like `name-framework-migrate-step-1`. If you need to document the migration process more explicitly, consider improving `zavod/docs/extract/names.md#migrating-to-the-name-cleaning-helpers` rather than adding detail to the skill.

Let's make a second skill `name-framework-migrate-step-3` in a separate PR which replaces existing `h.apply_name` and `Entity.add(name_prop` calls with `h.apply_reviewed_...`

.claude/skills/name-framework-migration/SKILL.md
```diff
@@ -0,0 +1,82 @@
...
+```python
+name = name_raw.replace("(Acting)", "").strip()
+parts = name.split(",", 1)
+if len(parts) == 2:
+    h.apply_name(entity, first_name=parts[1].strip(), last_name=parts[0].strip())
+else:
+    h.apply_name(entity, full=name)
+# Submit the unmodified source string for review; existing name properties are unchanged.
+h.review_names(context, entity, original=h.Names(name=name_raw))
+```
```
jbothma: If it's a sanction crawler, I'd argue that we should also supply a suggested value here with the rearranged name, and introduce a helper (separate issue, PR) for rearranging names if that's something we'll be doing more than once.

That will create reviews for the cases where suggested differs from original, and also for any other random irregularity.

I think automating the rearrangement will be less error-prone than doing that by hand.

```suggestion
name = name_raw.replace("(Acting)", "").strip()
parts = name.split(",", 1)
if len(parts) == 2:
    h.apply_name(entity, first_name=parts[1].strip(), last_name=parts[0].strip())
    full_name = h.make_name(first_name=parts[1], last_name=parts[0])
    suggested = h.Names(name=full_name)
else:
    h.apply_name(entity, full=name)
    suggested = h.Names(name=name)
# Submit the unmodified source string for review; existing name properties are unchanged.
is_irregular, suggested = h.check_names_regularity(entity, names)
h.review_names(context, entity, original=h.Names(name=name_raw), suggested=suggested, is_irregular=is_irregular)
```

I think for non-sanctions crawlers, it should just be as you proposed but with llm_cleaning=True.

I think either you should document the skill to only be applicable to sanctions crawlers right up top, or you should add examples for non-sanctions crawlers.

datasets/nz/designated_terrorists/crawler.py
```diff
@@ -49,6 +49,7 @@ def parse_table(
 
 def crawl_item(input_dict: dict, context: Context):
     # aliases will be either a list of size one or None if there is no aliases
+    raw_name: str = input_dict["terrorist-entity"][0]
     name, *aliases = h.multi_split(input_dict.pop("terrorist-entity")[0], ALIAS_SPLITS)
```
jbothma: let's use raw_names

datasets/nz/designated_terrorists/crawler.py
```diff
@@ -59,6 +60,11 @@ def crawl_item(input_dict: dict, context: Context):
 
     organization.add("name", name)
     organization.add("alias", aliases)
+    h.review_names(
+        context,
+        organization,
+        original=h.Names(name=raw_name),
+    )
```
jbothma: pass in suggested

datasets/us/ddtc_debarred/crawler.py
```diff
@@ -51,10 +51,12 @@ def crawl_debarment(
         schema = "LegalEntity"
 
     entity = context.make(schema)
+    raw_name = row[name_field]
     name, aliases = split_names(row.pop(name_field))
     entity.id = context.make_slug(name, date_of_birth, strict=False)
     entity.add("name", name)
     entity.add("alias", aliases)
+    h.review_names(context, entity, original=h.Names(name=raw_name))
```
jbothma: enable llm cleaning

.claude/skills/name-framework-migration/SKILL.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Do not
+
+- Do not modify or remove the existing name cleaning logic
+- Do not pass the cleaned/intermediate name as `original` — always use the unmodified raw source string
+- Do not use `llm_cleaning=True` (sanctions crawler)
+- Do not construct `Names` by guessing — read the source first
+- Do not add explanatory comments beyond what the code requires
```
jbothma: should this be in root level CLAUDE.md "Add comments to explain **why** we're doing something if it isn't obvious. Do not add comments explaining **what** we're doing unless it's easy to misunderstand."?

.claude/skills/name-framework-migration/SKILL.md
```diff
@@ -0,0 +1,82 @@
...
+    h.apply_name(entity, full=name)
+# Submit the unmodified source string for review; existing name properties are unchanged.
+h.review_names(context, entity, original=h.Names(name=name_raw))
+```
+
+## Do not
+
+- Do not modify or remove the existing name cleaning logic
+- Do not pass the cleaned/intermediate name as `original` — always use the unmodified raw source string
+- Do not use `llm_cleaning=True` (sanctions crawler)
```
jbothma: are these still needed? If so, can we fix them by improving the name cleaning docs page?

.claude/skills/name-framework-migration/SKILL.md
jbothma: Do we still need this example in the skill file itself now that we have a concrete migration example? https://github.com/opensanctions/opensanctions/pull/4075/changes

.claude/skills/name-framework-migration-first-step/examples/migrations.md
```diff
@@ -0,0 +1,151 @@
...
+
+```python
+# AFTER
+def crawl_entity(context: Context, data: Dict[str, Any]):
+    name_raw = data.pop("title")
+    if not name_raw:
+        return
+    ...
+    entity.add("name", RE_NAME_SPLIT.split(name_raw))
+    h.review_names(context, entity, original=h.Names(name=name_raw))
```
jbothma: llm_cleaning=true and custom check_name_irregularity until the additional split phrases have been added to rigour like

```
h.review_names(
    context,
    entity,
    originals,
    is_irregular=RE_NAME_SPLIT.search(...) is not None, # or something?
)

Comment (jbothma) [reply .claude/skills/name-framework-migration-first-step/examples/migrations.md]: otherwise we'd be missing out irregular cases, won't we

File: .claude/skills/name-framework-migration-first-step/examples/migrations.md [hunk 21]
```diff
@@ -0,0 +1,151 @@
...
+
+**Crawler:** `datasets/ee/international_sanctions/crawler.py`, function `crawl_item_human_rights`  
+**Trigger:** three transformations stack on the raw string:
+1. Numeric prefix stripped by regex (`^\d+\.\d*\.?\s*`)
+2. Alias text extracted from `(also known as ...)` brackets
+3. Remaining name split on `", "` to feed `h.apply_name(first_name=..., last_name=...)`
+
+**Capture approach:** none needed — `raw_name` is the function parameter; no mutation happens to it.
+
+```python
```
jbothma, line 109: @pudo would it be better for this migration to replace the regex with cleaning 433 items by hand? Or maybe just simplify the regex into a "also" split and separate numbering strip, followed by review?

@nvmbrasserie I see this one is also going down to name parts - I think you should simplify the example to get rid of the name parts part. it's complicating things and if we're cleaning by hand, we don't yet support adding clean name parts to the entity so this crawler will have to wait.

If we're cleaning by hand rather than regex, we want dots in names to be irregular so add `.` to reject_chars https://zavod.opensanctions.org/extract/names/#whats-a-dirty-name

.claude/skills/name-framework-migration-first-step/examples/migrations.md
```diff
@@ -0,0 +1,151 @@
...
+        entity.add("alias", alias)
+    h.review_names(context, entity, original=h.Names(name=raw_name))
+```
+
+---
+
+## Key decisions consistent across all three
+
+- `h.Names(name=<raw>)` — `Names` only has `name`, `alias`, `weakAlias`, `previousName`, `abbreviation` fields; no `middleName`. Use `name` for unsplit source strings.
+- `h.Names` is re-exported via `zavod.helpers` — verify it appears in `zavod/zavod/helpers/__init__.py` before assuming no extra import is needed.
```
jbothma, line 170: maybe this should be replaced by something like 

> Access `zavod.extract.names.clean.Names` as `h.Names` using our standard shortcut import `from zavod import helpers as h`

nvmbrasserie:
As discussed in today’s eng meeting regarding the next steps for Claude Skills, I’ve made a few updates and aligned everything with the latest Claude Code docs. It looks good from my side, so let’s proceed with testing in production. I’ll merge this now, and we can iterate through follow-up reviews as we go

PR #3972, opened 2026-04-10
Add type annotations to ar/parliament and ar/repet crawlers

leonhandreke:
Two commits:

**1. Add type annotations to ar/parliament and ar/repet crawlers**

Fixes `mypy --strict --explicit-package-bases datasets/ar`.

- `ar/parliament`: removed the untyped `_extract_text` helper (replaced with `h.xpath_string` directly, same thing), added missing return types
- `ar/repet`: typed `values`, `parse_date`, `parse_alias`, `fetch`, `crawl_common` and the top-level `crawl_*` functions; dropped deprecated `typing.Dict`/`Optional` in favour of `dict`/`X | None`; added docstring + example input to `parse_alias`

**2. [ar_repet] Fix aliases not being added with no quality spec**

Found this while fixing the types — the type checker flagged that `name_prop: str | None` was being passed to `apply_name(name_prop: str)`. Turns out this was a real bug: when QUALITY was absent, `name_prop=None` was passed, and `apply_name` would call `entity.add(None, full, quiet=True)`, silently dropping the name.

Fixed by defaulting to `"alias"` when QUALITY is missing or unrecognised (with a warning in the latter case). Also changed `alias.pop("ALIAS_NAME", None)` to `alias.pop("ALIAS_NAME", "")` to avoid an AttributeError when ALIAS_NAME is absent.

### DIFF
--- a/.pre-commit-config.yml
+++ b/.pre-commit-config.yml
@@ -78,7 +78,6 @@ repos:
           datasets/_wikidata|
           datasets/ae|
           datasets/am|
-          datasets/ar|
           datasets/at|
           datasets/ba|
           datasets/be|
--- a/datasets/ar/parliament/crawler.py
+++ b/datasets/ar/parliament/crawler.py
@@ -4,6 +4,7 @@ from rigour.mime.types import CSV
 from zavod import Context, helpers as h
 from zavod.extract import zyte_api
 from zavod.stateful.positions import categorise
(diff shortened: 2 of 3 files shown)


leonhandreke:
Verified `ar_repet` using `ftm-stmt diff`. Loads of new names, yay! Not the first bug we found by typing...

PR #3996, opened 2026-04-15
Modularize datasets/CLAUDE.md into routing layer + crawler skills

herrhellana:
issue: https://github.com/opensanctions/opensanctions/issues/3952


herrhellana:
Replacing the monolithic `datasets/CLAUDE.md` with a modular skill system that agents can invoke via `/crawler-pep` and `/crawler-sanctions` + the url argument (can be either the data source link or link to the Github PR). 

The root `CLAUDE.md` becomes a slim routing layer (~35 lines) that points to the right skill based on task. 

  ```
### Architecture

  CLAUDE.md                                — routing layer (always loaded)
  .claude/docs/crawler-guide.md            — shared patterns (loaded on demand)
  .claude/skills/crawler-pep/
    SKILL.md                               — PEP rules & decisions (~170 lines)
    examples.md                            — full code examples (~180 lines)
  .claude/skills/crawler-sanctions/
    SKILL.md                               — sanctions rules & decisions (~230 lines)
    examples.md                            — full code examples (~290 lines)
  ```


Code examples live in separate `examples.md` files following Anthropic's recommendation to keep `SKILL.md` under 500 lines. This avoids context compaction issues (Claude Code preserves only the first 5,000 tokens of each skill after the first response).


Other notes: 
- stop and warns users if Zyte is needed (eg bot detection).
- agents performed not bad at skill invocation without explicitly reading the guide (see test eval below) -- i'm oscillating between forcing them to read it (which adds up ~600 lines each time), or trimming down the guide as we grow skills, or keeping it as is but then enhance the skills a bit. the only problem i saw so far is with the agent not passing the cols to ignore into context.audit_data() arg but popping them into the void, -- the correct approach lives in the guide that the agent didnt read. 
- to keep the skills self-contained, i added instructions to prevent agents from searching the entire repo, and it worked (looking at their thinking process)
- i also trimmed down the guide / skills and got rid of repetitions found in `datasets/CLAUDE.md` 
- i didnt touch the `datasets/CLAUDE.md`, as it's good to still use it for human cross-reference while we're building up the vibes; but the idea of the PR is to get rid of it and migrate its content into the overarching guide + skills

herrhellana:
[lu_bourgmestres] claude's eval of another agent's thinking process for` /crawler-pep https://github.com/opensanctions/crawler-planning/issues/652 ` (skill runtime: 4m 30s): 

issue: https://github.com/opensanctions/crawler-planning/issues/652

```
⏺ This is a clean run. Here's my evaluation:

  What went well:
  - Read the GitHub issue first (as instructed)
  - Did NOT read datasets/CLAUDE.md or other crawler source files
  - Did NOT spawn Explore/Plan subagents
  - Fetched the data source directly, inspected structure with quick Python one-liners
  - Correctly followed Pattern C (multi-position with is_pep=True)
  - Caught its own bug (wikidata_id collapsing positions) and fixed it
  - Ran zavod crawl + zavod validate to verify
  - 4m 30s total — reasonable

  One unnecessary step:
  - Searched for 1 pattern — likely a grep for existing Luxembourg datasets. The skill says "Do NOT search the repository for similar crawlers" but it still did one search. Minor token burn.

  One thing it skipped:
  - It says "I already have the crawler guide loaded" — meaning it didn't explicitly Read the guide file. It relied on it being loaded via the CLAUDE.md context. This worked fine here, but it means the "read
  these files" instruction wasn't followed literally. The examples file also wasn't read — the agent didn't need it since the skill's inline rules were sufficient for this straightforward case.

  Wikidata issue is interesting feedback for the skill:
  The agent correctly found wikidata IDs for "bourgmestre" and "échevin", used them, then discovered they collapsed per-commune positions into 2 entities. The skill could add a note like: "Do not use
  wikidata_id for per-municipality/per-region positions — it collapses all instances into one entity. Only use wikidata_id for nationally unique positions (e.g. 'Member of Parliament')." <<-- this is added. 
```


[ph_gppb_debarred] claude's eval of another agent's thinking process for`/crawler-sanctions https://github.com/opensanctions/crawler-planning/issues/654` (skill runtime: 9m 1s): 

issue: https://github.com/opensanctions/crawler-planning/issues/654

```
⏺ Here's my evaluation:

  What went well:
  - Fetched GitHub issue first, extracted data URL
  - Did NOT read datasets/CLAUDE.md
  - Handled a tricky source: reverse-engineered a Next.js SPA to find the public API endpoint — impressive and necessary
  - Correctly used make_sanction() with program keys, h.lookup_sanction_program_key()
  - Self-corrected: caught duplicate reg numbers, added key=start_date for disambiguation, handled N/A reg numbers
  - Ran crawl + validate, fixed assertion issue
  - Only created .yml + crawler.py (didn't touch collections)

  Problems:

  1. Too many exploratory bash calls — 12+ Python one-liners to reverse-engineer the API. This was necessary for this source (SPA with no obvious API), but it's a lot of token burn. Not a skill issue though —
   the source was genuinely hard to inspect.
  2. Searched for 2 patterns (appeared twice) — still grepping the repo despite the "do not search" instruction. The skill instruction isn't being followed perfectly.
  3. Didn't read the guide or examples file — same as the PEP run. The agent relied on the skill's inline content and whatever was already in context. This means the guide's rules about "only create .yml and
  crawler.py" were followed by coincidence, not because it read them.
  4. Validation assertion issue — the agent added Sanction: 70 to assertions, then discovered the validator doesn't count Sanction entities for debarment lists and had to remove it. This is domain knowledge
  that could be in the guide.

  Feedback for skills:
  - The "do not search the repo" instruction is being ignored ~50% of the time. Could strengthen it or accept that agents will occasionally search.
  - Both runs skipped reading the guide/examples files — the instruction to read them should be more forceful, or we accept the agent will use them only when it needs patterns it doesn't already know.
```

datasets/lu/bourgmestres/lu_bourgmestres.yml
```diff
@@ -0,0 +1,59 @@
...
+entry_point: crawler.py
+prefix: lu-bourg
+coverage:
+  frequency: monthly
+  start: 2026-04-14
+load_statements: true
+ci_test: true
+summary: >-
+  Members of the collège échevinal (bourgmestres and échevins) of Luxembourg
+  municipalities since the 2023 communal elections.
```
herrhellana, line 11: this should be more broad, like "Data updated biannually after communal elections."

datasets/lu/bourgmestres/lu_bourgmestres.yml
```diff
@@ -0,0 +1,59 @@
...
+  The dataset lists all bourgmestres (mayors) and échevins (aldermen) of
+  Luxembourg's 100 municipalities for the 2023–2029 electoral period, including
+  those who have since left office.
+
+  Source: [data.public.lu](https://data.public.lu/en/datasets/college-des-bourgmestre-et-echevins-depuis-les-elections-communales-2023/)
+
+url: https://data.public.lu/en/datasets/college-des-bourgmestre-et-echevins-depuis-les-elections-communales-2023/
+
+publisher:
+  name: Portail Open Data Luxembourg
```
herrhellana, line 23: this looks a lil bit off

datasets/lu/bourgmestres/crawler.py
```diff
@@ -0,0 +1,69 @@
...
+    pos_key = f"{man_label}-{com_code}"
+    if pos_key not in positions:
+        position = h.make_position(
+            context,
+            name=f"{man_label} de {com_label}",
+            country="lu",
+            subnational_area=com_label,
+            lang="fra",
+        )
+        cat = categorise(context, position, is_pep=True)
```
herrhellana, line 36: actually do we want to drop the pep flag as we categorize it in the UI?

datasets/ph/gppb_debarred/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+    record.pop("stage_violation", None)
+    record.pop("project_location", None)
+    record.pop("additional_saction", None)
+    record.pop("bo_signatory", None)
+    record.pop("temp_start_date", None)
+    record.pop("temp_end_date", None)
+    record.pop("is_removed_by_system", None)
+    record.pop("is_temp_removed_by_system", None)
+    record.pop("updated_date", None)
+    record.pop("created_date", None)
```
herrhellana, line 49: ^ this happened bc the agent skipped reading the guide; i'll try forcing it to read it and see if there is improvement esp in the yml file generation

jbothma:
this is really exciting stuff

this doesn't need to be part of this skill, but I've been meaning to see if there are nice ways to let claude fetch data the way we do - both context.fetch_* and zyte_api.fetch_* so that it can troubleshoot things more by itself. 

is that "tool use"? "MCP"? just a skill? Perhaps it needs a wrapper so it doesn't have to write python snippets to invoke the calls? It'll need access to the response body, headers, etc. It shouldn't see the zyte api key but it can assume it's configured in the environment.

But maybe we should make a separate ticket for this.

I'll review the detail properly tomorrow

herrhellana:
ok sounds good! those are skills for now but i want to play around more, and if we keep using it as skills, the MD files must be under ~300 lines

jbothma:
We should wait with my fetch_ idea until we understand the risk of prompt injection

.claude/skills/crawler-pep/SKILL.md
```diff
@@ -0,0 +1,173 @@
...
+**The `categorisation` result must be passed to `make_occupancy()`.**
+
+### Critical rules
+
+- Set ALL person props (birthDate, deathDate, etc.) BEFORE calling `make_occupancy()` —
+  it reads them from the entity to determine PEP status.
+- `make_occupancy()` returns `None` if the occupancy doesn't meet PEP criteria (e.g. ended
+  too long ago). Only emit persons who have at least one valid occupancy.
+- Emit the person AFTER `make_occupancy` — it mutates `person.topics`.
+- Always pass `propagate_country=True` so the person inherits the position's country.
```
jbothma, line 103: this results in all the examples passing propagate_country=True. Firstly, that's the default, so we don't normally pass it explicitly. Secondly, there's actually a warning in there about this being deprecated. It might not have been merged at the point where this was written.

So I don't think we want to be telling authors to pass it, and we do want to tell authors to set the country on the person before passing it in.

jbothma:
> i also trimmed down the guide / skills and got rid of repetitions found in datasets/CLAUDE.md
> i didnt touch the datasets/CLAUDE.md, as it's good to still use it for human cross-reference while we're building up the vibes; but the idea of the PR is to get rid of it and migrate its content into the overarching guide + skills

these two statements are contradictory :)

I don't entirely understand why we are keeping datasets/CLAUDE.md - isn't the point to get rid of it with this PR?

Also, let's not duplicate how to write crawlers - we're already documenting that in the zavod docs. Until we split "zavod the framework" from "how opensanctions uses zavod", let's work improvements into those docs, and structure those docs so they're easy for both claude and humans to navigate and make sense of.

https://opensanctions-corp.slack.com/archives/C066UMQDNPQ/p1776441464731939

jbothma:
I've just merged https://github.com/opensanctions/opensanctions/pull/4042 

with this, `/plan` mode and the prompt 

> How should we update the crawler-pep and crawler-sanctions skills based on the content guidelines in CLAUDE.md?

claude seems to do a promising move of the content from the skill files to the documentation, but I've only glanced at it. I'm sure it could help to get rid of what remains in datasets/CLAUDE.md unless I'm missing any reason for keeping it around.

I then deployed /crawler-pep on `es_parliament` and it identified some potential improvements, but it didn't propose adding the country.

Perhaps it's a good idea to work the ideas in https://github.com/opensanctions/opensanctions/issues/3805 into the docs so that the skills can help with the alignment work. It's fine if that's a second PR after just getting what we have here so far correct and merged.

herrhellana:
> I don't entirely understand why we are keeping datasets/CLAUDE.md - isn't the point to get rid of it with this PR?

yes, of course, the point is to get rid of it, i kept it solely so we still have it somewhere on file to check new stuff against it and spot if there is something we're missing, and then delete the datasets/CLAUDE.md upon merging the new PR with skills. sorry if i did a poor job explaining it, but the idea is of course to delete it

PR #4003, opened 2026-04-15
Claude skill to fix common typechecker errors

leonhandreke:
- **Add typechecker-fixes Claude skill**
- **typechecker-fixes skill: drop two patterns, add tuple docstring note**
- **Fix mypy strict errors in cy/companies, remove from pre-commit exclusions**
- **Enable mypy for datasets/_testing**
- **Enable mypy for datasets/me**
- **Enable mypy for datasets/il**
- **Enable mypy for datasets/ve**
- **typechecker-fixes skill: mention --explicit-package-bases and .find/.findall**
- **Random type fixes that the typechecker skill did**

### DIFF
--- a/.claude/skills/typechecker-fixes/SKILL.md
+++ b/.claude/skills/typechecker-fixes/SKILL.md
@@ -12,8 +12,6 @@ Fix mypy strict-mode errors in opensanctions crawler files. Run `mypy --strict -
 
 - `zavod/zavod/helpers/html.py` — typed HTML helpers (`xpath_strings`, `xpath_elements`, `xpath_element`, `element_text`)
 - `zavod/zavod/util.py` — `Element` and `ElementOrTree` type aliases
-- `zavod/zavod/extract/zyte_api.py` — typed Zyte API helpers
-
 ## Workflow
 
 1. Run `mypy --strict <crawler.py>` to see current errors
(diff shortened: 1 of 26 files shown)


leonhandreke:
@jbothma could you look over the skill just to see if you agree with the ruleset?

leonhandreke:
Thinking about this some more, maybe we shouldn't actually merge this skill, as it's not really "production quality" and we should avoid the skills dir becoming a low-quality graveyard. We can just keep it in a branch and use it whenever we need it.

.claude/skills/typechecker-fixes/SKILL.md
```diff
@@ -0,0 +1,206 @@
...
+### 7. Import zyte_api as a module, not individual functions
+
+```python
+# Before
+from zavod.extract.zyte_api import fetch_html, fetch_json, fetch_resource
+
+# After
+from zavod.extract import zyte_api
+# then: zyte_api.fetch_html(...), zyte_api.fetch_json(...), zyte_api.fetch_resource(...)
+```
```
jbothma: this is delving into internal code style which should maybe be left to a different skill. 

I'm for standardising though. Out of interest, why is importing as module more desirable, just so it's more explicit that it's zyte?

jbothma (commented):
I like this, honestly. What were the issues making you feel it isn't prod quality?

.claude/skills/typechecker-fixes/SKILL.md
leonhandreke: Yeah, I really like seeing what is going on when reading the code. I removed it from here. What do you think about the pattern though?

jbothma: I like it for the same reason

Issue #4013, opened 2026-04-16
PEP alignment #5


herrhellana:
datasets: 
us/congress
us/navy
us/plum_book # it seems that some positions can be potentially open to non-US citizens, although most are still occupied by US citizens in practice (so i put "country")
ve/asamblea_nacional

^ the rest requires citizenship

herrhellana:
and seems like nothing to add in us/plum_book

PR #4016, opened 2026-04-16
[ve_asamblea_nacional] PEP5

herrhellana:
issue: https://github.com/opensanctions/opensanctions/issues/4013

### DIFF
--- a/datasets/ve/asamblea_nacional/crawler.py
+++ b/datasets/ve/asamblea_nacional/crawler.py
@@ -129,6 +129,7 @@ def crawl_member(
     person.id = context.make_id(member_name)
     # https://venezuela.justia.com/federales/constitucion-de-la-republica-bolivariana-de-venezuela/titulo-v/capitulo-i/
     person.add("citizenship", "ve")
+    person.add("political", party)
     context.log.debug(f"Unique ID {person.id}")
     h.apply_name(person, full=member_name, lang="esp")
 
@@ -143,7 +144,6 @@ def crawl_member(
         context, person, position, True, categorisation=categorisation
     )
     if occupancy is not None:
-        occupancy.add("politicalGroup", party)
         occupancy.add("constituency", state)
         context.emit(person)
         context.emit(position)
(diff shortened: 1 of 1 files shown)


datasets/ve/asamblea_nacional/crawler.py
```diff
@@ -156,10 +160,16 @@ def crawl_members(context: Context, page: ElementOrTree) -> None:
         if "text-diputado-slider" not in classes.split():
             continue
         member_link = el.find(".//a")
+
+        party_el, state_el = h.xpath_elements(el, ".//small")
+        party = h.xpath_string(party_el, ".//b/text()")
+        state = h.xpath_string(state_el, "./text()")
+        state = state.split("Estado: ")[1].strip()
```
leonhandreke, line 167: This will crash when party & state el's are swapped, which is good, but maybe we could make it more explicit?

nvmbrasserie (approved):

datasets/ve/asamblea_nacional/crawler.py
```diff
@@ -156,10 +161,18 @@ def crawl_members(context: Context, page: ElementOrTree) -> None:
         if "text-diputado-slider" not in classes.split():
             continue
         member_link = el.find(".//a")
+
+        party_el, state_el = h.xpath_elements(el, ".//small")
+        party = h.xpath_string(party_el, ".//b/text()")
+        state = h.xpath_string(state_el, "./text()")
+        assert "Partido" not in party, f"Unexpected party format: {party}"
+        assert "Estado:" in state, f"Unexpected state format: {state}"
```
leonhandreke, line 169: Does this make sense? Can a party not have the name "Partido" in its name? Seems rather likely.

Checking for `Estado` makes a lot of sense, so we notice if it ever changes position.

leonhandreke (approved):

datasets/ve/asamblea_nacional/crawler.py
nvmbrasserie: we'll know our cleaning is off this way:
Screenshot 2026-04-24 at 4 32 35 PM

Issue #4011, opened 2026-04-16
PEP alignment #3


herrhellana:
linked issue: https://github.com/opensanctions/opensanctions/issues/3805

datasets: 
es/mayors_councillors # municipal elections are open to other nationals
is/althingi
ky/parliament
lt/seimas

^ the rest requires citizenship

PR #4019, opened 2026-04-16
[ky_parliament] PEP3

herrhellana:
issue: https://github.com/opensanctions/opensanctions/issues/4011

### DIFF
--- a/datasets/ky/parliament/crawler.py
+++ b/datasets/ky/parliament/crawler.py
@@ -1,13 +1,13 @@
 import csv
-from normality import collapse_spaces
-from rigour.mime.types import CSV
-from typing import Dict
 import re
+from typing import Dict, Optional
+from rigour.mime.types import CSV
 
 from zavod import Context
 from zavod import helpers as h
+from zavod.util import Element
+from zavod.entity import Entity
 from zavod.stateful.positions import categorise
-from zavod.util import ElementOrTree
 
 
 HISTORICAL_DATA_CSV = "https://docs.google.com/spreadsheets/d/e/2PACX-1vRM-hl1drkQMn4wHXxHkHXHZb2TkUyIFWxGXwJ_UqZNRpH00DGWrdHk_zDrHUsZ4YCeVYYojqSMmZuX/pub?output=csv"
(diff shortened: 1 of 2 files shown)


nvmbrasserie (approved):

datasets/ky/parliament/crawler.py
```diff
@@ -38,94 +39,123 @@
...
         context.emit(position)
         context.emit(occupancy)
         return entity
+    return None
+
 
+def crawl_row(context: Context, row: Dict[str, str]) -> None:
+    start_date = row.pop("Start date", None)
+    # Some start dates are just years, so we compare the first 4 characters to the earliest term start year
+    if start_date and start_date[:4] < h.earliest_term_start(TOPICS)[:4]:
```
jbothma: no need for the substring

datasets/ky/parliament/crawler.py
```diff
@@ -38,94 +39,123 @@
...
+    if not categorisation.is_pep:
+        return
+
+    occupancy = h.make_occupancy(
+        context,
+        entity,
+        position,
+        no_end_implies_current=False,
+        period_start=start_date,
+        period_end=row.pop("End date", None),
```
jbothma: I don't think we should just replace start/end_date with period_start/end here - see the documentation below:

```
    # Former members were added to the historical Google Sheet
    # based on https://parliament.ky/members/former-members/.
    # Since many individuals in Parliament rotate through different roles,
    # some entries are duplicated to reflect multiple positions held over
    # time. Where exact dates are provided (e.g., “November 23, 2023–end of term”),
    # I used 2025 as the assumed end date. For positions without specific timeframes,
    # the overall parliamentary term (2021–2025) was used.
```

So I think you should add period start/end columns to the spreadsheet and use the appropriate attribute for the appropriate column

datasets/ky/parliament/crawler.py
```diff
@@ -137,22 +169,22 @@ def crawl(context: Context):
...
-    # time. Where exact dates are provided (e.g., “November 23, 2023–end of term”),
-    # I used 2025 as the assumed end date. For positions without specific timeframes,
-    # the overall parliamentary term (2021–2025) was used.
+    # Former members sourced from https://parliament.ky/members/former-members/ via a
+    # historical Google Sheet. Some individuals appear multiple times to reflect distinct
+    # roles held across different periods.
+    #
+    # Date precision varies: start/end dates are exact where provided; for entries with
+    # only a term reference (2021–2025), period start/end dates are used. Where a role
+    # ended at "end of term", 2025 is assumed as the end year.
```
jbothma:
```suggestion
    # ended at "end of term", e.g. “November 23, 2023–end of term”, the end of that term,
    # i.e. 2025 is assumed as the `end_date`.
```

PR #4031, opened 2026-04-17
eu_fsf finish names migration

### LINKED ISSUE #3603: Migrate eu_fsf to multi-names input

jbothma:
Fixes https://github.com/opensanctions/opensanctions/issues/3603

✅  ~Let's wait until name->alias moves are confirmed/fixed in eu_travel_bans, and eu_fsf is fully reviewed, before merging this.~ 

✅  Let's also see if the eu_fsf delta makes this disruptive. It probably shouldn't, because we don't anticipate a lot of changes, if we're reviewing consistently.

```
Delta export complete [eu_fsf] added=0 deleted=0 metric=delta_counts modified=105
```

eu_fsf most common change seems to be emitting names with language when not doing so before. Some name prop demotions.

```
Delta export complete [eu_travel_bans] added=0 deleted=0 metric=delta_counts modified=61 version=20260417131040-mlp
```

eu_travel_bans most changes seem to be good name prop demotions

- Replaces direct `apply_name` call and single-name `apply_reviewed_name_string` with one `apply_reviewed_names` for all names for one entity
- Fixes bug where we were deriving original_value as one of the strings containing a name when the name occurred exactly in the source, merely because of the ordering of original name values.

### DIFF
--- a/zavod/zavod/helpers/names.py
+++ b/zavod/zavod/helpers/names.py
@@ -417,10 +417,11 @@ def derive_original_values(original: Names, extracted: Names) -> Dict[str, str]:
         (3) If some value in original contains it, we can use that the value from original as original_value.
         Otherwise leave blank - this is best-effort only.
     """
-    original_values: set[str] = set()
+    original_values: list[str] = []
     for _prop, values in original.as_langtexts():
         for value in values:
-            original_values.add(value.text)
+            original_values.append(value.text)
+    original_values.sort()  # sort to pick deterministically regardless of input order
 
     derived_originals: dict[str, str] = {}
     for _prop, extracted_values in extracted.as_langtexts():
@@ -429,7 +430,7 @@ def derive_original_values(original: Names, extracted: Names) -> Dict[str, str]:
 
             if len(original_values) == 1:
                 # (1) If there's exactly one value in original, use that for all names.
-                derived_originals[extracted_string] = next(iter(original_values))
+                derived_originals[extracted_string] = original_values[0]
             elif extracted_string in original_values:
                 # (2) Exact match exists, no original_value needed.
                 continue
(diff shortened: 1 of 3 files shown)


jbothma:
I'm feeling 170 sanction mods, mostly name property demotions, isn't disruptive.

pudo:
I would like to validate your feelings. They are non disruptive.

zavod/zavod/tests/helpers/names/test_derive_originals.py
```diff
@@ -27,28 +27,15 @@ def test_derive_original_values_exact_match():
...
 
     result = derive_original_values(original, extracted)
 
-    # John Brandon Doe isn't contained exactly. We're not getting more fancy with this.
-    assert result == {
-        "Brandon Doe": "John Doe; Brandon Doe",
-    }
-
-
-def test_derive_original_values_first_match_wins():
```
leonhandreke, line 41: That you're removing this test makes me question whether `original_value` selection is still stable? Should there be a `sorted` in `next(iter(original_values))`?

zavod/zavod/helpers/names.py
```diff
@@ -417,10 +417,10 @@ def derive_original_values(original: Names, extracted: Names) -> Dict[str, str]:
         (3) If some value in original contains it, we can use that the value from original as original_value.
         Otherwise leave blank - this is best-effort only.
     """
-    original_values: List[str] = []
+    original_values: set[str] = set()
```
leonhandreke: There was a dumb comment here about how this could be a two layered list comprehension but then I realized it was dumb so I deleted it.

zavod/zavod/tests/helpers/names/test_derive_originals.py
jbothma: nice catch, thank you! Now I remember why it was a list, although changing input order might still have caused surprise changes in derivations.

I changed back to a list so it's sorted once. per call.

leonhandreke (approved):
LGTM!

PR #4033, opened 2026-04-17
fix program dataset values

nvmbrasserie:
The dataset values were incorrect; one with two values and the other referencing the program key instead of the dataset slug.

### DIFF
--- a/meta/programs/UN-SC1737.yml
+++ b/meta/programs/UN-SC1737.yml
@@ -9,4 +9,4 @@ summary: UN Security Council resolution 1737 (2006) imposes sanctions on Iran fo
 issuer: zz_unsc
 dataset: un_sc_sanctions
 target_territories:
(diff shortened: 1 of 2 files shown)


meta/programs/UN-SC1737.yml
```diff
@@ -7,6 +7,6 @@ summary: UN Security Council resolution 1737 (2006) imposes sanctions on Iran fo
   a ban the supply of nuclear-related technology and materials and imposes assets
   freeze on key individuals and companies related to the enrichment programme.
 issuer: zz_unsc
-dataset: un_sc_sanctions,fr_tresor_gels_avoir
+dataset: un_sc_sanctions
 target_territories:
-- ir
+  - ir
```
nvmbrasserie:
```suggestion
- ir
```

Issue #2996, opened 2025-10-27
[us_cftc_enforcement_actions] Additional docs on `status` field


leonhandreke:
https://review.opensanctions.org/review/dataset/us_cftc_enforcement_actions/9133-25

<img width="1223" height="566" alt="Image" src="https://github.com/user-attachments/assets/7caa3ed3-8949-4542-ad73-3ede04e93ed6" />

The press release announces: "[the court] entered a consent order [...]". So what's the `status` here? "Supplemental consent order" is tempting, but I think the right answer is "Settled"? But IANAL. I feel like if processing these is a job that we should all be able to do, we need a bit more documentation on the workings of the US judicial system. I'm not sure where this documentation should be, but right now I don't feel I can process these with confidence.

Assigning to @jbothma so that he can take a look when he's back from vacation. Feel free to just give some advice to me for how to handle this in the future, unassign yourself and put this bug to the large pile of ideas :)

jbothma:
@pudo I think we should just drop the status field because of the risk/harm if we get it wrong - do you reckon it's useful enough to put more effort in?

I had a quick go at addressing this in http://github.com/opensanctions/opensanctions/pull/4048

I don't like the direction this is going - there's so much variation in how this is written, and "Resolved" for "The court found them guilty a couple of weeks ago" doesn't feel like a status that really captures the severity of the outcome at a glance. And the idea with capturing status was to capture whether charges are abandoned, settled, filed, or ruled guilty, at a glance.

pudo:
@jbothma Honestly I like getting rid of it if it's too chaotic. Nobody will clean this up in the review system, so it ends up a liability.

pudo:
What's the status here?

jbothma:
it's waiting for the field to be done. 

I think that's either leaving the field in there for existing reviews but telling prompt not to use it, or a backward incompatible model change implying re-review or some kind of migration.

PR #4048, opened 2026-04-20
Try expanding statuses and status docs

jbothma:
Not to be merged as it stands - it's just for discussion in  https://github.com/opensanctions/opensanctions/issues/2996


jbothma:
I think ideally the prompt should be authoritative as documentation. The way the prompt is built right now isn't very readable.

PR #4107, opened 2026-04-29
[hr_public_officials] PEP alignment

herrhellana:
issue https://github.com/opensanctions/opensanctions/issues/3805

### DIFF
--- a/datasets/hr/public_officials/crawler.py
+++ b/datasets/hr/public_officials/crawler.py
@@ -83,9 +83,10 @@ def make_affiliation_entities(
         return []
 
     start_date = data.pop("position_start_date")
-    if start_date < h.earliest_term_start(TOPICS):
+    parsed_dates = h.extract_date(context.dataset, start_date)
+    if parsed_dates[0] < h.earliest_term_start(TOPICS):
         context.log.info(
-            f"Skipping row with start date {start_date} outside coverage window"
+            f"Skipping row with start date {parsed_dates[0]} outside coverage window"
         )
         return []
 
--- a/datasets/hr/public_officials/crawler.py
+++ b/datasets/hr/public_officials/crawler.py
@@ -52,6 +52,8 @@ FIELDS = [
     "secondary_position_end_date",
 ]
 
+TOPICS = ["gov.soe"]
+
 
 def make_position_name(data: Dict[str, str]) -> Optional[str]:
     title = data.pop("position")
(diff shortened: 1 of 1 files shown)


herrhellana:
note that we're fetching less PEPs in `hr_public_officials` now that we're using the earliest start date helper on the raw source dates (1360 count for Person). However, if we wrap the raw source dates into a date format lookup, the count looks fine (2667 for Person). 
```
start_date = data.pop("position_start_date")
parsed_dates = h.extract_date(context.dataset, start_date)
if parsed_dates[0] < h.earliest_term_start(TOPICS):
   ...
```

nvmbrasserie:
we agreed to do one PR per issue, didn't we?

herrhellana:
ok let me cherry pick, sorry forgot about it.

datasets/hr/public_officials/crawler.py
```diff
@@ -84,7 +86,7 @@ def make_affiliation_entities(
     end_date = data.pop("position_end_date")
     context.audit_data(data)
 
-    position = h.make_position(context, position_name, country="HR")
+    position = h.make_position(context, position_name, topics=TOPICS, country="HR")
```
jbothma: not all these positions are in state owned enterprises. Some are in state organisations which are not enterprises. The position categorisation UI exists very much for cases like this, so best to leave position topics to it.

Image

jbothma (changes requested):

datasets/hr/public_officials/crawler.py
herrhellana: understood, rolled it back!

jbothma (approved):
maybe just give it a sanity check that the delta exporter says add/mod/del is all zero before merging, since it'd be a fair amount of MODs otherwise.

herrhellana:
[hr_public_officials] added=23 dataset=hr_public_officials deleted=22 metric=delta_counts modified=0 version=20260506120000-dqy

some folks + their positions got added and removed (out of ~3k)

i wonder why we dont get any statements modified if we're adding a country prop though? shouldnt we see # MOD == # new country props?

jbothma:
I'm happy you asked! it's not a mod because make_occupancy already adds it (although we want to remove that side-effect, hence this task!)

https://github.com/opensanctions/opensanctions/commit/f63572e1e359c891ba825730b21265f2e8507ee2

But getting the deltas working was important for other PRs making more changes anyway, and now it's easy to see that we really aren't modifying anything we don't mean to - no surprises.

jbothma (approved):

Issue #4010, opened 2026-04-16
PEP alignment #2


herrhellana:
linked issue: https://github.com/opensanctions/opensanctions/issues/3805

datasets: 
co/funcion_publica  # various PEPs, incl academics or clinical stuff that don't require citizenship
co/join_dots
de/abgeordnetenwatch
ee/riigikogu

^ the rest requires citizenship

brrttwrks:
@herrhellana, I don't think the value must be citizenship, but any affiliated country. See code here where `zavod` checks that any country type property exists:

```python
            if country not in person.get_type_values(registry.country, matchable=True):
                _tmp_warn_propagate_country(context)
                person.add("country", country, origin=ORIGIN_INFERRED)
```

 So it could be the country in which the position occurs, so that could be using the property `country`. For example, a `country` value for `co/funcion_publica` could be `co`, since the position occurs in `co`. If the property in question was specifically `citizenship`, then that would be  `citizenship` property. You can do both and if both exist, probably should do both. Use the country-specific properties where we are certain. At the very least, in these, we know in which country the position occurs or is affiliated, so we can use `country`.

Did I understand the issue properly?

brrttwrks:
In any case, this is checking to see if the position entity country exists in the person entity and if it doesn't it creates and creates a warning and by default (propagate_country = True) assigns it.

herrhellana:
ok i just found a little bug in the Colombian crawler that was still firing this warning despite Position and Person having country prop. There was an additional function using `make_occupancy()` relying on another instance of Person entity which I overlooked. Fixed that, and you're right that the warning then should not fire.

PR #4108, opened 2026-04-29
[de_abgeordnetenwatch] PEP alignment 2

herrhellana:
issue: https://github.com/opensanctions/opensanctions/issues/4010

### DIFF
--- a/datasets/de/abgeordnetenwatch/crawler.py
+++ b/datasets/de/abgeordnetenwatch/crawler.py
@@ -63,10 +63,7 @@ def crawl_mandate(context: Context, mandate: Dict[str, Any]) -> None:
     )
     person.add("political", politician_detail["party"]["label"])
     person.add("sourceUrl", politician_detail.pop("abgeordnetenwatch_url"))
-    person.add(
-        "biography",
-        f"Past parties: {politician_detail.pop('party_past')}, education: {politician_detail.pop('education')}, residence: {politician_detail.pop('residence')}, occupation: {politician_detail.pop('occupation')}",
-    )
+    person.add("education", politician_detail.pop("education"))
 
     # Occupancy
     categorisation = categorise(context, position)
@@ -116,7 +113,6 @@ def crawl_mandate(context: Context, mandate: Dict[str, Any]) -> None:
             "sex",
             "party",
             "party_past",
-            "education",
             "residence",
             "occupation",
             "statistic_questions",
--- a/datasets/de/abgeordnetenwatch/crawler.py
+++ b/datasets/de/abgeordnetenwatch/crawler.py
@@ -69,8 +69,6 @@ def crawl_mandate(context: Context, mandate: Dict[str, Any]) -> None:
     )
 
     # Occupancy
-    mandate_start_date = mandate.pop("start_date") or period_detail["start_date_period"]
-    mandate_end_date = mandate.pop("end_date") or period_detail["end_date_period"]
     categorisation = categorise(context, position)
     if not categorisation.is_pep:
         return
@@ -79,8 +77,10 @@ def crawl_mandate(context: Context, mandate: Dict[str, Any]) -> None:
         person,
         position,
         True,
-        start_date=mandate_start_date,
-        end_date=mandate_end_date,
+        start_date=period_detail["start_date_period"],
+        period_start=mandate.pop("start_date"),
+        end_date=period_detail["end_date_period"],
+        period_end=mandate.pop("end_date"),
         categorisation=categorisation,
     )
     if occupancy:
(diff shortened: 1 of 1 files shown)


herrhellana:
`Delta export complete          [de_abgeordnetenwatch] added=4854 dataset=de_abgeordnetenwatch deleted=0 metric=delta_counts modified=0 version=20260429124106-iiv`

datasets/de/abgeordnetenwatch/crawler.py
```diff
@@ -74,8 +77,10 @@ def crawl_mandate(context: Context, mandate: Dict[str, Any]) -> None:
         person,
         position,
         True,
-        start_date=mandate_start_date,
-        end_date=mandate_end_date,
+        start_date=period_detail["start_date_period"],
+        period_start=mandate.pop("start_date"),
+        end_date=period_detail["end_date_period"],
+        period_end=mandate.pop("end_date"),
```
jbothma: it looks like the properties are flipped: period_start is for the fixed term of the office, while start_date is the time this human held that office.

Is this unclear in the documentation of the properties? Would you propose an improvement so that the next person gets it right straight away?

@pudo in fact the documentation really seems to say periodStart is when the human occupied the position - that's quite a change from how the property has been used in the past. Did I understand the new property incorrectly or is the documentation confusing? https://github.com/opensanctions/followthemoney/blob/main/followthemoney/schema/Occupancy.yaml#L56

herrhellana: the documentation is really confusing as i mentioned in another PR. ill update the dict description of these props once i get the final go from @pudo

herrhellana: ok from friedrich on slack: 

> So the US house of reps started it's 118th session on January 3, 2023 (periodStart). A member of it was George Anthony Devolder Santos (startDate), who had been elected on November 8, 2022 (electionDate ). He then was indicted for several counts (no fun in life without wire fraud) and expelled from Congress on December 1st, 2023 (endDate ). The 118th House of Reps, however - now devoid of any finanical criminals whatsoever - continued it's 118th session until January 3, 2025 (periodEnd ).

datasets/de/abgeordnetenwatch/crawler.py
```diff
@@ -61,11 +61,14 @@ def crawl_mandate(context: Context, mandate: Dict[str, Any]) -> None:
         last_name=politician_detail.pop("last_name"),
         maiden_name=politician_detail.pop("birth_name"),
     )
+    person.add("political", politician_detail["party"]["label"])
     person.add("sourceUrl", politician_detail.pop("abgeordnetenwatch_url"))
+    person.add(
+        "biography",
+        f"Past parties: {politician_detail.pop('party_past')}, education: {politician_detail.pop('education')}, residence: {politician_detail.pop('residence')}, occupation: {politician_detail.pop('occupation')}",
+    )
```
jbothma: nice idea, but it's high time we documented that we prefer not building profile strings from structured data like this https://github.com/opensanctions/opensanctions/pull/4150/changes

we actually do have an education prop. I think you can leave residence out just because as a value in the `address` prop, it's a bit weak, and I think there's enough other deduplicating information birth date, last 5+ years of political posts

datasets/de/abgeordnetenwatch/crawler.py
```diff
@@ -74,8 +77,10 @@ def crawl_mandate(context: Context, mandate: Dict[str, Any]) -> None:
         person,
         position,
         True,
-        start_date=mandate_start_date,
-        end_date=mandate_end_date,
+        period_start=period_detail["start_date_period"],
+        start_date=mandate.pop("start_date"),
+        period_end=period_detail["end_date_period"],
+        end_date=mandate.pop("end_date"),
```
jbothma, line 80: Image

But (and it's a big but), let's make really sure the delta on this is sane. How many entities do you expect to be added, modified, or removed by this PR, how does that compare to the numbers you actually get?

jbothma: Meme in reference to

```diff
-    mandate_start_date = mandate.pop("start_date") or period_detail["start_date_period"]
-    mandate_end_date = mandate.pop("end_date") or period_detail["end_date_period"]
```

datasets/de/abgeordnetenwatch/crawler.py
herrhellana: ah damn yes, we've been through this already, i forgot. let me rewrite it.

herrhellana:
Delta export complete          [de_abgeordnetenwatch] added=2420 dataset=de_abgeordnetenwatch deleted=2420 metric=delta_counts modified=2418 version=20260506121306-cor

looks legit. all Persons got modified bc of new props like education + all Occupancies got deleted and added bc of the id change for Occupancy that depends on start/end dates i'm editing here.

jbothma (approved):

jbothma:
Looks like for PEPs, this order of magnitude of MOD adding [a property that wouldn't be used for determining relevance](https://www.opensanctions.org/docs/monitoring/)  is ok

PR #4110, opened 2026-04-29
[ee_riigikogu] PEP alignment 2

herrhellana:
issue: https://github.com/opensanctions/opensanctions/issues/4010

### DIFF
--- a/datasets/ee/riigikogu/crawler.py
+++ b/datasets/ee/riigikogu/crawler.py
@@ -6,10 +6,10 @@ def get_members_urls(context: Context) -> list:
     main_website = context.fetch_html(context.data_url)
 
     # this XPath corresponds to the a tags wit the links for a member page
-    xpath_to_a_tags = "//*[@id='main']/section/div[2]/div/ul/li/div/div/ul/li[1]/h3/a"
-    return [
-        a_tag.get("href") for a_tag in h.xpath_elements(main_website, xpath_to_a_tags)
-    ]
+    xpath_to_a_tags = (
+        "//*[@id='main']/section/div[2]/div/ul/li/div/div/ul/li[1]/h3/a/@href"
+    )
+    return h.xpath_strings(main_website, xpath_to_a_tags)
 
 
 def crawl_item(member_url: str, context: Context):
@@ -20,16 +20,29 @@ def crawl_item(member_url: str, context: Context):
     except IndexError:
         context.log.info("Couldn't find name. Skipping person.", url=member_url)
         return
-    bio = " ".join(
-        h.xpath_strings(
-            member_page_html,
-            '//*[@class="col-xs-6 profile-desc bg-white content"]/p[text()][1]/text()',
+    try:
+        bio = " ".join(
+            h.xpath_strings(
+                member_page_html,
+                '//*[@class="col-xs-6 profile-desc bg-white content"]/p[text()][1]/text()',
+            )
+        ).strip()
+    except ValueError:
+        context.log.warning(
+            "Couldn't find biography. Skipping biography.", url=member_url
         )
-    ).strip()
-    electoral_district = h.xpath_string(
-        member_page_html,
-        '//*[@class="col-xs-6 profile-desc bg-white content"]//a[contains(@href, "searchByConstituency")]/text()',
-    )
+        return
+    try:
+        electoral_district = h.xpath_string(
+            member_page_html,
+            '//*[@class="col-xs-6 profile-desc bg-white content"]//a[contains(@href, "searchByConstituency")]/text()',
+        )
+    except ValueError:
+        context.log.warning(
+            "Couldn't find electoral district. Skipping electoral district.",
+            url=member_url,
+        )
+        return
 
     entity = context.make("Person")
     entity.id = context.make_id(name)
@@ -38,14 +51,9 @@ def crawl_item(member_url: str, context: Context):
     context.log.info(f"Full name {name} unique id {entity.id}")
     h.apply_name(entity, full=name)
 
-    try:
-        # base the xpath on the email icon to extract the email
-        email = h.xpath_string(
-            member_page_html, '//*[@class="icon icon-mail"]/../text()'
-        )
-        entity.add("email", email)
-    except IndexError:
-        context.log.warning("Couldn't find e-mail")
+    # base the xpath on the email icon to extract the email
+    email = h.xpath_string(member_page_html, '//*[@class="icon icon-mail"]/../text()')
+    entity.add("email", email)
 
     position = h.make_position(
         context,
--- a/datasets/ee/riigikogu/crawler.py
+++ b/datasets/ee/riigikogu/crawler.py
@@ -15,34 +15,16 @@ def get_members_urls(context: Context) -> list:
 def crawl_item(member_url: str, context: Context):
     member_page_html = context.fetch_html(member_url)
 
-    try:
-        name = h.xpath_string(member_page_html, '//*[@id="main"]/header/div/h1/text()')
-    except IndexError:
-        context.log.info("Couldn't find name. Skipping person.", url=member_url)
-        return
-    try:
-        bio = " ".join(
-            h.xpath_strings(
-                member_page_html,
-                '//*[@class="col-xs-6 profile-desc bg-white content"]/p[text()][1]/text()',
-            )
-        ).strip()
-    except ValueError:
-        context.log.warning(
-            "Couldn't find biography. Skipping biography.", url=member_url
-        )
-        return
-    try:
-        electoral_district = h.xpath_string(
-            member_page_html,
-            '//*[@class="col-xs-6 profile-desc bg-white content"]//a[contains(@href, "searchByConstituency")]/text()',
-        )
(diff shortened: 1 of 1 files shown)


herrhellana:
`Delta export complete          [ee_riigikogu] added=203 dataset=ee_riigikogu deleted=0 metric=delta_counts modified=0 version=20260429124215-hlf`

datasets/ee/riigikogu/crawler.py
```diff
@@ -7,31 +7,44 @@ def get_members_urls(context: Context) -> list:
...
     try:
         # the phone number apparently is the first contact in the list
         # but to make sure we correctly get it, we based our xpath on the icon
-        phone_els = member_page_html.xpath('//*[@class="icon icon-tel"]/../text()')
-        phone_number = phone_els[0].replace(" ", "")
-        entity.add("phone", phone_number)
+        phone_number = h.xpath_string(
+            member_page_html, '//*[@class="icon icon-tel"]/../text()'
+        )
+        entity.add("phone", phone_number.replace(" ", ""))
```
pudo: Let's actually remove the phone number unless we have a specific reason to keep it. It raises some concern around OS being used as a marketing database, which we don't want. Emails are a little less sensitive and work as identifiers/matching criteria.

pudo (dismissed):
Overall looks innocent. I assume the delta is because you ran it in dry-run mode?

datasets/ee/riigikogu/crawler.py
jbothma: this would be a good addition to https://zavod.opensanctions.org/best_practices/priorities/#:~:text=Won%27t%20%2D%20don%27t%20extract
The name of the page could improve.

herrhellana: fixed: https://github.com/opensanctions/opensanctions/pull/4141

datasets/ee/riigikogu/crawler.py
```diff
@@ -7,44 +7,42 @@ def get_members_urls(context: Context) -> list:
...
-        if name not in [
-            "Hele Everaus",
-            "Kaja Kallas",
-            "Mart Võrklaev",
-            "Toomas Kivimägi",
-        ]:
-            context.log.warning("Couldn't find phone number for", name=name)
-
     try:
         # we do the same as the phone number
```
goose-life: if you nuke the whole phone number block this comment points at nothing now

herrhellana: good catch thank you!

datasets/ee/riigikogu/crawler.py
```diff
@@ -7,44 +7,42 @@ def get_members_urls(context: Context) -> list:
 
     # this XPath corresponds to the a tags wit the links for a member page
     xpath_to_a_tags = "//*[@id='main']/section/div[2]/div/ul/li/div/div/ul/li[1]/h3/a"
-
-    return [a_tag.get("href") for a_tag in main_website.xpath(xpath_to_a_tags)]
+    return [
+        a_tag.get("href") for a_tag in h.xpath_elements(main_website, xpath_to_a_tags)
+    ]
```
jbothma: These changes have now made the wrapping awkward. It'd be a ton cleaner, but would it work, to skip the list comprehension and return h.xpath_strings(...@href)?

datasets/ee/riigikogu/crawler.py
```diff
@@ -7,44 +7,42 @@ def get_members_urls(context: Context) -> list:
...
-            context.log.warning("Couldn't find phone number for", name=name)
-
-    try:
-        # we do the same as the phone number
-        email = member_page_html.xpath('//*[@class="icon icon-mail"]/../text()')[0]
+        # base the xpath on the email icon to extract the email
+        email = h.xpath_string(
+            member_page_html, '//*[@class="icon icon-mail"]/../text()'
+        )
         entity.add("email", email)
```
jbothma: with h.xpath_string it'll raise a different exception than IndexError. But since this isn't a sanctions dataset, it's ok to just remove the try...except and "let it crash"

jbothma (approved):
a couple of style nit picks, otherwise ship it.

datasets/ee/riigikogu/crawler.py
herrhellana: yep!

herrhellana:
Delta export complete          [ee_riigikogu] added=0 dataset=ee_riigikogu deleted=4 metric=delta_counts modified=198 version=20260506125733-bve

two folks + their positions got deleted (Rene Kokk and Andrus Seeme) + expected mods

datasets/ee/riigikogu/crawler.py
```diff
@@ -6,48 +6,54 @@ def get_members_urls(context: Context) -> list:
...
+        electoral_district = h.xpath_string(
+            member_page_html,
+            '//*[@class="col-xs-6 profile-desc bg-white content"]//a[contains(@href, "searchByConstituency")]/text()',
+        )
+    except ValueError:
+        context.log.warning(
+            "Couldn't find electoral district. Skipping electoral district.",
+            url=member_url,
+        )
+        return
```
jbothma: the try catch isn't really a pattern we want to continue. Part of https://github.com/opensanctions/opensanctions/pull/4110#discussion_r3188896634 was really about it being ok for this crawler to error if we can't find some data. These properties are really not particularly important. Definitely not important enough for this PEP to drop out of our database until we fix it, which is what this logic does. Another thing to bear in mind - Exceptions are for exceptional circumstances, not for "control flow" (aka normal logic). Here we're doing normal logic - did we find this thing or not. And h.xpath_element raises so that you can easily detect an error and let it crash (with an already clear message why)

So it's ok if we're adding this, but "let it crash" (i.e. raise an uncaught exception) if it doesn't find the field, the we'll fix the crawler the next day and run it again.

The other thing here is the wrapping - especially the bio - two nested function calls that are being broken into bits is not fun to read. you should consider that very smelly. Just assign the result of the first function call to a variable, then use that in the second function call, and it's beautiful code.

datasets/ee/riigikogu/crawler.py
```diff
@@ -6,48 +6,54 @@ def get_members_urls(context: Context) -> list:
...
-        name = member_page_html.xpath('//*[@id="main"]/header/div/h1/text()')[0]
+        name = h.xpath_string(member_page_html, '//*[@id="main"]/header/div/h1/text()')
     except IndexError:
         context.log.info("Couldn't find name. Skipping person.", url=member_url)
         return
+    try:
+        bio = " ".join(
+            h.xpath_strings(
+                member_page_html,
+                '//*[@class="col-xs-6 profile-desc bg-white content"]/p[text()][1]/text()',
```
jbothma: this index-based selector `p[text()][1]` means "the first p tag which contains text".

This feels a bit brittle to me and I'm back and forth about saying "Person:biography" isn't really worth it. I guess the worst case here is that we lose a bit of text or gain a bit of nonsense, so maybe it's ok. What saves it is that we have two semantic classes: profile-desc and content.

What you probably want to do is replace `@class="col-xs-6 profile-desc bg-white content"` with `contains(@class, 'profile-des') and contains(@class, 'content') so that stylistic changes like changing col-xs-6 to col-xs-5 doesn't break it.

perhaps just note that you should feel uncomfortable as soon as you're using index-based lookup or a lookup without any semantic classes.

jbothma (commented):
Let's clean the try...catches up before merging.

@pudo this will need an approve from you before github will allow a merge - shout if you still have reservations.

datasets/ee/riigikogu/crawler.py
herrhellana: yeah, if we scrape not just the first p tag, then we get a bunch of gibberish like:

`'I studied to become a teacher. I worked in school for more than ten years, but I have spent most of my working life in civil service. I have been Mayor of Võhma, Managing Director of the Association of Estonian Cities, member of three Riigikogus, Minister of Social Affairs, and Minister of Public Administration. I have five grandchildren. I want to stand for a fair and solidary state. I have the experience and knowledge to have my say in the areas of the national social and health policy, regional policy, and the state reform. I definitely stand for the interests of my electoral district – Järva county and Viljandi county. \n                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              \n                                                          \n\n                                                            Electoral district:'`

which is harder to clean up. i decided to go with `p[text()][1]` because the bio is consistently stored in this first  tag, although, yes, it can change in the future... or is brittle in general... so i don't know, should we leave it as is or remove it? or maybe there is a better solution? 

def makes sense re: kicking the index and using `contains()`! didn't think of it immediately.

datasets/ee/riigikogu/crawler.py
herrhellana: ah yes ok, i.e. use `h.xpath_strings()` for optional fields instead of `h.xpath_string()` which expects 1 element and raises otherwise, and use `h.xpath_string()` for names

and yes, the bio bit makes sense, too. sorry, i feel like we've been over this a few times now, but i still recycle my old logic sometimes.

jbothma (approved):

PR #4111, opened 2026-04-29
[co_funcion_publica] Pep alignment

herrhellana:
issue https://github.com/opensanctions/opensanctions/issues/4010

### DIFF
--- a/datasets/co/funcion_publica/crawler.py
+++ b/datasets/co/funcion_publica/crawler.py
@@ -117,8 +117,10 @@ def crawl_sheet_row(context: Context, row: Dict[str, str]):
         # but that's not the case. I have an issue open with their support.
         # In the meantime:
         status=OccupancyStatus.UNKNOWN,
-        start_date=row.pop("FECHA_VINCULACION"),
-        end_date=row.pop("FECHA_DESVINCULACION"),
+        # "date of appointment"
+        period_start=row.pop("FECHA_VINCULACION"),
+        # "date of termination"
+        period_end=row.pop("FECHA_DESVINCULACION"),
         categorisation=categorisation,
     )
     context.emit(person)
--- a/datasets/co/funcion_publica/crawler.py
+++ b/datasets/co/funcion_publica/crawler.py
@@ -118,9 +118,9 @@ def crawl_sheet_row(context: Context, row: Dict[str, str]):
         # In the meantime:
         status=OccupancyStatus.UNKNOWN,
         # "date of appointment"
-        period_start=row.pop("FECHA_VINCULACION"),
+        start_date=row.pop("FECHA_VINCULACION"),
         # "date of termination"
-        period_end=row.pop("FECHA_DESVINCULACION"),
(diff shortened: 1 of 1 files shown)


herrhellana:
`Delta export complete          [co_funcion_publica] added=48896 dataset=co_funcion_publica deleted=0 metric=delta_counts modified=0 version=20260429131311-krd`

alas

datasets/co/funcion_publica/crawler.py
```diff
@@ -117,8 +117,10 @@ def crawl_sheet_row(context: Context, row: Dict[str, str]):
         # but that's not the case. I have an issue open with their support.
         # In the meantime:
         status=OccupancyStatus.UNKNOWN,
-        start_date=row.pop("FECHA_VINCULACION"),
-        end_date=row.pop("FECHA_DESVINCULACION"),
+        # "date of appointment"
+        period_start=row.pop("FECHA_VINCULACION"),
+        # "date of termination"
+        period_end=row.pop("FECHA_DESVINCULACION"),
```
jbothma: why period_start and period_end? the translation sounds very much like the field is about when this person entered the position and when they left it, rather than the position having some fixed calendar term, and the person potentially only occupied part of it.

jbothma (commented):
> added=48896 

Looks like your zavod isn't backfilling existing data hashes to determine which entities are new and which were modified.

Perhaps you need to set up zavod to use GoogleCloudBackend. These docs have recently improved https://www.notion.so/opensanctions/Engineer-onboarding-20f1d6e948e04ec1a1df864ac6693d8f?source=copy_link#3039721113008011b5abc85c00770e3a

datasets/co/funcion_publica/crawler.py
herrhellana: yes, this is because the data dict confuses me, and i thought period_start is about the office dates and not mandate dates, ie flipped the logic. ill fix it now.

jbothma (approved):
we expect the delta to be all zeros here, right?

This would be a big fat delta and cause changes on a lot of spanish names if we missed something - maybe worth just getting your delta exporter happy and confirm that before merging

herrhellana:
Delta export complete          [co_funcion_publica] added=3771 dataset=co_funcion_publica deleted=16 metric=delta_counts modified=39 version=20260506112907-ott

added people + associated positions are mostly new ppl that dont exist in the database yet; one person got deleted, the rest is positions, and mods im not quite sure what has changed there tbh bc data looks similar to what we have in prod

jbothma (approved):
yo could use the statement diffing tools to learn more about the deltas.

given the changes in the PR, if this was going horribly wrong, we'd see a delta on basically everyone here - 18,534 people.

We're just seeing a tiny change - that's probably in part because of differences in the positions database in production compared to your local one. No need to load the prod data locally - we can explain the changes.

Ship it.

herrhellana:
ok good, @brrttwrks  wanted to organize a little workshop on deltas next week -- would be nice to talk about the diffing tools, too, in this context.

PR #4123, opened 2026-04-30
[lu_bourgmestres] Luxembourg bourgmestres PEPs w/ Agent Skills

herrhellana:
Agent skills issue: https://github.com/opensanctions/opensanctions/pull/3996

[lu_bourgmestres] claude's eval of another agent's thinking process for/crawler-pep https://github.com/opensanctions/crawler-planning/issues/652 (skill runtime: 4m 30s):

issue: https://github.com/opensanctions/crawler-planning/issues/652


datasets/lu/bourgmestres/lu_bourgmestres.yml
```diff
@@ -0,0 +1,59 @@
...
+summary: >-
+  Members of the collège échevinal (bourgmestres and échevins) of Luxembourg
+  municipalities since the 2023 communal elections.
+
+description: |
+  The dataset lists all bourgmestres (mayors) and échevins (aldermen) of
+  Luxembourg's 100 municipalities for the 2023–2029 electoral period, including
+  those who have since left office.
+
+  Source: [data.public.lu](https://data.public.lu/en/datasets/college-des-bourgmestre-et-echevins-depuis-les-elections-communales-2023/)
```
leonhandreke: I think the source is in the `url` prop already, so maybe no need to put it here?

datasets/lu/bourgmestres/lu_bourgmestres.yml
```diff
@@ -0,0 +1,59 @@
+title: Luxembourg Bourgmestres and Échevins
```
leonhandreke: I don't know what our current policy is around this, but I feel something like "municipal leaders", "municipal executive" or "mayors" (even though technically it's not just mayors) would be more speaking.

datasets/lu/bourgmestres/lu_bourgmestres.yml
```diff
@@ -0,0 +1,59 @@
...
+entry_point: crawler.py
+prefix: lu-bourg
+coverage:
+  frequency: monthly
+  start: 2026-04-14
+load_statements: true
+ci_test: true
+summary: >-
+  Members of the collège échevinal (bourgmestres and échevins) of Luxembourg
+  municipalities since the 2023 communal elections.
```
leonhandreke, line 11: I'm not 100% sure on this, but would it make sense to just set the `coverage` start to the 2023 election date?

datasets/lu/bourgmestres/crawler.py
```diff
@@ -0,0 +1,69 @@
...
+    )
+    if occupancy is not None:
+        context.emit(occupancy)
+        context.emit(person)
+
+    context.audit_data(record)
+
+
+def crawl(context: Context) -> None:
+    data = context.fetch_json(context.data_url, cache_days=1)
```
leonhandreke: I think we want some sort of mechanism here to alert us about a new election having happened or new data being available (after 2029)

datasets/lu/bourgmestres/crawler.py
```diff
@@ -0,0 +1,69 @@
...
+    start_date = record.pop("COAC_START_DATE")
+    end_date = record.pop("COAC_END_DATE")
+
+    assert man_label in POSITION_LABELS, f"Unknown position: {man_label!r}"
+
+    pos_key = f"{man_label}-{com_code}"
+    if pos_key not in positions:
+        position = h.make_position(
+            context,
+            name=f"{man_label} de {com_label}",
```
leonhandreke: We generally try to do positions in English if we can.

leonhandreke (changes requested):

datasets/lu/bourgmestres/lu_bourgmestres.yml
pudo: No, I think coverage start should always reflect inclusion in `default`.

herrhellana: coverage how-to is not included in the skill, note to myself to update it.

leonhandreke: Done!

datasets/lu/bourgmestres/crawler.py
leonhandreke: Done.

datasets/lu/bourgmestres/crawler.py
leonhandreke: Done.!

leonhandreke (approved):

leonhandreke:
Hi, I pushed a commit that uses their JSON API to discover their datasets and verify that no new election happened. It also gets the most recent JSON, since it seems they do update it (the current one is from April).

For the future: I find it very helpful in code reviews if you add a little reply to my comments describing what you did do address it (just "Done" is fine), and resolve it if you're sure it's resolved. If you're unsure that it's resolved and would like me to take a look, then just leave it open. Merci!

herrhellana:
this looks really nice, thank you!

PR #4124, opened 2026-04-30
[ph_gppb_debarred] Philippines GPPB blacklisting report w/ Agent Skills

herrhellana:
Agent skills issue: https://github.com/opensanctions/opensanctions/pull/3996

[ph_gppb_debarred] claude's eval of another agent's thinking process for/crawler-sanctions https://github.com/opensanctions/crawler-planning/issues/654 (skill runtime: 9m 1s):

issue: https://github.com/opensanctions/crawler-planning/issues/654


datasets/ph/gppb_debarred/ph_gppb_debarred.yml
```diff
@@ -0,0 +1,65 @@
...
+  The Consolidated Blacklisting Report (CBR) lists entities that have been
+  blacklisted from participating in Philippine government procurement under
+  Republic Act 9184 (Government Procurement Reform Act). Entities are
+  blacklisted for offenses such as failure to comply with contractual
+  obligations, unsatisfactory progress, or submission of falsified documents.
+
+  Blacklisting periods are typically one year (first offense) or two years
+  (second offense).
+
+  Source: [Online Blacklisting Portal](https://onlineblacklistingportal.gppb.gov.ph/)
```
leonhandreke: We should list the source in the URL. I think we typically only name a source here if we're quoting very directly.

datasets/ph/gppb_debarred/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+    program_name = SANCTION_LABELS[sanction_type]
+    reason = format_offenses(offenses)
+
+    sanction = h.make_sanction(
+        context,
+        entity,
+        key=start_date,
+        program_name=program_name,
+        source_program_key=sanction_type,
+        program_key=h.lookup_sanction_program_key(context, sanction_type),
```
leonhandreke: Not sure we want a sanction program here? If removing, also clean up metadata and the `SANCTION_LABELS` check as it will be superfluous.

datasets/ph/gppb_debarred/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+    if procuring_entity is not None:
+        sanction.add("summary", f"Blacklisted by {procuring_entity}")
+
+    if project is not None:
+        sanction.add("description", f"Project: {project}")
+
+    context.emit(entity)
+    context.emit(sanction)
+
+    if managing_officer is not None:
```
leonhandreke: Who is this? If this is just the person who's handling the case, I don't think we want them in our dataset.

datasets/ph/gppb_debarred/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+        program_key=h.lookup_sanction_program_key(context, sanction_type),
+        start_date=start_date,
+        end_date=end_date,
+    )
+    sanction.add("reason", reason)
+    sanction.add("provisions", sanction_type)
+    if procuring_entity is not None:
+        sanction.add("summary", f"Blacklisted by {procuring_entity}")
+
+    if project is not None:
```
leonhandreke: Depending on what `project` contains, maybe it makes more sense to stuff this into one field with `procuring_entity`? Something like "Listed for work on {project} procured by {procuring entity}" (we also try to avoid the word "blacklist" unless we're quoting directly).

Thinking about this some more, I'm not even sure we want that much info? We could also put budget, location, whatever in there, but maybe `offenses` in `reason` is enough?

datasets/ph/gppb_debarred/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+    record.pop("bo_signatory", None)
+    record.pop("temp_start_date", None)
+    record.pop("temp_end_date", None)
+    record.pop("is_removed_by_system", None)
+    record.pop("is_temp_removed_by_system", None)
+    record.pop("updated_date", None)
+    record.pop("created_date", None)
+
+    entity = context.make("Organization")
+    if reg_no is None or reg_no == "N/A":
```
leonhandreke: let's process `reg_no` once only. We can also use rigour `is_nullword` to capture possible future variations of `N/A`

datasets/ph/gppb_debarred/ph_gppb_debarred.yml
```diff
@@ -0,0 +1,65 @@
...
+  Government Procurement Policy Board (GPPB).
+
+description: |
+  The Consolidated Blacklisting Report (CBR) lists entities that have been
+  blacklisted from participating in Philippine government procurement under
+  Republic Act 9184 (Government Procurement Reform Act). Entities are
+  blacklisted for offenses such as failure to comply with contractual
+  obligations, unsatisfactory progress, or submission of falsified documents.
+
+  Blacklisting periods are typically one year (first offense) or two years
```
leonhandreke: Let's use "Debarment" instead of "Blacklisting" where possible, as it's both more speaking and in line with our own comms policy.

datasets/ph/gppb_debarred/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+    address = record.pop("address")
+    managing_officer = record.pop("managing_officer")
+    procuring_entity = record.pop("procuring_entity")
+    project = record.pop("project")
+    offenses = record.pop("offenses")
+    sanction_type = record.pop("saction_imposed")
+    start_date = record.pop("start_date")
+    end_date = record.pop("end_date")
+
+    # Consumed but not used
```
leonhandreke: Then let's add them to `audit_data` `ignore`.

datasets/ph/gppb_debarred/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+    record.pop("category", None)
+    record.pop("blacklisting_order_no", None)
+    record.pop("date_blacklisting", None)
```
leonhandreke: How do the three `start_date` and `created_date` in this record relate? could one be a `listing_date` and the other a `start_date` or something?

datasets/ph/gppb_debarred/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+    record.pop("approved_budget_cost", None)
+    record.pop("stage_violation", None)
+    record.pop("project_location", None)
+    record.pop("additional_saction", None)
```
leonhandreke: Does this also want to go in provisions?

leonhandreke (changes requested):

datasets/ph/gppb_debarred/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+        context,
+        entity,
+        key=start_date,
+        program_name=program_name,
+        source_program_key=sanction_type,
+        program_key=h.lookup_sanction_program_key(context, sanction_type),
+        start_date=start_date,
+        end_date=end_date,
+    )
+    sanction.add("reason", reason)
```
jbothma: format_offenses concatenates them into one string, but we're really happy with adding each distinct value as a string.

datasets/ph/gppb_debarred/crawler.py
jbothma: yeah I'd be careful with constructing strings (this case and summary), reformatting strings (sanction_labels) etc. We generally try to represent source values as verbatim as we can. When we clean, we are normally cleaning punctuation. When we clean names, we try to construct distinct names as true to source as possible. I've just added these docs https://github.com/opensanctions/opensanctions/pull/4150/changes but the principle goes beyond structured data. Building sentences is probably a dangerous path.

datasets/ph/gppb_debarred/ph_gppb_debarred.yml
```diff
@@ -0,0 +1,65 @@
...
+  description: >
+    The GPPB is the single policy-making body on public procurement in the
+    Philippines, created under Republic Act 9184. It oversees procurement
+    rules, blacklisting, and compliance.
+  country: ph
+  url: https://www.gppb.gov.ph
+  official: true
+
+data:
+  url: https://onlineblacklistingportal.gppb.gov.ph/obp-backend/cbr/cbr_public/?category=BLACKLISTED_ENTITIES
```
jbothma: we might be missing permanent bans here

datasets/ph/gppb_debarred/crawler.py
herrhellana: a senior exec of the blacklisted entity, as specified [here](https://www.gppb.gov.ph/wp-content/uploads/2023/05/GPPB-Resolution-No.-09-2004.pdf):

"Blacklisting Order duly signed by the Head of the Procuring Entity/appellate
authority containing, among others, Department/Office Order or Board
Resolution number, name and address of the blacklisted person/entity, license
number, if applicable, Authorized Managing Officer (“AMO”), name of Uniform Guidelines for Blacklisting
project/contract and location/amount, specific ground(s)/offense(s) committed as
provided in Section 4 hereof, sanction imposed and its date of start and
completion, date of issuance of the order to blacklist, and other conditions which
can extend duration of sanctions in Sections 4 hereof."

do we want to emit them with an `UknonwnLink` to the entity?

datasets/ph/gppb_debarred/crawler.py
herrhellana: `start_date` and `end_date` := blacklist period, `created_date`, `updated_date` := record modifications i believe. i also decided to add `updated_date` as `modifiedAt` to the sanctions object.

datasets/ph/gppb_debarred/ph_gppb_debarred.yml
herrhellana: i wonder what's the best way to record `project` as? should it go as an extra field to `reason` instead of `description` as it's the contract the entity got blacklisted over?

datasets/ph/gppb_debarred/crawler.py
herrhellana: those are all null, so not including

herrhellana:
Screenshot 2026-06-03 at 15 25 43
also note that this entry looks funky -- see the entity name. added a skip to .yml.

datasets/ph/gppb_debarred/crawler.py
```diff
@@ -0,0 +1,80 @@
+from typing import Any
+
+from zavod import Context
+from zavod import helpers as h
+from rigour.text.stopwords import is_nullword
+
+
+CATEGORIES = {
+    "BLACKLISTED_ENTITIES": "active",
+    "PERMANENT_BLACKLISTED_ENTITIES": "permanent debarment",
```
leonhandreke: These end up in `Sanction:status`. `active` in this case means temporary debarment? Let's make it clear that these end up in the data by naming and commenting well. Is there any sort of source-supplied string that we could use here, even if we just copy it from the website?

datasets/ph/gppb_debarred/crawler.py
```diff
@@ -0,0 +1,80 @@
...
+    entity.add("country", "ph")
+    entity.add("topics", "debarment")
+
+    sanction = h.make_sanction(context, entity)
+    sanction.add("status", status)
+    h.apply_date(sanction, "startDate", record.pop("start_date"))
+    h.apply_date(sanction, "modifiedAt", record.pop("updated_date"))
+
+    end_date = record.pop("end_date")
+    if status != "permanent debarment":
```
leonhandreke: If we're comparing values, let's use some sort of constant (enum?)

datasets/ph/gppb_debarred/crawler.py
herrhellana: the source categories are either `BLACKLISTED_ENTITIES` (one- or two- years blacklists), or `PERMANENT_BLACKLISTED_ENTITIES` (no expiration date). those are a bit obscure as is? i switched from `active` to `temporary debarment` in categories.

datasets/ph/gppb_debarred/crawler.py
herrhellana: done!

datasets/ph/gppb_debarred/crawler.py
```diff
@@ -0,0 +1,89 @@
...
+from zavod import helpers as h
+from rigour.text.stopwords import is_nullword
+
+
+class DebarmentStatus(Enum):
+    TEMPORARY = "temporary debarment"
+    PERMANENT = "permanent debarment"
+
+
+CATEGORIES = {
```
brrttwrks: Maybe a nitpick, but for Enum, you could do something like:

```python
class DebarmentStatus(Enum):
    BLACKLISTED_ENTITIES = "temporary debarment"
    PERMANENT_BLACKLISTED_ENTITIES = "permanent debarment"
```

And leave out the CATEGORIES dictionary altogether.

Then, just do:

```python
    for category in DebarmentStatus:
        context.log.info(f"Crawling category {category.name}")
        print("CATEGORY", category.name)
        data = context.fetch_json(
            context.data_url + f"?category={category.name}", cache_days=1
        )
        assert isinstance(data, list), f"Expected list, got {type(data)}"
        for record in data:
            crawl_record(context, record, category)
```

AND

```python
    sanction.add("status", category.value)
    h.apply_date(sanction, "startDate", record.pop("start_date"))
    h.apply_date(sanction, "modifiedAt", record.pop("updated_date"))

    end_date = record.pop("end_date")
    if category.value != category.PERMANENT_BLACKLISTED_ENTITIES.value:
        h.apply_date(sanction, "endDate", end_date)
```

It's less complicated and uses the Enum better IMO.

herrhellana: thank you, this makes much more sense! committed.

datasets/ph/gppb_debarred/ph_gppb_debarred.yml
leonhandreke: We seem to have decided to drop it.

leonhandreke (approved):

PR #4126, opened 2026-04-30
[ir_uani_business_registry] canonicalise the path before hashing

nvmbrasserie:


leonhandreke:
Nice! Are we sure-ish that they flip-flop and they didn't just permanently change it and we're getting the changes bit by bit because of caching?

nvmbrasserie:
it's hard to tell, but given the website's historical instability, flip-flop is the most probable cause. it's been going back and forth for a while now

datasets/ir/uani_business_registry/crawler.py
```diff
@@ -146,17 +155,18 @@ def crawl_row(context: Context, row: dict[str, HtmlElement]) -> None:
 
     company_elem = row.pop("company_sort_descending")
     company_link = h.xpath_string(company_elem, ".//a/@href")
+    company_link_clean = normalize_company_url(company_link)
```
leonhandreke: Would not hurt to inline this here I think instead of creating a separate function, but as you prefer.

leonhandreke (approved):

PR #4141, opened 2026-05-04
Docs improvements

herrhellana:


goose-life (approved):
lgtm but I obviously don't know the policy — I assume this is the outcome of some discussion in the data team?

herrhellana:
yeah, in the PR mentioned. i also think we should add Occupancy:politicalGroup [here](https://www.opensanctions.org/reference/#schema.Occupancy). I'll add a commit

goose-life:
I think `Occupancy:politicalGroup` will show up once it's in the data — that page only reflects what we have in the `entities` export iirc

PR #4147, opened 2026-05-05
[de_bfv_extremism] Add Ownership 

herrhellana:
as discussed on slack: https://opensanctions-corp.slack.com/archives/C0AFBST4R4M/p1777890450556499
issue: https://github.com/opensanctions/crawler-planning/issues/389

### DIFF
--- a/datasets/de/bfv_extremism/crawler.py
+++ b/datasets/de/bfv_extremism/crawler.py
@@ -14,7 +14,7 @@ class Organization(BaseModel):
     aliases: list[str] = []
     abbreviations: list[str] = []
     previous_names: list[str] = []
-    asset: list[str] = []
+    child_organizations: list[str] = []
 
 
 class Organizations(BaseModel):
@@ -65,12 +65,20 @@ def crawl_row(context: Context, row: dict[str, _Element]) -> None:
         entity.add("topics", "sanction")
         entity.add("country", "de")
 
-        if item.asset is not None:
-            for asset in item.asset:
-                ownership = context.make("Ownership")
-                ownership.id = context.make_id(item.name, asset)
-                ownership.add("asset", asset)
-                ownership.add("owner", item.name)
+        if item.child_organizations is not None:
+            for child in item.child_organizations:
+                child_entity = context.make("Organization")
+                child_entity.id = context.make_id(child)
+                child_entity.add("name", child)
+                entity.add("topics", "sanction")
+                child_entity.add("country", "de")
+                context.emit(child_entity)
+
+                link = context.make("UnknownLink")
+                link.id = context.make_id(child_entity.id, entity.id)
+                link.add("subject", entity)
+                link.add("object", child)
+                context.emit(link)
 
         sanction = h.make_sanction(context, entity)
         h.apply_date(sanction, "listingDate", extracted_data.start_date)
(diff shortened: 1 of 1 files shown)


herrhellana:
since the `BaseModel` is getting updated here, does it mean it will null the currently accepted reviews for the dataset? or will it just update the review schema in the UI by adding the new prop, and i can go edit the single case where i need to move org name to asset prop?

jbothma (changes requested):
whatever we use to represent this link, you'll have to emit it otherwise the "interstitial" entity won't be in the exported data. Make sure you see the relationship in the exported data. Search around the codebase a bit to see other examples of relationships between entities.

When I run this I get 
```
2026-05-05 12:17:11 [warning  ] Rejected property value [owner]: „Samidoun – Palestinian Prisoner Solidarity Network“, Teilorganisation „Samidoun Deutschland“, [zavod.runtime.cleaning] dataset=de_bfv_extremism entity_id=de-bfv-3fe129368919665829d52828dbf14455013aec8a prop=owner value='„Samidoun – Palestinian Prisoner Solidarity Network“, Teilorganisation „Samidoun Deutschland“,'
```

Ownership asset and owner props take entities as values so you can't pass arbitrary strings like the entity names as values to those props. It has to be the related entity instance, or its id. 

So I'm not really sure what the right way is to model this.

Ownership doesn't quite work because it wants the `asset` to be an `Asset`, but currently the asset is an Organization, and I think it's correctly surfacing that this isn't really an ownership relationship, but just a control one.

Image

https://www.opensanctions.org/reference/#schema.Ownership

Perhaps UnknownLink? with the string they use for the relationship as the `role`?
Another option is just to leave role blank, keep the pydantic model as it was, and link each of extracted_data.organizations[1:] to extracted_data.organization[0]. Since this dataset is so small, I don't want us to over-engineer the data model when we have no idea what the rest of the pattern will be one day. But adding some kind of link is nice.

herrhellana:
we can't do the proposed solution (2), because there is only a single case where a parent-child relationship is presented. other cases feature either a list of separate organizations or same org with aliases. so it's a bit tricky...

i proposed to emit an `UnknownLink` if a child prop is populated in `BaseModel`. Wyt?

jbothma:
what I'm suggesting is just this - so any time there's more than one org, we end up with a link between one of the orgs, and the rest, and I think that's entirely sufficient for this use case. We don't know what other relationships are gonna come up, and I don't think we need to get any more fancy than this. I haven't actually run this, so try it and see if it works, and shout if I missed something important.

```diff
+ first_org = None
for item in extracted_data.organizations:
        entity = context.make("Organization")
        entity.id = context.make_id(item.name)

        entity.add("name", item.name)
        entity.add("alias", item.aliases)
        entity.add("previousName", item.previous_names)
        entity.add("abbreviation", item.abbreviations)
        entity.add("topics", "sanction")
        entity.add("country", "de")

        sanction = h.make_sanction(context, entity)
        h.apply_date(sanction, "listingDate", extracted_data.start_date)
        sanction.add("reason", extracted_data.reason)

+        if first_org is None:
+            first_org = entity
+        else:
+            rel = context.make("UnknownLink")
+            rel.add("subject", first_org)
+            rel.add("object", entity)
+            context.emit(rel)

        context.emit(entity)
        context.emit(sanction)
```

jbothma (approved):
maybe just remove the print :)

herrhellana:
wooops! done:)

PR #4152, opened 2026-05-05
[us_ga_med_exclusions] name migration

nvmbrasserie:


leonhandreke (approved):

jbothma (commented):
I guess the idea would be to get rid of split_middle_name() but we don't support name parts in the names framework yet. How are we selecting what we're working on? Can we leave the datasets with name parts and do a few simpler ones first?

nvmbrasserie:
closing this for now, since the name cleaning model doesn't support name parts yet

PR #4160, opened 2026-05-06
[jp mof sanctions] Name migration

### LINKED ISSUE #2928: [jp_mof_sanctions] Review use of weakAlias field
Lots of things in there that look more like notes/description.

nvmbrasserie:
fixes https://github.com/opensanctions/opensanctions/issues/2928


jbothma (commented):
ok I thought the person name splitting was on more japanese characters than this.

I think it's worth seeing if we can get rid of parse_names in step 3, so same as ebrd - step 1 just builds an original and suggested and accepts, and step 3 switches to h.apply_reviewed_names.

What do you think of doing these with a step 1 and step 3 branch (based on step 1) right off the bat so we can all see where we're heading with it, we can see the deltas are as expected, etc.

nvmbrasserie:
exploratory step 3 is ready for review here: https://github.com/opensanctions/opensanctions/pull/4716

datasets/jp/mof_sanctions/crawler.py
```diff
@@ -157,19 +158,51 @@ def emit_row(
...
-    # )
-    entity.add("previousName", parse_names(row.pop("past_alias", [])))
-    entity.add("previousName", parse_names(row.pop("old_name", [])))
+    entity.add("alias", parse_names(h.multi_split(raw_alias, ALIAS_SPLITS)))
+    entity.add("alias", parse_names(raw_known_alias))
+    entity.add("previousName", parse_names(raw_past_alias))
+    entity.add("previousName", parse_names(raw_old_name))
+    original = h.Names()
+    for n in chain(name_english, name_japanese):
+        original.add("name", n)
```
jbothma: both of these can pass a language to original.add, although check, maybe there's a reason name_japanese doesn't pass lang

datasets/jp/mof_sanctions/crawler.py
```diff
@@ -157,19 +158,51 @@ def emit_row(
...
+        parse_names(raw_known_alias),
+    ):
+        suggested.add("alias", n)
+    for n in chain(parse_names(raw_past_alias), parse_names(raw_old_name)):
+        suggested.add("previousName", n)
+    for n in chain(
+        parse_names(h.multi_split(raw_weak_alias, ALIAS_SPLITS)),
+        parse_names(h.multi_split(raw_nickname, ALIAS_SPLITS)),
+    ):
+        suggested.add("weakAlias", n)
```
jbothma, line 200: I thought we wanted to give these more attention. So I think for entities for whom either of these values exist, we should not auto-accept.

jbothma (approved):

PR #4162, opened 2026-05-06
[nz_designated_terrorists] name migration

nvmbrasserie:


datasets/nz/designated_terrorists/crawler.py
```diff
@@ -49,7 +49,8 @@ def parse_table(
 
 def crawl_item(input_dict: dict, context: Context):
     # aliases will be either a list of size one or None if there is no aliases
-    name, *aliases = h.multi_split(input_dict.pop("terrorist-entity")[0], ALIAS_SPLITS)
+    raw_name = input_dict.pop("terrorist-entity")[0]
+    name, *aliases = h.multi_split(raw_name, ALIAS_SPLITS)
```
jbothma: in this step we want to add `default_accepted=True` and I think also gracefully rekey on raw_name so that in step 3 we can get rid of all this splitting logic and switch to `h.apply_reviewed_names`

leonhandreke: Whoops, I would have never caught `default_accepted=True`. Just for my understanding, the reason is that the current splitting logic works quite well, we just want to get rid of it?

Maybe I should leave these name migration reviews to you for now and just watch...

jbothma:
we had a quick conversation on that earlier this week when the data team was discussing it. I assumed we'd be reviewing these by hand, in which case trivial splitting is an excellent start. But Friedrich pointed out that we could pre-fill with default_accepted=True, and then most datasets don't change so much that maintaining splitting logic wins over hand-splitting new resutls.

I haven't updated the docs to that effect yet, but basically, except for extreme cases, we want to remove all splitting logic in the name cleaning migration step 3.

jbothma:
I've now updated the docs accordingly

datasets/nz/designated_terrorists/crawler.py
jbothma:
```suggestion
    original_names = input_dict.pop("terrorist-entity")[0]
    old_key = context.make_slug(name)
    new_key = context.make_id(original_names)
    context.rekey(old_key, new_key)  # TODO: Remove in name migration step 3
    organization.id = new_key
```

datasets/nz/designated_terrorists/crawler.py
```diff
@@ -47,14 +47,32 @@ def _cell_dates(cell: Element) -> list[str]:
 def crawl_item(context: Context, row: dict[str, Element], expired: bool) -> None:
     name_cell = row.pop("terrorist_entity")
     # aliases will be either a list of size one or None if there is no aliases
-    name, *aliases = h.multi_split(h.element_text(name_cell), ALIAS_SPLITS)
+    raw_name = h.element_text(name_cell)
+    name, *aliases = h.multi_split(raw_name, ALIAS_SPLITS)
     alias_lists = [h.multi_split(alias, [", and", ","]) for alias in aliases]
```
jbothma, line 52: just note https://github.com/opensanctions/opensanctions/pull/4166#discussion_r3443270255

jbothma (approved):

PR #4164, opened 2026-05-06
[ebrd_ineligible] name migration

nvmbrasserie:
[This is not even its final form.](https://github.com/opensanctions/opensanctions/pull/4302)

### DIFF
--- a/datasets/_global/ebrd_ineligible/crawler.py
+++ b/datasets/_global/ebrd_ineligible/crawler.py
@@ -29,14 +29,29 @@ def crawl_entity(context: Context, data: Dict[str, Any]) -> None:
     country = collapse_spaces(data.pop("nationality"))
     entity = context.make("LegalEntity")
     entity.id = context.make_id(name_raw, address, country)
-    entity.add("name", RE_NAME_SPLIT.split(name_raw))
+    names = RE_NAME_SPLIT.split(name_raw)
+    entity.add("name", names)
     subtitle = data.pop("subtitle", "")
+    original = h.Names(name=name_raw)
+    suggested = h.Names()
+    for name in names:
+        suggested.add("name", name)
     if subtitle:
         res = context.lookup("subtitle", subtitle, warn_unmatched=True)
         if res:
             entity.add("alias", res.value)
+            suggested.add("alias", res.value)
             for alias in res.values:
                 entity.add("alias", alias)
+                suggested.add("alias", alias)
+    is_irregular, suggested = h.check_names_regularity(entity, suggested)
+    h.review_names(
+        context,
+        entity,
+        original=original,
+        suggested=suggested,
+        is_irregular=is_irregular,
+    )
     entity.add("address", address.split("$"))
     entity.add("country", country)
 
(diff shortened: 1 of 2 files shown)


leonhandreke (approved):
Does it make sense to attach a cover tracking bug in a comment to all of these half-migrated ones so that we remember to clean them up?

datasets/_global/ebrd_ineligible/crawler.py
```diff
@@ -29,14 +29,29 @@ def crawl_entity(context: Context, data: Dict[str, Any]) -> None:
...
     entity.id = context.make_id(name_raw, address, country)
-    entity.add("name", RE_NAME_SPLIT.split(name_raw))
+    names = RE_NAME_SPLIT.split(name_raw)
+    entity.add("name", names)
     subtitle = data.pop("subtitle", "")
+    original = h.Names(name=name_raw)
+    suggested = h.Names()
+    for name in names:
+        suggested.add("name", name)
     if subtitle:
```
jbothma, line 35: also add subtitle to original

datasets/_global/ebrd_ineligible/crawler.py
```diff
@@ -29,14 +29,29 @@ def crawl_entity(context: Context, data: Dict[str, Any]) -> None:
...
                 entity.add("alias", alias)
+                suggested.add("alias", alias)
+    is_irregular, suggested = h.check_names_regularity(entity, suggested)
+    h.review_names(
+        context,
+        entity,
+        original=original,
+        suggested=suggested,
+        is_irregular=is_irregular,
+    )
```
jbothma: we can add default_accept=True here

datasets/_global/ebrd_ineligible/crawler.py
```diff
@@ -29,14 +29,29 @@ def crawl_entity(context: Context, data: Dict[str, Any]) -> None:
...
     if subtitle:
         res = context.lookup("subtitle", subtitle, warn_unmatched=True)
         if res:
             entity.add("alias", res.value)
+            suggested.add("alias", res.value)
             for alias in res.values:
                 entity.add("alias", alias)
+                suggested.add("alias", alias)
+    is_irregular, suggested = h.check_names_regularity(entity, suggested)
+    h.review_names(
```
jbothma: looking at the delta, it doesn't catch all the irregularity when we do step 3 and we'll need some extra logic to catch the phrases not handled by rigour, e.g. https://github.com/opensanctions/opensanctions/issues/4177

This makes me feel we need to create the step 3 branch right away and try it out before merging the step 1 branch.

jbothma: this is now available https://zavod.opensanctions.org/extract/names/#zavod.meta.names.CleaningSpec.reject_strings

datasets/_global/ebrd_ineligible/crawler.py
jbothma: actually we'll use llm_cleaning which we always want to review

PR #4165, opened 2026-05-06
[za_fic_sanctions] name migration

nvmbrasserie:


datasets/za/fic_sanctions/crawler.py
jbothma, line 115: should add to original

jbothma, line 96: should add to original

jbothma (commented):
I think we can remove the alias cleaning in step 3 - could you make that PR based on this so we see where we end up? I find that when I do that, I realise things like that we need to add `f.n.a` to `names.schema_rules.Person.reject_strings` (and LegalEntity). Maybe ? and `(` too?

we can grandfather in the current set of names using `default_accepted=True` in step 1.

nvmbrasserie:
`[za_fic_sanctions] name migration step 3 (exploratory)` is here https://github.com/opensanctions/opensanctions/pull/4620

jbothma (approved):
just add the names.schema_rules from https://github.com/opensanctions/opensanctions/pull/4620/changes#diff-b42fe10735b3e0c77382a8795dc7be9fe6cd791629081ae9b28906c8160aab8b  to this PR so the affcted entities are already considered irregular, then you can ship

PR #4166, opened 2026-05-06
[ee_international_sanctions] name migration

nvmbrasserie:


leonhandreke (approved):
Does it make sense to attach a cover tracking bug in a comment to all of these half-migrated ones so that we remember to clean them up?

jbothma (commented):
in this step we want to add `default_accepted=True` and I think also gracefully rekey on raw_name so that in step 3 we can get rid of all this splitting logic and switch to `h.apply_reviewed_names`

datasets/ee/international_sanctions/crawler.py
jbothma, line 57: there's now a `reject_leading_digit` to make these automatically be considered irregular

I think if the crawler is cleaning perfectly at the moment, we don't have to add the name spec stuff in the step 1 PR for sanctions PRs because the difference between original and suggested would trigger the review. But in this case I do see one case of a name with ) and [ in names.txt so it's already worth adding them. And it helps to show where we're going with this.

datasets/ee/international_sanctions/crawler.py
```diff
@@ -49,12 +65,28 @@ def crawl_item_human_rights(context: Context, source_url: str, raw_name: str) ->
...
     entity = context.make("Person")
-    entity.id = context.make_id(name)
+    old_key = context.make_id(name)
+    new_key = context.make_id(raw_name)
+    context.rekey(old_key, new_key)  # TODO: Remove in name migration step 3
+    entity.id = new_key
+    original = h.Names(name=raw_name)
+    suggested = h.Names()
     h.apply_name(entity, first_name=first_name, last_name=last_name, lang="eng")
+    suggested.add("name", h.make_name(first_name=first_name, last_name=last_name))
```
jbothma, line 75: @pudo the name cleaning model doesn't support name parts yet so at the moment we'd be dropping first/last name props in step 3.

in this case case we get first/last from splitting on comma, and all names will be considered irregular just by the difference with raw_name.

do we have any plans on rolling out splitting `last, first` more broadly? alternatively is this something we shouldn't be doing more of?

If we don't want to do more of it, I think we should take this excuse to remove the splitting from this dataset.

datasets/ee/international_sanctions/crawler.py
jbothma: to be clear, name spec will be needed in step 3 so that they're still considered irregular (and the reviewed value used) in step 3. So it's ok to only add the name spec in step 3 when we remove the cleaning/splitting code, but it's really bad if we forget to add it then.

nvmbrasserie:
closing this for now, since the name cleaning model doesn't support name parts yet

PR #4173, opened 2026-05-07
[fr_maires] added periodStart and citizenship

herrhellana:
from daily issue board. 

PEP align issue: https://github.com/opensanctions/opensanctions/issues/3805

### DIFF
--- a/datasets/fr/maires/crawler.py
+++ b/datasets/fr/maires/crawler.py
@@ -27,9 +27,7 @@ def crawl_row(
     munid = row.pop("Code de la commune").strip()
     municipality = row.pop("Libellé de la commune").strip()
     birth_date = row.pop("Date de naissance").strip()
-    office_term_start_date = row.pop(
-        "Date de début du mandat"
-    ).strip()  # position's start date
+    office_term_start_date = row.pop("Date de début du mandat").strip()
     function_start = row.pop(
         "Date de début de la fonction"
     ).strip()  # individual's mandate start date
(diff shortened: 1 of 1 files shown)


herrhellana:
Delta export complete          [fr_maires] added=71347 dataset=fr_maires deleted=64952 metric=delta_counts modified=85 version=20260507094947-klv

weird delta for this one? Occupancy gets rekeyed here because of updated dates. ran export with `--rebuilt-store` and `$ZAVOD_ARCHIVE_BACKEND` is set to `GoogleCloudBackend`

datasets/fr/maires/crawler.py
```diff
@@ -28,7 +27,12 @@ def crawl_row(
     munid = row.pop("Code de la commune").strip()
     municipality = row.pop("Libellé de la commune").strip()
     birth_date = row.pop("Date de naissance").strip()
-    start_date = row.pop("Date de début du mandat").strip()
+    position_start_date = row.pop(
+        "Date de début du mandat"
+    ).strip()  # position's start date
```
leonhandreke: Let's be exact in our nomenclature here, cause this is a jungle already. It's "office term start date". Positions don't have start dates. Let's change the variable name as well.

leonhandreke (changes requested):
Only have thoughts about comments/naming things. Otherwise LGTM.

datasets/fr/maires/crawler.py
herrhellana: agreed and changed!

leonhandreke:
We recently had an election, so the new data just came out and that's why the delta is big :) I just reran the live crawler, now the delta looks sane (checked with `ftm-stmt`)

leonhandreke (approved):

herrhellana:
ah that makes so much sense, i totally forgot about it. nice and thank you :)

PR #4190, opened 2026-05-08
[adb_sanctions] name migration

nvmbrasserie:
I've expanded the SPLITS, since we were notified about a few unhandled cases. Taking into account a very poor quality of the data, I'd suggest to run the reviews on all names for this source


datasets/_global/adb_sanctions/crawler.py
```diff
@@ -112,4 +147,4 @@ def crawl(context: Context) -> None:
             crawl_row(context, item["attributes"])
 
         pages += 1
-        assert pages <= 500, "More pages than expected."
+        assert pages <= 500, "More pages than expected."
```
nvmbrasserie:
```suggestion
        assert pages <= 500, "More pages than expected."
 
```

jbothma (commented):
hey!

I get this

```
zavod crawl datasets/_global/adb_sanctions/adb_sanctions.yml
2026-05-12 15:15:44 [info     ] Initialized Google Cloud archive backend acting as etl-crawlerteam-sa@opensanctions-ops.iam.gserviceaccount.com [zavod.archive.backend] dataset=adb_sanctions
2026-05-12 15:15:45 [info     ] Running dataset                [adb_sanctions] data_path=datasets/adb_sanctions data_time=2026-05-12T15:15:39 dataset=adb_sanctions version=20260512151539-jyk
2026-05-12 15:15:46 [info     ] Building timestamp index...    [zavod.runtime.timestamps] dataset=adb_sanctions
2026-05-12 15:15:46 [info     ] Streaming backfilled statements... [zavod.archive] current=adb_sanctions dataset=adb_sanctions object=artifacts/adb_sanctions/20260512063002-lex/statements.pack
2026-05-12 15:15:47 [info     ] Index ready.                   [zavod.runtime.timestamps] count=18762 dataset=adb_sanctions
2026-05-12 15:15:48 [error    ] Runner failed: 'Names' object has no attribute 'incorporationDate' [adb_sanctions] dataset=adb_sanctions
  File "/Users/jdb/projects/opensanctions/opensanctions/zavod/zavod/crawl.py", line 36, in crawl_dataset
    entry_point(context)
    ~~~~~~~~~~~^^^^^^^^^
  File "/Users/jdb/projects/opensanctions/opensanctions/datasets/_global/adb_sanctions/crawler.py", line 147, in crawl
    crawl_row(context, item["attributes"])
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jdb/projects/opensanctions/opensanctions/datasets/_global/adb_sanctions/crawler.py", line 98, in crawl_row
    suggested.add(item["prop"], item["value"])
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jdb/projects/opensanctions/opensanctions/zavod/zavod/extract/names/clean.py", line 126, in add
    current = getattr(self, prop)
  File "/Users/jdb/projects/opensanctions/opensanctions/env/lib/python3.13/site-packages/pydantic/main.py", line 1026, in __getattr__
    raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
AttributeError: 'Names' object has no attribute 'incorporationDate'
```

datasets/_global/adb_sanctions/crawler.py
```diff
@@ -61,22 +76,42 @@ def crawl_row(context: Context, row: Dict[str, str | None]) -> None:
...
             entity.add("alias", other_names)
+            suggested.add("alias", other_names)
+
+        is_irregular, suggested = h.check_names_regularity(entity, suggested)
+        h.review_names(
+            context,
+            entity,
+            original=original,
+            suggested=suggested,
+            is_irregular=is_irregular,
```
jbothma: non-sanctions and we want to re-review so let's enable llm cleaning for this and do the reviews without the suggestion from the original

datasets/_global/adb_sanctions/crawler.py
```diff
@@ -4,21 +4,36 @@
...
     "also known as",
+    "Also known as",
     "also doing business as",
     "formerly operating as",
     "also",
     "formerly",
     "f/k/a",
     "(AKA",
+    "CURRENTLY KNOWN AS",
+    " (currently doing business as",
```
jbothma: those that are not part of https://github.com/opensanctions/rigour/blob/main/resources/names/stopwords.yml#L182 should be added to names.LegalEntity.reject_strings so that the framework knows these are irregular and sends them for review/llm cleaning. Only the ones that "look irregular" will get reviews and be sent to the llm.

jbothma: / is already part of reject_chars_baseline so no need to add

jbothma (commented):
I honestly think doig this using using the names framework is still gonna be too messy because a range of kinds of registration numbers and other text is often thrown in both the primary name and other name columns, and on a few occasions it's listing more than one entity.

I think we need a data model like the yaml

```
entities:
  - name: []
    alias: []
    weakAlias: []
    abbreviation: []
    registrationNumer: []
    uscCode: []
    ...
```

then basically it `rigour.name.contains_split_phrase()` or contains `(` or `Reg no` send it for review with a basic prompt for distinct entities, names in the right places, registration numbers in the right places. Try setting model version as 5.4 before getting too fancy on the prompt. It'd damn good.

Image

nvmbrasserie:
cc @brrttwrks for the future developments of this PR

nvmbrasserie:
## `adb_sanctions` crawler rewrite

### Problem

The previous crawler used a layered cascade of heuristics to parse ADB's two name fields (`name`, `otherName`): hardcoded split lists (`REG_NRS`, `NAME_SPLITS`, `ENTITY_SPLITS`), regex matching, and a 350-line `other_names` YAML lookup table that had to be manually extended every time a new pattern appeared in the source data. The source data mixes registration numbers, tax IDs, USC codes, aliases, former names, and occasionally multiple entities into a single row with no consistent structure.

### Solution

Replace the heuristic cascade with a direct LLM extraction step using `gpt-5.4`, gated by the zavod names framework's irregularity detection.

**New data model** — `EntityData` is a Pydantic model covering all the identifier types found in this dataset (`name`, `alias`, `weakAlias`, `previousName`, `abbreviation`, `registrationNumber`, `uscCode`, `taxNumber`, `idNumber`). `EntityExtractionResult` wraps a list of these to handle the occasional row that encodes multiple entities (e.g. `"Company A; Company B"`). The LLM response is posted to the review system via `review_extraction` so analysts can accept or correct the output.

**Trigger logic** — rows are sent to the LLM when `h.is_name_irregular()` flags either name field, or when `other_name` contains digits (the only case with no framework equivalent: bare identifiers like `"TIN No. 304332224"` carry no split phrase but reliably contain digits). Clean rows — plain name, plain alias — skip the LLM entirely.

**Dataset names config** — `reject_chars: "("` routes any parenthetical content (Chinese transliterations, bracketed aliases, registration info) to the LLM. `reject_strings: ["currently known as", "(aka"]` covers the two original `NAME_SPLITS` phrases that rigour's `contains_split_phrase` does not match; all other split indicators (`"also known as"`, `"formerly"`, `"f/k/a"`, etc.) are already handled by `contains_split_phrase` internally.

### Removed

- `REG_NRS`, `NAME_SPLITS`, `ENTITY_SPLITS`, `REGEX_ALIAS_REGNO` constants
- `other_names` YAML lookup (347 lines of manual overrides, now handled by the LLM)
- The `"reg"` substring trigger (redundant: any registration number contains digits)

datasets/_global/adb_sanctions/crawler.py
```diff
@@ -46,40 +105,41 @@ def crawl_row(context: Context, row: Dict[str, str | None]) -> None:
...
+
+    # Occasionally a single source row encodes more than one entity (e.g.
+    # "Company A; Company B"). The LLM returns one EntityData per distinct
+    # entity; for those rows we generate a stable per-entity ID using the
+    # loop index. Single-entity rows reuse base_entity to avoid a redundant
+    # context.make() call.
+    for i, entity_data in enumerate(entities_data):
+        if len(entities_data) > 1:
+            entity = context.make("LegalEntity")
+            entity.id = context.make_id(full_name, str(i), country)
```
brrttwrks: Using a an index number as an input to an ID worries me - it is fragile and I can see a situation where this leads to a rekeying issue down the road.  What are your thoughts? @jbothma @nvmbrasserie

datasets/_global/adb_sanctions/crawler.py
```diff
@@ -46,40 +87,45 @@ def crawl_row(context: Context, row: Dict[str, str | None]) -> None:
...
-            if res:
-                for item in res.items:
-                    entity.add(item["prop"], item["value"])
-            else:
-                context.log.warning(
-                    f'Unhandled other_names "{other_names}"', value=other_names
-                )
-        else:
-            entity.add("alias", other_names)
+        entity.id = context.make_id(full_name, country)
```
jbothma: unfortunately we have to use `*entity_data.name` here otherwise each related entity gets the same id

jbothma: as a bonus we can also use this pattern https://github.com/opensanctions/opensanctions/pull/4147#issuecomment-4398059150 to link the related entities to one another. But this doesn't have to be blocking - it doesn't change the extraction data model - it just adds to the emitted data.

brrttwrks: I've updated to reference `entity_data.name` instead of `full_name`  - that was just an oversight and I'm glad you caught it.

Cool. I've gone ahead and created the `UnknownLink` to preserve those relationships. Thanks for pointing that out.

PR #4191, opened 2026-05-08
Rename categorise(is_pep=) → default_is_pep, make keyword-only

leonhandreke:
## Summary

- Renames `categorise(is_pep=)` to `categorise(default_is_pep=)` in `zavod.stateful.positions`, and makes the parameter keyword-only.
- Updates ~73 crawlers, the wikidata position helper, and tests.

The supplied value is only used the very first time a position is seen — after that, `categorise()` returns whatever `is_pep` value was set via the human review UI. Calling the param `is_pep` made it look like an assertion ("this position *is* a PEP") when it's really just a default while the position hasn't been reviewed yet. Keyword-only forces every call site to spell it out.

### DIFF
--- a/datasets/_global/everypolitician/crawler.py
+++ b/datasets/_global/everypolitician/crawler.py
@@ -192,7 +192,7 @@ def parse_membership(
             country=country,
             topics=["gov.national", "gov.legislative"],
         )
-        categorisation = categorise(context, position, True)
+        categorisation = categorise(context, position, default_is_pep=True)
 
         position_property = post_summary(org_name, role, starts, ends, [])
         person = context.make("Person")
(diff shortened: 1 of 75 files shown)


pudo:
That is actually clarifying, thank you. Can you do a quick check on how many of the currently-open PEP alignment PRs get conflicted on this? I guess most of them are pre-review anyway and will need an edit, so it could also be OK to just merge...

leonhandreke:
Only two PRs with "alignment" in the name, and those don't seem affected. Anyway, I hope that git would deal fine with the merge in most/all cases.

jbothma:
very nice

Issue #4148, opened 2026-05-05
name review keys change when source changes name order


jbothma:
this is an issue for eu_fsf where the order of the names in the source data isn't stable:

e.g. euReferenceNumber="EU.10618.49"

22 April:

```xml
        <subjectType code="person" classificationCode="P"/>
        <nameAlias firstName="" middleName="" lastName="" wholeName="AMIKWE LEPIC" function="" gender="M" title="" nameLanguage="" strong="true" regulationLanguage="en" logicalId="165400">
            <regulationSummary regulationType="amendment" publicationDate="2024-04-30" numberTitle="2024/1271 (OJ L30042024)" publicationUrl="https://eur-lex.europa.eu/legal-content/EN/TXT/PDF/?uri=OJ:L_202401271"/>
        </nameAlias>
        <nameAlias firstName="" middleName="" lastName="" wholeName="Adonia" function="" gender="M" title="" nameLanguage="" strong="true" regulationLanguage="en" logicalId="156585">
            <regulationSummary regulationType="amendment" publicationDate="2023-07-28" numberTitle="2023/1564 (OJ L190 I)" publicationUrl="https://eur-lex.europa.eu/legal-content/EN/TXT/PDF/?uri=CELEX:32023R1564"/>
        </nameAlias>
```

24 April (last run of the day):

```xml
        <nameAlias firstName="" middleName="" lastName="" wholeName="Adonia" function="" gender="M" title="" nameLanguage="" strong="true" regulationLanguage="en" logicalId="156585">
            <regulationSummary regulationType="amendment" publicationDate="2023-07-28" numberTitle="2023/1564 (OJ L190 I)" publicationUrl="https://eur-lex.europa.eu/legal-content/EN/TXT/PDF/?uri=CELEX:32023R1564"/>
        </nameAlias>
        <nameAlias firstName="" middleName="" lastName="" wholeName="AMIKWE LEPIC" function="" gender="M" title="" nameLanguage="" strong="true" regulationLanguage="en" logicalId="165400">
            <regulationSummary regulationType="amendment" publicationDate="2024-04-30" numberTitle="2024/1271 (OJ L30042024)" publicationUrl="https://eur-lex.europa.eu/legal-content/EN/TXT/PDF/?uri=OJ:L_202401271"/>
        </nameAlias>
```

```
[...]

Issue #3206, opened 2025-12-09
Migrate to hashes as data review keys


jbothma:
https://github.com/opensanctions/opensanctions/pull/3133

pudo:
This can close, right?

PR #4193, opened 2026-05-08
Consistent hash review keys

jbothma:
This is Step 1, migrating legacy review keys to new has-based keys
Step 2 is running each crawler with reviews
Step 3 is removing the legacy key/migration code.

It addresses these main issues:

- The review key is currently dependent on the order names are added. We don't want crawlers to be responsible for that. https://github.com/opensanctions/opensanctions/issues/4148
- We'd like review keys to be hashes rather than unicode-sensitive slugs https://github.com/opensanctions/opensanctions/issues/3206

It does this with the following changes:

- review keys are now sha1 hashes, and parts are only stripped for surrounding whitespace, not slugified.
- if a review is found with a legacy key, we update the key on all entries with that legacy key
- a legacy key can be from the key parts as given (names in a prop now ordered), or in current source order. We try both.

Claude picked up that name order changes can appear to be source value changes since the JsonSourceValue comparison assumes array values have stable ordering. This sorts names before adding to the JsonSourceValue but doesn't migrate old source values. If it looks like the source value changed, crawlers that were pre-filled with reviews with the crawler name cleaning subsequently removed will have their acceptance invalidated.

### DIFF
--- a/datasets/tw/shtc/crawler.py
+++ b/datasets/tw/shtc/crawler.py
@@ -249,5 +249,4 @@ def crawl(context: Context):
         for row in csv.DictReader(infh):
             crawl_row(context, row)
 
-    # TODO: Stop raising once we're through the initial bunch of reviews.
-    assert_all_accepted(context, raise_on_unaccepted=True)
+    assert_all_accepted(context, raise_on_unaccepted=False)
--- a/zavod/zavod/helpers/names.py
+++ b/zavod/zavod/helpers/names.py
@@ -480,10 +480,11 @@ def apply_names(
 def review_key_parts(entity: Entity, original: Names) -> List[str]:
     # Only use the non-empty props in the key so that adding props in
     # future doesn't change the key unless they're actually populated.
+    # Names within each prop are sorted so the key is stable regardless of source order.
     key_parts = [entity.schema.name]
     for prop, names_values in original.as_langtexts():
         key_parts.append(prop)
-        for names_value in names_values:
+        for names_value in sorted(names_values, key=lambda n: (n.lang or "", n.text)):
             if names_value.lang is not None:
                 key_parts.append(names_value.lang)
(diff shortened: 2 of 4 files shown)


jbothma:
vibes log:
Damn claude is good. The main things I've needed to tweak are getting it to unpack nested list comprehensions into something readable. And adding some comments for quirks of the logic it identified are important, but I'd forgotten about.

jbothma:
why are http://localhost:3000/review/dataset/us_fed_enforcements/362cb0d35351071fee0f3b72cf9d75c8cc95c5d0 and http://localhost:3000/review/dataset/us_fed_enforcements/66c66ffbb5b36ecf35e99fb5dab92adbc6ce03de different keys?

because previously slugify on the key parts would merge these two to the same key, and the latter overwrote the former. TextSourceValue is still using slugify for equality, so it doesn't see that the source value on the former has changed now that the key is different.

```
 'Silvergate Capital Corporation, La Jolla, California; and Silvergate Bank, La Jolla, California'
 'Silvergate Capital Corporation, La Jolla, California and Silvergate Bank, La Jolla, California'
```

The consequence is 30 unaccepted in us_fed_enforcements after migration.

Similarly, tw has 9 unaccepted reviews because this isn't stripping punctuation and then squashing spaces, like unicode_slug used to.

I this this is a feature - I don't like the idea of two reviews, one perhaps missing an apostrophe in an arabic name or some other punctuation relevant to names in a writing system, being swallowed and merged with a different review which is missing that value. Maybe it's not that serious, but I don't know when it will be serious, and I don't think the precision comes at a big cost.

pudo (approved):
Thank you for doing this. Can't wait to remove most of this, but it's seems like a surprisingly safe operation as a migration. 

One day we will go away from SHA1 for everything, but it seems safe for this in particular.

PR #4242, opened 2026-05-14
[_global_afdb_sanctions] name migration step 3

jbothma:
## Summary

- Step 3 of the name framework migration for the AFDB sanctions crawler
- Removes the old apply_clean_name function (regex splits, alias detection, lookup logic) and all supporting constants
- Replaces with h.apply_reviewed_name_string(..., llm_cleaning=True)
- Adds reject_strings to the dataset names spec for split-phrase variants not covered by rigour: aka, d.b.a., ' or ', operating, previously

Depends on #4240 (step 1).

## Test plan

- [x] Verify crawler runs and entities are emitted correctly
- [x] Confirm names with the added reject_strings phrases are flagged for LLM review

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/datasets/_global/afdb_sanctions/afdb_sanctions.yml
+++ b/datasets/_global/afdb_sanctions/afdb_sanctions.yml
@@ -60,6 +60,11 @@ assertions:
       LegalEntity: 1400
       Person: 500
 
+names:
+  schema_rules:
+    LegalEntity:
+      reject_strings: ["aka", "d.b.a.", " or ", "operating", "previously"]
+
 lookups:
   names:
     options:
--- a/datasets/_global/afdb_sanctions/crawler.py
+++ b/datasets/_global/afdb_sanctions/crawler.py
@@ -1,52 +1,7 @@
-import re
-
 from zavod.extract.zyte_api import fetch_html
 
 from zavod import Context
 from zavod import helpers as h
-
-NAME_SPLITS = [
-    "doing business as",
-    "also known as",
-    " or ",
-    "a.k.a.",
-    "aka ",
-    "f/k/a",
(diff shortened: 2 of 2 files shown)


jbothma:
## Conversation notes

This PR was developed in the following steps:

**Step 3 approach:**
- Initially considered `h.apply_reviewed_names(context, entity, original=h.Names(name=name), llm_cleaning=True)` but since there is a single raw name string per entity, `h.apply_reviewed_name_string` is the simpler and more appropriate helper.
- Final: `h.apply_reviewed_name_string(context, entity, string=name, llm_cleaning=True)`

**reject_strings analysis:**
Compared the original `NAME_SPLITS` regex against rigour's `name_split_phrases_list()` to find what isn't already detected by `contains_split_phrase`:

| Phrase | Covered by rigour? |
|--------|-------------------|
| `doing business as` | yes |
| `also known as` | yes |
| `a.k.a.` | yes |
| `f/k/a` | yes |
| `d/b/a` | yes |
| `dba` | yes |
| `a/k/a` | yes (as `A/K/A`) |
| `formerly` | yes |
| `previously known as` | yes |
| `"aka "` (bare, no punctuation) | **no** |
| `d.b.a.` | **no** (only `d/b/a`) |
| `" or "` | **no** |
| `operating` | **no** ("formerly" catches "formerly operating as" but not bare "operating") |
| `previously` | **no** (only the full phrase "previously known as" is in rigour) |

The five missing phrases were added as `reject_strings` under a `LegalEntity` schema rule, which covers both `LegalEntity` and `Person` entities (since `Person.is_a("LegalEntity")` is true and no more-specific schema rule exists).

PR #4267, opened 2026-05-18
[ru_egrul] parse and emit OKVED activity codes

### LINKED ISSUE #4238: [ru_egrul] add OKVED code to companies
This is now `LegalEntity:okvedCode` in FtM, and it would be great to add it to both stages of the EGRUL exporter. 

closes #4202.

leonhandreke:
## Summary

- Parse OKVED economic-activity codes from EGRUL/EGRIP XML in `contrib/egrul/egrul_xml.py` via a new `parse_okved_codes` helper, store as a `;`-separated string in a new `okved_codes` column on `legal_entity_schema`.
- `ru_egrul` crawler splits it back out and emits `okvedCode` on the FtM entity. `emit_person` deliberately unchanged — source has no OKVED for persons.
- Helper is already 4.08-compatible: returns the set union of codes from `СвОКВЭД` (declarative type) and the new `СвОКВЭДОтч` (reporting type, added in EGRUL 4.08 / EGRIP 4.07 effective 2026-02-01). Russian terms in comments have inline English translations.
- Vendor the two FNS XSD zips under `contrib/egrul/docs/` (`13673441xsd.zip` for 4.07/4.06, `16493030_xsd.zip` for 4.08/4.07) plus a `docs/README.md` documenting where they came from and what's inside each.
- Add `contrib/egrul/CLAUDE.md` that `@README.md`-imports the main README so Claude picks up the format-docs context when working in this directory.

Closes #4238.


leonhandreke:
@pudo are we okay pulling in 50kB of XSD in zip files as LLM fodder? Or do you think there is a better place to put this? I'm asking because we're not including a lot of other documentation in this repo, so this is maybe a bit of a departure...

leonhandreke:
I ran it, it seems to have worked.

PR #4278, opened 2026-05-20
add draft crawler Australia Tasmania Parliament

brrttwrks:
Sourced from https://github.com/opensanctions/crawler-planning/issues/667


datasets/au/tas_parliament/au_tas_parliament.yml
```diff
@@ -0,0 +1,46 @@
+title: Tasmania Parliament
```
pudo: Can we start with a country name (Australian State Parliament in Tasmania maybe?)

PR #4304, opened 2026-05-23
Add new roles to wd_peps.yml

confirmordeny:
Three prominent positions added: central bank governor, auditor general and public prosecutor general.

Note: Within the file, central bank governors are in the list of posts that are prioritised.


pudo:
Let's hope this brings in a few more stragglers :)

PR #4312, opened 2026-05-26
crawler-pep skill updates

herrhellana:
Tighten the skill + incorporate updates from the reviewed PR with the skill invoked against a Luxembourg dataset: https://github.com/opensanctions/opensanctions/pull/4123


herrhellana:
Note that i also committed changes to `zavod/docs` (as discussed in https://github.com/opensanctions/opensanctions/pull/4123):
- `coverage.start` refers to when a dataset is added to collection
- we default to position names in english and not local language

herrhellana:
(not sure why the jobs are failing here?)

PR #4315, opened 2026-05-26
dataset: dynamically generate children in catalog building

leonhandreke:
This fixes a bug where for a collection containing a collection, `dataset.children` would temporarily be out of sync with `datasets` because `dataset.datasets` gets built from the latest exported sub-collection `index.json`.

https://github.com/opensanctions/operations/issues/2517


leonhandreke:
After seeing the test fail, I'm pretty sure this does not fix it and also don't know what's going on anymore. I will have to investigate more.

pudo:
I don't think we should mess with this without necessity.

PR #4396, opened 2026-06-05
Feature/topic gated enrichment

jbothma:
https://github.com/opensanctions/operations/issues/2583

The delta looks good other than the entire entity add/removes which I think are due to https://github.com/opensanctions/opensanctions/issues/4399 rather than this

image

The change to external is what we're hoping to see here, and covers the risk-topic-less entities while leaving the risky entities

```
 (feature/topic-gated-enrichment) $ grep 'CNOOC INTERNATIONAL LIMITED' ext_icij_offshoreleaks-ofac-gated-2026-06-05t1541/statements.pack ext_icij_offshoreleaks-ofac-not-gated-2026-06-05t1433/statements.pack
ext_icij_offshoreleaks-ofac-gated-2026-06-05t1541/statements.pack:icijol-129829,Company:name,CNOOC INTERNATIONAL LIMITED,ext_icij_offshoreleaks,,,,,2025-12-30T13:13:05,2026-06-05T14:33:49,aa20e6bc7799145a30ee529176acfb8966090e2a
ext_icij_offshoreleaks-ofac-not-gated-2026-06-05t1433/statements.pack:icijol-129829,Company:name,CNOOC INTERNATIONAL LIMITED,ext_icij_offshoreleaks,,,,,2025-12-30T13:13:05,2026-06-05T13:07:11,aa20e6bc7799145a30ee529176acfb8966090e2a
```

### DIFF
--- a/datasets/_analysis/ann_graph_topics/analyzer.py
+++ b/datasets/_analysis/ann_graph_topics/analyzer.py
@@ -29,10 +29,11 @@ def emit_patch(
         existing_topics=list(existing_topics),
     )
 
-    patch = context.make(related_entity.schema)
+    has_internal = any(not s.external for s in related_entity.statements)
+    patch = context.make("Thing")
     patch.id = related_entity.id
     patch.add("topics", topic)
-    context.emit(patch)
+    context.emit(patch, external=not has_internal)
 
 
 def analyze_entity(context: Context, view: View, entity: Entity) -> None:
@@ -120,7 +121,7 @@ def crawl(context: Context) -> None:
     linker = get_dataset_linker(scope)
     store = get_store(scope, linker)
     store.sync()
-    view = store.view(scope)
+    view = store.view(scope, external=True)
 
     for entity_idx, entity in enumerate(view.entities()):
         if entity_idx > 0 and entity_idx % 1000 == 0:
--- a/datasets/_analysis/ann_graph_topics/analyzer.py
+++ b/datasets/_analysis/ann_graph_topics/analyzer.py
@@ -29,11 +29,10 @@ def emit_patch(
         existing_topics=list(existing_topics),
     )
 
-    has_internal = any(not s.external for s in related_entity.statements)
     patch = context.make("Thing")
     patch.id = related_entity.id
     patch.add("topics", topic)
-    context.emit(patch, external=not has_internal)
+    context.emit(patch, external=related_entity.external)
 
 
 def analyze_entity(context: Context, view: View, entity: Entity) -> None:
--- a/zavod/zavod/runner/local_enricher.py
+++ b/zavod/zavod/runner/local_enricher.py
@@ -169,17 +169,17 @@ class LocalEnricher(BaseEnricher[Dataset]):
         yield from self.expand(match)
 
 
-def has_risk_topic(entity_id: str, view: View) -> bool:
-    """Return True if and only if the entity exists in the view and carries a risk topic."""
+def has_enrich_topic(entity_id: str, view: View, enrich_topics: frozenset[str]) -> bool:
+    """True only if the entity exists in the view and carries one of the enrich topics."""
     canonical_id = view.store.linker.get_canonical(entity_id)
     entity = view.get_entity(canonical_id)
     if entity is None:
         return False
-    return len(registry.topic.RISKS.intersection(entity.get("topics"))) > 0
+    return bool(enrich_topics.intersection(entity.get("topics")))
 
 
-def is_edge_internal(entity: Entity, view: View) -> bool:
-    """For an edge entity, return True if and only if every endpoint has a risk topic."""
+def is_edge_internal(entity: Entity, view: View, enrich_topics: frozenset[str]) -> bool:
+    """For an edge entity, return True only if every endpoint carries an enrich topic."""
     endpoint_ids: set[str] = set()
     if entity.schema.source_prop is not None:
         endpoint_ids.update(entity.get(entity.schema.source_prop.name))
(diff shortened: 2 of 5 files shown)


datasets/_analysis/ann_graph_topics/analyzer.py
```diff
@@ -29,10 +29,11 @@ def emit_patch(
         existing_topics=list(existing_topics),
     )
 
-    patch = context.make(related_entity.schema)
+    has_internal = any(not s.external for s in related_entity.statements)
```
pudo: can I vote for `related_entity.external` as a helper?

zavod/zavod/runner/local_enricher.py
```diff
@@ -170,12 +170,35 @@ def expand_wrapped(
         yield from self.expand(match)
 
 
+def has_topic(entity_id: str, view: View) -> bool:
+    """Return True iff the entity exists in the view and carries a risk topic."""
+    canonical_id = view.store.linker.get_canonical(entity_id)
+    entity = view.get_entity(canonical_id)
+    if entity is None:
+        return False
+    return bool(entity.get("topics"))
```
pudo: `registry.topic.RISKS` :)

pudo (commented):
Some notes

datasets/_analysis/ann_graph_topics/analyzer.py
jbothma: is this something that makes sense in ftm or shall we start its life in zavod and promote later?

zavod/zavod/runner/local_enricher.py
```diff
@@ -170,12 +169,35 @@ def expand_wrapped(
         yield from self.expand(match)
 
 
+def has_risk_topic(entity_id: str, view: View) -> bool:
+    """Return True if and only if the entity exists in the view and carries a risk topic."""
+    canonical_id = view.store.linker.get_canonical(entity_id)
+    entity = view.get_entity(canonical_id)
+    if entity is None:
+        return False
+    return len(registry.topic.RISKS.intersection(entity.get("topics"))) > 0
```
pudo: I'm sorry but I actually got this wrong: the enricher itself has a list of topics it is tuned to run on, I think we should use that list.

zavod/zavod/runner/local_enricher.py
```diff
@@ -170,12 +169,35 @@ def expand_wrapped(
...
+def is_edge_internal(entity: Entity, view: View) -> bool:
+    """For an edge entity, return True if and only if every endpoint has a risk topic."""
+    endpoint_ids: set[str] = set()
+    if entity.schema.source_prop is not None:
+        endpoint_ids.update(entity.get(entity.schema.source_prop.name))
+    if entity.schema.target_prop is not None:
+        endpoint_ids.update(entity.get(entity.schema.target_prop.name))
+    if not endpoint_ids:
+        return False
+    return all(has_risk_topic(eid, view) for eid in endpoint_ids)
```
pudo: This does the right thing, but it results in a lot of store `get_entity` calls when we already have all of the relevant entities loaded, I think. Is there a way that we can key the adjacency list on entity ID and then do a second pass on it to pick the `entity.schema.edge` we have both ends for?

zavod/zavod/runner/local_enricher.py
```diff
@@ -193,7 +215,16 @@ def save_match(
...
-            context.emit(adjacent)
+            if topic_gated:
+                if adjacent.schema.edge:
+                    ext = not is_edge_internal(adjacent, subject_view)
+                elif adjacent.id is None:
+                    ext = True
+                else:
+                    ext = not has_risk_topic(adjacent.id, subject_view)
+            else:
+                ext = False
```
pudo: A bit hard to read. Maybe a function in waiting? Or `not topic_gated` is just an override somewhere?

pudo (changes requested):
I agree that this is the minimal change needed, let's maybe look at the code structure a bit and then ship today.

zavod/zavod/runner/local_enricher.py
```diff
@@ -170,15 +168,74 @@ def expand_wrapped(
...
     entity: Entity,
     match: Entity,
+    subject_view: View,
+    topic_gated: bool,
+    enrich_topics: frozenset[str],
 ) -> None:
-    if match.id is None or entity.id is None:
-        return None
+    assert match.id is not None
+    assert entity.id is not None
```
jbothma, line 230: do we need to support entity.id == None gracefully?

I changed these checks added for to make mypy happy to asserts because the new logic reads weird when it's inferring that we're trying to handle missing id gracefully, when I don't think it's something we need to support.

pudo: It's not possible for it to happen in reality because of the store, so just asserting it away seems totally fine

jbothma:
when I run this on main and this branch with just ann_graph_topics, ext_icij and us_ofac_sdn as inputs, these and similar entities are completely missing, rather than just becoming external.

I'm just trying to understand why quick

https://www.opensanctions.org/entities/icijol-10194743/
https://www.opensanctions.org/entities/icijol-10197068/

zavod/zavod/runner/local_enricher.py
```diff
@@ -206,9 +261,17 @@ def enrich(context: Context) -> None:
     subject_store = get_store(scope, resolver)
     subject_store.sync()
     subject_view = subject_store.view(scope)
```
pudo: does this need to go `external=True`?

jbothma: nice catch, thanks.

Confirmed, patching external in the test like the analyzer will do fails until I load the subjest store with external=True

pudo (approved):
YOLO

jbothma:
> when I run this on main and this branch with just ann_graph_topics, ext_icij and us_ofac_sdn as inputs, these and similar entities are completely missing, rather than just becoming external.
> 
> I'm just trying to understand why quick
> 
> https://www.opensanctions.org/entities/icijol-10194743/ https://www.opensanctions.org/entities/icijol-10197068/

The issue was that a handful of entities were emitted as both internal and external. The internal dropped out and showed up as a statement diff. The external remains, as expected. And I must have made a mistake when double-checking them in the raw statements file

PR #4411, opened 2026-06-07
ua_war_sanctions: public-website crawler (crawler_zyte.py)

pudo:
## What

Re-implements the GUR War & Sanctions crawler to source from the **public website** (Zyte-fetched, server-rendered HTML) instead of the confidential JSON API, and wires it to run **in parallel** with the existing API crawler during migration.

- `crawler_zyte.py` — the new public-site crawler.
- `crawler_transition.py` — a thin entry point that runs **both** `crawler.py` (API) and `crawler_zyte.py` (public) in one pass; the dataset `entry_point` now points here.

## Why

- The public site is the **more current** source. The private API serves vessel owner/manager assignments that lag the website by ~a year (e.g. ship 869 PASIPHAE showed Namor/Silverfin from 03.2025 via the API while the site shows Elyrion/Zeythra from 03.2026).
- Ingesting only **publicly-published** data is preferable to depending on a confidentially-shared API.
- Retires an over-engineered abstraction (`WSAPILink`/dispatch) in favour of an explicit, GLEIF-style structure.

## Migration strategy (parallel-emit → dedupe → cutover)

Rather than rekey, both crawlers emit into the dataset for an interim period and dedupe links the old (`ua-ws-*`) and new (`imo-*`) entities by shared IMO / name. `crawler_transition.py` is the interim entry point. Once clusters are stable: switch `entry_point` to `crawler_zyte.py` and delete `crawler_transition.py` + `crawler.py`.

(zavod loads entry-point files by path with no parent package, so the transition wrapper loads its siblings by file path via importlib — relative/bare imports don't resolve, and there's no `__init__.py`.)

## Design of the public crawler

- **Explicit structure**: two reusable parsers (`crawl_entity_page`, `crawl_person_page`) driven by `ENTITY_SECTIONS` / `PERSON_SECTIONS` config tables; maritime via `crawl_vessels` / `crawl_shadow_fleet` / `crawl_tools`.
- **IDs**: IMO microformat for maritime entities (`imo-vsl-<imo>`, `imo-org-<imo>`); `ua-ws-entity-`/`ua-ws-person-<url_id>` elsewhere, matching the API crawler's ids so dedupe links them.
- **Coverage**: vessels (owner/managers/P&I/operator/`mare.shadow`), ships-company, tools factories, rostec (incl. ownership structure), components, uav, kidnappers, stolen, sport, propaganda, executives, partner sanctions.
- **Names** carry script-detected `lang` (eng/ukr/rus). Partner sanctions lists are emitted **without a risk topic** — they're GUR's re-publication of third-country (US/EU/UK) sanctions we already ingest directly and more authoritatively.
- **Defensive parsing**: typed `h.xpath_*` helpers, label maps with a fail-loud audit on unmapped labels, empty-entity/person guards. `mypy --strict` clean.

## Follow-ups (not in this PR)

- **Captains** omitted by design.
- **Assertions** (`.yml` min/max counts) will need retuning against the first full parallel run.
- Validated per-section and via a bounded end-to-end run; a full `zavod crawl` (~15k Zyte fetches for the public side) hasn't been run yet.
- Coverage item to confirm: public `sanctions/persons` lists ~2,196 individuals vs ~4,385 via the API — the site appears to publish only ~half (likely excluding lifted/historical).

### DIFF
--- a/datasets/ua/war_sanctions/crawler_zyte.py
+++ b/datasets/ua/war_sanctions/crawler_zyte.py
@@ -16,7 +16,6 @@ from zavod import Context, Entity, helpers as h
 from zavod.util import Element
 from zavod.extract.zyte_api import fetch_html
 
-BASE = "https://war-sanctions.gur.gov.ua"
 # Success marker for Zyte unblocking: the language switcher is on every real page.
 UNBLOCK = ".//div[contains(@class, 'lang')]"
 CACHE_DAYS = 7
@@ -370,7 +369,7 @@ def crawl_vessel_page(context: Context, url: str) -> None:
                 continue
             company_id = crawl_entity_page(
                 context,
-                f"{BASE}/en/transport/ships-company/{match.group(1)}",
+                f"{context.data_url}/transport/ships-company/{match.group(1)}",
                 program_key="UA-WS-MARE",
                 topic="poi",
             )
@@ -387,7 +386,7 @@ def crawl_vessel_page(context: Context, url: str) -> None:
 
 
 def fetch_listing(context: Context, path: str, page: int) -> Element:
-    url = f"{BASE}/en/{path}?page={page}&per-page={LISTING_PER_PAGE}"
+    url = f"{context.data_url}/{path}?page={page}&per-page={LISTING_PER_PAGE}"
     return fetch_html(
         context, url, UNBLOCK, html_source="httpResponseBody", cache_days=CACHE_DAYS
     )
@@ -408,8 +407,8 @@ def listing_max_page(doc: Element) -> int:
     return max(pages)
 
 
-def listing_detail_urls(doc: Element, path: str) -> list[str]:
-    """The /en/<path>/<id> detail URLs linked from one listing page, de-duplicated."""
+def listing_detail_urls(doc: Element, base: str, path: str) -> list[str]:
+    """The <base>/<path>/<id> detail URLs linked from one listing page, de-duplicated."""
     urls: list[str] = []
     seen: set[str] = set()
     pattern = re.compile(rf"/{re.escape(path)}/(\d+)")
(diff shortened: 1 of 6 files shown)


datasets/ua/war_sanctions/crawler_zyte.py
```diff
@@ -0,0 +1,637 @@
...
+
+from normality import squash_spaces
+
+from zavod import Context, Entity, helpers as h
+from zavod.util import Element
+from zavod.extract.zyte_api import fetch_html
+
+# Success marker for Zyte unblocking: the language switcher is on every real page.
+UNBLOCK = ".//div[contains(@class, 'lang')]"
+CACHE_DAYS = 7
```
nvmbrasserie, line 21: is it safe to cache for 7 days if we run it daily?

datasets/ua/war_sanctions/crawler_zyte.py
```diff
@@ -0,0 +1,637 @@
...
+            rel.add("role", "subsidiary of")
+            context.emit(rel)
+
+    for label in pairs:
+        if label not in COMPANY_SKIP_LABELS:
+            context.log.warning("Unmapped entity label", label=label, url=url)
+    return entity_id
+
+
+def crawl_vessel_page(context: Context, url: str) -> None:
```
nvmbrasserie, line 261: do we want to also emit captains from here?

Image

datasets/ua/war_sanctions/crawler_zyte.py
```diff
@@ -0,0 +1,637 @@
...
+    listing section. Keyed `ua-ws-entity-<url_id>` to match the retiring API crawler.
+    """
+    doc = fetch_html(
+        context, url, UNBLOCK, html_source="httpResponseBody", cache_days=CACHE_DAYS
+    )
+    pairs = entity_label_map(doc)
+
+    def take(label: str) -> Optional[str]:
+        el = pairs.pop(label, None)
+        return h.element_text(el) or None if el is not None else None
```
nvmbrasserie: `take()` closure defined twice, identically (265-267 and 200-202)

pudo: Thanks, fixed :)

PR #4421, opened 2026-06-08
[au_nsw_parliament] first draft

brrttwrks:
From crawler request https://github.com/opensanctions/crawler-planning/issues/663

### DIFF
--- a/datasets/au/nsw_parliament/au_nsw_parliament.yml
+++ b/datasets/au/nsw_parliament/au_nsw_parliament.yml
@@ -27,18 +27,13 @@ publisher:
   url: https://www.parliament.nsw.gov.au/
   country: au
   official: true
-
 data:
   url: https://www.parliament.nsw.gov.au/members/Pages/all-members.aspx
   format: HTML
   lang: eng
-
 tags:
   - list.pep
-  - role.pep
-  - gov.legislative
-  - gov.state
-
+  
 assertions:
   min:
     schema_entities:
--- a/datasets/au/nsw_parliament/au_nsw_parliament.yml
+++ b/datasets/au/nsw_parliament/au_nsw_parliament.yml
@@ -1,0 +1,52 @@
+name: au_nsw_parliament
+title: Australia State Parliament of New South Wales
+prefix: au-nswparl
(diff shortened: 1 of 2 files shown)


datasets/au/nsw_parliament/au_nsw_parliament.yml
```diff
@@ -0,0 +1,52 @@
...
+  official: true
+
+data:
+  url: https://www.parliament.nsw.gov.au/members/Pages/all-members.aspx
+  format: HTML
+  lang: eng
+
+tags:
+  - list.pep
+  - role.pep
```
pudo: These are topics, not tags - I think we should avoid the mix-up a bit?

brrttwrks: Yes, for sure. Fixing.

pudo:
Rebased on `main`, fixed a blocker, and enriched from the member detail pages. Ran clean (272 entities, no warnings, assertions pass).

**Fixes**
- `make_occupancy()` no longer accepts `propagate_country` / `no_end_implies_current` on `main`; dropped both (the crawler crashed on the first row otherwise). Persons keep `citizenship=au`, which still satisfies the `country_entities: au` assertion.

**Detail-page enrichment** (each profile fetched with `cache_days=14`)
- **Occupancy `startDate`** — pulled from the member's current-positions table. All 135 occupancies are now dated (1996–2025, peak at the 2023 election). Added `dates.formats: ["%d %b %Y"]` so the raw `"03 May 2017"` parses to ISO.
  - The positions table's accent colour differs by chamber (Assembly = green, Council = maroon), so it's matched on the `[Position, Start, Notes]` header rather than the CSS class, and the membership row is keyed off the house code (the table says "NSW" while our position name says "New South Wales").
- **`Person:biography`** — from the `div.biopar` sections (111/135 members; the rest have no biography published). `<br>`-separated items are spaced so they don't run together.

**Out of scope (left for a follow-up):** the second table on each profile lists ministerial / shadow-ministerial roles with start+end dates — these could become additional Position/Occupancy entities, but that's a larger change.

PR #4429, opened 2026-06-08
al_kuvendi: PEP crawler for members of the Albanian parliament (Kuvendi)

### LINKED ISSUE #694: Albania Kuvendi
- Data URL: https://kuvendiapi.azurewebsites.net/api/Anetaret?$top=500
- Publisher: https://www.parlament.al/
- Publisher country/territory code: al
- Type of data: PEPs (Politically Exposed Persons)
- Coverage region: region:Europe

Undocumented JSON API backing the parlament.al SPA. Filter `status eq 0` returns exactly 140 current MPs with `emer`/`atesi`/`mbiemer`, DOB, district code, party, photo. Caveat: API hosted on `*.azurewebsites.net` could move/rename — pin the host and sanity-check the count. Tier-3 fallback: scrape SPA at `https://www.parlament.al/deputet` via headless browser. `opendata.gov.al` is WAF-blocked.

pudo:
Adds a new PEP crawler for the 140 sitting members of the Kuvendi (Assembly of the Republic of Albania). Closes opensanctions/crawler-planning#694.

## Source

Undocumented OData API backing the parlament.al SPA: `https://kuvendiapi.azurewebsites.net/api/Anetaret?$top=1000`. `$top` is server-capped at 1000, so a single request returns the full member set (236 rows) — no pagination. `status` is a .NET enum so server-side `$filter` doesn't work; we filter client-side to `status == 0`, the 140 sitting members (matches the Kuvendi's 140 seats).

## Data cleaning

The `ditlindje` (date of birth) field needs two fixes, both handled in `clean_birth_date`:

- **Import-timestamp artifacts** — when a DOB was never entered it defaults to the row's creation timestamp (year 2022/2025), plus a couple of placeholder `0001` years. A plausible-birth-year filter (1925–2007, given MPs must be 18+) cleanly separates these from real dates. ~31% of rows are dropped this way.
- **Timezone off-by-one** — legitimate dates are stored as local midnight expressed in UTC (e.g. a 5 July birthday appears as `…-07-04T23:00:00`). Converting to `Europe/Tirane` recovers the intended calendar day across both standard and DST offsets. (Verified against Iris Luarasi: `1973-07-04T23:00:00` → 1973-07-05.)

A single source email domain typo (`parlamentd.al`) is corrected via a `type.email` lookup. `atesi` (patronymic) `"-"` placeholders are dropped. Other endpoints (`strukturat` committee/group memberships) were evaluated but carry no role information, so they add no PEP-qualifying position beyond "Member of Parliament" and are not crawled.

## Output (verified with qsv)

- 140 Person, 140 Occupancy, 1 Position — 0 issues, `mypy --strict` clean.
- citizenship 100% `al` (Albanian citizenship is a legal eligibility requirement to hold a seat); occupancy status 100% `current`.
- 96/140 birth dates emitted, range 1944–2000 (no artifacts leaked); 140 distinct person ids, no duplicate names.
- Position: "Member of the Parliament of Albania", Wikidata `Q20177772`, topics `gov.national` + `gov.legislative`.

Note: `political` (party) is left as inconsistent source free-text (`PS` / `Partia Socialiste` / `Partia Socialiste e Shqipërisë` all refer to the Socialist Party). It is source-faithful and not used for matching; canonicalizing it would be a separate optional change.

### DIFF
--- a/datasets/al/kuvendi/al_kuvendi.yml
+++ b/datasets/al/kuvendi/al_kuvendi.yml
@@ -1,0 +1,47 @@
+title: Albania Members of Parliament (Kuvendi)
+entry_point: crawler.py
+prefix: al-kuvendi
+coverage:
+  frequency: weekly
+  start: 2026-06-08
+load_statements: true
+summary: >-
+  Current members of the Kuvendi, the unicameral parliament of the Republic of Albania
+description: |
+  This dataset contains the current members of the Kuvendi i Shqipërisë (Assembly of
+  the Republic of Albania), the country's unicameral legislature of 140 seats.
+
+  The data is sourced from the JSON API that backs the official parlament.al website.
+  It lists each member's name, date of birth (where available), email address and
+  political party.
+publisher:
+  name: Kuvendi i Shqipërisë
+  name_en: Assembly of the Republic of Albania
+  description: |
+    The Kuvendi is the unicameral parliament of Albania. It has 140 members elected
+    for four-year terms by proportional representation across the country's twelve
+    electoral districts (qarqe).
+  country: al
+  url: https://www.parlament.al/
+  official: true
+data:
+  # Undocumented OData API backing the parlament.al single-page app. $top is capped
+  # at 1000 by the server, returning the full member set in one request.
+  url: https://kuvendiapi.azurewebsites.net/api/Anetaret?$top=1000
+  format: JSON
+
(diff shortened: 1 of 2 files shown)


herrhellana (approved):
- changed frequency to monthly
- edited dataset description
- simplified the crawler

PR #4439, opened 2026-06-09
[gb_commons] PEP month

herrhellana:
issue: https://github.com/opensanctions/crawler-planning/issues/687

### DIFF
--- a/datasets/gb/commons/crawler.py
+++ b/datasets/gb/commons/crawler.py
@@ -61,6 +61,20 @@ def crawl_member(
         context.emit(position)
         context.emit(occupancy)
 
+        # previous government posts
+        bio_url = value.pop("thumbnailUrl").replace("Thumbnail", "biography")
+        data_bio = context.fetch_json(bio_url)
+        bio = data_bio["value"]
+        gov_posts = bio["governmentPosts"]
+        if gov_posts != [] and gov_posts is not None:
+            for post in gov_posts:
+                previous_position = h.make_position(
+                    context,
+                    name=post["name"],
+                    country="gb",
+                )
+                context.emit(previous_position)
+
 
 def crawl(context: Context) -> None:
     skip = 0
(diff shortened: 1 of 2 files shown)


datasets/gb/commons/crawler.py
```diff
@@ -0,0 +1,93 @@
...
+def crawl_member(
+    context: Context,
+    value: dict[str, Any],
+) -> None:
+    name = value.pop("nameDisplayAs")
+    person = context.make("Person")
+    person.id = context.make_id(name, value.pop("id"))
+
+    person.add("name", name)
+    person.add("title", value.pop("nameFullTitle"))
```
pudo: Does `nameFullTitle` also have the name in it? If so, it should be an `alias`

datasets/gb/commons/crawler.py
```diff
@@ -0,0 +1,93 @@
...
+    person.add("country", "gb")
+
+    party = value.pop("latestParty")
+    if party is not None:
+        person.add("political", party.pop("name"))
+
+    membership = value.pop("latestHouseMembership")
+    start_date = None
+    if membership is not None:
+        start_date = membership.pop("membershipStartDate")
```
pudo: Can we get all the previous terms?

pudo (commented):
Wooo rule Britannia!

datasets/gb/commons/crawler.py
herrhellana: ah, is it an alias? i recorded it as another name in the lords crawler

datasets/gb/commons/crawler.py
herrhellana: it just lists `latestHouseMembership`, but we can also fetch historic members -- https://members-api.parliament.uk/api/Members/Search?House=Commons

datasets/gb/commons/crawler.py
```diff
@@ -0,0 +1,109 @@
...
+        bio = data_bio["value"]
+        gov_posts = bio["governmentPosts"]
+        if gov_posts != [] and gov_posts is not None:
+            for post in gov_posts:
+                previous_position = h.make_position(
+                    context,
+                    name=post["name"],
+                    country="gb",
+                )
+                context.emit(previous_position)
```
pudo: This means they will not be emitted even if they held a ministerial position that would still be in the retention period. Can we broaden the logic there?

herrhellana: fixed!

PR #4442, opened 2026-06-09
[gb_lords] PEP month

herrhellana:
issue: https://github.com/opensanctions/crawler-planning/issues/688

### DIFF
--- a/datasets/gb/lords/crawler.py
+++ b/datasets/gb/lords/crawler.py
@@ -2,73 +2,44 @@ from typing import Any
 
 from zavod import Context
 from zavod import helpers as h
-from zavod.entity import Entity
-from zavod.stateful.positions import PositionCategorisation, categorise
+from zavod.stateful.positions import categorise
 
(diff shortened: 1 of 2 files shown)


herrhellana:
some lords have previous posts listed (e.g. https://members-api.parliament.uk/api/Members/4210/biography) under link/member_id/biography. do we want these?

pudo:
I don't think so. Do the commons emit ministerial posts?

herrhellana:
the commons API lists various positions in prev gov terms, see e.g. https://members-api.parliament.uk/api/Members/1535/biography

pudo:
Oh yeah we defs want the `governmentPosts`

pudo:
This might lead to a person being retained longer because of an occupancy that is not their commons membership. Can we make the crawler handle that

herrhellana:
not a problem at all, the previous positions are listed in a separate url, and the `start_date` always depends on the commons position, so all good!

pudo (approved):
Made this run against `main`, data looks good.

PR #4450, opened 2026-06-09
[ua_rada] PEP month

herrhellana:
issue: https://github.com/opensanctions/crawler-planning/issues/689

### DIFF
--- a/datasets/ua/rada/crawler.py
+++ b/datasets/ua/rada/crawler.py
@@ -1,0 +1,147 @@
+from typing import Any
+
+from zavod import Context
+from zavod import helpers as h
+from zavod.entity import Entity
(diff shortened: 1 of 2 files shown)


datasets/ua/rada/crawler.py
```diff
@@ -0,0 +1,87 @@
...
+        return
+    occupancy = h.make_occupancy(
+        context,
+        person,
+        position,
+        categorisation=categorisation,
+        start_date=start_date,
+        end_date=row.pop("date_end", None),
+        no_end_implies_current=True,
+        propagate_country=True,
```
pudo: These two can drop, then it should work

pudo (approved):

PR #4456, opened 2026-06-09
Fix ca dfatd schema and date warning

jbothma:
@pudo mainly requesting for the helper change


pudo (approved):
Oh I like that helper change a lot

Issue #3781, opened 2026-03-12
Figure out why ancient `statistics.json` exist in `/datasets`


leonhandreke:
```
curl https://data.opensanctions.org/datasets/latest/default/statistics.json | jq . | grep last_change
  "last_change": "2024-05-27T13:13:01",
```

Spun off from https://github.com/opensanctions/opensanctions/issues/2541

leonhandreke:
I think the garbage collection would have to happen here:

https://github.com/opensanctions/opensanctions/blob/97dfb36a9777ba49fe697c43751cef6d241f45bc/zavod/zavod/publish.py#L40-L58

That's where we have a list of all resourced to be published (aka uploaded to `datasets`), so that's where we can clean the old ones up. Would you agree?

I'm not sure https://github.com/opensanctions/opensanctions/issues/3912 is exactly the same. This one here is about garbage collecting files from a `/datasets/latest/dataset_name/`, where dataset is currently being exported. https://github.com/opensanctions/opensanctions/issues/3912 is about removing a directory entirely from `/dataset/latest` after it is no longer being published. I would expect the mechanism to be different (and possibly manual) for the latter.

pudo:
I think removing the `statistics.json` (and other remnants) from the latest dir for living crawlers/collections would be a great idea. Regarding older datasets: is it perhaps enough if we show a little warning if a dataset the user is viewing is not contained in `default` (and/or also not contained in `kyb`)?

leonhandreke:
I think manual removal + a warning would make the most sense. Deleting from the bucket in zavod code makes me uneasy.

leonhandreke:
```
statistics.json mtime              index.json mtime                    statistics.json path

2023-05-26T20:43:54Z	2023-05-26T20:43:54Z	gs://data.opensanctions.org/datasets/latest/ru_fedsfm_terror/statistics.json
2023-05-26T20:43:56Z	2023-05-26T20:43:56Z	gs://data.opensanctions.org/datasets/latest/ru_fedsfm_wmd/statistics.json
2023-06-09T04:21:29Z	2023-06-09T04:21:29Z	gs://data.opensanctions.org/datasets/latest/ubo/statistics.json
2023-09-01T07:27:32Z	2023-09-01T07:27:32Z	gs://data.opensanctions.org/datasets/latest/interpol_yellow_notices/statistics.json
[...]

PR #4477, opened 2026-06-10
Warn about stale files in datasets/latest/

leonhandreke:
## Summary

Refs https://github.com/opensanctions/opensanctions/issues/3781.

After republishing a dataset to `datasets/latest/`, list the files already there and emit a warning for anything we no longer publish so we can clean them up by hand. We don't delete automatically because deleting from the bucket is scary.

Adds `ArchiveBackend.list_objects(prefix)` and `ArchiveObject.updated_at()` so the warning can include the file's last-modified timestamp. Also renames the `latest` kwarg on `publish_dataset`/`publish_resource` to `republish_to_latest` for clarity and introduces a `LATEST` constant.


jbothma (approved):
nice

PR #4479, opened 2026-06-10
[ext_ (smaller ones)] More topic gating for small local enrichers

jbothma:
https://github.com/opensanctions/operations/issues/2583


pudo (approved):
Thank you, exactly what we discussed on Slack

PR #4485, opened 2026-06-10
[ua_war_sanctions] remove vessel crawling code from the API crawler

nvmbrasserie:


pudo (approved):
Thanks for cleaning that up, let's applu

Issue #2874, opened 2025-09-29
Translate all position names from the PEPs crawlers?


nvmbrasserie:
We currently use GPT for translation in following datasets:

- bg_jud_declarations (PEP positions)
- ge_declarations (PEP names and positions)
- me_acp_peps (PEP positions)
- sk_public_officials (PEP positions)
- th_cabinet (PEP positions)

Maybe we want to switch to the Deepl API (or similar) and translate all positions?

leonhandreke:
```sql
SELECT COUNT(*)
FROM "statement"
WHERE ("dataset" IN ('wd_peps', 'wd_categories'))
AND "schema" in ('Position')
AND "lang" IS NOT NULL
AND ("lang" NOT IN ('eng')

4862
```

leonhandreke:
Turns out an LLM will likely be the better option here as well. "conseiller communautaire des Forêts du Perche" should be translated to something like "community councillor of Forêts du Perche". Google Translate spews out "community councillor for the Perche Forests".

leonhandreke:
> I would really like to have it for Wikidata positions, but it needs to be done a bit deeper in the intestines of that code (near: https://github.com/opensanctions/opensanctions/blob/main/zavod/zavod/shed/wikidata/position.py#L87-L88) - but careful, the Wikidata is passing around a lot of `LangText`, which is a special object to handle text + language metadata.

Did a little sketch here: https://github.com/opensanctions/opensanctions/pull/4514

leonhandreke:
```sql
SELECT COUNT(*)
FROM "statement"
WHERE ("dataset" IN ('wd_peps', 'wd_categories'))
AND "schema" in ('Position')
AND "lang" IS NULL
AND "prop" = 'name'
AND entity_id LIKE 'Q%'

29
```

So that's all the `mul` entries. Not worth detecting language for.

leonhandreke:
I went down the list in the position review tool, and that's all I think. After these three PRs have gone in, we can call this done.

pudo:
I'm +1 on closing this. Should we file a follow-on ticket for Wikidata position names?

leonhandreke:
Wikidata is done, I think? https://github.com/opensanctions/opensanctions/pull/4514/ did it. No language detection on `mul`, but as highlighted above, only 29 `mul` items for now, and maybe we can even go through them manually and fix them at the source if they're not as `mul` as they claim to be?

pudo:
[...]

PR #4514, opened 2026-06-12
Translate Wikidata position labels

leonhandreke:
## Summary

- Translates non-English Wikidata position labels into English using the existing LLM helper (`run_text_prompt`) with a position-specific prompt instead of Google Translate.
- Adds a small `shed/wikidata/debug.py` click tool that prints the statements a single QID would emit through the `_wikidata/peps` helpers — useful for inspecting individual items without running the full crawler.

## Why not Google Translate?

The task is a literal translation of role/office words while keeping place names intact (or transliterated to Latin script when the source isn't Latin). Google Translate treats place names as ordinary words and translates them by meaning — e.g. for `conseiller communautaire des Forêts du Perche` it returns `community councillor for the Perche Forests`, when what we want is closer to `community councillor of Forêts du Perche`. The LLM prompt can express the "keep place names intact, use English exonyms only for very common cases" rule directly, and in spot checks does the right thing.

Statement origin for translated names is the model identifier (matching `apply_translit_names`), so we can re-run when we change models.

Related: https://github.com/opensanctions/opensanctions/issues/2874

### DIFF
--- a/zavod/zavod/shed/trans.py
+++ b/zavod/zavod/shed/trans.py
@@ -45,8 +45,17 @@ names, do not separate them in the output but adhere strictly to the output spec
 """
 POSITION_TRANS_PROMPT = """
 Translate the following public office position label from the language denoted by the
- ISO 639-2 code {code}, returning a JSON object where the key 'eng' has the value in
- English.
+ ISO 639-2 code {code} into English, returning a JSON object where the key 'eng' has
+ the value in English.
+
+Keep place names (countries, cities, regions, administrative areas) intact rather than
+ translating their meaning. Only replace a place name with its established English
+ exonym for very common, widely recognised cases — e.g. "Россия" → "Russia", "Москва"
+ → "Moscow", "Wien" → "Vienna", "München" → "Munich". For anything less obvious, err
+ on the side of keeping the original place name: if it is already in Latin script,
+ keep it verbatim (e.g. "São Paulo" stays "São Paulo", not "Saint Paul"); if it is in
+ a non-Latin script, transliterate it to Latin script rather than translating it
+ (e.g. "Карачаево-Черкесия" → "Karachayevo-Cherkesiya", not "Karachay-Cherkessia").
 """
 
 
--- a/zavod/zavod/shed/wikidata/position.py
+++ b/zavod/zavod/shed/wikidata/position.py
@@ -134,23 +134,28 @@ def wikidata_position(
         # else is the next-best language Wikidata gave us, and we translate it to
         # English. Picking what to translate may get more complex in the future,
         # but for now translating the next-best pick after English works for us.
-        if item.label.lang in ("eng", MULTI_LANG):
+        if item.label.lang in ("eng", MULTI_LANG, None):
             item.label.apply(position, "name", clean=clean_wikidata_name)
-        elif item.label.lang is not None:
-            prompt = make_position_translation_prompt(item.label.lang)
-            for translated in run_translation_prompt(
-                context, prompt, item.label.text
-            ).texts:
-                clean_text = clean_wikidata_name(translated.text)
-                if clean_text is None or clean_text.strip() == "":
-                    continue
-                position.add(
-                    "name",
-                    clean_text,
-                    lang=translated.lang,
-                    original_value=item.label.text,
-                    origin=DEFAULT_MODEL,
-                )
+        else:
+            clean_label_text = clean_wikidata_name(item.label.text)
+            if clean_label_text is not None and clean_label_text.strip() != "":
+                assert item.label.lang is not None
+                prompt = make_position_translation_prompt(item.label.lang)
+                translations = run_translation_prompt(
+                    context, prompt=prompt, text=item.label.text
+                ).texts
+                # if for some reason the translation fails, fall back to the original
+                if len(translations) == 0:
+                    item.label.apply(position, "name", clean=clean_wikidata_name)
+                else:
+                    for translated in translations:
+                        position.add(
+                            "name",
+                            translated.text,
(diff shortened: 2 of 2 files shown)


zavod/zavod/shed/wikidata/position.py
```diff
@@ -121,8 +127,30 @@ def wikidata_position(
...
+                clean_text = clean_wikidata_name(translated.text)
+                if clean_text is None or clean_text.strip() == "":
+                    continue
+                position.add(
+                    "name",
+                    clean_text,
+                    lang=translated.lang,
+                    original_value=item.label.text,
+                    origin=DEFAULT_MODEL,
+                )
```
pudo: Weirdly, this isn't where I was imagining it: I thought we'd add it to `make_position` for all of `zavod`, not just Wikidata. Of course, we can very much roll it out here first - but a lot of the benefit for me was some of the weird chaotic position name lists we're getting from Chile, Montenegro, Brazil.

pudo (commented):
Question: do we need/want a language detection thing to determine the input language? We've been pushing this down the road for a long while now, and I could imagine that there's a nifty rust lib somewhere to do language detection that we can bake into rigour and which could equip us with that extra capability across the stack.

zavod/zavod/shed/trans.py
```diff
@@ -45,8 +45,17 @@ class TransliterationLanguageSpec(NamedTuple):
...
 Translate the following public office position label from the language denoted by the
- ISO 639-2 code {code}, returning a JSON object where the key 'eng' has the value in
- English.
+ ISO 639-2 code {code} into English, returning a JSON object where the key 'eng' has
+ the value in English.
+
+Keep place names (countries, cities, regions, administrative areas) intact rather than
+ translating their meaning. Only replace a place name with its established English
+ exonym for very common, widely recognised cases — e.g. "Россия" → "Russia", "Москва"
+ → "Moscow", "Wien" → "Vienna", "München" → "Munich". For anything less obvious, err
```
jbothma, line 54:
```suggestion
 → "Moscow", "Wien" → "Vienna", "München" → "Munich". For anything less obvious, er
```

zavod/zavod/shed/wikidata/position.py
```diff
@@ -121,8 +127,30 @@ def wikidata_position(
...
+    if item.label is not None and item.label.text is not None:
+        # item.label is picked from the available labels in PREFERRED_WD_LANGS order
+        # (English first, then "mul"/multilingual, then the next-best preferred
+        # language). English and multilingual labels can be used as-is; anything
+        # else is the next-best language Wikidata gave us, and we translate it to
+        # English. Picking what to translate may get more complex in the future,
+        # but for now translating the next-best pick after English works for us.
+        if item.label.lang in ("eng", MULTI_LANG):
+            item.label.apply(position, "name", clean=clean_wikidata_name)
+        elif item.label.lang is not None:
```
jbothma: am I right in thinking we're applying either the english/multi label, or the translated label, but never both?

Would it make sense to apply both? I half expected that pick_name considers language, but it looks like it only considers script, so maybe picking both wouldn't be ideal for presentation purposes. I just wonder if keeping original values is useful for customers in non-english contexts.

zavod/zavod/shed/wikidata/position.py
```diff
@@ -121,8 +127,30 @@ def wikidata_position(
...
+            prompt = make_position_translation_prompt(item.label.lang)
+            for translated in run_translation_prompt(
+                context, prompt, item.label.text
+            ).texts:
+                clean_text = clean_wikidata_name(translated.text)
+                if clean_text is None or clean_text.strip() == "":
+                    continue
```
jbothma: shouldn't this fall back to original?

jbothma (commented):
nice prompt improvement!

zavod/zavod/shed/wikidata/position.py
leonhandreke: Ah, interesting, makes sense. So we'd have a new `LangText` type argument to `make_position` that. Decision tree as I see it right now:

(1) `if name.lang in ("eng", "mul")` → use name
(2) `elif name.lang` is another known language → translate from `name.lang` → English
(3) else (no language known): detect language & translate to English

Questions for me:
- How many (3) are there? How many crawlers do we have where we don't know the source language? Wikidata, maybe, in few cases?
- Do we need to detect language or can we let the translation LLM do it?
- If we decide we need detection: 
  - Various libraries based on https://fasttext.cc/ - large-ish model, 176 languages, apparently better for longer text.
  - https://github.com/pemistahl/lingua is apparently better for short things, since it combines a rule-based engine with a smaller probablistic engine. 75 languages
  - Google Translate has a language detection API.

zavod/zavod/shed/trans.py
leonhandreke: I think "err" is correct here?

jbothma: oh you're right! I thought er was caution and err was to make a mistake

zavod/zavod/shed/wikidata/position.py
leonhandreke: Actually, we want to run the cleaning and drop empty values before feeding the empty value to the translation tool. Thanks for flagging!

zavod/zavod/shed/wikidata/position.py
leonhandreke: I agree. In particular, I think it would make sense to figure out a position country -> language mapping and then keep that language. I think I will pass in this PR, I'm trying to keep it minimal here.

jbothma (approved):

PR #4515, opened 2026-06-12
Factor out run_translation_prompt for LLM translation calls

leonhandreke:
## Summary

Adds `run_translation_prompt` in `zavod/zavod/shed/trans.py`. It wraps the existing `run_text_prompt` call plus the JSON parse, cache invalidation on bad responses, and unexpected-keys warning that's been duplicated across the `apply_translit_*` helpers. It returns `zavod.extract.names.clean.LangText` values, leaving entity application to the caller.

`apply_translit_full_name` and `apply_translit_names` are refactored to delegate to it, so the cache/parse/retry behaviour lives in one place. Behaviour for existing crawlers should be unchanged — the one small drop is the per-lang "did not return a value" warning inside `apply_translit_names` (the unexpected-keys warning is now centralised; per-spec misses silently skip, matching what `apply_translit_full_name` already did).

The motivating use case is a wikidata position-label translation path in a follow-up PR — that caller wants `LangText` so it can write the result to `Position.name` itself (with `original_value` traceability) rather than go through `h.apply_name`.

Follow-up (not in this PR): `LangText` currently lives in `zavod.extract.names.clean`, which is a slightly odd home for a generic translation/transliteration result type. Worth moving it somewhere more obvious (e.g. alongside other shared types).

### DIFF
--- a/zavod/zavod/shed/trans.py
+++ b/zavod/zavod/shed/trans.py
@@ -226,11 +226,11 @@ def apply_translit_full_name(
     result = run_translation_prompt(
         context, prompt, name, output_langs=output_langs, model=model
     )
-    for lt in result.texts:
+    for lang_text in result.texts:
         h.apply_name(
             entity,
-            full=lt.text,
-            lang=lt.lang,
+            full=lang_text.text,
+            lang=lang_text.lang,
             alias=alias,
             origin=model,
         )
--- a/zavod/zavod/shed/trans.py
+++ b/zavod/zavod/shed/trans.py
@@ -78,6 +78,7 @@ class TranslationResult(NamedTuple):
 
 def run_translation_prompt(
     context: Context,
+    *,
     prompt: str,
     text: str,
     output_langs: List[str] = ["eng"],
@@ -160,10 +161,18 @@ def apply_translit_names(
     prompt = make_name_translit_prompt(input_code, output_spec)
     output_langs = [spec.language_code for spec in output_spec]
     first_result = run_translation_prompt(
-        context, prompt, first_name, output_langs=output_langs, model=model
+        context,
+        prompt=prompt,
+        text=first_name,
+        output_langs=output_langs,
+        model=model,
     )
     last_result = run_translation_prompt(
-        context, prompt, last_name, output_langs=output_langs, model=model
+        context,
+        prompt=prompt,
+        text=last_name,
+        output_langs=output_langs,
+        model=model,
     )
     first_by_lang = {lt.lang: lt.text for lt in first_result.texts}
     last_by_lang = {lt.lang: lt.text for lt in last_result.texts}
@@ -224,7 +233,7 @@ def apply_translit_full_name(
         prompt = make_name_translit_prompt(input_code, output)
     output_langs = [spec.language_code for spec in output]
     result = run_translation_prompt(
-        context, prompt, name, output_langs=output_langs, model=model
+        context, prompt=prompt, text=name, output_langs=output_langs, model=model
     )
     for lang_text in result.texts:
         h.apply_name(
--- a/zavod/zavod/shed/trans.py
+++ b/zavod/zavod/shed/trans.py
@@ -67,15 +67,24 @@ def make_position_translation_prompt(input_code: str) -> str:
     return POSITION_TRANS_PROMPT.format(code=input_code)
 
 
+class TranslationResult(NamedTuple):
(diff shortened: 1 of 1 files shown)


leonhandreke:
@jbothma adding you since you wrote most of the code touched by this. WDYT of moving `LangText` to a more general location in zavod?

zavod/zavod/shed/trans.py
```diff
@@ -165,43 +220,17 @@ def apply_translit_full_name(
...
+        prompt = make_name_translit_prompt(input_code, output)
+    output_langs = [spec.language_code for spec in output]
+    result = run_translation_prompt(
+        context, prompt, name, output_langs=output_langs, model=model
+    )
+    for lt in result.texts:
+        h.apply_name(
+            entity,
+            full=lt.text,
+            lang=lt.lang,
```
jbothma: total nit, but maybe lt could be lang_text?

jbothma (approved):
This looks fine, but as an alternative it could make sense to move this to run_typed_text_prompt to get rid of the cache invalidation stuff entirely. Would need some good evaluation to make sure it's not doing anything weird with a generic data model. 

re LangText, I have to confess my sins now. There's another [LangText](https://github.com/opensanctions/nomenklatura/blob/main/nomenklatura/wikidata/lang.py#L16). The cleaning LangText makes sense as a name and concept that borrows from the wikidata LangText in that it's text with an associated language code. Are the weights and floats confusing these two? Marrying the two is something that maybe never needs to occur, but if it does, it's worth some careful thought.

leonhandreke:
> re LangText, I have to confess my sins now. There's another [LangText](https://github.com/opensanctions/nomenklatura/blob/main/nomenklatura/wikidata/lang.py#L16). The cleaning LangText makes sense as a name and concept that borrows from the wikidata LangText in that it's text with an associated language code. Are the weights and floats confusing these two? Marrying the two is something that maybe never needs to occur, but if it does, it's worth some careful thought.

I intentionally chose this one, as it's the one we have inside zavod. I guess we could unify them, but given how simple it is, I'm not too worried about it being duplicated. Since translation is a zavod concept and only happens to be applied to Wikidata in this specific PR, I think it makes sense to depend on the zavod one. WDYT?

jbothma:
> I intentionally chose this one, as it's the one we have inside zavod. I guess we could unify them, but given how simple it is, I'm not too worried about it being duplicated. Since translation is a zavod concept and only happens to be applied to Wikidata in this specific PR, I think it makes sense to depend on the zavod one. WDYT?

oh my langtext comment was in reply to

> Follow-up (not in this PR): LangText currently lives in zavod.extract.names.clean, which is a slightly odd home for a generic translation/transliteration result type. Worth moving it somewhere more obvious (e.g. alongside other shared types).

> WDYT of moving LangText to a more general location in zavod?

I haven't thought much about it - it's where it is because it was until now not a particularly generic thing - it was specific to name cleaning. Currently it's part of the model used in name cleaning reviews. Changes to it could have implications for those reviews. So it's not necessarily wrong, but should be done carefully.

leonhandreke:
> I haven't thought much about it - it's where it is because it was until now not a particularly generic thing - it was specific to name cleaning. Currently it's part of the model used in name cleaning reviews. Changes to it could have implications for those reviews. So it's not necessarily wrong, but should be done carefully.

Ah, in this case maybe the right thing to do would be to introduce a third `LangText` namedtuple somewhere in zavod and keep the `LangText` that is written to the database in `extract` hidden behind its API.

leonhandreke:
> This looks fine, but as an alternative it could make sense to move this to run_typed_text_prompt to get rid of the cache invalidation stuff entirely. Would need some good evaluation to make sure it's not doing anything weird with a generic data model.

I drafted this in https://github.com/opensanctions/opensanctions/pull/4583 and realized that if we don't want to put the target languages statically in the model, the validation that `run_typed_text_prompt` offers is not enough. We want to check if we got all the languages back (we know what we expect) and if not, invalidate the cache. Currently, that doesn't seem to be possible with what `run_typed_text_prompt` returns.

I would shelve it for now, also since it's a bigger (possibly `disruptive`?) change that requires more thought than this re-work.

jbothma:
Thanks for trying it - yeah it feels like your refactor is then the best step forward.

zavod/zavod/shed/trans.py
leonhandreke: Done.

PR #4516, opened 2026-06-12
Fixes 12 06

nvmbrasserie:


datasets/cz/pep_declarations/crawler.py
```diff
@@ -122,7 +122,14 @@ def crawl(context: Context) -> None:
             # The list endpoint only returns flattened summary fields; the
             # structured workingPositions array lives on the detail endpoint.
             detail_url = f"{context.data_url}/{item['id']}"
-            detail = context.fetch_json(detail_url, cache_days=14)
+            try:
+                detail = context.fetch_json(detail_url, cache_days=14)
+            except requests.exceptions.HTTPError as e:
+                if e.response is not None and e.response.status_code == 404:
+                    # Person appeared in list but was removed before detail fetch.
```
leonhandreke: (nit) is this really what happened? Sounds extremely unlikely to me. More likely maybe just an error in the webpage? Could we just add a manual skiplist?

The assertions are going to catch us anyway if something goes large-scale wrong, but it's better to make the crawler defensive.

leonhandreke (commented):
Let's maybe discuss the one `cz_pep_declarations` change. LGTM for the rest!

PR #4519, opened 2026-06-12
Add shed/wikidata debug CLI for a single QID

leonhandreke:
## Summary

Adds a small click-based debug tool that runs `wikidata_basic_human` / `wikidata_position` / `wikidata_occupancy` for a single Wikidata QID and prints the resulting statements in pack-CSV format (same format zavod writes during a real crawl). Useful for inspecting what `_wikidata/peps` would emit for one item without running the full crawler. Output can be piped to `csvlens` for interactive browsing.

## Where should this live?

I dropped this under `zavod/zavod/shed/wikidata/debug.py` because that's where the helpers it exercises live. I'm not 100% sure that's the right home though — should it go under `contrib/` (or somewhere similar) instead, since it's a developer-only debug tool rather than a building block other crawlers compose with? Happy to move it.


leonhandreke:
How do we feel about something like this? It was useful for me in debugging, but maybe it's too single-use? Putting it out there as a suggestion - wikidata is always a bit hard to debug.

jbothma:
I think contrib is the right home for it, unless it becomes part of zavod.
Here's a sibling contrib/wd_categories_paths.py 

I don't like that contrib things are out of CI, although we discussed moving dspy and its bits to contrib and setting up CI to test it if any of its files or dependencies are modified. And also maybe it's ok to have to fix some bits in contrib when they're needed once a quarter, rather than to keep them moving along with the rest of the codebase every change.

leonhandreke:
> I think contrib is the right home for it, unless it becomes part of zavod.

Cool, makes sense — moved it to `contrib/wikidata/` next to `wd_categories_paths.py` (renamed to `wd_item_debug.py` while there). Also dropped a little `CLAUDE.md` in `zavod/zavod/shed/wikidata/` pointing at both tools so future-me (and Claude) actually find them when debugging in that area.

pudo (approved):
I like this a lot! Maybe it's `wd_item_debug.py human Qxxxx`? So we can also do positions in the future?

PR #4522, opened 2026-06-12
remove iso9362_bic from default collection

### LINKED ISSUE #3698: Change iso9362_bic to enrichment dataset
We try to keep the `default` collection focused on datasets representing risk, and enrichment of those and related entities.

iso9362_bic was added to default because it's relatively small, but since it's purely reference data, let's give it the same treatment so that entities purely in there because they have a BIC number don't get questioned about their existence in a risk dataset.
 
- [x] Create [changelog entry for effectively a dataset removal](https://www.opensanctions.org/docs/data/changes/)
- [ ] add ext_iso9362_bic
- [ ] remove iso9362_bic from default
- [ ] add iso9362 to kyb

goose-life:
closes #3698

ONLY MERGE ON MONDAY 2026-06-15

### DIFF
--- a/datasets/_collections/default.yml
+++ b/datasets/_collections/default.yml
@@ -50,7 +50,6 @@ children:
   - securities
 
   # banks
-  - iso9362_bic
   - ru_cbr_banks
 
   # misc
--- a/datasets/_externals/ext_gleif.yml
+++ b/datasets/_externals/ext_gleif.yml
@@ -58,12 +58,12 @@ inputs:
   - ext_gb_coh_psc
   - ext_gleif
   - ext_icij_offshoreleaks
+  - ext_iso9362_bic
   - ext_lv_business_register
   - ext_md_companies
   - ext_ru_egrul
   - ext_ua_edr
   - ext_us_irs_ffi
-  - iso9362_bic
   - opencorporates
   - permid
   - regulatory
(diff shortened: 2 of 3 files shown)


datasets/_collections/default.yml
```diff
@@ -50,7 +50,6 @@ children:
   - securities
 
   # banks
-  - iso9362_bic
```
goose-life, line 53: is it really this simple?

jbothma: looks like it should remain in banks.yml - it's defined as reference data.
looks like it should be replaced in ext_gleif and ext_us_irs_ffi with ext_iso9362_bic

goose-life: ah okay! I saw those and thought they'd still be using the full one … is the logic that they'll only be interested in risk-y entities from the full dataset, which would all be in ext_, and it's more efficient not to look through the full dataset again?

pudo:
This is correct.

datasets/_collections/default.yml
jbothma: yeah exactly. those enrichers are only going to consider the enities matching a list of risky topics in their inputs. So there's no point loading the whole dataset including entities without risk topics. So the ext_ version is entirely sufficient.

jbothma (approved):

PR #4524, opened 2026-06-13
Add Pharaoh role to wd_peps.yml as a non-PEP role

confirmordeny:


pudo:
let my people go!

confirmordeny:
Very good!

PR #4539, opened 2026-06-15
Rebuild store by default in zavod export and validate

### LINKED ISSUE #3836: Rebuild store by default in zavod export and validate
zavod export and zavod validate currently reuse an existing store if available. It should rebuild the store by default with an option like `--keep-store` (similar to crawl's --keep-data). The current behaviour is inconsistent with prod and makes it easy to look at stale results by mistake.

leonhandreke:
## Summary

- Flip the default for `zavod export` and `zavod validate` to rebuild the store from scratch (matching prod / `zavod run` behavior).
- Use `--keep-store` to opt out, mirroring the `--clear-data/--keep-data` pattern on `crawl`.
- Drop the now-redundant `--rebuild-store` from `zavod/docs/metadata.md` and the `debug-crawler` skill doc.

Closes https://github.com/opensanctions/opensanctions/issues/3836.

Note: I left the explicit `--rebuild-store` in `.github/workflows/build.yml` — redundant now but arguably clearer about intent. Happy to drop it too.

### DIFF
--- a/.claude/skills/debug-crawler/SKILL.md
+++ b/.claude/skills/debug-crawler/SKILL.md
@@ -98,7 +98,7 @@ Check `data/datasets/<dataset_name>/issues.log` for remaining warnings. Then exp
 and confirm the delta is plausible:
 
 ```bash
-zavod export --rebuild-store datasets/<path>/<dataset_name>.yml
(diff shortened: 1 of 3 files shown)


jbothma (approved):
yay

PR #4544, opened 2026-06-15
[kr_assembly] Add South Korea National Assembly PEP crawler

pudo:
## Summary

Adds a PEP crawler for the **National Assembly of the Republic of Korea** (대한민국 국회), the country's 300-seat unicameral legislature (254 single-member districts + 46 proportional-representation), closing opensanctions/crawler-planning#741.

Emits, for the sitting Assembly:
- one **Person** per member (Korean/romanized/Hanja names, gender, birthDate, party, email, website, sourceUrl, `citizenship: kr`),
- one **Occupancy** per member (with constituency),
- a shared **Position** — *Member of the National Assembly of the Republic of Korea* (`Q14850694`, `gov.legislative`/`gov.national`).

## Source

The issue proposed the official 열린국회정보 OpenAPI, but its API key requires a Korean national ID / public certificate — not viable for us. Instead this uses the **public member-search endpoint** backing `memberSchPage.do`:

- `POST /portal/assm/search/searchAssmMemberSch.do` with `page=1&rows=500` → all members as JSON in **one keyless, cookieless request** (verified). Served under a `text/html` content-type, so `fetch_json` parses the body directly.
- Default scope is the **current** Assembly, so it tracks the sitting body across terms (live roster, not a term-bounded archive). A `len(rows) == total` assertion fails loudly if paging is silently reintroduced or the seat count outgrows `rows`.

## Notes

- **Citizenship** is set per Public Official Election Act (공직선거법) Art. 16(2), cited in a code comment.
- A few members pack two emails into one field with Korean location labels (회관/지역); handled via a `type.email` lookup rather than a parser.
- `telNo` (office phone) and `assemAddr` are deliberately not extracted per PEP guidance.
- `ci_test: true` — single lightweight keyless request; flip to false only if GitHub Actions' US IPs get geo-blocked.

## Validation

- 601 entities (300 Person + 300 Occupancy + 1 Position), **0 crawler issues**, `mypy --strict` clean, `zavod validate` passes.
- All qsv integrity checks empty (no dangling Occupancy refs; every PEP has an occupancy + citizenship); `Occupancy:status` 100% `current`.

### DIFF
--- a/datasets/kr/assembly/crawler.py
+++ b/datasets/kr/assembly/crawler.py
@@ -4,12 +4,6 @@ from zavod import Context, Entity
 from zavod import helpers as h
 from zavod.stateful.positions import PositionCategorisation, categorise
 
-# The endpoint caps the page size silently; request all seats in one call and assert
-# the returned count matches the reported total (see crawl()).
-ROWS = "500"
-
-POSITION_NAME = "Member of the National Assembly of the Republic of Korea"
-
 
 def crawl_member(
     context: Context,
@@ -39,6 +33,7 @@ def crawl_member(
     if occupancy is not None:
         occupancy.add("constituency", row.pop("origNm", None))
         context.emit(occupancy)
+        context.emit(person)
 
     context.audit_data(
         row,
@@ -61,13 +56,12 @@ def crawl_member(
             "openNaId",
         ],
     )
-    context.emit(person)
 
 
 def crawl(context: Context) -> None:
     position = h.make_position(
         context,
-        name=POSITION_NAME,
+        name="Member of the National Assembly of the Republic of Korea",
         wikidata_id="Q14850694",
         country="kr",
         topics=["gov.legislative", "gov.national"],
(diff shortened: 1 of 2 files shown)


datasets/kr/assembly/crawler.py
```diff
@@ -66,7 +65,7 @@ def crawl(context: Context) -> None:
         country="kr",
         topics=["gov.legislative", "gov.national"],
     )
-    categorisation = categorise(context, position, default_is_pep=True)
```
pudo: Why?

datasets/kr/assembly/kr_assembly.yml
```diff
@@ -39,13 +38,9 @@ assertions:
   min:
     schema_entities:
       Person: 250
-      Position: 1
```
pudo: Why?

datasets/kr/assembly/crawler.py
herrhellana: because it's `True` by default?

datasets/kr/assembly/kr_assembly.yml
herrhellana: excessive? or do we want to check against the position count in case we don't emit any or emit garbage?

datasets/kr/assembly/crawler.py
pudo: Good reason!

datasets/kr/assembly/kr_assembly.yml
pudo: It does provide a little safety check and does no harm

PR #4553, opened 2026-06-15
[ca_sema_sanctions_news] resolve stale data and compare entries against consolidated list

herrhellana:
issue: https://github.com/opensanctions/crawler-planning/issues/580


pudo:
I like that approach a lot. Would it make sense to lift the helper function from `enforcements` to `zavod.helpers.dates`?

herrhellana:
I think logically it would -- I'm sure we use the helper not only in the context of enforcements.

pudo:
well this is not an enforcement source. what is your next action on this source?

herrhellana:
With the crawler? I left a comment to @jbothma here in the [issue](https://github.com/opensanctions/crawler-planning/issues/580), asking about the approach we should take regarding any duplicates. What's your opinion?

herrhellana:
checked the assertions: we're fetching the exact 69 total of Persons and Organizations within the retainment scope as published in the [source](https://www.international.gc.ca/world-monde/issues_development-enjeux_developpement/response_conflict-reponse_conflits/crisis-crises/ukraine-sanctions.aspx?lang=eng).

PR #4572, opened 2026-06-16
topic-gate more local enrichers; update assertions

goose-life:


pudo (approved):
Let's see where these guesses get us to!

datasets/_externals/ext_us_irs_ffi.yml
jbothma:
```suggestion
      Company: 80
```

got to ignore Financial services and Bank topics

jbothma (approved):

datasets/_externals/ext_iso9362_bic.yml
```diff
@@ -89,7 +90,12 @@ config:
     - Organization
     - LegalEntity
 
+# estimates June 2026
+# Organizations: 3664
```
goose-life, line 94: whoops, this should have been closer to 250 as well. at least it shouldn't fail on min! thanks for catching the other one @jbothma 🙏

PR #4575, opened 2026-06-16
Fetch fresh CAPTCHA on each black_sea_mou_psc login retry

nvmbrasserie:
fetch_resource skips the download when the target file already exists, so login attempts 2-5 reused the stale captcha.png from attempt 1 while fetch_html had already rotated the server-side CAPTCHA. Every retry was therefore guaranteed to fail. Use a per-attempt filename so each retry gets a fresh image matching the current session.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/datasets/_global/black_sea_mou/psc/crawler.py
+++ b/datasets/_global/black_sea_mou/psc/crawler.py
@@ -40,12 +40,15 @@ Preserve leading zeros. The answer is always exactly 5 digits."""
 MAX_LOGIN_ATTEMPTS = 5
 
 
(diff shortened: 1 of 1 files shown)


leonhandreke:
Alternatively, could we just delete the image in the function or request it with a generic `fetch_response` and write to python `tempfile` supplied tempfile and pass that on to the LLM helper? Then we don't need the `attempt` as the arg, that's a bit ugly.

leonhandreke (approved):
Sorry, missed this. Looks great!

PR #4583, opened 2026-06-16
Use run_typed_text_prompt for the translation helper

leonhandreke:
Stacked on https://github.com/opensanctions/opensanctions/pull/4515.

Migrates `run_translation_prompt` to call `run_typed_text_prompt` internally so the response shape is enforced by OpenAI structured outputs rather than hand-rolled `orjson` parsing and key-subset checks.

The response model follows the `seco_sanctions` pattern — a `List[Translation]` of `{lang, text}` pairs — rather than enumerating each language as a named field. This keeps the schema stable when adding new output languages (which is important because `run_typed_text_prompt` bakes the model's JSON schema into the cache key).

One thing to flag: this drops the `cache_key` return that the previous commit added for caller-side invalidation. `run_typed_text_prompt` doesn't expose it, and with structured outputs the per-result retry it enabled feels less load-bearing. Happy to add it back upstream in `run_typed_text_prompt` if you'd rather keep that capability.

### DIFF
--- a/zavod/zavod/shed/trans.py
+++ b/zavod/zavod/shed/trans.py
@@ -1,10 +1,11 @@
-from typing import Optional, NamedTuple, List
-import orjson
+from typing import NamedTuple, Optional, List
+
+from pydantic import BaseModel
 
 from zavod.context import Context
 from zavod.entity import Entity
 from zavod.exc import ConfigurationException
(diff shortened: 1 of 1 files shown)


leonhandreke:
See https://github.com/opensanctions/opensanctions/pull/4515#issuecomment-4720394884 for why this is not a smart idea for now.

Issue #3843, opened 2026-03-19
[ui] Show subnationalArea in positions classification table


pudo:
This probably requires adding a column to that table, but we are currently falsely classifying national/subnational positions quite a bit because we cannot see that eg. a `Minister of Health` has a `subnationalArea` for a state scope.

PR #4585, opened 2026-06-16
Persist subnationalArea on position records

leonhandreke:
## Summary

Groundwork for https://github.com/opensanctions/opensanctions/issues/3843. Adds a `subnational_areas` JSON column to the `position` table and persists/refreshes it in `categorise()` alongside `countries`, so the positions classification UI can later surface it.

## Rollout

- The `subnational_areas` column has to be added to the prod `position` table (manual `ALTER TABLE`) **before** deploying this — `meta.create_all` only creates missing tables, it won't add columns to an existing one.
- Once the column exists, the UI needs to land in lockstep: `ui/lib/db.ts` does an exact-set schema check on startup (throws on both missing and unexpected columns), so adding the column on the DB without updating `expectedPositionColumns` + the `PositionTable` interface will break the UI, and vice versa. UI PR to follow.

### DIFF
--- a/ui/lib/db.ts
+++ b/ui/lib/db.ts
@@ -39,6 +39,7 @@ const expectedPositionColumns = new Set<string>([
   'entity_id',
   'caption',
   'countries',
+  'subnational_areas',
   'is_pep',
   'topics',
   'dataset',
@@ -82,6 +83,7 @@ export interface PositionTable {
   entity_id: string
   caption: string
   countries: JSONColumnType<string[], string[], string[]>
+  subnational_areas: JSONColumnType<string[], string[], string[]> | null
   is_pep: boolean | null
   topics: JSONColumnType<string[], string[], string[]>
   dataset: string
@@ -510,6 +512,7 @@ export async function softDeleteAndCreatePosition({ entityId, positionUpdate, mo
         // NOTE(Leon Handreke): I don't know how else to make it work than to stringify
         // the lists manually.
         countries: JSON.stringify(updatedData.countries) as unknown as string[],
+        subnational_areas: JSON.stringify(updatedData.subnational_areas ?? []) as unknown as string[],
         topics: JSON.stringify(updatedData.topics) as unknown as string[],
       })
       .returningAll()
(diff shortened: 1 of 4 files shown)


pudo:
Should we consider just having a `properties` JSONB on the position table? I feel like `subnationalArea` is one thing, there is also `inception`/`dissolution`, and I think we may want to introduce a text-only sibling to `organization`, soon.

leonhandreke:
Draft here: https://github.com/opensanctions/opensanctions/pull/4593

Although honestly, maybe getting our ducks in a row and adopting Alembic for etldb as well might be the better path? Or is there a strong reason not to go with nullable columns here, except deployment convenience?

leonhandreke:
Some more thoughts on this:

- I feel that since we're using Postgres, colums are sort of the default concept that everything is centered around. The query language, renames and any sort of database-level migration and validation, indices... so I think that should be the default
- Changing the schema is a deployment headache, but that's only because we don't have something like Alembic. Alembic + Claude really are a dream team of doing complex database migration work, and would be a good idea anyway?

Disadvantages of using a column:
- for list-value fields, we're already using a JSON value, so who cares whether they're in a JSONB bag
- we support both Postgres and sqlite, and this may be easier to get right across both.

jbothma (approved):
I guess one argument for a jsonb blob is that sqlite doesn't support arrays, so we're already using jsonb for the countries column.

If we go that way, I would propose that we keep the categorisation columns (topics) out of the properties blob so that `properties` is what we crawl and `topics` and `is_pep` is what we categorise.

But I think adding alembic and just adding/migrating columns when we need them is a nice improvement and once it's there, adding the next column when we need it is easy.

I don't lean strongly either way.

leonhandreke:
Rollout plan, since this needs ordering across DB / backend / UI:

1. On the prod position DB, run:
   ```sql
   ALTER TABLE position ADD COLUMN subnational_areas JSON;
   ```
   Nullable on purpose so existing rows are fine — the backend coalesces `NULL` to `[]` when comparing.

2. Deploy this PR — zavod starts writing the column on every `categorise()` call, and the UI's startup schema check learns about it.

Doing step 2 without step 1 will break both sides: the backend insert fails on the missing column, and the UI startup throws on "Missing columns: subnational_areas".

Surfacing subnationalArea in the classification table itself (https://github.com/opensanctions/opensanctions/issues/3843) is the follow-up after this is bedded in.

PR #4595, opened 2026-06-17
[gb_commons] Add contact details (email, phone, website)

pudo:
## Summary

The UK House of Commons crawler previously only read `governmentPosts` out of the Members API. This adds the member contact details published on the separate `/Contact` endpoint:

- `email` → `Person:email`
- `phone` → `Person:phone` (normalized to E.164, multiple per MP where both parliamentary and constituency offices are listed)
- the `Website` contact type → `Person:website`

Social media handles (X/Twitter) and office contact-form URLs are skipped — FollowTheMoney has no property to hold them. The contact request runs only after the PEP categorisation check, so it's never issued for non-PEP rows.

### Note on date of birth

The original motivation was missing DOB. The Members API does not publish birth dates on any endpoint (search, overview, biography, contact, synopsis), so this can't be sourced here — it would only come from Wikidata enrichment downstream.

## Test plan

- Full local crawl: exit 0, empty `issues.log`, 8,376 entities.
- Emitted: 676 `email`, 676 `phone`, 349 `website` statements; `birthDate` remains 0 (confirming the source gap).
- `mypy --strict` clean.

### DIFF
--- a/datasets/gb/commons/crawler.py
+++ b/datasets/gb/commons/crawler.py
@@ -2,11 +2,28 @@ from typing import Any
 
 from zavod import Context
 from zavod import helpers as h
+from zavod.entity import Entity
 from zavod.stateful.positions import categorise
 
 # Page size accepted by the Members API. The API caps `take` at 20.
 PAGE_SIZE = 20
 TOPICS = ["gov.national", "gov.legislative"]
+
+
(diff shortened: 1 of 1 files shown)


leonhandreke:
I thought usually we don't do phone numbers of politicians? (I realize they're not personal in this case)

PR #4597, opened 2026-06-17
[pg_parliament crawler] 

### LINKED ISSUE #671: Papua New Guinea National Parliament
### Data URL

https://www.parliament.gov.pg

### Publisher

_No response_

pg

### Type of data

### Coverage region

region:Asia/Pacific

### Can you tell us more?

_No response_

- [x] I understand.

brrttwrks:
Closes https://github.com/opensanctions/crawler-planning/issues/671

### DIFF
--- a/datasets/pg/parliament/crawler.py
+++ b/datasets/pg/parliament/crawler.py
@@ -151,22 +151,6 @@ def crawl_member(
 
 
 def crawl(context: Context) -> None:
-    doc = context.fetch_html(context.data_url, cache_days=1)
-
-    bio_links: list[tuple[str, str, bool]] = []
-    for a_el in h.xpath_elements(doc, ".//a[contains(@href,'parliament/bio/view/')]"):
-        href_val = a_el.get("href", "")
-        if "cat_url_title" in href_val:
-            continue  # skip template placeholder
-        if href_val.startswith("https"):
-            url = href_val
-        else:
-            url = context.dataset.url + "index.php" + href_val
-
-        slug = url.rstrip("/").split("/")[-1]
-        is_governor = slug.startswith("governor-")
-        bio_links.append((url, slug, is_governor))
-
     mp_position = h.make_position(
         context,
         name="Member of the National Parliament of Papua New Guinea",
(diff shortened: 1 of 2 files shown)


datasets/pg/parliament/crawler.py
```diff
@@ -0,0 +1,246 @@
...
+from typing import Any
+
+from lxml.etree import _Element as Element
+
+from zavod import Context
+from zavod import helpers as h
+from zavod.entity import Entity
+from zavod.stateful.positions import PositionCategorisation, categorise
+
+BASE_URL = "https://www.parliament.gov.pg"
```
pudo: is this also `dataset.url`? Can we avoid duplicates?

datasets/pg/parliament/crawler.py
```diff
@@ -0,0 +1,246 @@
...
+
+# The term segment embedded in every member bio URL.
+# A new parliament changes this to "twelfth-parliament" etc.; raise so the
+# crawler is updated rather than silently producing stale data.
+PARLIAMENT_TERM = "eleventh-parliament"
```
pudo: The problem is that hard-coding this makes it not a working crawler for our purposes if it fails to pick up new sessions/terms. That should be a really high priority to determine automatically.

datasets/pg/parliament/crawler.py
```diff
@@ -0,0 +1,246 @@
...
+
+# Q21294916 — "Member of the National Parliament of Papua New Guinea"
+MP_POSITION_QID = "Q21294916"
```
pudo: inline

datasets/pg/parliament/crawler.py
```diff
@@ -0,0 +1,246 @@
...
+            if value:
+                fields[key] = value
+    return fields
+
+
+def clean_name(raw: str) -> str:
+    name = re.sub(r",.*$", "", raw)
+    for title in ["Hon.", "Hon.Sir", "Rt.", "Sir.", "Sir", "Dr.", "Hon.Mai"]:
+        name = name.replace(title, "", 1)
+    return name.strip()
```
pudo: Do we need this? As in: are we sure we want to be cleaning the names like this?

datasets/pg/parliament/pg_parliament.yml
```diff
@@ -0,0 +1,55 @@
...
+entry_point: crawler.py
+coverage:
+  frequency: weekly
+  start: 2026-06-17
+load_statements: true
+ci_test: false
+
+summary: >-
+  Current members of the National Parliament of Papua New Guinea, comprising
+  89 Open electorate members and 22 Provincial Governors.
```
pudo: Can we bring it down to 50-80 chars so it fits in one line and doesn't overlap with the title.

pudo (changes requested):
Some notes :)

datasets/pg/parliament/crawler.py
brrttwrks: I thought it would be simple cleaning, but it makes sense to use reviews I think, as this will grow with new sessions.

datasets/pg/parliament/crawler.py
brrttwrks: We get the bio links of the current term period from the page menu. These links update per term if I understand how the site works, so when it does change, these links should also update and the crawler will automatically get the new term period members. I was unable to find on their site any listing of term periods and their start/end dates.

pudo (approved):
Looks good, I like the names rules!

PR #4620, opened 2026-06-18
[za_fic_sanctions] name migration step 3

nvmbrasserie:
Step 3 of the name-framework migration for `za_fic_sanctions`, stacked on top of the Step 1 PR (#4165) so the diff shows only the Step 1 → Step 3 change. **Exploratory** — Step 3 normally runs only after the Step 1 reviews are completed; this is to see the end state and surface which `names.schema_rules` we need.

## What changed

**`crawler.py`** — remove the custom alias splitting/cleaning and hand the raw name + alias strings to `h.apply_reviewed_names`:
- Drop `ALIAS_SPLITS`, the `multi_split` alias loops, the `?`/space filtering, and the `?` warnings.
- Build a single `h.Names(name=full_name)` and `.add("alias", ...)` the raw `IndividualAlias`/`EntityAlias` field, then one `h.apply_reviewed_names` call per entity.
- Address splitting (`ADDRESS_SPLITS`) is unchanged — that's not name data.

**`za_fic_sanctions.yml`** — add `names.schema_rules` so the markers the old splitter relied on still flag irregularity once the custom logic is gone. Reconciled the old split markers against `rigour.names.name_split_phrases_list()`:

| old marker | covered by rigour? | action |
|---|---|---|
| `a.k.a.` / `Formerly known as` | yes | nothing |
| `f.k.a` (dotted) | no (only `f/k/a`) | `reject_strings` |
| `f.n.a` | no | `reject_strings` |
| `?` (transliteration artefacts) | n/a (char) | `reject_chars` both |
| `(` `)` (LegalEntity) | already in Person baseline | `reject_chars` LegalEntity |

**Incidental** (not name-related, but needed for `mypy --strict` on the touched file): fix `clean_passports`' return annotation to a tuple, and replace the deprecated lxml `getchildren()` with `iterchildren()` (skipping empty elements).


jbothma (approved):
looks good - to be merged only after reviews are completed following step 1's merge

PR #4640, opened 2026-06-21
Rare examples where religious roles result in PEP status

confirmordeny:
(1) Holders of religious offices are not normally regarded as PEPs as a result of holding those positions in this case however the Vatican recognises that Members of the College of Cardinals are PEPs of the Holy See.

https://www.asif.va/ENG/pdf/Regolamenti/EN-Instruction_no.5_Annex.pdf

(2) Andorra has two joint heads of state known as "Co-princes" - both are PEPs. One is the President of France, the other is the Episcopal Co-Prince (the Bishop of Urgel).

https://en.wikipedia.org/wiki/Co-princes_of_Andorra


zucolg92-star:
ok

PR #4658, opened 2026-06-23
[ar_senado] edits

herrhellana:
issue: https://github.com/opensanctions/crawler-planning/issues/634

### DIFF
--- a/datasets/ar/senado/ar_senado.yml
+++ b/datasets/ar/senado/ar_senado.yml
@@ -7,25 +7,23 @@ coverage:
 load_statements: true
 ci_test: false # Live external government source, not available in CI
 summary: >-
-  Upper house of Argentina's National Congress: three senators per province,
-  serving six-year terms
+  Memebers of the Senate of Argentina, the upper house of the National Congress.
 description: |
-  Senators of the Honorable Senado de la Nación (Senate of the Nation), the upper
-  house of the bicameral National Congress of Argentina. Its 72 seats are elected
-  directly, three per province across the 23 provinces and the Autonomous City of
-  Buenos Aires, for six-year terms.
+  Current and historical senators of the Senate of the
+  Nation (Honorable Senado de la Nación), the upper house of the bicameral National Congress of Argentina. Its 72
+  senators are directly elected, three per province across the 23 provinces and the
+  Autonomous City of Buenos Aires, for six-year terms, and together with the Chamber
+  of Deputies hold the country's legislative power, including passing laws and
+  approving the budget.
 
-  The dataset covers the chamber's full historical roster, spanning many
-  legislative terms: both sitting senators and those who left office recently
-  enough to remain politically exposed. Each senator is recorded with their name,
-  party, province and term dates; sitting senators additionally carry an email
-  address and their parliamentary group (bloque). It complements the ar_parliament
-  dataset, which covers the lower house (Chamber of Deputies).
+  This dataset records senators with their name, party, province, and term
+  dates; sitting senators additionally carry an email address and their parliamentary
+  group. It complements the [Argentina Members of Parliament](https://www.opensanctions.org/datasets/ar_parliament/) dataset, which covers the lower
+  house.
 url: https://www.senado.gob.ar/micrositios/DatosAbiertos/
 publisher:
   name: Honorable Senado de la Nación Argentina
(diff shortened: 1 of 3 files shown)


herrhellana:
i also changed the dataset description for the parliament dataset to mirror the Senate description (complements...) + set up frequency there to monthly.

datasets/ar/senado/ar_senado.yml
```diff
@@ -1,4 +1,4 @@
-title: Argentina Members of the Senate
+title: Argentina Members of Senate
```
pudo, line 1: Why???

datasets/ar/senado/ar_senado.yml
```diff
@@ -7,38 +7,42 @@ coverage:
...
-  dataset, which covers the lower house (Chamber of Deputies).
+  This dataset records senators with their name, party, province, and term
+  dates; sitting senators additionally carry an email address and their parliamentary
+  group. It complements the [Argentina Members of Parliament](https://www.opensanctions.org/datasets/ar_parliament/) dataset, which covers the lower
+  house.
 url: https://www.senado.gob.ar/micrositios/DatosAbiertos/
 publisher:
   name: Honorable Senado de la Nación Argentina
   name_en: Honorable Senate of the Argentine Nation
-  acronym: HSN
```
pudo, line 28: Why?

datasets/ar/senado/ar_senado.yml
```diff
@@ -7,38 +7,42 @@ coverage:
...
   description: >
     The Honorable Senate of the Nation is the upper house of the National
     Congress, the bicameral national legislature of Argentina.
   url: https://www.senado.gob.ar/
   country: ar
   official: true
 data:
-  url: https://www.senado.gob.ar/micrositios/DatosAbiertos/ExportarListadoSenadoresHistorico/json
+  url: https://www.senado.gob.ar/micrositios/DatosAbiertos/ExportarListadoSenadores/json
   format: JSON
```
pudo: Should we be doing both?

datasets/ar/senado/crawler.py
```diff
@@ -22,42 +14,24 @@ def fetch_rows(context: Context, url: str) -> list[dict[str, Any]]:
...
-def clean_date(context: Context, value: str | None) -> str | None:
-    """Normalize a source date to an ISO string, or None for empty/placeholder values.
-
-    Parses through the dataset's configured date formats so an unexpected format
-    raises rather than silently corrupting the downstream cutoff comparison.
-    """
-    if value is None or value == "" or value == NO_DATE:
-        return None
-    dates = h.extract_date(context.dataset, value, fallback_to_original=False)
-    return dates[0]
```
pudo, line 34: This function was used to make sure we're only pulling in to `cutoff`. Does that still work afterwards?

pudo (changes requested):
Looks like a couple of gremlins got in?

datasets/ar/senado/ar_senado.yml
herrhellana: those acronyms often look like Claude's inventions and not something that is widely used

datasets/ar/senado/ar_senado.yml
herrhellana: a stylistic choice to match the sister dataset's title, `Argentina Members of Parliament`

datasets/ar/senado/ar_senado.yml
herrhellana: the historical URL doesn't have `BLOQUE` (but its values are almost exactly the same as in `PARTIDO O ALIANZA` or emails, but i guess we can survive without it. i'll roll it into a single url call

datasets/ar/senado/crawler.py
herrhellana: we don't need to rely on `cutoff` here as `make_occupancy()` already drops PEPs outside the coverage window, and we're not gaining much here in terms of processing efficiency -- we're skipping around 1k historical JSON rows outside the coverage window, but with the cutoff present, we still first need to process the rows to extract the clean dates with the `clean_date() `function. so i decided against using the cutoff helper for simplicity and readability.

pudo (approved):
Let's go, made some direct edits :)

PR #4660, opened 2026-06-23
[xk_assembly] edits

herrhellana:
issue: https://github.com/opensanctions/crawler-planning/issues/693

### DIFF
--- a/datasets/xk/assembly/crawler.py
+++ b/datasets/xk/assembly/crawler.py
@@ -97,7 +97,7 @@ def crawl_member(
     name = h.element_text(h.xpath_element(doc, ".//h1[@class='name']"))
 
     person = context.make("Person")
-    person.id = context.make_id(deputy_id, name)
+    person.id = context.make_slug(deputy_id)
     person.add("name", name)
     person.add("sourceUrl", url)
     # Deputies of the Assembly must be citizens of Kosovo: Constitution Art. 71(1)
@@ -126,6 +126,7 @@ def crawl_member(
     h.apply_date(person, "birthDate", dob)
     for bio_label, prop in BIO_PROPS.items():
         person.add(prop, bio.pop(bio_label, ""))
+    context.audit_data(bio, ignore=BIO_IGNORE)
 
     occupancy = h.make_occupancy(
         context,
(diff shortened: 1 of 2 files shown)


datasets/xk/assembly/crawler.py
```diff
@@ -94,10 +94,11 @@ def crawl_member(
         cache_days=20,
     )
     deputy_id = parse_qs(urlparse(url).query)["deputy"][0]
+    name = h.element_text(h.xpath_element(doc, ".//h1[@class='name']"))
 
     person = context.make("Person")
-    person.id = context.make_slug(deputy_id)
```
pudo, line 99: Is there a specific reason we're re-keying a prod dataset here?

datasets/xk/assembly/crawler.py
```diff
@@ -126,7 +126,6 @@ def crawl_member(
     h.apply_date(person, "birthDate", dob)
     for bio_label, prop in BIO_PROPS.items():
         person.add(prop, bio.pop(bio_label, ""))
-    context.audit_data(bio, ignore=BIO_IGNORE)
```
pudo, line 129: Why?

datasets/xk/assembly/xk_assembly.yml
```diff
@@ -1,23 +1,23 @@
-title: Assembly of the Republic of Kosovo (Kuvendi i Kosovës)
+title: Members of Assembly of the Republic of Kosovo
```
pudo: I guess we should start with country name

pudo (commented):
Curious about the reason for re-keying this? It's already merged and through de-dupe.

datasets/xk/assembly/xk_assembly.yml
herrhellana: yes good flag!

datasets/xk/assembly/crawler.py
herrhellana: i looked again into how `make_slug()` works, and no, there is not... i'll roll this edit back.

datasets/xk/assembly/crawler.py
herrhellana: hm it seems like i removed it by mistake, sorry!

pudo:
Thanks for the fixes!

PR #4498, opened 2026-06-11
zavod: serve /artifacts as canonical URLs; clarify archive vs publish

jbothma:
Main part of https://github.com/opensanctions/operations/issues/2587

## Summary

- **Terminology**: settle on **archive** for `/artifacts/` and **publish** for `/datasets/`. The archive is the immutable, per-version store of everything we persist about a run; the published locations are server-side copies of the public subset, kept for back-compat with customers using stable URLs.
  - Module docstring on `zavod/archive/__init__.py` documents this.
  - `publish_artifact` (upload to `/artifacts/`) → renamed to `archive_artifact`.
  - `republish_resource_from_artifact` (server-side copy `/artifacts/` → `/datasets/`) → renamed to `publish_artifact`.
  - `publish_dataset_version` → renamed to `archive_version_history`.
  - `ARTIFACT_FILES` → renamed to `EXTRA_ARTIFACTS` (the persisted files that aren't registered DatasetResources). New `UNLISTED_RESOURCES` names the registered DatasetResources we don't list publicly (statistics, delta).
  - `publish_resource` (the old "upload directly to `/datasets/`" path) is removed — superseded by `archive_artifact` + `publish_artifact`.

- **Behavior change** (implements [opensanctions/operations#2587](https://github.com/opensanctions/operations/issues/2587)): every per-dataset file is now uploaded to `/artifacts/{dataset}/{version}/{file}` first and then server-side copied into `/datasets/{RELEASE}/{dataset}/` and `/datasets/latest/{dataset}/`. The `/artifacts/` URL is the canonical, immutable one we surface in metadata (`index_url`, `resources[*].url`, `statements_url`); `/datasets/` copies are preserved so customers using hardcoded URLs keep working without 3xx redirects. Re-runs on the same day no longer overwrite a unique URL — the artifact has a per-version path, while `/datasets/{date_stamp}/...` and `/datasets/latest/...` are intentionally the "latest successful run" view.

- **GCS cache headers on copies**: `GoogleCloudObject.republish` now always patches `cache_control` on the destination so a `/datasets/` copy doesn't inherit the long TTL of the artifact it was copied from.

- **Metadata responsibility split**: `get_base_dataset_metadata` no longer takes a `version` argument or stamps any URLs. `write_dataset_index` is the single place that stamps `version`, `index_url`, and the per-resource URLs.

## Test plan

- [x] `cd opensanctions/zavod && ../env/bin/ruff check zavod/` — passes
[...]

PR #4682, opened 2026-06-25
Comprehensively test publishing, remove missing file log

jbothma:
Fix for https://github.com/opensanctions/operations/issues/2587#issuecomment-4776749157

Before https://github.com/opensanctions/opensanctions/pull/4498/changes#diff-3756d4832a5703849acd40d473a2a5dbd8e8d9817ba734b9cbcbf3c7f5fe740dL65 we were logging for missing meta files and resources. After it, we started logging for any missing archived or published files.

This change drops logging for all archived files, but tests every expected archived file.

I think the broader fix is for the [publish step to know exactly all the files expected to exist](https://github.com/opensanctions/operations/issues/2647), but that's out of scope for https://github.com/opensanctions/operations/issues/2587

### DIFF
--- a/zavod/zavod/publish.py
+++ b/zavod/zavod/publish.py
@@ -50,7 +50,6 @@ def _archive_artifacts(dataset: Dataset, extra_artifacts: list[str] = []) -> Non
     for artifact in extra_artifacts:
         path = dataset_resource_path(dataset.name, artifact)
         if not path.is_file():
-            log.error("Resource not found: %s" % path, dataset=dataset.name)
             continue
         archive_artifact(
             path,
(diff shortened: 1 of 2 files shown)


leonhandreke (approved):
Nice improvement for the test coverage! For my understanding: `default` for example does not write a `statements.pack`, and if there are no issues, no `issues.log` either, but they are still iterated as part of `EXTRA_ARTIFACTS`, and that's why they are showing an log message now, correct?

jbothma:
Correct. And I believe we've always been attempting to write these files as part of ARTIFACT_FILES, and just failing silently when they don't exist. 

I started on a rampage to bring more order to it, but doing so for the delta exporter which has non-resource files and is itself optional, it goes way beyond scope and needs a better think. Tests should cover our backsides. Raising during the publish stage would just be another layer of safety.

PR #4694, opened 2026-06-25
[cz_national_sanctions] xlsx is back

### LINKED ISSUE #4596: [cz_national_sanctions] assert len(anchors) > 1
excel file has been replaced by PDF. CSV page still says it'll be updated soon since 2025.

I phoned - they said they're actively working on improving the format.

I emailed requests, specifically excel or CSV, separating delisting information from name column, and removing corrected values.

I also asked for the excel version to replace the PDF version in the short term.

We really ought to switch this to a google sheet that we manually update, but then we need to detect what changed etc etc.

The recent changes were delisting `ROGACH/ROGAČ/РОГАЧ Konstantin Valerievych / Konstantin
Valerjevič / Константин Валериевич` because they're listed in EU sanctions, so still sanctioned. And date corrections to `Al-Qard Al-Hassan Association`.

Let's give them a couple of days - if we're lucky we get the excel file back and push the rework out until they fix the format more properly.

nvmbrasserie:


nvmbrasserie:
fixes https://github.com/opensanctions/opensanctions/issues/4596

jbothma (approved):
thanks - it's not glamorous, but it's good work

PR #4695, opened 2026-06-25
[wd_categories] Tag our position names as
  English

leonhandreke:
The position names we define ourselves in the wd_categories config are always
written in English, so we now tag them with `lang="eng"` in `make_position`.

While doing that I went through all 273 position names we define and found
exactly one that was actually written in Russian:
`Глава парламента субъекта Российской Федерации`. Translated it to
"Chairman of a Regional Parliament of Russia" — matches the common Wikipedia
term [Regional parliaments of Russia](https://en.wikipedia.org/wiki/Regional_parliaments_of_Russia)
and the "Chairman" convention used by the other federal-level entries.


jbothma (approved):
might be worth adding this as a comment where `categories` starts in the yml but who knows if anyone will see it - the file is pretty unwieldy already.

PR #4696, opened 2026-06-25
[ua_war_sanctions] crawl only the public website, retire the API crawler

nvmbrasserie:
Point entry_point at crawler_zyte.py and remove the transitional dual-crawler setup: crawler_transition.py (ran both in parallel) and crawler.py (the confidential JSON API crawler) are deleted. Switch data.format JSON -> HTML to match the server-rendered site, and re-enable the max assertions now that entities are no longer emitted twice under parallel ua-ws-* / imo-* ids.


pudo (approved):
Thank you!

PR #4697, opened 2026-06-25
make_position translate_if_required (+ util.LangText, position translate helper)

leonhandreke:
Stacked on #4514 — based on the `leonhandreke/wikidata-position-llm-translation` branch, not main, so review/merge that one first.

Pulls the Wikidata position-label translation logic out into something reusable and uses it in another crawler:

- **`zavod.util.LangText`** — a general-purpose `NamedTuple` (text + lang) for translated/transliterated strings, kept separate from the pydantic `LangText` in `zavod.extract.names.clean` (that one's part of the persisted name-cleaning review model and has to stay stable). `TranslationResult` uses this one now.
- **`translate_position_name`** in `shed/trans.py` — takes a `LangText`, builds the position prompt, runs the translation, and returns a `TranslationResult`. Callers read the English value via `result.get_english()` and use `result.origin` (the model) as the statement origin. `wikidata/position.py` just calls it.
- **`make_position(..., translate_if_required=True)`** — translates a non-English position name to English (original kept as `original_value`), but keys the entity id on the *untranslated* name so ids stay stable and independent of the LLM output. `cz_pep_declarations` opts in, so Czech labels surface in English.

### On the `translate_if_required` design

An alternative would have been to let `make_position` take `LangText`s instead, and call `translate_position_name` at the callsite. That would have been the better design from a composition-over-inheritance standpoint — `make_position` stays dumb about translation, and crawlers compose the translate step themselves. But here I opted for the minimal change to the widely-used `make_position` instead, since threading translation through every callsite (and re-keying concerns) felt like more churn than it's worth right now. WDYT?

### DIFF
--- a/zavod/zavod/helpers/positions.py
+++ b/zavod/zavod/helpers/positions.py
@@ -8,6 +8,7 @@ from zavod import settings
 from zavod.constants import ORIGIN_INFERRED
 from zavod.context import Context
 from zavod.entity import Entity
+from zavod.util import LangText
 from zavod.stateful.positions import (
     DEFAULT_AFTER_OFFICE,
     OccupancyStatus,
@@ -33,6 +34,7 @@ def make_position(
     source_url: Optional[str] = None,
     lang: Optional[str] = None,
     id_hash_prefix: Optional[str] = None,
+    translate_name: bool = False,
 ) -> Entity:
     """Creates a Position entity.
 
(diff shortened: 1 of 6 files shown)


zavod/zavod/util.py
```diff
@@ -12,6 +12,21 @@
 ID_SEP = "-"
 
 
+class LangText(NamedTuple):
```
pudo: I wonder if our lives want more @dataclasses ?

zavod/zavod/helpers/positions.py
```diff
@@ -33,6 +34,7 @@ def make_position(
     source_url: Optional[str] = None,
     lang: Optional[str] = None,
     id_hash_prefix: Optional[str] = None,
+    translate_if_required: bool = False,
```
pudo: `translate_name`? Not doing unnecessary work should "go without saying" :)

pudo (commented):
Love where this is going!

zavod/zavod/helpers/positions.py
leonhandreke: Yes!

I still think doing the translation at the caller might be the nicer interface in the long run, but then we also need to transfer the original to key on uncleaned data (although with position IDs we aren't as strict anyway I think?)

zavod/zavod/util.py
leonhandreke: Done.

PR #4700, opened 2026-06-25
[np_parliament] Add Nepal Federal Parliament PEP crawler

### LINKED ISSUE #771: Nepal Federal Parliament (House of Representatives + National Assembly)
### Data URL

https://hr.parliament.gov.np/en/members (House) / https://na.parliament.gov.np/en/members (National Assembly)

### Publisher

Federal Parliament of Nepal (Sanghiya Sansad)

np

### Type of data

PEP (Members of Parliament)

### Coverage region

region:Asia/Pacific

### Can you tell us more?

Nepal's bicameral Federal Parliament exposes clean public JSON APIs on both chambers — no HTML scraping needed.

**House of Representatives** (lower house, 275 members = 165 FPTP + 110 PR)
- API: `https://hr.parliament.gov.np/api/v1/members`
- Bulk export: `https://hr.parliament.gov.np/api/v1/members?getAllMemberDownload=true` returns all members in one ~2MB call (~833 records, no pagination)
- The `/en/members` page is just a jQuery/AJAX shell over the API

**National Assembly** (upper house, 59 members = 56 STV + 3 presidential appointees)
- API: `https://na.parliament.gov.np/api/v1/members` (~173 records, same field set)
- Staggered 6-year terms, one-third rotating every 2 years — never fully dissolved, so track per-member `tenure_end_date` rather than whole-body dissolution.

**Fields** (bilingual EN/Nepali): name, district/constituency, party, election type, gender, DOB, `tenure_end_date`, photo (original + thumbnail).

**Implementation notes:**
- **TLS cert-chain error** ("unable to verify the first certificate") on all `*.parliament.gov.np` hosts. `curl -k` / relaxed verification works. The crawler must handle this.
- Records mix `member_type` (member, staff, secretariat, speaker, vicespeaker) — **filter to MP types**, drop secretariat staff.
- `tenure_end_date` cleanly separates historical (past) from current (null/future), so a single source covers both current and historical PEP coverage. Supersedes the rationale behind the closed #105 (Nepal MHA) for parliamentary PEPs.

Found via research sweep (June 2026).

- [X] I understand.

pudo:
## Summary

Adds a PEP crawler for both chambers of Nepal's Federal Parliament (Sanghiya Sansad):
- **House of Representatives** (Pratinidhi Sabha) — `hr.parliament.gov.np`
- **National Assembly** (Rastriya Sabha) — `na.parliament.gov.np`

Both expose the same JSON shape; the bulk export returns members of every term, so the dataset covers current and former parliamentarians in one pass. One YAML, one crawler; entity IDs are prefixed by chamber (`np-pa-hr-…` / `np-pa-na-…`).

Closes opensanctions/crawler-planning#771

## Output

950 Persons, 2 Positions, 950 Occupancies (337 current / 613 ended). Verified: `mypy --strict` clean, crawl with no warnings, `zavod validate` passes, and all four entity-integrity checks empty.

## Notable source-handling decisions

- **TLS**: `*.parliament.gov.np` serves the leaf cert but omits the Sectigo intermediate, so the chain can't be verified — the crawler sets `context.http.verify = False` (the source's own docs rely on `curl -k`).
- **Member filtering**: keeps `member`/`speaker`/`vicespeaker`; drops secretariat/staff. Placeholder `-` name records are skipped.
- **Gender / sentinel dates**: `0`/`1` mapped via a `type.gender` lookup; the `0000-00-00` tenure sentinel nulled via a `type.date` lookup.
- **Citizenship** `np`, per Constitution of Nepal art. 87.
- **Positions**: HR uses Wikidata `Q61704026`; the National Assembly has no Wikidata position item, so it gets a name-derived id.
- **Birth dates**: the source `dob` field is messy (mixed Gregorian / Bikram Sambat calendars, Excel serials, Devanagari digits). Emitted on a best-effort basis at year precision (BS→Gregorian). Excel-serial values were investigated and deliberately **not** parsed — cross-checking against sourced Wikidata DOBs showed the serial encoding is unreliable (off by up to a decade), so decoding them would inject wrong birth years.

### DIFF
--- a/datasets/np/parliament/crawler.py
+++ b/datasets/np/parliament/crawler.py
@@ -75,20 +75,14 @@ IGNORE = [
 
 
 def parse_dob(raw: str | None) -> str | None:
-    """Best-effort Gregorian birth date from the messy ``dob`` field.
+    """Best-effort Gregorian birth *year* from the ``dob`` field.
 
-    Reach for this because the source mixes Gregorian dates, Bikram Sambat (BS, ~57
-    years ahead of the Gregorian calendar), Excel serial numbers and Devanagari digits
-    in one field. Only dashed ``YYYY-MM-DD`` values are trusted. In this data Gregorian
-    birth years stop at 1973 while BS values start at 2002, so a year from 2000 on is
-    BS and is folded to a Gregorian *year* — BS months don't map onto Gregorian ones,
-    so the day and month are dropped. Earlier values are already Gregorian and kept in
-    full. The result is returned only if it is a plausible adult birth year.
-
-    The BS new year falls in mid-April (month 1 = Baishakh), so a single BS year spans
-    two Gregorian ones: months 1–9 land in BS−57, months 10–12 (Magh/Falgun/Chaitra,
-    ≈ Jan–Apr) in BS−56. A flat −57 would put anyone born in those late months a year
-    too early.
+    Every parseable member value is a Bikram Sambat (BS) date, so this always converts
+    from BS; only strict ``YYYY-MM-DD`` values are trusted (the field also carries Excel
+    serials, Devanagari digits and placeholders, which are dropped). BS runs ~57 years
+    ahead and its new year is in mid-April, so months 1–9 map to BS−57 and months 10–12
+    to BS−56. BS months don't align with Gregorian ones, so only the year is kept, and
+    only when it falls in a plausible adult birth range.
     """
     if raw is None:
         return None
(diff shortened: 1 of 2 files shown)


datasets/np/parliament/crawler.py
```diff
@@ -0,0 +1,180 @@
...
+    "district",
+    "representation_type",
+    "election_type",
+    "name",  # top-level convenience copy of the "en" translation name
+    "designation",
+    "description",
+]
+
+
+def parse_dob(raw: str | None) -> str | None:
```
nvmbrasserie: parse_dob turns the dob field into a Gregorian birth year. It assumes every date is in the Bikram Sambat calendar, which is true now but isn't guaranteed — the source never says which calendar a date uses. If a normal Western date ever shows up in the same format, it'll be quietly shifted back ~57 years and come out wrong, with no error.

A couple of smaller things, all silent: dates in the ninth Nepali month can land a year off, and anything outside the expected year range or not in YYYY-MM-DD form is just dropped without warning.

We might consider dropping `dob`s entirely. @brrttwrks what do you think?

nvmbrasserie (commented):
should we drop `dob`s considering the fragility of our current approach? everything else looks good

pudo:
maybe we can cut off under some magic threshold (people over 100 years old)?

pudo:
updated plan: drop DOB (unless there's an error-free heuristic)

herrhellana:
dropped DOBs and pushed some cosmetic changes @pudo @nvmbrasserie

PR #4702, opened 2026-06-25
[eu_journal_sanctions] Resolve consolidated acts via CELLAR instead of EUR-Lex

pudo:
## Problem

The published run surfaced 3 warnings — `Could not find original act in amendment relations table` for `CELEX:32025R1982` (the Iran implementing regulation). The consolidated-text cross-verification (which checks that a listed name actually appears in the consolidated regulation text) scraped `eur-lex.europa.eu` through a headless browser. That front end sits behind an AWS WAF and renders the amendment-relations table via a post-load AJAX call, so the fetch was occasionally snapshotted **and cached** before the table populated → no relations parsed → warning. The path never affected emitted data, only the verification.

## Change

Move the entire verification path off the WAF-protected front end onto the Publications Office **CELLAR** repository:

- **`get_consolidated_url` → `get_consolidated_celex`**: one SPARQL query to the CELLAR CDM graph resolves both the framework act a notice amends *and* that framework's consolidated versions — replacing two browser-rendered scrapes (the relations table and the `consLegVersions` nav).
  - Uses the `amends` predicate only (`based_on` on CFSP decisions points at the TEU article, not the framework).
  - Picks the latest consolidation in **Python** (`max()`), because the endpoint's Virtuoso `MAX()` aggregate does **not** return the lexicographic maximum for `xsd:string` (it returned a 2005 version instead of 2026).
- **`get_consolidated_text` / `extract_program_code`**: fetch document XHTML from CELLAR over plain HTTP (parsing re-encoded bytes to tolerate the XML encoding declaration).

The crawler **no longer depends on Zyte** at all. All XPath now goes through the typed `h.xpath_element(s)` / `h.element_text` helpers.

## Verification

- Ran `zavod crawl` against the latest published version (`20260625202503-ifq`): **`changed=0`** — emitted data is byte-for-byte identical.
- Issues: **3 published warnings → 0**. (An intermediate run surfaced one stale-text false positive that pinned down the Virtuoso `MAX()` quirk, now fixed.)
- `ruff` clean; `mypy --strict` down to 1 pre-existing error (untouched code), from 20.

### DIFF
--- a/datasets/eu/journal_sanctions/crawler.py
+++ b/datasets/eu/journal_sanctions/crawler.py
@@ -3,11 +3,12 @@ import re
 from functools import cache
 from datetime import timedelta, datetime
 
-from normality import normalize, squash_spaces
+from lxml import html
+from lxml.etree import _Element
+from normality import normalize
 from nomenklatura.resolver import Linker
 from rigour.ids.ogrn import OGRN
 
-from zavod.extract.zyte_api import fetch_html
 from zavod.integration import get_dataset_linker
 from zavod import Context, Entity, helpers as h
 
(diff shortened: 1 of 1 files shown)


jbothma (approved):
Nice improvement of the previous version. Everything sounds sane.

I don't love claude telling us what it's replacing in docstrings. It's like it forgot that PRs and git commit messages exist.

I look forward to getting rid of this script entirely.

pudo:
Sam Altman got ya covered.

PR #4712, opened 2026-06-26
[pl_mswia_sanctions] Fix dirty identifiers in details parsing

nvmbrasserie:
Add REGON and colon variants to CHOPSKA so compound identifier strings like "KRS X, REGON Y, NIP Z" are split cleanly without one keyword swallowing another. Dirty values that still contain punctuation are resolved via the details lookup rather than emitted as-is.


nvmbrasserie:
fixes https://github.com/opensanctions/operations/issues/2646

PR #4714, opened 2026-06-26
Migrate sk/bg/me PEP crawlers to make_position(translate_name=True)

leonhandreke:
Follow-up to https://github.com/opensanctions/opensanctions/pull/4697 and part of https://github.com/opensanctions/opensanctions/issues/2874.

Now that `make_position(..., translate_name=True)` exists, these three crawlers can drop the old `make_position` + `apply_translit_full_name(POSITION_PROMPT)` dance and just pass the flag:

- `sk/public_officials` (`slk`)
- `bg/jud_declarations` (`bul`)
- `me/acp_peps` (`cnr`) — also adds the previously-missing `lang="cnr"` on the `make_position` call

The prompt and model are identical to what these crawlers used before (both route through `make_position_translation_prompt` + the default model, English-only output), so the translation itself doesn't change.

What *does* change is the shape: instead of two parallel `name` values (original + English), the position now carries a single English `name` with the original kept as `original_value` and the model as `origin` — i.e. the cleaner cz_pep_declarations behavior. So expect the position names to shift on the next crawl.

### DIFF
--- a/datasets/bg/jud_declarations/crawler.py
+++ b/datasets/bg/jud_declarations/crawler.py
@@ -8,15 +8,8 @@ from typing import Dict, Generator, List, Optional
 
 from zavod import Context, helpers as h
 from zavod.stateful.positions import categorise, OccupancyStatus
-from zavod.shed.trans import (
-    apply_translit_full_name,
-    make_position_translation_prompt,
-    ENGLISH,
-)
 
 
-TRANSLIT_OUTPUT = [ENGLISH]
-POSITION_PROMPT = prompt = make_position_translation_prompt("bul")
 # e.g. (name_html, name_pdf)
 # {
 #     ("Георги Иванов Иванов", "ВАНЯ СТОЯНОВА ЛАКОВСКА"),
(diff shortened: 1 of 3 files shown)


leonhandreke:
Some API design questions here:

As highlighted in the description, this drops the original name in favor of only English. Do we want to add the original name back? Always? Configurable?

I think my preferred solution would still be to add a `names : list[Union[str, LangText]]` argument to `make_position` and do the translation call at the callsite. But then we'd have to add a `key_name` as well, to build the key, which might be a good idea anyway since we're often using cleaned data to build the id now?

Also, in `ge_declarations`, we translate to Russian and Arabic for some reason. Does anyone remember why? This would also work better with this pattern.

jbothma (approved):
❤️

jbothma:
I don't think I thought about this before, but it's great that positions were originally keyed on the original name, and will now continue to be keyed on original name.

pudo (approved):
Excited for this :)

leonhandreke:
Since no opinions on non-English names, will merge as is.

PR #4716, opened 2026-06-26
[jp_mof_sanctions] Name migration step 3

nvmbrasserie:


datasets/jp/mof_sanctions/crawler.py
jbothma: I think we should be removing these too, while making sure they're covered by the namespecs

sorry, I see I only mentioned parse_names in the original review.

datasets/jp/mof_sanctions/crawler.py
```diff
@@ -157,19 +120,22 @@ def emit_row(
...
+    for n in chain(
+        h.multi_split(row.pop("alias", []), ALIAS_SPLITS),
+        row.pop("known_alias", []),
+    ):
+        original.add("alias", n)
+    for n in chain(row.pop("past_alias", []), row.pop("old_name", [])):
+        original.add("previousName", n)
+    for n in chain(
+        h.multi_split(row.pop("weak_alias", []), ALIAS_SPLITS),
+        h.multi_split(row.pop("nickname", []), ALIAS_SPLITS),
```
pudo: We're sure we want to split before sending to reviews here? If it's thousands of extra fixes, it may be worth it.

pudo (commented):
There's a little devil inside of me that wonders if enabling LLM cleaning on this (with GPT-5.5/5.6) would be a savy move. The point is that none of us speak Japanese, so we'd be shit out of luck if there was a "here's a comment" text hidden in one of the names...

datasets/jp/mof_sanctions/crawler.py
```diff
@@ -176,13 +137,6 @@ def emit_row(
     raw_weak_alias = row.pop("weak_alias", [])
     raw_nickname = row.pop("nickname", [])
     english_first, english_rest = split_english_names(name_english)
```
jbothma: I thought this was going away in step 3 - I don't think having some latin script transliterations tagged as `eng` is valuable enough to keep this in here. I don't think we want this to live on in this crawler, so we might need one more quick step1 PR to get rid of split_english_names, do those reviews, then we can merge this.

datasets/jp/mof_sanctions/crawler.py
```diff
@@ -26,7 +24,6 @@
 SPLITS = SPLITS + ["(i)", "(ii)", "(iii)", "(iv)", "(v)", "(vi)", "(vii)", "(viii)"]
 SPLITS = SPLITS + ["; a.k.a.", "; a.k.a ", ", a.k.a.", ", f.k.a."]
```
jbothma, line 28: I see that this is being used way at the root of this crawler where we're splitting basically every cell on SPLITS

removing this kinda means going back to square one on this crawler a bit, so I think let's keep this for now - we can come back to it if we really want to one day.

datasets/jp/mof_sanctions/crawler.py
nvmbrasserie: I prepared the PR [here](https://github.com/opensanctions/opensanctions/pull/4971), but with the `eng` tag removed, those names now fall to the dataset default `jpn`. We'll be mislabelling all English names, do you think we should keep this code in to prevent it?

jbothma: so that's been a bug since before we started the name migration, right?

I think fixing it is nice - one way to fix it I think is the to remove data.lang: jpn from the yaml and make sure that every other `add` that really contains japanese passes the arg?

datasets/jp/mof_sanctions/crawler.py
```diff
@@ -14,18 +14,19 @@
...
 from zavod import helpers as h
+from zavod.stateful.review import assert_all_accepted
 
-# Match non-brackets inside an opening and closing pair of brackets
-BRACKETED = re.compile(r"([(（][^\(\)]*[)）]|\[[^\[\]]*\])")
-
+# Cells throughout the sheet pack multiple values into one cell, numbered or
+# newline-separated, so every cell is split on SPLITS when the rows are read.
+# Names thus reach review already split, not as the raw cell. Kept anyway:
+# undoing it takes the crawler back to square one.
```
jbothma, line 22:
```suggestion
# Cells throughout the sheet pack multiple values into one cell, numbered or
# newline-separated, so every cell (incl. names) is split on SPLITS when the rows are read.
```

"Kept anyway" is claude describing the change, while we normally want comments to describe the state at that commit.

"Back to square one" was about taking name cleaning back to square one :)

datasets/jp/mof_sanctions/crawler.py
jbothma: thanks for the comment above - handy hint

jbothma:
`Delta export complete          [jp_mof_sanctions] added=0 dataset=jp_mof_sanctions deleted=0 metric=delta_counts modified=1178`

`ftm-stmt diff $(zavod archive url statements.pack --last-successful datasets/jp/mof_sanctions/jp_mof_sanctions.yml) data/datasets/jp_mof_sanctions/statements.pack`

Screenshot 2026-08-06 at 14 15 18

jbothma (approved):
Looks good, let's just decide whether this is disruptive or not

jbothma:
We're deciding it's not disruptive on the basis of the number of modified entities, and that this is not one of the critical datasets at the moment

Issue #4134, opened 2026-05-01
We're merging occupancies with only period* dates


jbothma:
<img width="845" height="579" alt="Image" src="https://github.com/user-attachments/assets/90bde8e1-daba-4acd-8f89-0d673c6b8233" />

Probably edge-dedupe. 

I think we should also find the IDs where this has happened and explode them.
https://www.opensanctions.org/entities/NK-JRLS8Fktty57ge5Kr4soZY/

pudo:
FIxed in #4718.

PR #4718, opened 2026-06-26
Redesign edge deduplication

### LINKED ISSUE #3710: Make edge deduper more aggressive
Current: 

> Rules implemented:
> 
> - Same source, target, schema and temporal extent (and extra keys specified for some schemata)
> - Merge edges with same temporal start and no end, into one of the edges with temporal end
>   - Only for allow-listed schemata - see that for why
> 
> Rules explored but likely problematic without further investigation:
> 
> - Merge edges without temporal extent into one of the edges with temporal extent
>   - Family - more specific fiancee might end less specific spouse prematurely.
>   - Directorship - might merge "is signatory for" without temporal extent into "is director of" with temporal extent
> 
> This could also propose candidates for human decisions.

I would like to propose: 

* Merge edges where one has dates and the other doesn't and there are no opposing featured props
* Merge edges where both have dates, but one's dates are a prefix of the others (eg. 2025 -> 2025-10-01)
* We can also define a few props that are always irrelevant: sourceUrl, recordId, etc.

pudo:
## Summary

- replace exact-key/common-start edge dedupe with endpoint/schema buckets, temporal compatibility, and protected-property conflict checks
- support partial ISO date overlap and skip ambiguous temporal bridge merges
- add regression coverage for Occupancy period edges and protected-property behavior

## Mechanism

The deduper processes edge entities in three conservative stages:

1. Bucket edges by exact schema and endpoints. Directed edges preserve endpoint order; undirected edges canonicalize it. Edges without exactly one source and one target are skipped.
2. Compare all schema-declared temporal start values against starts and end values against ends. Partial ISO dates are ranges, so `2025` is compatible with `2025-10-01`. Missing dates are treated as unspecified. A vague edge that is compatible with multiple mutually incompatible groups is considered ambiguous and excluded from all of them.
3. Reject candidate groups whose schema-specific protected properties conflict. Protected values are slug-normalized; empty values act as wildcards and unlisted provenance/noise properties are ignored.

Only groups passing all three stages produce positive resolver decisions.

## Coverage

Tests cover endpoint direction, malformed/multi-ended edges, exact schema separation, exact and partial temporal compatibility, ambiguous temporal bridges, normalized protected values, ignored provenance fields, and the historical multi-period Occupancy shape from #4134.

Fixes #3710.

Prevents recurrence of the over-merge described in #4134. Repairing its already-persisted resolver decisions remains separate work because resolver decisions are durable and this dedupe pass cannot split them.


pudo:
## Revised once-off cleanup: skip complex components

This supersedes the star-only cleanup proposal above.

Production analysis found 425,284 generated components. Of these, 424,301 (99.77%) contain exactly one NK ID and form ordinary NK-centred stars. The remaining 983 components contain multiple NK IDs and are deliberately excluded from this cleanup.

The guarded preflight produced:

```text
generated_rows       921570
ordinary_bridges          0
shared_members          386
mixed_nks                  0
eligible_components   424301
eligible_rows         915819
rows_to_expire        491518
```

The cleanup retains the lexically first ordinary-member binding for each eligible NK ID and soft-deletes the other generated bindings. It excludes every NK involved in an NK-to-NK bridge, every NK sharing an ordinary member with another NK, and every generated component touching another user's positive decision.

```sql
BEGIN;

WITH
generated AS MATERIALIZED (
    SELECT id, source, target
    FROM resolver
    WHERE deleted_at IS NULL
      AND judgement = 'positive'
      AND "user" = 'edge-dedupe'
),
star_edges AS (
    SELECT
        id,
        CASE WHEN source LIKE 'NK-%' THEN source ELSE target END AS nk_id,
        CASE WHEN source LIKE 'NK-%' THEN target ELSE source END AS member_id
    FROM generated
    WHERE (source LIKE 'NK-%') <> (target LIKE 'NK-%')
),
bridge_nks AS (
    SELECT source AS nk_id FROM generated
    WHERE source LIKE 'NK-%' AND target LIKE 'NK-%'
    UNION
    SELECT target FROM generated
    WHERE source LIKE 'NK-%' AND target LIKE 'NK-%'
),
shared_members AS (
    SELECT member_id
    FROM star_edges
    GROUP BY member_id
    HAVING count(DISTINCT nk_id) > 1
),
shared_nks AS (
    SELECT DISTINCT nk_id
    FROM star_edges
    WHERE member_id IN (SELECT member_id FROM shared_members)
),
generated_nodes AS (
    SELECT nk_id, nk_id AS node_id FROM star_edges
    UNION
    SELECT nk_id, member_id FROM star_edges
),
other_positive_nodes AS (
    SELECT source AS node_id
    FROM resolver
    WHERE deleted_at IS NULL
      AND judgement = 'positive'
      AND "user" <> 'edge-dedupe'
    UNION
    SELECT target
    FROM resolver
    WHERE deleted_at IS NULL
      AND judgement = 'positive'
      AND "user" <> 'edge-dedupe'
),
mixed_nks AS (
    SELECT DISTINCT generated_nodes.nk_id
    FROM generated_nodes
    JOIN other_positive_nodes USING (node_id)
),
unsafe_nks AS (
    SELECT nk_id FROM bridge_nks
    UNION
    SELECT nk_id FROM shared_nks
    UNION
    SELECT nk_id FROM mixed_nks
),
ranked AS (
    SELECT
        star_edges.*,
        row_number() OVER (
            PARTITION BY nk_id
            ORDER BY member_id, id
        ) AS position
    FROM star_edges
    WHERE nk_id NOT IN (SELECT nk_id FROM unsafe_nks)
),
params AS (
    SELECT left(
        to_char(
            transaction_timestamp() AT TIME ZONE 'UTC',
            'YYYY-MM-DD"T"HH24:MI:SS.US'
        ) || '+00:00',
        28
    ) AS cleanup_timestamp
),
updated AS (
    UPDATE resolver
    SET deleted_at = params.cleanup_timestamp
    FROM ranked, params
    WHERE resolver.id = ranked.id
      AND ranked.position > 1
    RETURNING resolver.id
)
SELECT
    params.cleanup_timestamp,
    count(updated.id) AS expired_rows
FROM params
LEFT JOIN updated ON true
GROUP BY params.cleanup_timestamp;
```

Before committing:

1. `expired_rows` must equal the preflight result: **491,518**.
2. Rerunning the preflight in the same transaction must report `rows_to_expire = 0`.
3. The 895 NK-to-NK rows and 386 shared members must remain unchanged.
4. Active generated rows must equal `921570 - 491518 = 430052`.

Commit only after those checks. Then run edge dedupe with a fresh aggregation store:

```shell
zavod dedupe-edges --rebuild-store datasets/_collections/default.yml
```

Exports and other resolver writers should remain paused between the SQL cleanup and the rebuilt dedupe run. This repairs the simple 99.77% of generated components while leaving the complex 0.23% for future transient-overlay work.

zavod/zavod/integration/edges.py
```diff
@@ -1,228 +1,358 @@
...
-                "Skipping common start grouping for unapproved schema %s (source=%s target=%s)"
-                % (
-                    common_start_key.schema.name,
-                    common_start_key.source.id,
-                    common_start_key.target.id,
-                )
-            )
+def date_ranges_compatible(left: DateRange, right: DateRange) -> bool:
+    """Check whether two parsed partial-date ranges overlap."""
+    return left.start <= right.end and right.start <= left.end
```
jbothma, line 157: using overlap is cool - nice

zavod/zavod/integration/edges.py
```diff
@@ -1,228 +1,358 @@
...
+
+    if left_ranges and right_ranges:
+        return any(
+            date_ranges_compatible(left_range, right_range)
+            for left_range in left_ranges
+            for right_range in right_ranges
+        )
+
+    # Date-typed values should normally be clean ISO dates. Fall back to exact
+    # equality for any legacy values that cannot be parsed as partial ISO dates.
```
jbothma: is claude imagining things a little here? intersecting start makes sense, but I think the comment doesn't

jbothma (approved):
cool! Glazed over a bit around connected_components, but generally I think I get it it looks sound.

It's maybe a little contradictory to merge a precise and imprecise date, except when two incompatible candidates are present. It might just be that the incompatible edge shows up in a few months. Maybe this is being too careful, but it could be interesting to have a warning if such an incompatible edge shows up later so we learn if it's a real or imagined scenario.

pudo:
Re merging precise and imprecise dates: yeah it's a bit icky, but in the end the data ends up representing the right thing: two precise links, one of which happens to have an imprecise merged into it. I am not sure we'd ever be able to divine which precise entry was "meant to be" described by the imprecise one, so in a weird way I'm not so sure it matters in practice. The thing that shouldn't happen: merging two incompatible precise entries. And I think that's avoided now.

Issue #3488, opened 2026-02-09
[wd_categories,peps] Next steps on data freshness


pudo:
We now have a nice shared codebase between `wd_categories` and `wd_peps`, so I think we could advance the two crawlers a bit:

* Wikidata merges are quite rare, but they do occur. The problem is that the entire OS ETL infra has hard boundaries on trying to merge two entities with different QIDs, so we'll never reflect that. Our best solution is to update all of our internal references to the new QID and pretend the other one never existed. Do we need to add a `resolver.rename_node` method for this? 
* With EP.org going live, we will see more pressure to bring down the latency between Wikidata edits and fresh data showing up in OS/PEPs. We can do a few things here:
  * Bring down the cache lifespan on the SPARQL queries we run in both of the crawlers. 
  * Include the `modifiedAt` for the item in the query. If the item has been modified in the last N (days? weeks?), we then force a cache refresh? Or we compare the `modifiedAt` in the SPARQL response to the one on the item we get from the cache? 

Ultimately, the cleanest thing to do would be to build a sort of event system that listens to Wikidata edits and runs cache purges on our end, but that seems like a significant piece of infrastructure, so I'd like to steer around that if at all possible.

PR #4720, opened 2026-06-26
Use modification date to invalidate modified wd items in cache

jbothma:
Part of https://github.com/opensanctions/opensanctions/issues/3488

Depends on https://github.com/opensanctions/nomenklatura/pull/332

Queries for the modification date of position holder items in sparql queries, then passes the modification date to the cache to use that as the cache cut-off and invalidate items cached before that date.

This has the effect that something that if a person was modified more recently than when we last fetched them, we invalidate the cache for their wikidata item and refetch.

### DIFF
--- a/datasets/_wikidata/categories/crawler.py
+++ b/datasets/_wikidata/categories/crawler.py
@@ -9,6 +9,7 @@ from nomenklatura.wikidata import Claim, WikidataClient
 from zavod.shed.wikidata.human import wikidata_basic_human
 from nomenklatura.wikidata.value import clean_wikidata_name
 from zavod.shed.wikidata.position import (
+    days_since_modified,
     position_holders,
     wikidata_occupancy,
     wikidata_position,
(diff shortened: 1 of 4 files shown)


pudo (commented):
I'm trying to wrap my head around the logic here: if an entity was modified 3 days ago, max cache age should be less than 3 days. But then the same entity is being crawled by the next wd_categories run 2 days later, and refreshed again, and again, etc. So this basically implements something more like incremental back-off, not "refresh once". 

What if we instead set a simple deadline: if the entity was modified in the last 7 days, cache_days=6, otherwise cache_days=default. That way we guarantee under normal ops that we'll refetch after an edit, but don't do it for a longer timeframe?

jbothma:
That wasn't very clever ... I only now wrote it down and either it's always invalidating, or it's never invalidating depending on offset.

- 2026-07-06: QID last modified 2026-06-22 (20 days ago)
  -  fetch(cache_days=20) -> cache HIT
- 2026-07-07: QID last modified 2026-06-22 (1 day ago
  - fetch(cache_Days=1) -> cache HIT

now I understand this point

> Or we compare the modifiedAt in the SPARQL response to the one on the item we get from the cache?

passing modification date to fetch_item is actually so simple - I'd like to try that quick

- 2026-07-06: QID last modified 2026-06-22 (14 days ago)
  - cached 2026-06-20 (6 days ago)
  - `last_modified >= cached` -> MISS
- 2026-07-07: QID last modified 2026-06-22 (1 day ago)
  - cached 2026-07-06
  - `last_modified < cached` -> HIT
- 2026-07-08: QID last modified 2026-07-08 (2 days ago)
  - cached 2026-07-07
  - `last_modified >= cached` -> MISS

PR #4727, opened 2026-06-29
[pl_mswia_sanctions] Replace regex details parsing with LLM extraction

nvmbrasserie:
The identification details field ("dane_identyfikacyjne_osoby/podmiotu") is free-form Polish text that previously required a brittle combination of suffix-stripping (CHOPSKA), regex date parsing, and an ever-growing manual lookup table to handle edge cases (bilingual addresses, non-Polish registry codes, citizenship notes, etc.).

Replace parse_details() and its helpers with extract_details(), which uses run_typed_text_prompt (gpt-5.4) to extract the same set of properties: taxNumber, innCode, registrationNumber, idNumber, address, birthDate, birthPlace, and citizenship. Results pass through review_extraction() so extracted values are human-reviewable before being accepted, and assert_all_accepted() enforces that all extractions have been reviewed at crawl end.

The details lookup in pl_mswia_sanctions.yml is now superseded and can be pruned once the LLM extractions have been reviewed and accepted.

### DIFF
--- a/datasets/pl/mswia_sanctions/crawler.py
+++ b/datasets/pl/mswia_sanctions/crawler.py
@@ -3,14 +3,13 @@ from pydantic import BaseModel
 
 from zavod import Context, Entity
 from zavod import helpers as h
-from zavod.extract.llm import run_typed_text_prompt
+from zavod.extract.llm import run_typed_text_prompt, DEFAULT_MODEL
 from zavod.stateful.review import (
     JSONSourceValue,
     assert_all_accepted,
     review_extraction,
 )
 
-LLM_MODEL_VERSION = "gpt-5.4"
 EXTRACT_PROMPT = """Extract structured data from an identification details string taken from a Polish sanctions list entry.
 
 The input is a Polish-language text that may contain various combinations of:
@@ -75,13 +74,13 @@ def extract_details(context: Context, entity: Entity, details: str) -> None:
         prompt=EXTRACT_PROMPT,
         string=details,
         response_type=DetailsExtractionResult,
-        model=LLM_MODEL_VERSION,
+        model=DEFAULT_MODEL,
     )
     review = review_extraction(
         context=context,
         source_value=source_value,
         original_extraction=result,
-        origin=LLM_MODEL_VERSION,
+        origin=DEFAULT_MODEL,
     )
     if not review.accepted:
         return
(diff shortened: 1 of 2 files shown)


nvmbrasserie:
fixes https://github.com/opensanctions/operations/issues/2646

datasets/pl/mswia_sanctions/crawler.py
```diff
@@ -198,3 +224,5 @@ def crawl(context: Context) -> None:
     )
     for row in h.parse_html_table(table, header_tag="td"):
         crawl_row(context, h.cells_to_str(row), "podmioty")
+
+    assert_all_accepted(context)
```
pudo: I don't think we should crash on un-reviewed. Does that make life better?

datasets/pl/mswia_sanctions/crawler.py
```diff
@@ -1,76 +1,102 @@
...
 from zavod import Context, Entity
 from zavod import helpers as h
+from zavod.extract.llm import run_typed_text_prompt
+from zavod.stateful.review import (
+    JSONSourceValue,
+    assert_all_accepted,
+    review_extraction,
+)
+
+LLM_MODEL_VERSION = "gpt-5.4"
```
pudo: Should this be specified here? I'd prefer to go with a framework-wide default.

pudo (commented):
I understand that gpt-5.4 is probably better results - what would be needed to make that our default for new invocations?

datasets/pl/mswia_sanctions/crawler.py
nvmbrasserie: my suggestion would be to crash on it once before we do the first full round of reviews to minimise the delta. and then remove it afterwards

datasets/pl/mswia_sanctions/crawler.py
nvmbrasserie: The default is the `DEFAULT_MODEL = "gpt-4o"` constant in `zavod/zavod/extract/llm.py`. To change the default for all new invocations, we could just update that constant to `"gpt-5.4"` and pass an explicit `model=` argument with 4o to all older crawlers

pudo (approved):
Nice! Let's see :)

brrttwrks (commented):
I think @pudo pudo raised the main things. I've added to those, but I don't have anything else.

datasets/pl/mswia_sanctions/crawler.py
brrttwrks: I had some concern that if we used a config declared model version that if we change that, it could cause a cascade of changes that all need new reviews. 

Is that a misunderstanding? Or a non-issue?

pudo:
I asked Mr. Claude yesterday on changing the model. Turns out: only cache is keyed on it, so it would not have any immediate consequences at all. Reviewed stay reviewed, unreviewed candidates remain until their cache ttl expires.

PR #4743, opened 2026-06-30
[us_plum_book] Widen Person assertion bounds for duplicate persons

leonhandreke:
The PLUM dataset suddenly grew a lot of new entries and they're largely duplicate persons — bumping the assertion bounds (min 6700→9000, max 9000→12500) so the crawler stays green, and documenting what's actually going on below.

## What's happening

The duplicates are the same individual showing up as several `Person` entities. The cause is in the ID:

`person.id = context.make_id(first_name, last_name, position_name)`

`position_name` includes the position title, so one human who held N positions becomes N separate `Person` entities. The source recently started listing each appointee's full chain of (often acting/career) roles during the 2025 administration transition — mostly at EPA — where people moved through a ladder of positions with consecutive, non-overlapping date ranges. E.g. Rafael Deleon at EPA:

| Position | Begin | Vacate |
|---|---|---|
| Principal Deputy Asst Administrator, Int'l & Tribal Affairs | 2022-02-27 | 2025-01-30 |
| Assistant Administrator, Int'l & Tribal Affairs (PAS) | 2025-01-20 | 2025-01-30 |
| Deputy Asst Administrator for Management (CA) | 2025-01-30 | 2025-07-26 |
| Senior Advisor (CA) | 2025-07-27 | 2025-12-31 |

Each row becomes its own `Person`, and we currently emit 12,165 of them.

## Why we can't key on a person ID

The [PLUM website](https://www.opm.gov/about-us/open-government/plum-reporting/plum-data/) lists 9,797 records and shows a unique person ID per incumbent — but that ID isn't in the CSV we download. The CSV's only columns are:

`AgencyName, OrganizationName, PositionTitle, PositionStatus, AppointmentTypeDescription, ExpirationDate, LevelGradePay, Location, IncumbentFirstName, IncumbentLastName, PaymentPlanDescription, Tenure, IncumbentBeginDate, IncumbentVacateDate`

So the only thing identifying a person in our source data is the name, which isn't unique either (there are two unrelated "William Burns"). We don't get the unique IDs from the source.

### DIFF
--- a/datasets/us/plum_book/us_plum_book.yml
+++ b/datasets/us/plum_book/us_plum_book.yml
@@ -58,10 +58,10 @@ dates:
 assertions:
   min:
     schema_entities:
-      Person: 6700
+      Person: 9000
   max:
     schema_entities:
-      Person: 9000
+      Person: 12500
 
 lookups:
   type.date:


leonhandreke:
Screenshot 2026-07-03 at 11 31 02

For now, I'll raise the assertions to make the warning go away.

PR #4754, opened 2026-07-01
[lt_seimas] Bio prefix, position dedup, politicalGroup cleanup

leonhandreke:
Some cleanups to lt_seimas, mostly from re-reading the PEP crawler docs.

- **Biography prefix**: the biography table always leads with "Date of birth …
  Place of birth …", which we already capture as `birthDate`/`birthPlace`. Skip
  those two rows so the biography starts at the actual narrative.
- **Position dedup**: current members emitted `Member of the Seimas` *without*
  the Wikidata QID, while the historical crawler emitted it *with* `Q18507240`,
  so the same role split into two Position entities. Shared a single
  `make_seimas_position()` helper; now dedupes to one entity.
- **politicalGroup**: the `frakcija` (current) and "Political Groups of the
  Seimas" (historical) values are per-tenure parliamentary factions (the
  historical ones even carry per-term dates), not general party membership.
  Moved from `Person:political` to `Occupancy:politicalGroup` per peps.md.
- **Assertions**: bumped the stale single-term counts (max was 150) now that we
  crawl historical legislatures back to the after-office cutoff (~486 persons).

Verified: `mypy --strict` clean, crawl runs, positions dedupe to `Q18507240`,
`zavod validate` passes with 0 issues.

Note: left the person ID (`make_slug(person_name)`, which embeds the name and
shouldn't) untouched for a later pass.

### DIFF
--- a/datasets/lt/seimas/crawler.py
+++ b/datasets/lt/seimas/crawler.py
@@ -266,7 +266,9 @@ def crawl(context: Context) -> None:
         absolute_links=True,
     )
 
-    for anchor_url in h.xpath_strings(doc, members_list_validator + "/@href"):
+    current_member_urls = h.xpath_strings(doc, members_list_validator + "/@href")
+    context.log.info("Current legislature", member_link_count=len(current_member_urls))
+    for anchor_url in current_member_urls:
         crawl_member_bio(context, anchor_url)
 
     ### === crawl older legislatures === ###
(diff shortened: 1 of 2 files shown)


jbothma (approved):
looks good to me - any concerns about merging @leonhandreke

leonhandreke:
I'm a bit surprised about it raising the assertions - we're getting lots of people we didn't get before? But maybe it was just broken before and we didn't notice. Anyway, also looks good to me.

jbothma:
funny story, https://github.com/opensanctions/opensanctions/pull/4417 was merged 2 hours after the last successful run. That's where the historical terms were added, I think. And am I right in thinking claud's run of this PR was without a resolver? That'd explain the high numbers.

leonhandreke:
I don't think Claude ran it, it's one of the crawlers that take ~forever without a cache

jbothma:
ultimately the issue with reproducing these bigger assertions was that the position now had a qid, but the qid-based position wasn't categorised as national yet.

PR #4755, opened 2026-07-01
russia kremlin catalog of persons

### LINKED ISSUE #778: kremlin.ru Persons directory
### Data URL

http://en.kremlin.ru/catalog/persons

### Publisher

http://en.kremlin.ru/

ru

### Type of data

### Coverage region

region:Global

### Can you tell us more?

Directory of ~700 persons referenced on the President of Russia's site, A-Z indexed, each with a numeric ID: `en.kremlin.ru/catalog/persons/{id}/events` or `.../{id}/biography`.

**Structure**

- Listing page: name + link only, no position/dates.
- Per-person page: name, position/title (e.g. "Director of the Foreign Intelligence Service"), sometimes a location line.
- Two page variants per person:
  - `biography`: definition-list timeline of career milestones (year → free text), used for senior ru officials.
  - `events`: just a feed of news mentions, no structured bio — most entries (including all foreign heads of state/government) only have this.

**Licensing**

Site content is CC BY 4.0, per footer.

- [x] I understand.

nvmbrasserie:
fixes https://github.com/opensanctions/crawler-planning/issues/778


nvmbrasserie:
kremlin's official "persons" directory mixes russian officials, foreign leaders, and ex-officials, with no flag to tell them apart. 

Our fix: only crawl `/biography` pages — the kremlin writes biographies only for russian-state people, so it's a clean structural filter (and those pages are the only ones with birth/career detail).

From each we emit:

- `poi `on everyone who has a biography page (always ru senior officials) — it's a research dataset;
- Position + Occupancy (which adds PEP `role.pep`) only for those holding a current state position;

pudo (approved):
Methodology sounds good - nice filter.

nvmbrasserie:
Two-sided crawl: associates + PEPs

This crawler now pulls from the Kremlin site in two complementary ways, unified by catalogue id:

`crawl_associate` — the person directory (/catalog/persons). Each linked biography is parsed for name, birth/death details (via the details lookup), and the timeline. These are recorded as `persons of interest`, and any single listed job title is enrolled as a review-gated position (default_is_pep=None) — we don't assume a biography title makes someone a PEP.

`crawl_pep` — the membership listings of three presidential bodies (State Council, Security Council, Presidential Administration), read surface-only (no detail-page fetches). Membership of these bodies is itself the PEP signal, so each member's listed role is emitted as a current PEP position by default (default_is_pep=True).

Because both sides reuse the `/catalog/persons/<id>` id as the entity slug, a person appearing in both (e.g. an aide who also sits on a council) merges into one entity — gaining their biography from one side and a PEP occupancy from the other. Both paths share a single emit_position step; the only difference is that default classification.

datasets/ru/kremlin_persons/crawler.py
```diff
@@ -0,0 +1,244 @@
...
+    for link in h.xpath_elements(doc, ".//a[contains(@href, '/catalog/persons/')]"):
+        match = link_pattern.search(link.get("href", ""))
+        if match is not None:
+            person_ids.add(match.group(1))
+
+    # An associate and a council/administration member can be the same person;
+    # they merge by catalogue id, so an entity may gain both a biography and a
+    # PEP occupancy.
+    for person_id in person_ids:
+        crawl_associate(context, person_id)
```
jbothma: I think calling them all associate is a little confusing in our context, where associate is a very specific classification - especially in a PEP dataset - one which results in a relationship entity and a role.rca topic. That's not what we mean here - they're not all getting a relationship to putin, so let's call them something else.

```suggestion
    # A person with a biography and a council/administration member can be the same person;
    # they merge by catalogue id, so an entity may gain both a biography and a
    # PEP occupancy.
    for person_id in person_ids:
        crawl_biography(context, person_id)
```

datasets/ru/kremlin_persons/crawler.py
```diff
@@ -0,0 +1,244 @@
...
+
+    # The role sits in a sibling ``jobTitle`` within the same card block.
+    block = card.getparent()
+    roles = (
+        h.xpath_elements(block, ".//*[@itemprop='jobTitle']")
+        if block is not None
+        else []
+    )
+    if len(roles) == 1:
+        emit_position(context, person, h.element_text(roles[0]), default_is_pep=True)
```
jbothma, line 193: I suspect since we haven't applied details to this person, make_occupancy won't see their death date, and will give their occupancy current status, which isn't right. Let's apply details in this path too - caching shouldn't make it cheap to do so.

jbothma (commented):
fun! Just a naming thing and making sure make_occupancy sees deathDate then I think we can ship

I'm curious what the notes with newlines between sections will look like on the website - let's just check it looks reasonable before we release this.

datasets/ru/kremlin_persons/crawler.py
nvmbrasserie: These people have no details and are all current administration employees. They are only associated with one position, which is why we always list them as PEPs. Some of them also have biographies (but those are coming in through `crawl_biography`), but none of them have a death date

nvmbrasserie: the assumption here is that they are all peps working for either one of the admin branches: state council, security council, presidential administration

jbothma: oh ok - makes sense that the death dates are relevant to the others

jbothma (approved):

PR #4760, opened 2026-07-02
[vn_terrorist_orgs] crawler draft

brrttwrks:
I ended up excluding the English as it was a completely separate DOM with nothing to connect it to the Vietnamese. I guess we could manually create a mapping, but maintaining it would be a thing. Since it was clear that the Vietnamese was the definitive/primary source and had the complete data, including notes which weren't in the English, I sourced it only and made sure we used lang params in case we want to translate it in the future.


nvmbrasserie (approved):

nvmbrasserie (changes requested):
I just found English version of the website: https://en.bocongan.gov.vn/terrorist-group. I'll switch the crawler to parse it

nvmbrasserie (approved):
We prefer Vietnamese portal (Portal-Id 22) over the English one (Portal-Id 28). Only the Vietnamese portal exposes structured member records as JSON.

pudo (approved):

PR #4769, opened 2026-07-02
[br_pep] Translate position names

leonhandreke:
Turns on English translation of position names for `br_pep` via `translate_name=True`.

Split into its own PR because of its size — it's the only one whose delta is over 10k on its own, so it gets released separately.

Refs https://github.com/opensanctions/opensanctions/issues/2874 (doesn't close it, just chips away).

Only Position `name`s change; position IDs are keyed on the untranslated name, so they stay stable and occupancies/persons aren't re-emitted. Expected delta:

| dataset | positions (≈ delta) |
|---|--:|
| br_pep | 13,516 |

### DIFF
--- a/datasets/br/pep/crawler.py
+++ b/datasets/br/pep/crawler.py
@@ -44,7 +44,9 @@ def create_entity(raw_entity: Dict[str, Any], context: Context) -> None:
     person.add("citizenship", "br")
 
     position_name = f"{raw_entity['Descrição_Função']}, {raw_entity['Nome_Órgão']}"
-    position = h.make_position(context, position_name, country="br")
+    position = h.make_position(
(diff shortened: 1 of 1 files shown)


leonhandreke:
@goose-life wdyt, does this dataset having >10k position names make this change `disruptive`? It really isn't, other than that the language of the position name changes.

goose-life:
I'm inclined to agree with you but I worry that this is exactly the kind of improvement that gets interpreted as noise — so please hold off until I've had time to look at this more closely! (not today 😅)

goose-life:
okay I'm going to say because Position is a [contextual schema type](https://www.opensanctions.org/reference/#contextual-schema-types) it's not `disruptive`, despite the size of the delta 🙏

PR #4774, opened 2026-07-02
[zavod] Default property_fill_rate assertions for source datasets

leonhandreke:
Adds baseline `property_fill_rate` assertions to every dataset so we get a name-coverage floor for free, and makes the metric's semantics play nicely with defaults across multiple schemata.

## What
- **Default assertions** (`meta/dataset.py`): `min.property_fill_rate` of `0.95` on `name` for `Person`, `Organization`, `Company`, applied to all datasets.
- **Overridable in YAML**: merged at the config-dict level via a new `merge_assertions_config` deep-merge (`meta/assertion.py`), so a dataset's `assertions:` block wins at the leaf. e.g. `min.property_fill_rate.Person.name: 0.8` replaces just that threshold; set it to `0` to disable a `min` default.
- **Semantics change** (`validators/assertions.py`): `PROPERTY_FILL_RATE` now skips any schema the dataset doesn't emit (zero entities). Without this, a default spanning three schemata would spuriously fail a Person-only dataset (the lookup defaults to `0.0`). A schema that *is* present but never fills the property still fails.
- **"No assertions" nudge** stays based on the maintainer's YAML config (read from `dataset._data`), not the merged-in defaults — so datasets without explicit assertions still get nudged to add real ones (e.g. entity counts).

## On collections
Defaults now apply to collections too. Note that collections aren't actually validated as part of `zavod run` today (`cli/etl.py` gates `validate_dataset` on `not is_collection`), so the defaults are inert for them in the normal pipeline and only fire under a manual `zavod validate <collection>`. But that gate is a practical performance decision, not a design stance — collections *should* conceptually be validatable — so I didn't want to bake "collections have no assertions" into the model as an axiom. Applying the defaults uniformly keeps the door open for whenever we do validate collections.

## Notes / open questions
- **These defaults are fatal.** `min` → `GTE` → aborts the export. A 95% name floor as a hard failure across all sources at once is a real policy change — happy to land it as `max` (warn-only) first if we'd rather ease it in.
- Fill rate is counted by *concrete* schema, so the `Organization` default only covers entities emitted literally as `Organization` (not `Company`/`PublicBody`/etc.).

### DIFF
--- a/zavod/zavod/meta/dataset.py
+++ b/zavod/zavod/meta/dataset.py
@@ -44,6 +44,7 @@ DEFAULT_ASSERTIONS: Dict[str, Any] = {
     "min": {
         "property_fill_rate": {
             "Person": {"name": 0.95},
+            "LegalEntity": {"name": 0.95},
             "Organization": {"name": 0.95},
             "Company": {"name": 0.95},
         }
--- a/zavod/zavod/meta/dataset.py
+++ b/zavod/zavod/meta/dataset.py
@@ -13,7 +13,11 @@ from followthemoney.dataset.resource import DataResource
 from zavod import settings
 from zavod.archive import dataset_data_path
 from zavod.logs import get_logger
-from zavod.meta.assertion import Assertion, parse_assertions
+from zavod.meta.assertion import (
+    Assertion,
+    merge_assertions_config,
+    parse_assertions,
+)
 from zavod.meta.dates import DatesSpec
 from zavod.meta.http import HTTP
 from zavod.meta.model import DataModel, ZavodDatasetModel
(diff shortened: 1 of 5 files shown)


zavod/zavod/meta/dataset.py
```diff
@@ -26,6 +30,26 @@
...
+# exact schema/property while keeping the rest. Lower a threshold to 0 to
+# effectively disable a `min` default.
+#
+# `property_fill_rate` only applies to schemata the dataset actually emits: the
+# validator skips any schema with zero entities, so naming schemata a dataset
+# doesn't produce here is harmless.
+DEFAULT_ASSERTIONS: Dict[str, Any] = {
+    "min": {
+        "property_fill_rate": {
+            "Person": {"name": 0.95},
```
jbothma, line 46: why not legalentity?

zavod/zavod/tests/test_validate.py
```diff
@@ -140,11 +140,43 @@ def test_countries_count_assertion(testdataset3) -> None:
 
 
 def test_no_assertions_error() -> None:
```
jbothma, line 142: very thorough ❤️

zavod/zavod/tests/test_validate.py
```diff
@@ -140,11 +140,43 @@ def test_countries_count_assertion(testdataset3) -> None:
...
+    assert ds.is_collection is False
+    assert len(ds.assertions) > 0
+    validator, _ = run_validator(StatisticsAssertionsValidator, ds)
+    assert validator.abort is False
+
+
+def test_default_property_fill_rate_skips_absent_schema() -> None:
+    # The default Person/Organization/Company name fill-rate assertions must not
+    # fail a dataset that simply doesn't emit some of those schemata.
+    ds = Dataset(BASE_DATASET_CONFIG)
```
jbothma, line 164: maybe a little test here asserting that Company is in there to maintain the test's validity in case we remove the default assertion on Company, for example

jbothma (approved):
Nice! I think it's a good change, but I'm pretty sure there are datasets emitting without names that may violate this. Let's merge on a Monday or Tuesday morning, not a friday afternoon :)

zavod/zavod/tests/test_validate.py
leonhandreke: I added a separate test case for Company.

zavod/zavod/meta/dataset.py
leonhandreke: Good idea, done.

PR #4780, opened 2026-07-03
[ext_ror] Lower min Organization assertion to 30

pudo:
Lowers the minimum Organization entity assertion for the ROR enricher from 210 to 30 orgs.


pudo:
The idea here is to get this read for xref again to see if we can generate more matches.

Issue #4632, opened 2026-06-19
[us_finra_actions] unstable crawl results from source pagination + caching


nvmbrasserie:
`us_finra_actions` returns different entity sets between runs — records appear, disappear, and get replaced without matching changes on FINRA's side. 

Two cases flagged: the Albert Gary Shilling entities (https://www.opensanctions.org/entities/us-finra-albert-gary-shilling/ and https://www.opensanctions.org/entities/us-finra-a-gary-shilling-co-inc/ replaced) and "Adam A Pflum" (dropped, still live on FINRA https://www.finra.org/rules-guidance/oversight-enforcement/finra-disciplinary-actions?search=Adam+A+Pflum&firms=&individuals=&field_fda_case_id_txt=&field_core_official_dt%5Bmin%5D=&field_core_official_dt%5Bmax%5D=&field_fda_document_type_tax=All).

Cause is most probably on the source side, not our parsing:

1. **Pagination instability** — sorted newest-first, ~15 rows over ~1,290 pages.
   A new action at the top shifts every record down, so sequential paging skips
   and duplicates records when the list changes mid-crawl.
2. **Inconsistent caching** — pages intermittently return empty / "No results
   found" then fill back in on a later fetch. Looks like stale per-page caching.

Reported to FINRA (proposed a bulk export, or stable sort + cache invalidation).

Crawl-side mitigations to consider:
- [x] Retry empty/"No results" pages before treating them as empty
- [x] Detect page-count drift mid-crawl and abort instead of emitting a partial run
- [x] Sort by a stable key (case ID / `field_core_official_dt` asc) if supported

Dataset: https://www.opensanctions.org/datasets/us_finra_actions/

jbothma:
demo

```
curl -s https://data.opensanctions.org/artifacts/us_finra_actions/20260613051235-nri/delta.json     | jq -r '.versions[]'     | while read -r url; do         curl -s "$url" | grep -i 'SHILLING' | sed "s|^|$url\t|";       done
```

jbothma:
I think claude is getting hung up on the empty pages thing. 

The main thing is the listing page cache invalidating inconsistently so movement in the listing due to additions/removals can hide entities (server-side version of https://zavod.opensanctions.org/best_practices/caching/#issue-with-pagination-and-cached-pages) and the ordering on their site meaning deletions while we're crawling can move an entity off of the page we're currently crawling. 

But now I spell it out, the latter option is so rare and won't account for the flip flopping.

[...]

PR #4784, opened 2026-07-03
fix: harden us_finra_actions crawler against pagination/caching instability

mvanhorn:
## Summary

`us_finra_actions` no longer emits partial datasets when FINRA's listing pages misbehave. Intermittently-empty pages are retried through the existing Zyte fetch backoff, and the crawl aborts on pagination drift instead of publishing a shifted, incomplete run.

## Why this matters

Between runs the crawler was returning different entity sets with no corresponding change on FINRA's side. Issue #4632 traced two mechanisms: listing pages that intermittently render the "No results found" empty state because of inconsistent server-side caching, and newest-first pagination that shifts every record down a slot when a new action lands mid-crawl. @jbothma confirmed the flaky listing-page cache as the dominant cause. Entities like the Albert Gary Shilling records and "Adam A Pflum" dropped out of runs while still live on FINRA.

## Changes

- The Zyte unblock validator (`RESULT_ROW_VALIDATOR`, `datasets/us/finra_actions/crawler.py:21`) now requires a populated data row and rejects the `view-empty` marker, so empty-state pages fail validation and ride `zavod.extract.zyte_api.fetch_html`'s existing retry-with-backoff and cache-invalidation path. Persistent empties raise `UnblockFailedException` and abort the run. The old silent-skip branches in `crawl()` are removed, so a partial run cannot be emitted quietly.
- `crawl()` pins `max_page` on first observation and raises if `get_max_page()` later disagrees, aborting on mid-crawl pagination drift rather than continuing against a shifted listing. The previous good export stays published.
- `sourceUrl` hrefs are normalized with `urljoin` before emission so pages recovered by a validator-triggered retry still produce absolute URLs.

## Testing

`black --check`, `ruff check`, and a syntax check pass on the changed file. A live crawl was not run here since it needs Zyte API access; the empty-page path was traced against `fetch_html`'s retry and cache-invalidation behavior in `zavod/zavod/extract/zyte_api.py`.

This covers the retry-empty-pages and abort-on-drift items from the issue checklist. The stable-sort option and the FINRA-side cache fix remain open, so this mitigates rather than closes the flip-flopping.

Refs #4632

### DIFF
--- a/datasets/us/finra_actions/crawler.py
+++ b/datasets/us/finra_actions/crawler.py
@@ -1,15 +1,15 @@
 """
 # Occasional issues:
 
-## Crawl completes but fewer than asserted entities are emitted.
+FINRA listing pages are newest-first and served through inconsistent caches.
+Intermediate pages can temporarily render the "No results found" empty state
+even though reruns later return rows; pagination can also drift if the listing
+changes while a crawl is in progress.
 
-Running the crawler locally the next day results in the expected number
-of entities being emitted.
-
-The crawl runs to the same number of pages (1231 in querystring) as usual.
-It's not clear whether entities are removed and then new entities added by
-the time we check the issue, or whether there's a bug. Keeping an eye on this
-for a bit longer (2024-07-31)
+The Zyte fetch validator requires a populated table and rejects the empty-state
+marker so these pages are retried and, if persistent, abort the crawl instead
+of emitting a partial run. The crawler also aborts if the advertised last page
+changes after pagination has been established.
 """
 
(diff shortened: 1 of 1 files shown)


leonhandreke:
Hi Matt and welcome! This looks like a very nice fix, also crashing in case the total number of pages changes is a very nice touch. Thanks for taking the time to investigate and fix this!

mvanhorn:
Glad this landed - thanks @leonhandreke! The FINRA crawler should ride out pagination hiccups now.

PR #4791, opened 2026-07-04
Fable skill compact experiment

pudo:
Stacked on #4788. Compacts the refactor-crawler skill and extracts its underlying principle into the docs:

- New `zavod/docs/best_practices/strict_interpretation.md`: every source value within the crawler's scope is handled, explicitly ignored, or raises a signal — with scope/mechanism decisions, loudness guidance, and anti-patterns. Added to the mkdocs nav and cross-linked from `patterns.md` and the crawler guide.
- `refactor-crawler` skill slimmed from 17 inline pattern recipes to procedure only, delegating rules to the best-practices docs. Output changes are now allowed when they trace to a named methodological gain (baseline diff required), instead of demanding byte-identical output.
- Warnings framed as work items for the daily issues rotation (fixed within days, never permanent) rather than as per-milestone zero counts.
- Name cleaning declared out of scope for a refactor pass — changing it means the multi-PR review-system migration in `zavod/docs/extract/names.md`.
- `entity_id.md` states when a dataset counts as published (added to a collection).

### DIFF
--- a/.claude/docs/crawler-guide.md
+++ b/.claude/docs/crawler-guide.md
@@ -89,7 +89,7 @@ def crawl(context: Context) -> None:
 
 `context.data_url` is from `data.url` in the YAML.
 
-**Fail loudly.** Crash or warn on unexpected data — never silently emit ambiguous output.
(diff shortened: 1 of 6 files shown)


nvmbrasserie:
I like this a lot. It also makes perfect sense not to rewrite best practices inside the skill

PR #4797, opened 2026-07-06
Re-write QIDs during wd_ prod runs in the resolver

pudo:
I am not certain this would get us style points. It basically moves the re-write we to in `contrib/rewrite_qids.py` into the crawler, and will get rid of the Wikidata moved QID warnings. Downside: resolver is being written from a crawler.


jbothma (approved):
I guess it's not too insane. I see the original user's name gets moved along with the edge. It might be nice if the user was zavod or the crawler. That and/or an info log entry here could be nice.

I'm really itching to add some pytest tests to wd_categories - there are about 4 changes I need to make to the crawler to be able to test a change, then remember to remove before committing. Maybe sensitive functionality like this add weight to the idea that the wd* crawlers are a big enough deal for tests.

Issue #4782, opened 2026-07-03
Fix the root causes for why we didn't get alerted about `ext_ror` being empty


leonhandreke:
`ext_ror` has emitted no statements since `20251205145121-mea`, yet we've been consistently saying the following in `index.json`

```
  "entity_count": 388,
  "target_count": 0,
  "thing_count": 266,
```

Some codepath is still backfilling these statistics.

But then, why did we suddenly get alerted about this?

leonhandreke:
Alright, I went on a bit of an odyssey through the backfill machinery and the archive. The answer is worse than "some codepath is backfilling statistics": we were silently *republishing the December data* as fresh successful runs for six months. Timeline from the artifacts:

- `20251205145121-mea` is the last version with a real `statements.pack` (entity_count 388).
- From Dec 12 to early June, every weekly run is `result: success` with **no statements.pack archived under it**, and entity_count drifting 388 → 385 → 384. What happened on each of those runs:
  - The enricher emitted nothing, and the statement writer only opens on the first emit — so no local statements file was ever written.
  - `store.sync` then silently fell back to streaming the newest archived statements from the archive (`_iter_scope_statements`) — i.e. the Dec 5 data.
  - Validation passed on the resurrected entities; export recomputed statistics from them each run. The count drift is the dedupe graph evolving around the zombie data.
  - The fallback streams without writing a local file, and extra artifacts missing locally are silently skipped at publish — hence no statements.pack under those versions. That's also what kept the loop stable (the version-walk landed on Dec 5 every week instead of compounding copies-of-copies), and why the gap is visible in the archive at all.
- Around June 19 the runs flip to `result: failure` — but not because the enricher came back to life. The runs were still hollow, and validation was still looking at the backfilled Dec 5 statements. My current theory for what changed — unconfirmed, but it seems likely: garbage collection on the resolver. The statements that `validate` kept checking never ended up in the export, so the garbage collection hasn't seen them since December — and without the match decisions, the resurrected entities no longer consolidate into what the assertions expect (July 2: `Assertion schema_entities failed for Organization: 34 is not >= threshold 210`). So the alert was an accident: the assertions were checking zombie data all along, and only fired once the resolver eroded underneath it.
- From `20260703142259-cuz` it's actually fixed: 409 orgs, real 15MB statements.pack.

[...]

PR #4799, opened 2026-07-06
Write an empty statements file for crawls that emit nothing

leonhandreke:
The statement writer only opens on the first emit, so a crawler that completes without emitting anything leaves no local `statements.pack`. `store.sync` then silently streams the *previous* version's statements from the archive: validation passes on the old data and export republishes it as a fresh successful version. This is how `ext_ror` kept republishing its December data for half a year (https://github.com/opensanctions/opensanctions/issues/4782).

`Context.flush()` now touches an empty statements file if nothing was emitted. This happens at the end of the crawl rather than in `begin()`, so the file keeps meaning "complete output of a finished crawl" — `begin()` is also used by the export/validate contexts, where an empty statements.pack would be actively harmful. An empty crawl now yields an empty store view, and min entity count assertions abort the run.

Caveats: `--keep-data` still sees the previous run's local statements, and datasets without min-count assertions will now publish honestly-empty data instead of resurrecting old data. The archive fallback itself stays — collection exports legitimately depend on it.

Both new tests fail without the fix.

### DIFF
--- a/zavod/zavod/context.py
+++ b/zavod/zavod/context.py
@@ -173,6 +173,13 @@ class Context:
         """Flush the context to ensure all data is written to disk."""
         if self._db is not None:
             self._db.checkpoint()
+        # The statement writer is only opened on the first emit, so a crawl
+        # that emitted nothing would leave no statements file behind - and
+        # downstream stages would silently fall back to streaming a previous
+        # version's statements from the archive, republishing old data as a
+        # new successful run. Record the empty output explicitly instead.
+        if not self.dry_run and self._writer is None:
+            self._writer_path.touch()
 
     def close(self) -> None:
         """Flush and tear down the context."""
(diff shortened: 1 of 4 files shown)


leonhandreke:
@jbothma just for general vibes, you have dealt with empty `statements.pack` before, but I don't remember the discussion - does this sound scary?

jbothma:
Another way would be to look up the last successful version in versions.json and then stream by specified version id, as opposed to continuing streaming from the latest found file (which will now just be an empty file).

I'm ok with this as an interim fix, but I think we should move to this metadata-based approach in parallel with steps in a job being based on a version stamped somewhere to represent the current job (https://github.com/opensanctions/operations/issues/2675)

leonhandreke:
I'd argue that publishing an empty `statements.pack` is more correct than publishing none, even once we eventually move to the versioned-everything approach.

jbothma:
Yeah that's fair!

PR #4805, opened 2026-07-06
docs: position names via lang= and translate_name, not YAML lookups

leonhandreke:
Updates the crawler-pep skill (and the conflicting bit in `zavod/docs/peps.md`) on how position names should be passed to `h.make_position`:

- Always pass `lang=` declaring the language of the position name.
- Crawler-supplied names (the standard "Member of the ... Parliament" case) are always written in English, so `lang="eng"`.
- Source-supplied names in a non-English language get passed through as-is with `translate_name=True`, instead of translating them via a `position` lookup in the YAML. The helper keys the entity ID on the untranslated name, so IDs stay stable.

Also updated the skill's examples to match (the subnational variant now mirrors what e.g. `datasets/hr/public_officials/crawler.py` actually does).

### DIFF
--- a/.claude/skills/crawler-pep/SKILL.md
+++ b/.claude/skills/crawler-pep/SKILL.md
@@ -63,7 +63,7 @@ assertions:
 
 - Include `Position` counts in assertions when the crawler creates multiple position types.
 - `frequency` matches source update cadence (daily/weekly/monthly). PEP crawlers do not have to be monthly.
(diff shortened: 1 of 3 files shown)


leonhandreke:
@nvmbrasserie since I think you have more experience writing skill files for LLMs and also if you think this makes sense. I did this in a couple of crawlers over the past week.

nvmbrasserie (approved):

PR #4811, opened 2026-07-07
Document archive layout and how to investigate past dataset runs

leonhandreke:
This is a take on https://github.com/opensanctions/operations/issues/2674 — when something unclear has gone wrong with a dataset, Claude should be able to go dig through past runs on the archive without having the file layout explained to it (or reading half of zavod) every time.

Two pieces:

- The `zavod.archive` module docstring is now a proper layout reference: `/artifacts/` vs `/datasets/`, the root `versions.json` (window of version IDs + `last_successful`), how to walk versions past the window, and the `result` field / failure semantics. Living in the module means it's in context whenever the archive code gets edited.
- A new `.claude/docs/archive-investigation.md` with the investigation procedure: gsutil first (it can list directories, HTTPS can't), a standalone version-walking snippet using `followthemoney.dataset.VersionHistory` (verified against the live archive — `zavod.archive.iter_dataset_versions` needs a configured backend, so it's no good for read-only poking), the index.json → statistics.json → deltas escalation, and a "stream large files instead of dumping them into context" example. Referenced from the root CLAUDE.md next to the crawler guide.

The investigation doc points at the docstring for the layout instead of duplicating it, and version IDs are treated as opaque throughout — the only documented property is that they sort chronologically.

The issue asks "Is this a claude skill, or better docs, or both?" — I went back and forth on that, and actually built two of the options before landing here:

a) a nested `zavod/archive/CLAUDE.md` — rejected because nested CLAUDE.md files auto-load when *editing* files in that directory, which is exactly backwards: you're rarely hacking on the archive module while investigating a dataset, and when you *are* investigating, nothing pulls it into context.
b) a reference doc in `.claude/docs/` like the crawler guide — what I went with.
c) human-readable docs in `zavod/docs` (I had a version that rendered the module docstring via mkdocstrings) — rejected because the audience is wrong: gsutil invocations and "don't dump large files into your context" tips are agent-directed, not something for the human zavod docs. The layout reference itself stays human-readable in the docstring anyway.
d) a skill (`/investigate-dataset-archive`, which this branch briefly was) — rejected because a skill is centered around one defined task, and investigating a dataset issue takes many forms; a reference doc that gets pulled in whenever the archive comes up fits better than a single scripted procedure.

Let me know if you disagree!

### DIFF
--- a/.claude/docs/archive-investigation.md
+++ b/.claude/docs/archive-investigation.md
@@ -1,0 +1,103 @@
+# Investigating Dataset Runs via the Data Archive
+
+How to investigate past versions of a dataset using the data.opensanctions.org
+archive — why an entity count changed, when an entity appeared or disappeared,
+whether runs failed, and what a past run actually produced.
+
+First read the module docstring of `zavod/zavod/archive/__init__.py` — it
+contains the full, authoritative documentation of the archive layout: the
+`/artifacts/` and `/datasets/` prefixes, the root `versions.json` file and its
+`last_successful` field, how version walking works, and the `result` field
+marking a run as success or failure. Everything below assumes that layout.
+Version IDs are opaque strings; their one guaranteed property is that they
+sort chronologically.
+
+## Accessing files
+
+Prefer `gsutil` against the production bucket — unlike HTTPS it can list
+directories, which is often the fastest way to see what a run produced:
+
+```
+gsutil ls gs://data.opensanctions.org/artifacts/{dataset}/
+gsutil ls gs://data.opensanctions.org/artifacts/{dataset}/{version}/
+gsutil cat gs://data.opensanctions.org/artifacts/{dataset}/{version}/index.json
+```
+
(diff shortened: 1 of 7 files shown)


leonhandreke:
@jbothma not asking you to review this line by line, but do you think this is a sane way to offer this context to Claude, like from the structure of where to put it and how to trigger it?

leonhandreke:
Superseded by https://github.com/opensanctions/opensanctions/pull/4840

pudo:
wait these are different animals

jbothma (commented):
Yeah - this looks like a big help. 

I realise now we don't really document the archive for crawler devs, so as "a platform" we're not documenting how to leverage that to see history which can be helpful for debugging. A simple page on that leveraging or taking on some of the archive module docstring might be good - that's probably where I should have put it in the fierst place. 

It does feel like examples like this make sense in `.claude` - I'm not super likely to throw together 15-line python snippets for debugging now claude is around. I would have wanted a CLI to script with in the past.

Issue #4644, opened 2026-06-22
warn on unaccepted reviews for any dataset


jbothma:
we currently have `zavod.stateful.review.assert_all_accepted` which we have to rememebr to call in crawlers using reviews.

1. let's make it warn, never raise - we don't really see the need any more for it to fail an export
2. let's make it warn as part of zavod, without the need for a crawler to call a function.

It isn't obvious to an external user what the impact of the warning is (e.g. sometimes it means some names haven't been cleaned but the raw names are exported. Sometimes it means that new entities on an enforcement action notice are missing in spite of the crawl occurring after it was published"

Should datasets be able to customise the warning? e.g. in metadata

```
unaccepted_reviews_warning: |
  Unreviewed entities may be missing from the exported data, 
  and should become available once reviews have been completed.
```

This might make sense to have on an enforcements dataset where this would be the case, while it might cause unnecessary alarm on a sanctions dataset where this should never be the case.

PR #4832, opened 2026-07-09
[docs] Note assert_all_accepted in names migration step 3

nvmbrasserie:
Adds a paragraph to Step 3 of the name framework migration documentation (`zavod/docs/extract/names.md`) noting that all sanctions crawlers should end with `assert_all_accepted(context, raise_on_unaccepted=False)`.

The paragraph explains that this call logs a warning with the count of any name reviews still unaccepted for the current dataset and version (rather than raising), so the crawler completes and emits output while flagging that reviews remain outstanding.

Note: the requested `assert_all_accepted(raise=False)` maps to the actual parameter name `raise_on_unaccepted=False`, and the helper is imported from `zavod.stateful.review`.

### DIFF
--- a/zavod/docs/extract/names.md
+++ b/zavod/docs/extract/names.md
@@ -182,7 +182,7 @@ entity.id = context.make_id(names_string, ...)
 h.apply_reviewed_name_string(context, entity, string=names_string)
 ```
 
-All sanctions crawlers should end with a call to `assert_all_accepted(context, raise_on_unaccepted=False)` (imported from `zavod.stateful.review`). This logs a warning with the count of any name reviews that are still unaccepted for the current dataset and version, rather than raising, so the crawler completes and emits output while flagging that reviews remain outstanding.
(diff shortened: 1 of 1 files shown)


zavod/docs/extract/names.md
```diff
@@ -182,6 +182,8 @@ entity.id = context.make_id(names_string, ...)
 h.apply_reviewed_name_string(context, entity, string=names_string)
 ```
 
+All sanctions crawlers should end with a call to `assert_all_accepted(context, raise_on_unaccepted=False)` (imported from `zavod.stateful.review`). This logs a warning with the count of any name reviews that are still unaccepted for the current dataset and version, rather than raising, so the crawler completes and emits output while flagging that reviews remain outstanding.
+
```
jbothma:
```suggestion
All crawlers calling `h.review_names` or `h.apply_reviewed_names` should end with a call to `assert_all_accepted(context, raise_on_unaccepted=False)` (imported from `zavod.stateful.review`). This logs a warning with the count of any name reviews that are still unaccepted for the current dataset and version, rather than raising, so the crawler completes and emits output while flagging that reviews remain outstanding.

```

jbothma (commented):
see also https://github.com/opensanctions/opensanctions/issues/4644

jbothma (approved):

PR #4868, opened 2026-07-12
[cn_npc] Add National People's Congress deputies PEP crawler

pudo:
Adds a primary-source PEP crawler for deputies of China's National People's Congress, crawled from the live official roster on www.npc.gov.cn. Research and scoping: opensanctions/crawler-planning#780.

## What it does

- Follows the term redirect stub (`/npc/c191/dbmd/`) to the current congress's delegation index, then crawls all 35 delegation pages (31 provincial units + PLA, Hong Kong, Macau, Taiwan).
- Emits ~2,860 `Person` entities with Chinese name, LLM transliteration to English, gender/ethnicity where the source annotates them, and `citizenship=cn` (Electoral Law Art. 3, cited in code).
- One shared `Position` ("Member of the National People's Congress", `Q10891456`, `default_is_pep=True`) and per-deputy `Occupancy` entities carrying the delegation in `constituency` — same modelling as `cn_peoples_congress_wikipedia`, which stays in place; persons dedupe downstream.
- The roster is maintained by the publisher (terminated deputies are removed), so this is a live-roster crawl: no occupancy dates, `no_end_implies_current=True`.

## Source quirks handled

- Plain HTTP only — the site rejects TLS handshakes from non-China clients.
- WZWS WAF: browser user agent set via `http.user_agent`; a served challenge page fails the structural asserts loudly.
- No charset header (UTF-8 declared in META only) — encoding forced on all fetches.
- Long names (mostly Uyghur/Kazakh/Mongolian) sit in a second `md_zi2` CSS class on both the index and delegation pages — a `md_zi`-only selector would silently drop ~35 of Xinjiang's 60 deputies.
- Names padded with mixed whitespace (U+2002/U+3000/spaces); annotation parens/commas appear in both half- and full-width forms.
- Tibet has two deputies both named 拉琼 (both Tibetan); the source disambiguates them by home district in the annotation, which is part of the entity ID key.
- Term-bounded freshness: when the stub flips to the 15th NPC (March 2028), the crawler fails with an explicit message to review the structure.

## Validation

- 2,858 deputies across 35 delegations; `zavod validate` passes, no issues logged, `mypy --strict` clean.
- Referential integrity checked (Occupancy↔Person/Position; all `role.pep` persons have an occupancy and citizenship).
- Spot checks: 孙绍骋 (credentials terminated 2026-02-26) correctly absent; 779 women (27.3%, matching the NPC's reported share); ethnicity distribution demographically plausible.

## Before deploy

All testing ran from a residential IP. Chinese government sites often block datacenter ranges — worth a quick fetch test from the production crawler infrastructure; fallback would be routing through Zyte.


datasets/cn/npc/cn_npc.yml
```diff
@@ -45,14 +45,130 @@ tags:
...
+      - match: 鄂伦春族
+        value: Oroqen
+      - match: 赫哲族
+        value: Hezhen
+      - match: 门巴族
+        value: Monba
+      - match: 珞巴族
+        value: Lhoba
+      - match: 基诺族
+        value: Jino
```
pudo: Why are we translating these?

pudo (commented):
Looks like a lot of simplification - very cool. We sure we're still capturing everything?

datasets/cn/npc/cn_npc.yml
herrhellana: kicked the translation

PR #4873, opened 2026-07-13
Version-prefix local working state for artifacts

leonhandreke:
Phase 1 of https://github.com/opensanctions/operations/issues/2675.

Getting an artifact always takes a version now, both in the archive (already true) and on local disk. That kills the implicit "closest thing available, locally or backfilled from the archive" resolution: each stage operates on a version that verifiably exists, and the logs show which one.

## What's in this PR (phase 1)

- Local run artifacts move from a flat `data/datasets/<name>/` to `data/datasets/<name>/<version>/`, mirroring `artifacts/<name>/<version>/` in the archive. New helpers: `dataset_version_path`, `latest_local_version`, `find_archive_artifact`.
- `Context` takes a required `Version`. Only a crawl (or the start of a collection run) mints one; validate/export/publish receive it and no longer invent versions.
- `get_dataset_artifact` requires an exact version. The "latest published" case is now explicit via `find_archive_artifact` / `latest_local_version`, used only where we actually want that (catalog building, `ArchiveBackedCatalog`). The `datasets/latest` legacy fallback is isolated there behind a `# REMOVE AFTER MIGRATION` warning.
- Statement `first_seen`/`last_seen` stamp from the run's `version.dt` rather than the process-global `RUN_TIME` (behaviour-identical today, correct once stages run in separate processes).
- `export_resource` copies working files (e.g. fetched sources) into the version dir; exporters write there directly via the new `context.get_artifact_path`.
- A failed run has no `statistics.json` at its version, so the failure index carries no stale entity counts — this removes the need for the `backfill_statistics` flag from the earlier `failure-index-no-stats-backfill` branch; that fix falls out for free here.

Tests are updated and green (`pytest zavod/tests/` 231 passed; `mypy --strict` clean apart from a pre-existing unrelated `http_.py` stub error). Behavioural changes reflected in the tests:

- `zavod export`/`validate` operate on the latest local run and error if there's been no crawl yet.
- A failed run still mints its version (so `versions.json` exists) but leaves `last_successful` unset.
- Backfilling references an exact version and no longer depends on a published version history.

## Not yet (follow-ups)

- Move fetched sources into `_state/` (they still land in the dataset root for now).
- `--version` CLI option (validate/export/publish currently default to the latest local version, which is the intended default) and the `--clear-data` → `--clear-state` rename.
- Coordinated `kombinat` change (it calls `get_dataset_artifact` / `get_catalog_datasets` directly) — **this must land before merge**, since the `get_dataset_artifact` signature changed. `zavod run`-based orchestration is unaffected.
- `zavod/docs` updates.

### DIFF
--- a/zavod/zavod/archive/__init__.py
+++ b/zavod/zavod/archive/__init__.py
@@ -24,7 +24,7 @@ import shutil
 from pathlib import Path
 from functools import lru_cache
 from typing import TYPE_CHECKING
-from typing import Optional, Generator, TextIO, Set
+from typing import Optional, Generator, TextIO, Set, Tuple
 from rigour.mime.types import JSON
 from followthemoney import Statement
 from followthemoney.statement.serialize import read_pack_statements_decoded
@@ -89,8 +89,9 @@ def dataset_data_path(dataset_name: str) -> Path:
 
 
 def dataset_state_path(dataset_name: str) -> Path:
-    """The state directory is outside the main data directory and is used for temporary
-    processing artifacts (like the materialised graph, and the timestamp index)."""
+    """The state directory is outside the versioned data directories and is used for
+    temporary processing artifacts (like the materialised graph, and the timestamp
+    index)."""
     path = dataset_data_path(dataset_name) / "_state"
     path.mkdir(parents=True, exist_ok=True)
     return path.resolve()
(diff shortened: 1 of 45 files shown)


zavod/zavod/archive/__init__.py
```diff
@@ -101,22 +102,57 @@ def clear_data_path(dataset_name: str) -> None:
...
+    versions = []
+    for child in dataset_data_path(dataset_name).iterdir():
+        if not child.is_dir():
+            continue
+        try:
+            version = Version.from_string(child.name)
+        except ValueError:
+            continue
+        if with_resource is not None and not child.joinpath(with_resource).is_file():
+            continue
```
jbothma, line 135: when do we need this? Are there any cases where zavod wants to consider an older version than either the version we're building, or the last successful version? I think a missing file in the intended version should be an error.

zavod/zavod/exporters/metadata.py
```diff
@@ -133,21 +139,34 @@ def write_dataset_index(dataset: Dataset, result: DatasetVersionResult) -> None:
...
+        version, legacy_object = find_archive_artifact(dataset.name, INDEX_FILE)
+        if version is None and legacy_object is not None:
+            # FIXME: legacy fallback, remove after migration.
+            with legacy_object.open() as fh:
+                index_data = json.load(fh)
+    if version is not None:
+        path = get_dataset_artifact(dataset.name, version, INDEX_FILE)
+        if path.is_file():
+            with open(path, "r") as fh:
+                index_data = json.load(fh)
```
jbothma, line 159: do we need to introduce reading the latest local, and backfilling here?

zavod/zavod/archive/__init__.py
jbothma: in the CLI, when we're running e.g. export locally, we don't want it to fall back to my run from yesterday, if my run from 5 mins ago which I thought I was exporting, didn't even write a statements file.

jbothma (commented):
just some thoughts while reading through for https://github.com/opensanctions/operations/issues/2762

zavod/zavod/archive/__init__.py
pudo: totally agree

zavod/zavod/exporters/metadata.py
pudo: my sense is that `export` should have a mandatory `version` argument if invoked separately going forward. There will be no discovery of the latest local version - it has to be 100% explicit.

jbothma:
To be clear, this is a very early version I asked @leonhandreke to share while I'm working on related things since he was probably already identifying relevant code.

PR #4876, opened 2026-07-13
Add property_fill_rate assertions for core sanctions datasets

leonhandreke:
Extends the [au_dfat_sanctions pattern](https://github.com/opensanctions/opensanctions/pull/4874) to our other core sanctions lists: US OFAC SDN & Consolidated, EU FSF, Canada SEMA, Switzerland SECO, and UK FCDO.

These are core sanctions datasets, so we deliberately err on the side of stability and getting a human involved. Thresholds sit well below current production fill rates — normal source fluctuation won't trip them, but a mass drop fails the export instead of quietly publishing degraded data. The expectation is that these *will* trip when a source changes its structure, and when they do we react to that change (fix the crawler or adjust the threshold) rather than shipping broken data.

A few notes:

- Nothing is asserted at 1.0, not even props that are always 1.0 in production — those get 0.98, so a single unparseable value doesn't fail the whole export. There's a comment on each `property_fill_rate` block saying so.
- eu_fsf `Person.name` is at 0.98 for a second reason too: manual review demotes some names to weakAlias, leaving those entities without a primary name.
- ch_seco `Vessel.imoNumber` is only ~0.25 (most SECO vessels have no IMO), so it's asserted low at 0.2 — mostly a guard against a total collapse.

Draft for review — happy to take feedback on the specific thresholds.

### DIFF
--- a/datasets/au/dfat_sanctions/au_dfat_sanctions.yml
+++ b/datasets/au/dfat_sanctions/au_dfat_sanctions.yml
@@ -49,22 +49,18 @@ assertions:
     schema_entities:
       Person: 2000
       LegalEntity: 700
+    # Set at roughly a quarter of the production fill rate: low enough that a
+    # shift in the shape of the source data doesn't hold up the release, high
+    # enough to catch a crawler that stopped parsing a field altogether.
     property_fill_rate:
       Person:
-        name: 1.0
-        birthDate: 0.85
-        citizenship: 0.7
-        createdAt: 1.0
-      LegalEntity:
-        name: 1.0
-        createdAt: 1.0
+        birthDate: 0.25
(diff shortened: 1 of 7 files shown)


leonhandreke:
We're getting into territory here where maybe CI wants a copy of the reviews database at some point?

leonhandreke:
@pudo not sure about this one - might save us from losing important stuff without noticing, but might also cause us to be late if the legitimately change something that trips them.

pudo:
What if we set them really low, eg. at 20-30% of current. That way we are alerted if we mess up the crawler (which is worth something) but not when the shape of the data shifts (which could break our latency requirements).

PR #4877, opened 2026-07-13
Saudi Arabia National Terrorism List (1373)

### LINKED ISSUE #824: Saudi National Terrorism List (1373)
### Data URL

https://pctc.pss.gov.sa/wps/portal/departments/pcct/lists/!ut/p/z1/04_Sj9CPykssy0xPLMnMz0vMAfIjo8ziDT0s3T2cTQz93d1CjA0CzYyCDY3NgwwszEz0w_EpMPAw0o8iRr-Bp4mJh6GJkbeBaZCzgaOZWai7iVmQgYmPIZH6cQBHA-L041EQhd_4cP0ofFaAQwBiBh4vErIkOLFIvyA3NDQ0wiAzIF1REQDZ712U/dz/d5/L0lHSkovd0RNQUZrQUVnQSEhLzROVkUvZW4!/

### Publisher

Permanent Committee of Counter Terrorism (PCTC) of Saudi Arabia

sa

### Type of data

Sanctions (Governments barring specific engagement e.g. financial transaction with these entities)

### Coverage region

region:Middle East

### Can you tell us more?

_No response_

- [x] I understand.

nvmbrasserie:
fixes https://github.com/opensanctions/crawler-planning/issues/824


nvmbrasserie:
In the `notes`, there are often previous names, aliases, and other details that could be parsed. I don't think this should be done within the original crawl, but it could be a future extension project

datasets/sa/pcct_terrorism_list/crawler.py
```diff
@@ -0,0 +1,283 @@
...
+def crawl(context: Context) -> None:
+    path = context.fetch_resource("source.xlsx", context.data_url)
+    context.export_resource(path, XLSX, title=context.SOURCE_TITLE)
+    workbook = load_workbook(path, read_only=True)
+
+    columns = context.get_lookup("columns")
+    for sheet in workbook.sheetnames:
+        schema = context.lookup_value("schema", sheet)
+        if schema is None:
+            context.log.warning("Unmapped sheet", sheet=sheet)
```
jbothma, line 283: it might be worth also asserting that the current sheet names exist, although data assertions should catch if so many entities drop out. This warning is a nice touch.

datasets/sa/pcct_terrorism_list/crawler.py
```diff
@@ -0,0 +1,283 @@
...
+    # Unambiguous: one number, one type, one country -> emit without review.
+    if len(numbers) == 1 and len(countries) == 1 and len(types) == 1:
+        passport = context.lookup_value("document.type", doc_type) == "passport"
+        identification = h.make_identification(
+            context,
+            entity,
+            numbers[0],
+            doc_type,
+            country=countries[0],
+            passport=passport,
```
jbothma:
```suggestion
        is_passport = context.lookup_value("document.type", doc_type) == "passport"
        identification = h.make_identification(
            context,
            entity,
            numbers[0],
            doc_type,
            country=countries[0],
            passport=is_passport,
```

would be clearer.

datasets/sa/pcct_terrorism_list/crawler.py
```diff
@@ -0,0 +1,283 @@
...
+    # its own type and country - no code-side pairing or fallback.
+    source_data: JsonValue = {
+        key: value
+        for key, value in (
+            ("document_type", doc_type),
+            ("document_number", raw_number),
+            ("issuing_country", raw_country),
+        )
+        if value is not None
+    }
```
jbothma, line 158: this could just be literal then it's much easier to read. The LLM will cope with a null.

e.g.

```json
{
  "documenttype": doc_type,
 ...,
 ... 
}

File: datasets/sa/pcct_terrorism_list/crawler.py [hunk 4]
```diff
@@ -0,0 +1,283 @@
...
+        response_type=IdentityDocuments,
+    )
+    review = review_extraction(
+        context,
+        source_value=source_value,
+        original_extraction=extraction,
+        origin=DEFAULT_MODEL,
+    )
+    if not review.accepted:
+        return
```
jbothma, line 181: I think we should put the value into `notes` until we accept the review rather than dropping the value.

datasets/sa/pcct_terrorism_list/crawler.py
```diff
@@ -0,0 +1,283 @@
...
+        proxy.id = context.make_id(entity.id, document.number, document.type)
+        # Reviewed values are tagged with the review origin so the statements
+        # are attributed to the model extraction rather than the crawler.
+        proxy.add("holder", entity, origin=review.origin)
+        proxy.add("number", document.number, origin=review.origin)
+        proxy.add("type", document.type, origin=review.origin)
+        proxy.add("country", document.country, origin=review.origin)
+        context.emit(proxy)
+        holder_prop = "passportNumber" if passport else "idNumber"
+        entity.add(holder_prop, document.number, origin=review.origin)
```
jbothma:
```suggestion
        is_passport = document.type == "passport"
        h.make_identification(...)
```

datasets/sa/pcct_terrorism_list/crawler.py
```diff
@@ -0,0 +1,283 @@
...
+    for org in h.multi_split(row.pop("linked_org", None), [";"]):
+        emit_related(context, entity, "terror-org", org)
+    # Emit owners of designated vessels
+    emit_related(
+        context,
+        entity,
+        "owner",
+        row.pop("owner_name", None),
+        row.pop("owner_info", None),
+    )
```
jbothma, line 256: arg names for the details really help readability here

```suggestion
    for org in h.multi_split(row.pop("linked_org", None), [";"]):
        emit_related(
          context, 
          entity, 
          kind="terror-org", 
          name=org
        )
    # Emit owners of designated vessels
    emit_related(
        context,
        entity,
        kind="owner",
        name=row.pop("owner_name", None),
        notes=row.pop("owner_info", None),
    )
```

datasets/sa/pcct_terrorism_list/crawler.py
```diff
@@ -0,0 +1,283 @@
...
+        "Organization",
+        "UnknownLink",
+        "subject",
+        "object",
+        "Linked terrorist organization",
+        "crime.terror",
+    ),
+    # The owner of a designated organization or vessel.
+    "owner": Relation("LegalEntity", "Ownership", "asset", "owner", "Owner", None),
+}
```
jbothma: Maybe a nitpick/personal prefernece but I think this is more readbale:

I can see the value of factoring this out. I don't love the string-based lookup ("terror-org")

Also I think one big benefit of using a named tuple to organise this long list of config is that we can provide field names when we instantiate it.

I think defining two constants which we then reference directly as arguments to emit_related is quite nice and readable:

```suggestion
TERROR_ORG_REL = Relation(
        schema="Organization",
        "UnknownLink",
        anchor_prop="subject",
        related_prop="object",
        ...="Linked terrorist organization",
        topics="crime.terror",
    )
    
OWNER_RELATION = Relation(
    scema="LegalEntity", 
    ...
)
```

datasets/sa/pcct_terrorism_list/crawler.py
```diff
@@ -0,0 +1,283 @@
...
+        row.pop("owner_name", None),
+        row.pop("owner_info", None),
+    )
+
+    sanction = h.make_sanction(
+        context,
+        entity,
+        program_key=PROGRAM_KEY,
+        start_date=designated_date,
+    )
```
jbothma, line 266: let's give these folks crime.terr and sanction topics

datasets/sa/pcct_terrorism_list/sa_pcct_terrorism_list.yml
```diff
@@ -0,0 +1,146 @@
+title: "Saudi Arabia National Terrorism List (1373)"
```
jbothma: maybe either qualify the number or skip entirely

```suggestion
title: "Saudi Arabia National Terrorism List (UNSCR 1373)"
```

jbothma (commented):
> In the notes, there are often previous names, aliases, and other details that could be parsed. I don't think this should be done within the original crawl, but it could be a future extension project

ok - let's just create a new issue in crawler-planning so we can prioritise that.

jbothma:
cc @pudo this is using LLM extraction and reviews to split multiple passport and or ID numbers into the right props. It's a sanctions dataset.

I think it's ok since it isn't other scripts that we easily misread, it isn't names, and having clean identifiers is quite valuable. Just wanted to flag in case we've been explicit about not using LLMs with sanctions data more broadly than just for names.

pudo:
For non-names were also doing that in Switzerland, so it's not unprecedented

jbothma (approved):
thanks for the changes!

Issue #4496, opened 2026-06-11
Introduce sanction.control; make sanction.linked non-transitive


pudo:
## Summary

Today the `ann_graph_topics` analyzer propagates `sanction.linked` transitively down ownership chains (`analyzer.py` Rule C). This conflates two different things: an entity *adjacent* to a sanctioned entity, and an entity *controlled* by one. We want a stricter, separate topic — `sanction.control` (already defined: "Sanction ownership or control") — that models the control chain precisely, and we want `sanction.linked` to go back to meaning plain adjacency.

This is staged in two phases with a measurement gap between them, because flipping it all at once would strip `sanction.linked` from entities mid-flight.

## `sanction.control` semantics

- **Seeds:** entities carrying `sanction` **or** `sanction.control` (transitive — control of a controlled entity is still control).
- **Propagates across only two edges, in the directed sense:**
  - `Ownership`: `owner → asset`
  - `Directorship`: `director → organization`
- **No percentage gate.** We don't have reliable stake data; we assert the structural link and leave 50%-rule arithmetic to the user at the point of a hit.
- **`endDate` terminates propagation** — a former owner/director doesn't propagate control (consistent with current behavior).
- **One hop per run**, converging over successive runs like the existing ownership walk.
- **Co-emit:** every entity tagged `sanction.control` is **also** tagged `sanction.linked` in lockstep. `sanction.linked` stays a superset of `sanction.control` so existing customer filters on `sanction.linked` never lose an entity.

> Note: `Directorship` is included deliberately. A single board seat isn't legally "control," but a sanctioned person on a board creates sanctions exposure for the whole corporate hierarchy, and we'd rather over-reach here than miss it. This is called out in a code comment so it isn't "fixed" later.

## Phase 1 — additive (ship now)

- Add the `sanction.control` rule above (with co-emit).
- **Leave the existing transitive `sanction.linked` rule (Rule C) running unchanged.**

Phase 1 changes no existing tags: `sanction.linked` behaves exactly as today, while `sanction.control` back-fills in parallel over ~weeks.

## Measurement gate (~1 month)

Before phase 2:

1. Confirm `sanction.control` back-fill has converged across multi-tier hierarchies.
[...]

PR #4890, opened 2026-07-14
[ann_graph_topics] Phase 2: remove transitive sanction.linked descent

jbothma:
Removes rule_ownership_descent and its tests. sanction.linked is now strictly non-transitive on its own: it means "direct broad-edge adjacency to a sanctioned entity" (rule_sanction_adjacency), plus every entity in a sanction.control chain via the lockstep co-emit (rule_sanction_control_descent). Multi-tier reach for sanction.linked now flows exclusively through sanction.control.

Adds a regression test test_sanction_linked_does_not_propagate_transitively that pins the new behavior: an entity carrying only sanction.linked must not push sanction.linked to its asset.

Phase 2 of #4496.

### DIFF
--- a/datasets/_analysis/ann_graph_topics/analyzer.py
+++ b/datasets/_analysis/ann_graph_topics/analyzer.py
@@ -282,6 +282,8 @@ def rule_sanction_control_descent(
         if target_topics & SANCTION_CONTROL_SEEDS:
             continue
         emit_patch(context, source, target, "sanction.control", target_topics)
+        if target_topics & SANCTION_SEEDS:
+            continue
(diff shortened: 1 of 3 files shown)


pudo (approved):
We do have this asymmetry now of `export.control.linked` and `sanction.linked` but let's  open this up the next time we go into this whole topic. I agree on the semantics here now.

jbothma:
we expect around 40,000 entities to drop out due to this change. First ann_graph_topics needs to run, then each of ext_ru_egrul, ext_gleif, ext_icij, and around 200 entities from others?

PR #4892, opened 2026-07-14
[jp_mof_sanctions] Raw names to review + LLM cleaning (follow-up to #4716)

nvmbrasserie:
Follow-up to #4716 addressing the reviewer feedback on name splitting. Stacked on the step-3 branch, so the diff here is only the incremental change; it will retarget to `main` once #4716 merges.

## What changed

1. **Stop pre-splitting names before review** (@pudo, @jbothma). `crawl_sheets` no longer runs `SPLITS` on name columns — they are split **only on newlines** (structural; unrecoverable later because whitespace is squashed before names reach review) and otherwise passed raw. Dates/IDs/addresses keep their splitting.
2. **Cover the split markers in the namespec** (@jbothma). `reject_chars` already flags brackets/colons/semicolons; added `a.k.a.` / `f.k.a.` to `names.schema_rules.reject_strings` for the comma-form alias markers.
3. **Dropped `split_english_names`** — the English column is passed as-is; original-script appendages are handled in review.
4. **Enabled `llm_cleaning` on the name fields** (@pudo's "little devil"). This is a **deliberate, inline-flagged exception** to the "no llm_cleaning on sanctions datasets" rule: the list mixes English and Japanese names and no reviewer reads Japanese, so a non-speaker can't reliably tell a name from embedded commentary. The LLM proposes splits/cleanups that a human still accepts — nothing is auto-applied. Implemented as per-value `apply_reviewed_name_string` calls per language (llm_cleaning returns plain strings and drops per-value language).

## Verification

`zavod crawl` runs clean locally (7206 entities, exit 0). LLM path is wired correctly — without a key it logs `LLM cleaning enabled but OPENAI_API_KEY not configured, falling back to non-LLM review` and degrades to raw review.

## Why draft — not merge-ready as-is

- Changing `original` changes the **review keys**, so the accepted step-2 reviews from #4716 no longer match → everything reverts to unaccepted (~3k items) and falls back to the raw string until re-reviewed. This effectively resets the migration to a step-1/step-2 cycle for this dataset.
- While unaccepted, long combined alias cells (e.g. `(a)…(z) a.k.a.…`) exceed the property length limit and are dropped; they return once reviews are accepted and split.
- `entity.id` is derived from the name values, so IDs shift for entities whose names contained split markers — needs a look before merge.
- The sanctions LLM-cleaning exception should get explicit sign-off and a note in `zavod/docs`.

Needs: policy sign-off on the LLM exception, a decision on the ID shift, and a fresh step-2 review pass (ideally with the LLM key configured so reviewers see suggestions).

### DIFF
--- a/datasets/jp/mof_sanctions/crawler.py
+++ b/datasets/jp/mof_sanctions/crawler.py
@@ -145,18 +145,37 @@ def emit_row(
     raw_old_name = row.pop("old_name", [])
     raw_weak_alias = row.pop("weak_alias", [])
     raw_nickname = row.pop("nickname", [])
-    original = h.Names()
-    for n in name_english:
-        original.add("name", n, lang="eng")
-    for n in name_japanese:
-        original.add("name", n, lang="jpn")
-    for n in chain(raw_alias, raw_known_alias):
-        original.add("alias", n)
-    for n in chain(raw_past_alias, raw_old_name):
-        original.add("previousName", n)
-    for n in chain(raw_weak_alias, raw_nickname):
-        original.add("weakAlias", n)
-    h.apply_reviewed_names(context, entity, original=original)
+    # Exception: we normally never enable llm_cleaning on a sanctions dataset, but
+    # this list mixes English and Japanese names and none of our reviewers read
+    # Japanese - so a non-speaker can't reliably tell a name from embedded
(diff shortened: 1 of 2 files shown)


nvmbrasserie:
no LLM cleaning in sanctions datasets. keeping it as is for now

PR #4949, opened 2026-07-17
helpers: fix duplicated state in make_address; normalize country_code casing in address IDs

### LINKED ISSUE #4934: helpers: make_address renders the state twice ("PA, PA"); address IDs are case-sensitive on country_code
Found in a systematic audit of the zavod helper library (2026-07-17). Two defects in `zavod/zavod/helpers/addresses.py`, both verified by repro.

## 1. State rendered twice in the formatted address

`make_address` passes `state` into the format data twice — once as `state=` and again folded into `state_district=join_text(region, state)` (`addresses.py:188-190`). rigour's US template (`generic4`) renders `state_district` in the city slot and `state` in the state slot, so both copies appear whenever `city` is absent:

```python
make_address(context, city=None, state="PA", country_code="us")
# full == "PA, PA"
make_address(context, street="1 Main St", state="California", region="Southern", country="United States")
# full == "1 Main St, Southern, California, California, United States"
```

The corrupted string becomes the entity's `full` property (displayed on opensanctions.org) and feeds the address entity's ID hash. Exposed callers: ~17 crawlers pass `state=` (`us/ofac` `ofac_advanced.py:207-219` passes both `region=` and `state=`; most `us/*/med_exclusions`; `us/sam_exclusions`; `us/cms_npi`).

Fix: `state_district=region` — drop `state` from the join.

## 2. Address entity IDs case-sensitive on `country_code` — cross-dataset dedup defeated

`_make_id` hashes the raw `country_code` (`addresses.py:154-164,200` → `make_entity_id`): `make_entity_id("US", slug) != make_entity_id("us", slug)`. `make_address` also adopts any 2-letter `country` string as `country_code` verbatim, without lowercasing or ISO validation. Inside `format_address` the code *is* lowercased, but that normalized value never reaches the ID path.

Mixed casing exists in the fleet today: `country_code="US"` in `us/sc|mn|nc|nj|wy` med_exclusions vs `"us"` in `us/ms|wv|ar|nd|oh`, `us/hhs_exclusions`, `us/fdic_failed_banks` — identical addresses across these datasets get distinct content-addressed `addr-<sha1>` entities, defeating the cross-dataset address merging the scheme exists for.

Fix: lowercase (and ideally ISO-validate) `country_code` before `_make_id`. **Note: this changes existing address entity IDs and needs team coordination.**

pudo:
Fixes #4934

Two fixes to `zavod/zavod/helpers/addresses.py`, deliberately split into two commits because the second one changes entity IDs and can be dropped independently.

## Commit 1 — state rendered twice in the formatted address

`make_address` passed `state` into `format_address` twice: once as `state=` and again folded into `state_district=join_text(region, state)`. Country templates that render `state_district` in the city slot (e.g. rigour's US `generic4`) printed the state twice whenever `city` was absent:

```python
make_address(context, city=None, state="PA", country_code="us")
# full == "PA, PA"
```

Fix: `state_district` now carries `region` only. New tests cover the `city=None, state="PA"` case and an address with both `region` and `state` set.

## Commit 2 — country_code casing in address entity IDs

`_make_id` hashes the raw `country_code`, and `make_address` adopts 2-letter `country` values verbatim, so `country_code="US"` and `"us"` produced distinct `addr-<sha1>` entities for identical addresses — defeating the cross-dataset address dedup the content-addressed ID scheme exists for (mixed casing exists in the fleet today, e.g. `us/sc|mn|nc|nj|wy` med_exclusions pass `"US"` while `us/ms|wv|ar|nd|oh`, `us/hhs_exclusions`, `us/fdic_failed_banks` pass `"us"`). The code is now lowercased once resolved — consistent with what `format_address` already does internally — before it reaches the ID hash and the stored `country` property. New test asserts `country_code="US"` and `"us"` yield the same entity ID.

> [!WARNING]
> **Commit 2 changes existing address entity IDs** for every crawler that passes an uppercase country code (or an uppercase 2-letter string in `country`). Downstream references via `addressEntity` are emitted in the same run so they stay consistent, but previously published address IDs from those datasets will be replaced. Please decide whether to merge both commits or drop the second one pending coordination.

Verification: `python -m pytest zavod/tests/helpers/test_addresses.py` (4 passed) and `mypy --strict` on the changed module (clean).

### DIFF
--- a/zavod/zavod/helpers/addresses.py
+++ b/zavod/zavod/helpers/addresses.py
@@ -186,7 +186,7 @@ def make_address(
             postal_code=postal_code,
             city=city,
             state=state,
-            state_district=join_text(region, state, sep=", "),
+            state_district=region,
             country=country,
             country_code=country_code,
         )
--- a/zavod/zavod/helpers/addresses.py
+++ b/zavod/zavod/helpers/addresses.py
@@ -163,6 +163,12 @@ def make_address(
             country_code = country
             country = None
 
+    # Normalize casing (as format_address does internally) so that the
+    # country code hashed into the address ID is stable across datasets
+    # passing e.g. "US" vs "us":
+    if country_code is not None:
+        country_code = country_code.lower().strip()
+
     if country is not None:
         parsed_code = registry.country.clean(country)
         if parsed_code is not None:
--- a/zavod/zavod/helpers/addresses.py
+++ b/zavod/zavod/helpers/addresses.py
@@ -183,6 +183,11 @@ def make_address(
     if country_code is None:
         country_code = registry.country.clean(full)
 
+    # If both fields carry the same value, keep only the state so that no
+    # rendering path can duplicate it (e.g. "Aleppo, Aleppo"):
+    if region is not None and state is not None and region == state:
+        region = None
+
     full_origin = origin
     if not full:
         full = format_address(
@@ -196,6 +201,22 @@ def make_address(
             country=country,
             country_code=country_code,
         )
+        if state is not None and state not in full:
+            # Some country templates (e.g. ae, sa, sy) have a state_district
+            # slot but no state slot, dropping the state from the rendered
+            # line. Fold it into state_district instead. format_address is
+            # cached, so the second render is cheap.
+            full = format_address(
+                summary=summary,
+                po_box=po_box,
+                street=street,
+                postal_code=postal_code,
+                city=city,
+                state=state,
+                state_district=join_text(region, state, sep=", "),
+                country=country,
+                country_code=country_code,
+            )
         full_origin = ORIGIN_INFERRED
 
     if full == country:
--- a/zavod/zavod/tests/helpers/test_addresses.py
+++ b/zavod/zavod/tests/helpers/test_addresses.py
@@ -76,6 +76,39 @@ def test_make_address_state_not_duplicated(vcontext: Context):
     assert "Southern" in full, full
 
 
+def test_make_address_state_without_state_slot(vcontext: Context):
+    # Some country templates (e.g. ae, sy) render {{state_district or state}}
+    # and have no dedicated state slot. When region occupies that slot, the
+    # state must be folded back in rather than silently dropped:
+    addr = make_address(
+        vcontext, city="Dubai", state="Sharjah", region="Deira", country_code="ae"
+    )
+    assert addr is not None, addr
+    full = addr.first("full")
+    assert full == "Dubai, Deira, Sharjah", full
+    assert full.count("Sharjah") == 1, full
+
+    # Without a region, the template's own fallback renders the state:
+    addr = make_address(vcontext, city="Dubai", state="Sharjah", country_code="ae")
+    assert addr is not None, addr
+    full = addr.first("full")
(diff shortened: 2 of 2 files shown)


pudo:
**Update after review discussion: commit 1 as written introduces a regression for 57 country templates — hold off merging until it's amended.**

An empirical sweep across all 249 two-letter country templates in rigour's `formats.yml` (rendering with `state="STATEX"`, `state_district="REGIONX"` under old vs. new semantics) found:

- **57 countries lose a passed `state` entirely** under this PR: their templates have a `state_district` slot but **no `state` slot at all**, so the old `join_text(region, state)` was accidentally load-bearing there. The list includes sanctions-relevant countries: **AE, SA, SY, LY**, BD, EC, GH, UG, ZW, and others.
- `region` is never lost anywhere, and the "PA, PA" duplication is fully fixed — no template duplicates after the change.

Exposure is real: `us/ofac` `ofac_advanced.py` passes both `region=` and `state=` for worldwide addresses, so an SDN address in e.g. Syria or the UAE with a province field would silently drop it. Since `full` feeds the address entity ID, two addresses differing only by state would collapse into one entity.

**Why no input-side check works:** whether `state` needs to ride along in `state_district` depends on whether the country's template has a `state` slot — unknowable to the caller without template introspection. Every cheap single-condition variant fails on one side (skip-when-equal misses the region=None case; state_district-only loses state for normal US addresses when city occupies the slot).

**Proposed amendment** (output-side check instead):

```python
summary = format_address(..., state=state, state_district=region, ...)
if state is not None and (summary is None or state not in summary):
    # this country's template has no state slot — fold state back in
    summary = format_address(..., state=state, state_district=join_text(region, state, sep=", "), ...)
```

`format_address` is lru_cached, so the second render is cheap and only fires for the 57 no-state-slot countries. State then appears exactly once across all 249 templates. Additionally, dedupe identical inputs before the join: `region == state` (plausible from ofac_advanced filling both from coinciding SDN fields) currently produces `"Aleppo, Aleppo"` inside the state_district value — a second duplication source not previously flagged.

Tests to add with the amendment: a no-state-slot country (e.g. `ae`) with `state=` must render the state exactly once; `region == state` must not duplicate.

Longer term this belongs in rigour (either `{{state}}` lines for those templates or an option to append supplied-but-unrendered admin components) — but the fallback keeps this PR self-contained.

pudo:
Implemented in bda13c648 (new commit on the branch; the country-casing commit is untouched):

- **Output-side fallback in `make_address`**: render with `state_district=region` first; if `state` is set and missing from the result, re-render once with `state_district=join_text(region, state)`. `format_address` is lru_cached, so the second render is cheap and only fires for the no-state-slot templates.
- **Input dedupe**: when `region == state`, region is dropped before rendering, so the folded join can never produce "Aleppo, Aleppo".

Verified against rigour's actual templates: `make_address(city="Dubai", state="Sharjah", region="Deira", country_code="ae")` now renders `"Dubai, Deira, Sharjah"` (was `"Dubai, Deira"` — state dropped). Tests added for the ae fallback path, the region==state dedupe, and the existing "PA"/California cases still pass unchanged (6 passed; mypy --strict clean).

PR #4953, opened 2026-07-17
helpers: stop split_comma_names silently splitting realistic single names

### LINKED ISSUE #4939: helpers: split_comma_names silently mis-splits hyphenated, multi-word, and middle-initial names
Found in a systematic audit of the zavod helper library (2026-07-17). Verified by repro against `zavod/zavod/helpers/names.py:34-39,274-312`.

The safety design of `split_comma_names` is: split confidently where safe, warn on ambiguity so a dataset lookup can patch it. Three defects undermine this:

## 1. The "safe single name" guard is far too narrow — mis-splits get **no warning**

`REGEX_LNAME_FNAME = ^\w+, \w+$` only protects single-token-surname + single-token-firstname. Realistic name shapes fall through to the confident comma-split path and are split into separate entities silently:

```python
split_comma_names(context, "Smith-Jones, Mary")   # → ['Smith-Jones', 'Mary']  (two entities)
split_comma_names(context, "O'Brien, John")       # → ["O'Brien", 'John']
split_comma_names(context, "Van Der Berg, Jan")   # → ['Van Der Berg', 'Jan']
split_comma_names(context, "Smith, Jane B.")      # → ['Smith', 'Jane B.']
```

Each half then becomes an entity via `context.make_id(name)`. Because no warning fires, the lookup-patching workflow never sees these.

## 2. Dead alternatives in `REGEX_CLEAN_COMMA`

The `\(?A/K/A`, `\(?N\.K\.A`, `\(?N/K/A`, `\(?F\.K\.A` branches can never match when the paren is present: the `\b` after `", "` cannot match before `(`. Verified: `"John Smith, (A/K/A Smitty)"` passes through unchanged and is comma-split into a bogus `"(A/K/A Smitty)"` entity.

## 3. Correctly-handled `Lastname, Firstname` still emits the scary warning

Text matching the safe pattern lands in the else-branch and logs `"Not sure how to split on comma or and."` on every crawl (e.g. `"Smith, John"` — returned correctly but warned), training maintainers to ignore the warning that matters. Also, path 1 keeps empty fragments (`"A,,B"` → `['A', '', 'B']`).

## Blast radius

`us/sec/harmed_investors/crawler.py:10` and `us/finra_actions/crawler.py:47` — US enforcement lists where "Lastname, Firstname [Middle]" defendant names are routine. Their ymls show mis-splits already being hand-patched, but only for the warned path.

pudo:
Fixes #4939

`split_comma_names` is designed to split confidently only where safe and warn on ambiguity so a dataset `comma_names` lookup can patch it. Three defects undermined this:

1. **Silent mis-splits of realistic single names.** `REGEX_LNAME_FNAME` (`^\w+, \w+$`) only protected single-token-surname + single-token-firstname, so "Smith-Jones, Mary", "O'Brien, John", "Van Der Berg, Jan" and "Smith, Jane B." were confidently comma-split into two entities with no warning. The pattern now allows hyphens, apostrophes, up to three surname tokens, and dotted middle initials. It stays conservative: the part after the comma must be a single given-name token (plus dotted initials), so comma-separated lists of full names ("John Smith, Mary Jones", "A B C, D E F") still split as before, and ambiguous `and`+comma strings ("Smith, John and Jones, Mary") still hit the warn/lookup path.

2. **Dead parenthesised alias alternatives in `REGEX_CLEAN_COMMA`.** The `\b` after `", "` could never match before `(`, so `\(?A/K/A` etc. never fired and "John Smith, (A/K/A Smitty)" was split into a bogus "(A/K/A Smitty)" entity. The optional paren is now captured separately and preserved in the substitution.

3. **Noise and empty fragments.** Safe "Lastname, Firstname" strings now return without the "Not sure how to split on comma or and." warning, and the confident-split path filters empty fragments from doubled/trailing commas ("A,,B" no longer yields an empty name).

Tests cover all shapes above, including that genuinely ambiguous multi-name strings still warn.

Verified: `pytest zavod/tests/helpers` (101 passed), `mypy --strict zavod/helpers/names.py`, black/ruff clean on changed lines.

### DIFF
--- a/zavod/zavod/helpers/names.py
+++ b/zavod/zavod/helpers/names.py
@@ -32,9 +32,20 @@ from zavod.stateful.review import (
 )
 
 REGEX_AND = re.compile(r"(\band\b|&|\+)", re.I)
-REGEX_LNAME_FNAME = re.compile(r"^\w+, \w+$", re.I)
+# A single person name written "Lastname(s), Firstname" — e.g. "Smith, John",
+# "Smith-Jones, Mary", "O'Brien, John", "Van Der Berg, Jan", "Smith, Jane B.".
+# Kept conservative: the surname may span at most three tokens, and the part
+# after the comma must be a single given name (optionally followed by dotted
+# initials), so that comma-separated lists of full names don't match and still
+# get split or warned about.
+REGEX_LNAME_FNAME = re.compile(
+    r"^[\w'’-]+(?: [\w'’-]+){0,2}, [\w'’-]+\.?(?: [^\W\d_]\.)*$",
+    re.I,
+)
+# The optional opening paren is captured separately so the alias markers also
+# match in parenthesised form, e.g. "John Smith, (A/K/A Smitty)".
 REGEX_CLEAN_COMMA = re.compile(
-    r", \b(LLC|L\.L\.C|Inc|Jr|INC|LLLP|L\.P|LP|Sr|III|II|IV|S\.A|LTD|USA INC|\(?A/K/A|\(?N\.K\.A|\(?N/K/A|\(?F\.K\.A|formerly known as|INCORPORATED)\b",  # noqa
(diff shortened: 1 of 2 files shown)


pudo:
I don't understand what the vibes did here, and somehow that's a sign that it's too complicated and we should focus on the review framework

PR #4957, opened 2026-07-17
helpers: don't store redundant original_value for exact-match reviewed names

### LINKED ISSUE #4954: helpers: derive_original_values stores original_value == value for every clean single-name entity
Found in a systematic audit of the zavod helper library (2026-07-17). Verified by repro.

`derive_original_values` (`zavod/zavod/helpers/names.py:466-499`) documents: "(2) If some value in original matches it exactly, leave blank - no original_value needed". But case (1) — `len(original_values) == 1` — is checked *first*, so with exactly one original value an exact match still gets `original_value` set to itself:

```python
derive_original_values(Names(name="Jim Doe"), Names(name="Jim Doe"))
# → {'Jim Doe': 'Jim Doe'}
```

Because `apply_reviewed_names`' fallback path calls `apply_names(entity, original=original, names=original)`, **every clean single-name entity in every reviewed-names crawler** emits a name statement with a redundant `original_value` identical to the value (`Entity.unsafe_add` passes any non-None `original_value` through unconditionally, `zavod/entity.py:72-83`).

Consequences: statement bloat, and "has original_value" no longer signals "value was modified" — inconsistent with the rest of the stack, where `original_value` is only set when the cleaned value differs.

## Blast radius

All ~19 crawlers using `apply_reviewed_names`/`apply_reviewed_name_string` (us/sam_exclusions, gb/fca_notices, ru/cbr_banks, afdb_sanctions, …), on essentially every entity with a single regular name.

Note: `test_derive_original_values_single_original` pins case-(1) behavior for the *differs* case but never covers the identical case — this looks like an unconsidered interaction rather than intent.

## Fix direction

Check the exact-match case before the single-original case (or add an equality guard inside case 1), and add the missing test.

pudo:
Fixes #4954

`derive_original_values` checked the single-original case (1) before the exact-match case (2), so with exactly one original value an extracted value identical to it still got `original_value` set to itself:

```python
derive_original_values(Names(name="Jim Doe"), Names(name="Jim Doe"))
# → {'Jim Doe': 'Jim Doe'}
```

Since `apply_reviewed_names`' fallback path calls `apply_names(entity, original=original, names=original)`, every clean single-name entity in reviewed-names crawlers emitted a name statement with a redundant self-referential `original_value` — bloating statements and breaking the invariant that `original_value` signals "value was modified".

## Changes

- Reorder the cases in `derive_original_values` so an exact match is checked first and never produces an `original_value`; the single-original case still applies when the extracted value genuinely differs (pinned by `test_derive_original_values_single_original`). Docstring updated to match the new priority.
- Add the missing tests: identical single original → no entry; single original with one exact-match and one differing extracted value → only the differing value gets an `original_value`.

Deterministic ordering (sorted originals, substring preference) is unchanged.

## Verification

- `pytest zavod/tests/helpers/names/` — 42 passed (includes the reviewed-names tests in `test_names.py`)
- `mypy --strict --exclude zavod/tests zavod/helpers/names.py` — clean

### DIFF
--- a/zavod/zavod/helpers/names.py
+++ b/zavod/zavod/helpers/names.py
@@ -468,8 +468,8 @@ def derive_original_values(original: Names, extracted: Names) -> Dict[str, str]:
     Derive an original_value for each value in extracted based on the values in original.
 
     For each value in extracted:
-        (1) If there's exactly one value in original, use that for all names.
-        (2) If some value in original matches it exactly, leave blank - no original_value needed.
+        (1) If some value in original matches it exactly, leave blank - no original_value needed.
+        (2) If there's exactly one value in original, use that for all names.
         (3) If some value in original contains it, we can use that the value from original as original_value.
         Otherwise leave blank - this is best-effort only.
     """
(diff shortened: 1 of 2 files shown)


pudo:
Really not sure about this one, @jbothma.

jbothma (approved):
yeah looks reasonable! Nice improvement for statements bloat

PR #4958, opened 2026-07-17
extract: include model in LLM prompt cache keys

### LINKED ISSUE #4956: extract: LLM prompt cache key omits the model — model bumps serve stale outputs under the new model's origin label

`run_typed_text_prompt` (`zavod/zavod/extract/llm.py`) caches responses under a key of sha1(input string + prompt + response schema) — the `model` parameter is not part of the key. `clean_names` passes `model=LLM_MODEL_VERSION` and `_review_names` records `origin = LLM_MODEL_VERSION` on the resulting review (`zavod/zavod/extract/names/clean.py:256-262`, `zavod/zavod/helpers/names.py:572-573`).

When `LLM_MODEL_VERSION` is bumped, cached responses from the *old* model keep being served for the full `cache_days=100`, and are attributed to the *new* model in review provenance — misattributed origins on human-facing review records, and no actual re-extraction despite the version bump.

`run_typed_image_prompt` has the same omission.

## Blast radius

Every crawler with `llm_cleaning=True` (gb/fca_notices, afdb_sanctions, ru/cbr_banks, …), on every model-version bump.

## Fix direction

Include the model identifier in the cache key for both text and image prompt functions. Note the deliberate side effect: the first run after deploying the fix re-fetches everything currently cached (one-time API cost) — worth stating in the PR/changelog.

pudo:
Fixes #4956

The LLM prompt functions in `zavod/extract/llm.py` built their response cache key from the input (string / image), prompt and — for the typed variants — response schema, but not the `model` parameter. After an `LLM_MODEL_VERSION` bump, cached responses generated by the *old* model kept being served for the full `cache_days=100`, while `_review_names` recorded the *new* model as `origin` on human-facing review records: misattributed provenance and no actual re-extraction despite the version bump.

This hashes the model identifier into the cache key of all four prompt functions (`run_text_prompt`, `run_typed_text_prompt`, `run_image_prompt`, `run_typed_image_prompt`), in the same slot after the existing parts. No cache-layer changes otherwise.

## Deliberate side effect: full cache invalidation

**Changing the key composition invalidates ALL currently-cached LLM responses.** The first crawl of each `llm_cleaning=True` dataset (gb/fca_notices, afdb_sanctions, ru/cbr_banks, …) after this merges will re-fetch everything from the OpenAI API — a one-time API cost. In exchange, the recorded review `origin` (`LLM_MODEL_VERSION`) is truthful again after future model bumps, and bumping the model actually triggers re-extraction.

## Tests

New `zavod/tests/extract/test_llm.py` mocks the OpenAI client and asserts, for all four prompt functions, that identical input/prompt(/schema) with two different `model` values produces two upstream fetches, while a repeat call with the same model is served from the cache.

- `python -m pytest zavod/tests/ -k llm -q` — 8 passed
- `python -m pytest zavod/tests/extract zavod/tests/test_tune.py zavod/tests/helpers/names/test_names.py -q` — passed
- `mypy --strict --exclude zavod/tests zavod/extract/llm.py` — clean

### DIFF
--- a/zavod/zavod/extract/llm.py
+++ b/zavod/zavod/extract/llm.py
@@ -99,6 +99,7 @@ def run_typed_image_prompt(
     cache_hash.update(prompt.encode("utf-8"))
     json_schema = response_type.model_json_schema()
     cache_hash.update(json.dumps(json_schema, sort_keys=True).encode("utf-8"))
+    cache_hash.update(model.encode("utf-8"))
     cache_key = cache_hash.hexdigest()
     cached_data = context.cache.get_json(cache_key, max_age=cache_days)
     if cached_data is not None:
@@ -197,6 +198,7 @@ def run_typed_text_prompt(
     cache_hash.update(prompt.encode("utf-8"))
     json_schema = response_type.model_json_schema()
     cache_hash.update(json.dumps(json_schema, sort_keys=True).encode("utf-8"))
+    cache_hash.update(model.encode("utf-8"))
     cache_key = cache_hash.hexdigest()
     cached_data = context.cache.get_json(cache_key, max_age=cache_days)
     if cached_data is not None:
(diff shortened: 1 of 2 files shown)


pudo:
Strong disagree.

Issue #4940, opened 2026-07-17
[be_fod_sanctions] Two-digit birth years pivot into the future — 5 sanctioned persons published with DOBs like 2068-07-16


pudo:
Found in a systematic audit of the zavod helper library (2026-07-17), verified against the **live** source.

`datasets/be/fod_sanctions/be_fod_sanctions.yml:47` configures `formats: ["%d-%m-%y"]`. The Belgian consolidated list (`https://sifi.minfin.fgov.be/public/api/consolidated-list`) publishes two-digit birth years, and `strptime`'s `%y` pivot (00–68 → 20xx, 69–99 → 19xx) sends pre-1969 birth dates into the future. The crawler applies this at `crawler.py:74` via `h.apply_dates(entity, "birthDate", ...)`, and prefixdate/FtM accept any year up to 2999 — so the wrong dates are stored as fully valid.

## Current damage (checked 2026-07-17)

5 of 137 two-digit DOB rows on the live list pivot into the future, e.g.:

- Abdelaziz BOULHIR, `16-07-68` → **2068-07-16**
- Oumar MAKHMOUDOV, `23-10-64` → **2064-10-23**

These are birth dates of sanctioned persons on a ~4000-person screening list; a wrong DOB directly degrades match quality.

## Fix

Handle the century explicitly in the crawler (e.g. parse the 2-digit year and subtract 100 when the result is in the future) or normalize the raw value before `apply_dates`. A general helper-side guardrail against `%y` future-dates is tracked in #4930 — this issue is the immediate dataset fix, since the corruption is live.

PR #4973, opened 2026-07-17
Guard against birth dates parsing into the future

### LINKED ISSUE #4930: helpers: no guardrail against %y date formats parsing birth dates into the future

`extract_date`/`apply_date` pass dataset-configured format strings straight to `datetime.strptime` (`zavod/zavod/helpers/dates.py:100`). `strptime`'s `%y` pivot maps 00–68 → 20xx and 69–99 → 19xx, so two-digit historical dates near the pivot silently parse into the future: `strptime("16-07-68", "%d-%m-%y")` → **2068**-07-16. prefixdate/FtM accept any year 1000–2999, so the wrong value is stored as fully valid — silently wrong is worse than crashing.

This is corrupting production data today in `be_fod_sanctions` (separate issue for the dataset fix). Ten more datasets configure `%y` and are exposed the moment a pre-1969 date appears in the field: `us_sc_med_exclusions`, `us_mo_med_exclusions`, `us_bis_antiboycott`, `c4ads_xinjiang` / `c4ads_xinjiang_mining`, `un_ga_protocol`, `un_1718_vessels`, `icij_offshoreleaks`, `in_nse_debarred`, `il_mod_crypto`.

## Fix direction

Two complementary guardrails in the helper:

1. For birth-date-like properties, reject (or warn loudly on) parse results later than RUN_TIME — a future `birthDate` is always wrong regardless of format.
2. Warn (at dataset validation or first use) whenever a configured format contains `%y`, nudging maintainers toward explicit century handling.

pudo:
`strptime`'s `%y` pivot maps two-digit years 00–68 to 20xx, so historical dates near the pivot silently parse into the future: `"16-07-68"` with format `%d-%m-%y` becomes **2068**-07-16, and is stored as a fully valid FtM date. Silently wrong is worse than crashing — this is corrupting production data in `be_fod_sanctions` today, with more datasets configuring `%y` and exposed.

Closes #4930.

## Changes

- `runtime/cleaning.py`: entity property cleaning rejects any `birthDate` value that starts after `RUN_TIME`, with a loud warning. Living in `value_clean` (rather than `h.apply_date`), the guard covers every path that sets a birth date — `apply_date`, direct `entity.add`, enrichment. Non-canonical values skip the check and fall through to the existing validity handling. Uses `starts_after` from `rigour.dates` (rigour 2.2.2).
- `zavod/pyproject.toml`: rigour pinned to 2.2.2.

The datasets currently configuring `%y` formats are inventoried in #4974; the dataset-level fix for `be_fod_sanctions` (explicit century handling in the crawler) is tracked separately in #4940.

### DIFF
--- a/zavod/pyproject.toml
+++ b/zavod/pyproject.toml
@@ -20,7 +20,7 @@ dependencies = [
     "followthemoney == 4.10.0",
     "nomenklatura == 4.11.4",
     "plyvel == 1.5.1", # Also update macOS install docs
-    "rigour == 2.2.0",
+    "rigour == 2.2.2",
     "datapatch >= 1.2.4, < 1.3",
     "banal >= 1.1.2, < 2.0",
     "certifi",
--- a/zavod/zavod/runtime/cleaning.py
+++ b/zavod/zavod/runtime/cleaning.py
@@ -1,6 +1,7 @@
 import unicodedata
 from typing import TYPE_CHECKING
 from typing import Optional, Generator, Tuple
+from rigour.dates import starts_after
 from rigour.ids import get_identifier_format
 from rigour.names import is_name
 from prefixdate.precision import Precision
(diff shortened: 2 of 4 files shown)


leonhandreke (approved):
I think this is a good idea.

leonhandreke:
Although I don't really understand the `cleaned=True` comment. Why doesn't rigour handle this case?

leonhandreke:
Rebased, merging this now. Let's see what we surface!

PR #4981, opened 2026-07-17
[meta] Remap OFAC and DDTC program attributions

pudo:
The data-affecting half of the US program-metadata review (#4978 carries the metadata-only half). Everything here changes `programId` values on entities:

- **ITAR debarment consolidation** — `us_ddtc_debarred` emits `US-AECA-DEBARRED` for both the statutory and administrative sheets (one legal regime, 22 CFR § 127.7, two entry paths); the source-list distinction stays on `Sanction:program`. `US-DDTC-SD`/`US-DDTC-AD` are deprecated with a header comment, deletable once published statements age out. Verified by full crawl: 1502 statements on the new key, 0 on the old.
- **CAATSA split** — the § 241 oligarchs report becomes the informational `US-CAATSA-241` (no measures), referenced by `wd_oligarchs`; `US-CAATSA` covers the OFAC designations and points at `us_ofac_sdn`.
- **`HRIT-SY` → `US-SYR-REL`** — OFAC regrouped EO 13606 under PAARSS after revoking the country-wide Syria program (1 affected entry on the live SDN list).
- **`HKAA` → `US-HONGKONG`** — the tag appeared on NS-MBS entries when the EO 13936 emergency expired (July 14, 2026). Mapped to the authority-shaped program per the direction in #4980, in both OFAC lookups.

### Known gap, follow-up on this branch

The cons crawler currently derives the program from the *list name* for non-SDN entries, so the `HKAA` mapping in `us_ofac_cons.yml` does not fire yet — the 39 HKAA entries still get `US-NS-MBS`, and their new `'HKAA Section 6 Information:'` feature is dropped with "Missing feature" warnings. Fixing this properly means letting cons entries use their Program-type `SanctionsMeasure` comment (the slot SDN entries already use), which starts the #4980 migration for the multi-authority lists.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #4980, opened 2026-07-17
[meta] Map OFAC programs by legal authority, not list membership


pudo:
OFAC's data model has two orthogonal dimensions that our flat program keys currently mix:

- **Authority** — the program tag on each entry (`CAATSA - RUSSIA`, `PEESA`, `UKRAINE-EO13662`, `GLOMAG`, …): *why* someone is sanctioned, under which legal regime.
- **List** — SDN vs. the Consolidated (non-SDN) lists (NS-MBS, SSI, FSE, CAPTA, NS-CMIC, NS-PLC): *which mechanism/severity* applies. A person can move between lists under the same authority (e.g. blocking gets selected from the CAATSA menu → NS-MBS entry moves to the SDN List).

Some of our keys are authority-shaped (`US-CAATSA`, `US-GLOMAG`, `US-RUSHAR`), others are list-shaped (`US-NS-MBS`, `US-CAPTA`, `US-SSI`, `US-FSE`, `US-NS-CMIC`). The inconsistency is visible in the wiring today: `us_ofac_sdn` maps entries by authority tag and nulls out list names, while `us_ofac_cons` maps entries by list name. The same CAATSA menu-based designation therefore surfaces as `US-CAATSA` via one dataset and `US-NS-MBS` via the other.

### Proposal

Adopt the rule: **a program identifies the legal authority/regime; list membership is a delivery mechanism**, expressed through the sanction's `measures`, description, and (where useful) the source list name — consistent with `meta/README.md`'s "programs are law- and policy-centred" principle.

Concretely:

- **Multi-authority lists are the real offenders.** `US-NS-MBS` and `US-CAPTA` bundle designations under several unrelated authorities (CAATSA, PEESA, UFSA; NKSR, IFSR, EO 14024 Directive 2, …). Map those entries by their per-entry authority tag instead (the Consolidated List carries the same program tags as the SDN List), then retire the two list-shaped keys.
- **1:1 lists can keep their keys.** `US-SSI` (EO 13662), `US-FSE` (EO 13608), `US-NS-CMIC` (EO 13959/14032), `US-NS-PLC` each correspond to a single authority, so the key already *is* the authority — only the framing in title/summary needs to make that clear.
- Program-level `measures` then stay truthful per regime (menu restrictions vs. blocking), instead of a list-shaped catch-all.

### Notes

[...]

PR #5000, opened 2026-07-20
[us_ofac] Fix Hong Kong program attribution

pudo:
## Summary

OFAC moved 39 Hong Kong Autonomy Act designations onto the Non-SDN Menu-Based Sanctions List after the EO 13936 emergency expired. The advanced XML identifies these entries with the source program tag `HKAA`, while the crawler currently attributes all CONS entries using only their list name. This assigned the generic `US-NS-MBS` program and produced 39 warnings for the new `HKAA Section 6 Information:` feature.

This change:

- extracts the Program-type sanctions measure for both SDN and CONS entries;
- uses `HKAA` as the program lookup key only when that exact source tag is present, while retaining `Non-SDN Menu-Based Sanctions List` as `Sanction:program`;
- maps `HKAA` to `US-HONGKONG` in both OFAC dataset configurations;
- treats `HKAA Section 6 Information:` as a sanction summary and registers it in the shared feature allowlist.

The special case deliberately avoids changing the attribution model for other multi-authority CONS entries. It is the Hong Kong-only subset of the broader program work in #4981.

## Validation

Both datasets were crawled in full using the current OFAC advanced XML sources. Both completed with zero issues.

### SDN

Compared the local output with production archive `20260720081029-iyc` using `ftm-stmt diff`:

- 698,331 statements parsed on each side;
- 0 statements added;
- 0 statements removed.

The additional `HKAA` lookup therefore has no effect on the current SDN output.

### CONS

Compared the local output with production archive `20260720063501-jnm` using `ftm-stmt diff`:

- exactly 39 source entries are tagged `HKAA`;
- exactly those 39 persons and their 39 linked sanctions change;
- 117 statements are removed and 156 are added.

For each of the 39 records, the complete diff is:

- Person `programId`: `US-NS-MBS` -> `US-HONGKONG`;
- Sanction `programId`: `US-NS-MBS` -> `US-HONGKONG`, with original value `HKAA`;
- Sanction `programUrl`: generic NS-MBS page -> Hong Kong program page;
- one `HKAA Section 6 Information` sanction summary added.

No `Sanction:program` statements change. All 39 sanction `authorityId` values were checked against the corresponding changed person IDs.

The source also contains 12 non-HKAA NS-MBS records across five other program tags. All 12 retain `US-NS-MBS`, none gain `US-HONGKONG`, and none appear in the production diff.

## Checks

- full `us_ofac_cons` crawl: passed, zero issues;
- full `us_ofac_sdn` crawl: passed, zero issues;
- `ftm-stmt` production comparisons: passed;
- exact statement-shape and source-membership assertions: passed;
- Ruff and formatting: passed;
- YAML lint and dataset mypy checks: passed;
- `git diff --check`: passed.

Issue: https://www.opensanctions.org/issues/us_ofac_cons/

### DIFF
--- a/datasets/us/ofac/ofac_advanced.py
+++ b/datasets/us/ofac/ofac_advanced.py
@@ -576,6 +576,9 @@ def emit_sanctions_entry(
     # HKAA entries are published on the multi-authority NS-MBS list. Use the
     # source program tag for their program ID without changing how other
     # consolidated-list entries are attributed.
+    # TODO(#4980): other CONS entries (SSI/CMIC/NS-PLC/NS-MBS) also carry Program-type
+    # source tags (e.g. RUSSIA-EO14024, CMIC-EO13959) but stay attributed to their list
+    # program. Generalize to prefer a resolvable source tag and drop this HKAA special case.
     program_lookup_key = "HKAA" if source_program == "HKAA" else program
     # For us_ofac_sdn, only process entries with list_id 'SDN List'
     if dataset == "us_ofac_sdn" and list_id != "SDN List":
--- a/datasets/us/ofac/ofac_advanced.py
+++ b/datasets/us/ofac/ofac_advanced.py
@@ -65,6 +65,7 @@ SANCTION_FEATURES = {
     # https://www.ecfr.gov/current/title-31/part-594/section-594.201
     # "affiliates of the IRGC are identified by a special reference to the “IRGC” at the end of their entries"
     "Additional Program Tags -": "program",
+    "HKAA Section 6 Information:": "summary",
     "Secondary sanctions risk:": "summary",
     "Transactions Prohibited For Persons Owned or Controlled By U.S. Financial Institutions:": "summary",  # noqa
     "IFCA Determination -": "summary",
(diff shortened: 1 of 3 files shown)


datasets/us/ofac/us_ofac_cons.yml
```diff
@@ -80,6 +80,8 @@ lookups:
         value: US-NS-PLC
       - match: Sectoral Sanctions Identifications List
         value: US-SSI
+      - match: HKAA
+        value: US-HONGKONG
```
nvmbrasserie, line 86: Suggest documenting why HKAA maps to US-HONGKONG (context is otherwise lost once #4981 drops its commented copy of this row):

```suggestion
      - match: HKAA
        # Hong Kong Autonomy Act designations moved to the NS-MBS list when the
        # EO 13936 national emergency expired in July 2026; the authority stays US-HONGKONG.
        value: US-HONGKONG
```

datasets/us/ofac/us_ofac_sdn.yml
```diff
@@ -191,7 +191,9 @@ lookups:
         value: US-FORINT
       - match: HOSTAGES-EO14078
         value: US-HOSTAGE
-      - match: HK-EO13936
+      - match:
+          - HKAA
+          - HK-EO13936
         value: US-HONGKONG
```
nvmbrasserie, line 199: Same suggestion here — mirror the explanatory comment #4981 uses so the mapping is self-documenting on `main`:

```suggestion
      - match:
          # HKAA designations moved to the NS-MBS list when the EO 13936 national
          # emergency expired in July 2026; the authority stays US-HONGKONG.
          - HKAA
          - HK-EO13936
        value: US-HONGKONG
```

datasets/us/ofac/ofac_advanced.py
```diff
@@ -580,11 +571,12 @@ def emit_sanctions_entry(
...
-        extract_sdn_sanctions_measure_name(entry, refs)
-        if list_id == "SDN List"
-        else list_id
-    )
+    source_program = extract_sanctions_measure_name(entry, refs)
+    program = source_program if list_id == "SDN List" else list_id
+    # HKAA entries are published on the multi-authority NS-MBS list. Use the
+    # source program tag for their program ID without changing how other
+    # consolidated-list entries are attributed.
+    program_lookup_key = "HKAA" if source_program == "HKAA" else program
```
nvmbrasserie, line 582: Worth marking this as a known special-case so the debt is visible in-code and tied to the generalization tracked in #4980 (the multi-authority list migration this is a subset of):

```suggestion
    # HKAA entries are published on the multi-authority NS-MBS list. Use the
    # source program tag for their program ID without changing how other
    # consolidated-list entries are attributed.
    # TODO(#4980): other CONS entries (SSI/CMIC/NS-PLC/NS-MBS) also carry Program-type
    # source tags (e.g. RUSSIA-EO14024, CMIC-EO13959) but stay attributed to their list
    # program. Generalize to prefer a resolvable source tag and drop this HKAA special case.
    program_lookup_key = "HKAA" if source_program == "HKAA" else program
```

nvmbrasserie (approved):

PR #5001, opened 2026-07-20
[jp_mof_sanctions] Convert _x000D_ carriage-return artifacts to newlines

leonhandreke:
`_x000D_` is the OOXML escape for a carriage return (0x0D). openpyxl surfaces it
verbatim instead of un-escaping it, so cells with in-cell line breaks were dragging
the literal artifact into names, addresses, notes and other fields.

Fix in `str_cell`: replace `_x000D_` with a newline so the existing `\n` token in
SPLITS separates multi-value cells cleanly (rather than dfat's approach of deleting
it and smushing values together).

Verified against a live crawl: `_x000D_` occurrences in statements.pack go 171 → 0,
and multi-value cells like `NAME_x000D_\n<arabic>` now split into separate name values.

Note: this re-keys some entities, since `make_id` derives from the name cells we're
now cleaning.

### DIFF
--- a/datasets/jp/mof_sanctions/crawler.py
+++ b/datasets/jp/mof_sanctions/crawler.py
@@ -59,7 +59,14 @@ def str_cell(cell: Cell | MergedCell) -> str | None:
         return value.isoformat()
     if isinstance(value, bool):
         return str(value).lower()
-    return stringify(value)
+    text = stringify(value)
+    if text is None:
+        return None
+    # `_x000D_` is the OOXML escape for a carriage return (0x0D). openpyxl surfaces
+    # it verbatim instead of un-escaping it, so cells with in-cell line breaks arrive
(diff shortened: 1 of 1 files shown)


jbothma (approved):
It feels like there could be a fix closer to `hconvert_excel_cell` but I'm not sure. 

> I looked into this, and it turns out that x000D is the UTF-8 code for "\r" (carriage return), not "\n". The difference is that "\n" jumps to the beginning of the next line, whereas "\r" jumps to the beginning of the current line.

at https://github.com/troldal/OpenXLSX/issues/16 makes it sound a bit more complicated than just `replace(..., "\n") in the context of left to right and right to left text, but looking at the output here, it's doing the right thing.

So let's go with the local fix and come back to a general fix.

leonhandreke:
I agree it could be more general, but I didn't want the blast radius of tackling the helper.

Issue #5017, opened 2026-07-21
last_change is incorrect for deletions which don't empty the dataset


leonhandreke:
Follow-up to https://github.com/opensanctions/opensanctions/issues/4643.

When a dataset exports with zero statements, there's no entity to derive `last_change` from, so it comes out null and produces catalog metadata that fails validation in kombinat. As a stopgap we now fall back to the current run time so the field is always populated.

The stopgap is semantically wrong though: `last_change` is meant to be when the dataset's entity content last changed, and it doesn't track deletions — so an empty run didn't "change" anything as of now, and reporting the run time overstates freshness.

**Permanent fix:** backfill `last_change` from the last successful run's value instead of using the current run time. This needs a reliable notion of "last successful run" plus versioned artifact reads, which is the machinery tracked in https://github.com/opensanctions/operations/issues/2675.

Edge case: if there's no prior successful run (genuinely first run *and* empty), falling back to run time is fine.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

leonhandreke:
@jbothma mentioned in https://github.com/opensanctions/opensanctions/pull/5018 that @pudo said that maybe what we actually want is for `last_change` to also cover deletions. In that case, does it make sense to redefine `last_chage` as "the last time the delta for this dataset was non-empty"?

jbothma:
specifically cover the last deletion when it became empty. your suggestion sounds technically correct. that'd be one way of implementing it.

jbothma:
- when the last entity is deleted, last_change is None - we currently fall back to RUN_TIME_ISO which is correct the firs time. From then onwards, using RUN_TIME_ISO is potentially confusing since nothing has changed.
- when the latest entity is deleted, last_change goes back in time
- when any other entity is deleted, last_change doesn’t update

jbothma:
eventually we might want to be thinking about last_change at the statement level if this solution doesn't already handle e.g. same value from two sources correctly.

pudo:
If we had a database table that we use for date tracking on `Sanction:startDate` for sources which don't state that, could we perhaps use that table also to store a corrective value to handle this fluctuation? 

I also wonder if we need to step back and look at which problems we're trying to solve for end-users. There's UI display, and alerting filters - anything else?

jbothma:
yeah I guess a DB that does that could be tracking when any entity was added/removed and answer this.

for the monitoring use case, 
[...]

Issue #4643, opened 2026-06-22
broken metadata when zero entities are emitted


jbothma:
Zavod currently produces invalid metadata when no entities are emitted.

I propose we

1. validate the metadata using this model before publishing
2. prevent publishing if no entities were emitted - I don't see a use case for zero entities being a valid export.

```
  File "/Users/jdb/projects/opensanctions/operations/etl/kombinat/kombinat/meta/catalogs/__init__.py", line 48, in get_opensanctions_catalog
    datasets.append(CatalogDatasetModel.model_validate(d))
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jdb/projects/opensanctions/operations/etl/kombinat/venv/lib/python3.12/site-packages/pydantic/main.py", line 732, in model_validate
    return cls.__pydantic_validator__.validate_python(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 4 validation errors for CatalogDatasetModel
entity_count
  Field required [type=missing, input_value={'issue_levels': {'error'...s': ['default', 'peps']}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.13/v/missing
thing_count
  Field required [type=missing, input_value={'issue_levels': {'error'...s': ['default', 'peps']}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.13/v/missing
target_count
  Field required [type=missing, input_value={'issue_levels': {'error'...s': ['default', 'peps']}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.13/v/missing
last_change
  Field required [type=missing, input_value={'issue_levels': {'error'...s': ['default', 'peps']}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.13/v/missing
```

jbothma:
I now see a use case for zero entities being a valid export: if a sanctions list continues to exist, it merely doesn't contain sanctions right now.
Concretely, if we say ca_sema_sanctions_news contains notices no older than 2 weeks because it's just covering an update gap, it should be ok to be empty.

jbothma:
as for last_change:

for a first run, run_time is correct.
for an emergency fix, run_time is fine.
if there are no statements, picking whatever was last_change in the last successful run is ideal ux and maybe not horrible

leonhandreke:
[...]

PR #5018, opened 2026-07-21
Populate last_change for empty exports

leonhandreke:
Empty exports (zero statements) produced `index.json` with no `last_change`, which fails `CatalogDatasetModel` validation in kombinat. Fall back to the run time so the field is always populated. Stopgap for https://github.com/opensanctions/opensanctions/issues/4643; permanent fix (backfill from the last successful run) tracked in https://github.com/opensanctions/opensanctions/issues/5017.

### DIFF
--- a/zavod/zavod/exporters/metadata.py
+++ b/zavod/zavod/exporters/metadata.py
@@ -53,8 +53,22 @@ def get_base_dataset_metadata(dataset: Dataset) -> Dict[str, Any]:
             things = stats.get("things", {})
             meta["thing_count"] = things.get("total", 0)
             last_change = stats.get("last_change")
-            if last_change is not None:
-                meta["last_change"] = last_change
+            # Stopgap: an empty export (no statements) has no entity to derive
+            # last_change from, so it stays null and fails catalog validation
+            # downstream in kombinat
+            # (https://github.com/opensanctions/opensanctions/issues/4643). Fall
+            # back to the run time so the field is always populated.
+            #
+            # TODO: The permanent fix is to backfill last_change from the last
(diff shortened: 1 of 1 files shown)


zavod/zavod/exporters/metadata.py
```diff
@@ -53,8 +53,22 @@ def get_base_dataset_metadata(dataset: Dataset) -> Dict[str, Any]:
...
+            # downstream in kombinat
+            # (https://github.com/opensanctions/opensanctions/issues/4643). Fall
+            # back to the run time so the field is always populated.
+            #
+            # TODO: The permanent fix is to backfill last_change from the last
+            # successful run, which needs the versioned-artifact semantics from
+            # https://github.com/opensanctions/operations/issues/2675 — tracked
+            # in https://github.com/opensanctions/opensanctions/issues/5017.
+            # Counter-intuitive but correct: last_change doesn't track deletions,
+            # so an emptied dataset's last change is whatever the last successful
```
jbothma, line 67: oh now I realised why @pudo said we actually want the run when the entities were dropped - because if the first time we have zero entities we use the previous run's run time, we're persisting the wrong last run until we get entities again.

jbothma (approved):

zavod/zavod/exporters/metadata.py
leonhandreke: Ah, that changes the semantics of the whole field? I think it makes more sense, but probably a bit hard to implement since we have to hook into the delta mechanism instead of just `MAX(first_seen)` over all statements. But haven't looked into it in more detail.

PR #5022, opened 2026-07-21
[iq_aml_list] Migrate to the new JSON API

nvmbrasserie:
The aml.iq source was rebuilt from a WordPress site (xlsx downloads) into a React SPA backed by a JSON API, breaking the crawler. Consume the local sanctions list from https://api.aml-iq.com/api/localsanctionslist instead.

- Keep the Person ID scheme (name + birth year) and the LegalEntity scheme identical to the pre-migration crawler so published IDs stay stable: 27/27 entities and 5658/5662 persons retain their IDs.
- Fetch the whole list in one request; the API's paginated ordering is unstable (pages overlap and drop records). Assert it isn't truncated.
- Scope to the local list only (the foreign list is login-gated).
- Retune assertions and lookups for the new source.

### DIFF
--- a/datasets/iq/aml_list/crawler.py
+++ b/datasets/iq/aml_list/crawler.py
@@ -1,94 +1,94 @@
 import re
-from typing import Optional, Set, Dict, List
+from typing import Any
 
 from normality import latinize_text
-from openpyxl import load_workbook
-from rigour.mime.types import XLSX
-from zavod.shed.trans import (
-    apply_translit_full_name,
-    ENGLISH,
-)
-from zavod.extract.zyte_api import fetch_html, fetch_resource
+
 from zavod import Context
 from zavod import helpers as h
+from zavod.shed.trans import apply_translit_full_name, ENGLISH
 
-# These sheets do not contain actual data but serve as reference sheets
-LOCAL_FILE_IGNORE_SHEETS = [
-    "الافراد",  # Individuals
-    "القوائم المحلية ",  # Local lists
-]
-NAME_SPLITS = [
-    "المكنى",  # Nickname
-    # It looks like what's in parentheses is always a nickname, sometimes with,
-    # sometimes without "المكنى" (nicknamed)
-    "(",
-    ")",
-]
+# The local sanctions list is served by the platform's JSON API. Its paginated
+# ordering is unstable — successive pages overlap and drop records — so we fetch
+# the whole list in a single request with a page size well above the row count
+# and assert it wasn't truncated.
+API_URL = "https://api.aml-iq.com/api/localsanctionslist/getall/1"
+PAGE_SIZE = 100_000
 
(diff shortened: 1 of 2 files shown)


datasets/iq/aml_list/crawler.py
```diff
@@ -1,165 +1,127 @@
...
-
-
-def crawl_row(row: Dict[str, str | None], context: Context) -> None:
-    row.pop("id", None)
-    # Skip empty rows
-    if not any(row.values()):
+    # Excluded (delisted) records.
+    is_excluded = item.pop("isExcluded")
+    if is_excluded:
+        context.log.warning(
```
leonhandreke: Should this be a warning, is it actionable for us? what does "representation not yet defined" mean?

datasets/iq/aml_list/crawler.py
```diff
@@ -1,165 +1,127 @@
...
+        )
         return
-    raw_entity_name = row.pop("entity_name", None)
-    decision_number = row.pop("decision_no")
-    entity_name = clean_entity_name(raw_entity_name)
 
-    if entity_name:
+    # The numeric `type` maps to a FollowTheMoney schema via the `type` lookup;
+    # an unmapped value signals a new category that needs handling.
+    schema = context.lookup_value("type", str(entity_type))
```
leonhandreke: I think I preferred the version with constants in Python, there is only two types and they will remain around forever.

datasets/iq/aml_list/crawler.py
```diff
@@ -1,165 +1,127 @@
...
-        entity.id = context.make_id(raw_person_name, birth_date)
-        entity.add("nationality", row.pop("nationality", None), lang="ara")
-        h.apply_date(entity, "birthDate", birth_date)
-        h.apply_name(
-            entity,
-            full=name,
-            matronymic=row.pop("matronymic"),
-            lang="ara",
-        )
+        # ID scheme kept identical to the pre-migration crawler (name + birth
```
leonhandreke: (nit) Don't need this comment about what was there previously.

datasets/iq/aml_list/crawler.py
```diff
@@ -1,165 +1,127 @@
...
-    filename: str,
-    titles: Set[str],
-    ignore_sheets: List[str] = [],
-) -> None:
-    context.log.info(f"Processing {filename} from {url}")
-    excel_link_xpath = (
-        '//article[contains(@id, "post-")]//a[contains(@href, "xlsx")]/@href'
+    context.audit_data(
+        item,
+        ignore=["id", "createdAt", "typeAr", "typeEn", "typeKu", "excludeAttachment"],
```
leonhandreke: How do `decisionYear` and `createdAt` relate? Is one maybe `startDate` and the other `listingDate`?

leonhandreke (commented):
Nice, this cleans things up!

leonhandreke (approved):

PR #5023, opened 2026-07-21
Translate position name later

goose-life:
- checks for explicit `is_pep = False` first — [`categorise_many()` is read-only](https://github.com/opensanctions/opensanctions/blob/main/zavod/zavod/stateful/positions.py#L111), so an untranslated caption should be fine here
- only translates position name after checking for things like `is_historical_country`


pudo (approved):
That's a really lovely compromise. I was thinking about the analyzer path, but it means that names in the position review UI are never translated, which is one of the cool features of the whole translation effort. This seems great.

PR #5025, opened 2026-07-21
Daily issues 21 07 part 2

nvmbrasserie:


datasets/im/disqualified_directors/im_disqualified_directors.yml
```diff
@@ -34,6 +34,11 @@ data:
   url: https://www.iomfsa.im/enforcement/disqualified-directors/
   format: HTML
   lang: eng
+http:
+  # The site's WAF rejects zavod's default user agent with a 200 "Request
+  # Rejected" page, yielding zero entities. A browser-like user agent gets
+  # served the real page.
+  user_agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/126.0.0.0 Safari/537.36 (zavod; opensanctions.org)
```
jbothma, line 41: without this, do we get an error or just missing entities?

if it's the latter, I think we should change the crawler to error? otherwise we're kinda trusting that it's behaving consistently based on user agent, which isn't always the case.

PR #5026, opened 2026-07-21
[us_cms_npi] Filter non-name values from name fields with is_name

nvmbrasserie:
The source misuses name fields for placeholders, phone numbers, postal codes, dates and bare digits, tripping zavod's "not a valid name" warnings. Add a clean_name() helper using rigour.names.is_name and apply it to organization name, first/last/middle name and other-name alias.


jbothma (approved):

leonhandreke:
Beautiful, thanks for doing this the right way!

PR #5027, opened 2026-07-21
expand on configured topic match or supporting schemata

### LINKED ISSUE #4566: topic gating shouldn't exclude supporting schemata
If we enabled topic gating globally, Articles would drop out. So would Address, etc.

Decision:
Refine topic-gating to be aware of supporting schemata and always include them (assuming they're linked to something that will be tagged / included? or does that happen automatically anyway).

jbothma:
Closes https://github.com/opensanctions/opensanctions/issues/4566

"supporting" schemata like Article, Identification, etc, never get risk topics. This defines a relevant set of superclasses which are always emitted when expanding related entities. Supporting edges (e.g. Documentation) are emitted if both vertices are "publishable" (topic and supporting)

@leonhandreke maybe good to get a sense whether this code is understandable. the intro in zavod/runner/README.md should help.

### DIFF
--- a/zavod/zavod/runner/local_enricher.py
+++ b/zavod/zavod/runner/local_enricher.py
@@ -223,8 +223,8 @@ def enrich(context: Context) -> None:
     enricher = LocalEnricher(context.dataset, context.cache, config)
     # The same resolved set gates expansion context (check_publishability) and
     # filters which subject entities are matched and expanded at all
-    # (_filter_entity). That coupling guarantees the confirmed match always
-    # passes the gate, so supporting entities never publish disconnected.
+    # (BaseEnricher._filter_entity). That coupling guarantees the confirmed match
+    # always passes the gate, so supporting entities never publish disconnected.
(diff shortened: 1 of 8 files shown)


zavod/zavod/runner/local_enricher.py
```diff
@@ -195,9 +195,12 @@ def save_match(
         expanded = [adj for adj in expanded if not check_person_cutoff(adj)]
 
         if topic_gated:
-            topic_matches = check_enrich_topics(expanded, subject_view, enrich_topics)
+            # The first expansion result is the confirmed match itself. Keep it
+            # visible; gate the graph context that follows it on risk topics assigned
+            # by the subject datasets and analyzers.
+            publishable = check_publishability(expanded, subject_view, enrich_topics)
```
jbothma, line 207: update comment - we're adding always-emit to supporting schemata.

check that we'll only emit supporting schemata if the thing they're connected to is publishable. Automatically covered for edges but not sure about non-edges now.

PR #5029, opened 2026-07-21
[wd_peps] Usage-driven position discovery

pudo:
## What

Rewrites how `wd_peps` discovers Wikidata positions: **by evidence of use** (someone holds the position via P39, or it names an officeholder via P1308) instead of traversing the subclass tree of Q4164871 ("position") per country. The ontology is consulted only *upward* (`Item.types`) to classify candidates. The stateful position review DB becomes the single curation surface and always feeds the crawl, so a failing WDQS query can delay new positions but never drop known ones.

- deletes the ~500-line `position_subclasses` triage lookup from the yml
- discovery drops from ~58k SPARQL queries (~290 transitive-closure class queries × ~200 territories) to **814 flat queries** (590 territory usage sweeps incl. alias QIDs like the RU/BY Union State + 224 citizenship/occupation sweeps for politicians, diplomats and judges)
- persons are emitted exactly once, at the end, with occupancies for **all** their accepted positions (previously a person was re-fetched per position, and P39 claims pointing at positions swept under other countries were dropped)
- `position_holders()` now also honors P1308 (officeholder) claims
- classification precedence made explicit: **DB verdict > ALLOW_TYPES > EXCLUDE_TYPES > type gate**; the new EXCLUDE list is deliberately conservative (priests, religious occupations, honorary titles) since an abolished or obscure position can still make living former holders PEPs; reviewed-PEP positions are exempt from the dissolution cutoff, now unified at 1990-12-26
- run-level memory is bounded: QID-keyed sets plus a 5k-LRU position builder that reconstructs entities from the SQL item cache on eviction; DB transaction is committed per territory / 500 candidates / 1000 persons instead of accumulating for hours

## Parity

New sweeps vs the positions emitted by the current pipeline (local full run, 2026-07-04), all misses individually verified against live Wikidata:

| country | baseline | new candidates | net missing | resolution |
|---|---|---|---|---|
| de | 329 | 3,312 | 1 | Ghana's ambassador to Germany — found via Ghana's sweep |
| ru | 1,876 | 2,356 | 4 | found via ua/ee/kp/by sweeps |
| us | 1,667 | 4,647 | 4 | EU/BG/OAS ambassadorships — found via holders' citizenship sweeps |
| bn | 33 | 55 | 0 | — |

Structurally: the old politician query is a strict subset of the new occupation sweep, and any old class-tree hit that was actually *emitted* had a P39 holder — exactly the new usage sweep's condition.

## Verification

- `mypy --strict` clean on the crawler and `zavod/shed/wikidata`; all 257 zavod tests pass
- bounded end-to-end run (bn+ee): 289 candidates discovered in 5s, 165 accepted, person pass emitting correctly incl. cross-border positions (Finnish/German parliament seats via Estonian citizens); still completing — full numbers to follow before undrafting
- known judgment call: elected religious offices (e.g. Archbishop of the Estonian Evangelical Lutheran Church) hit both priest-EXCLUDE and elective-office-ALLOW; allow wins, review DB is the correction path (same rule that correctly keeps the Pope includable)

## Follow-ups (out of scope)

- `wd_categories`: drop `crawl_position_seeds` descendant expansion + holder walking (now wd_peps' job); keep PetScan/declarator and an explicit IGO seed list (IGO positions have no FtM country and remain excluded here)
- consider enrolling unclassifiable-but-in-use positions as `is_pep=NULL` review candidates instead of dropping silently
- optional: batch `position_holders` lookups via VALUES to cut the ~60-80k holder queries per cold sweep


pudo:
## Bounded end-to-end verification (bn-only, completed)

Full `crawl()` run limited to Brunei, scratch data path, registry backfill disabled to keep it bounded: **51s, exit 0, zero issues**. 55 candidates discovered → 39 accepted → **31 positions, 71 persons, 81 occupancies emitted**, including the end-of-run position emission. Peak RSS 762MB, most of it the timestamp index (streams 3.86M archived statements — fixed per-run cost).

Reconciliation vs the 2026-07-04 baseline (33 bn positions):

- **4 gained**: Minister of Defence, ambassador of Brunei to France, ambassador to Japan, Pengiran Perdana Wazir — the occupation sweep + P1308 doing their job.
- **5 missing because the local review DB lacks prod's human verdicts**: Commander of the Armed Forces/Navy/Air Force, Speaker of the Legislative Council, State Mufti — all were discovered and enrolled (`is_pep=false`, no `role.pep` ancestry); in production the DB rows say `is_pep=true` and the registry backfill + DB-precedence emit them. Verified by inspecting the rows this run inserted.
- **1 missing due to retention-window drift**: Minister of Communications was accepted, but both holders' tenures fell out of the (unchanged) `make_occupancy` PEP-retention rules between July 4 and today; one holder still emits via another position.

An earlier ee+bn run (interrupted) additionally emitted 4,127 persons / 5,950 occupancies through the same person pass with an empty issues log, incl. cross-border positions via Estonian citizens (Finnish parliament, Bundestag, PACE).

Marking ready for review. First production run should be watched for: assertion counts, WDQS load during the cold holder-query sweep, and new review-queue volume.

PR #5034, opened 2026-07-22
Validate dataset catalog metadata in zavod before writing

leonhandreke:
Adds a `CatalogDatasetModel` (+ published-form `ResourceModel`) in the zavod `exporters.metadata` package and validates the assembled `index.json` against it in `write_dataset_index`.

The model **extends FtM `DatasetModel`** (the loose read schema) and tightens it into the published output contract: `version` / `updated_at` / `last_export` become required, `resources` use the stricter `ResourceModel`, and it adds the operational fields that only exist after a run (counts, `last_change`, issue levels, artifact URLs, `result`). The statistics-derived fields are required only for successful runs — a failed run legitimately drops its statistics.

For now this only **warns** on non-conformance so we can gauge how often runs actually violate the contract before making it fatal — https://github.com/opensanctions/opensanctions/issues/4643.

kombinat keeps its own catalog validation at assembly time for now; unifying the two (and eventually using this model to *build* the metadata rather than validating a hand-assembled dict) is follow-up work.

### DIFF
--- a/zavod/zavod/exporters/metadata/__init__.py
+++ b/zavod/zavod/exporters/metadata/__init__.py
@@ -2,9 +2,12 @@ from enum import StrEnum
 import json
 from typing import Any
 
+from pydantic import ValidationError
+
 from zavod import settings
 from zavod.logs import get_logger
 from zavod.meta import Dataset, get_catalog
+from zavod.exporters.metadata.model import CatalogDatasetModel
 from zavod.archive import INDEX_FILE, STATISTICS_FILE, ISSUES_FILE
 from zavod.archive import CATALOG_FILE, DELTA_INDEX_FILE, DELTA_EXPORT_FILE
 from zavod.archive import UNLISTED_RESOURCES
@@ -139,6 +142,21 @@ def write_dataset_index(dataset: Dataset, result: DatasetVersionResult) -> None:
                     dataset.name, version.id, DELTA_INDEX_FILE
                 )
                 break
+
+    # Validate against the published output contract before writing. The model
+    # knows a failed run legitimately lacks statistics, so a degraded failure
+    # index doesn't trip it.
+    # TODO: For now this only warns so we can gauge how often real runs violate
+    # the contract; tighten it to a hard failure once we're confident
+    # (https://github.com/opensanctions/opensanctions/issues/4643).
+    try:
+        CatalogDatasetModel.model_validate(meta)
+    except ValidationError as exc:
+        log.warning(
+            "Dataset metadata does not conform to the catalog model",
+            dataset=dataset.name,
+            errors=exc.errors(),
+        )
 
     with open(index_path, "wb") as fh:
         write_json(meta, fh)
--- a/zavod/zavod/exporters/metadata/model.py
+++ b/zavod/zavod/exporters/metadata/model.py
@@ -1,30 +1,24 @@
 """Typed model for the metadata zavod writes to a dataset's ``index.json``.
 
-This is the *output* contract for a single dataset entry in the catalog, and it
-is deliberately distinct from the two *input* models it sits downstream of:
+This is the *output* contract for a single dataset entry in the catalog. It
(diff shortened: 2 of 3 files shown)


leonhandreke:
@jbothma moved these from kombinat, so no need to review the model in detail. Maybe you could take a look at (a) the reasoning in the docstring and (b) where I put the files, how I use them? Merci!

zavod/zavod/exporters/metadata/model.py
```diff
@@ -0,0 +1,225 @@
...
+    """Internal representation of the datasets included in this collection. Prefer `datasets` for external use."""
+    deprecation: Optional[str] = None
+    """Markdown explanation of why the dataset was deprecated and what happened to it.
+    Only present when `deprecated` is True, e.g. 'The source API was shut down in April 2025...'"""
+    deprecated: bool
+    """Whether this dataset is deprecated (i.e. no longer actively maintained or updated)."""
+    entry_point: Optional[str] = None
+    disabled: bool
+    hidden: bool
+    resolve: bool = True
```
jbothma, line 97: I think hidden isn't always set:

```
grep -c hidden datasets/**/*.yml | grep 0
datasets/_collections/banks.yml:0
datasets/_collections/crime.yml:0
datasets/_collections/debarment.yml:0
datasets/_collections/default.yml:0
datasets/_collections/enforcement.yml:0
datasets/_collections/peps.yml:0
datasets/_collections/regulatory.yml:0
datasets/_collections/sanctions.yml:0
datasets/_collections/securities.yml:0
datasets/_externals/ext_eu_esma_firds.yml:0
datasets/_externals/ext_gb_fca_firds.yml:0
datasets/_externals/ext_lv_business_register.yml:0
datasets/_externals/opencorporates.yml:0
datasets/_externals/wikidata.yml:0
```

jbothma, line 97: The entirely separate model makes a lot of sense.

duplicating defaults doesn't feel ideal. can we get away with skipping optional fields, or do we miss out on important validation? I wonder if the defaults in ZavodDatasetModel are sufficiently accessible to reuse those instead of duplicating? Or does that just get really yucky?

Feel free to ignore this comment - I'm thinking aloud a bit

zavod/zavod/exporters/metadata/__init__.py
```diff
@@ -140,6 +143,21 @@ def write_dataset_index(dataset: Dataset, result: DatasetVersionResult) -> None:
                 )
                 break
 
+    # Validate against the published output contract before writing. The model
+    # knows a failed run legitimately lacks statistics, so a degraded failure
+    # index doesn't trip it.
+    # TODO: For now this only warns so we can gauge how often real runs violate
+    # the contract; tighten it to a hard failure once we're confident
+    # (https://github.com/opensanctions/opensanctions/issues/4643).
```
jbothma, line 151: what do you think of reopening this and putting it in the validating column for the couple of weeks before we make this raise?

jbothma (approved):
I think this is awesome

jbothma:
as for stable surface, I think our request and response models in our API are the same thing - maybe we should just add a nice big comment on this model in zavod so we don't forget that a change to this might be a Big Deal and should be taken seriously.

zavod/zavod/exporters/metadata/model.py
leonhandreke: Good catch, but it's fine: `Dataset.to_dict()` always sets `data["hidden"] = self.model.hidden` (default `False`), so it lands in every `index.json` even when the `.yml` omits it. Same for `disabled`.

leonhandreke: I inherited from FtM now, tightening where applicable.

I agree with you - I added a TODO that eventually, these should be passed through. But for now, I don't want to mess with the manual JSON building code (first want this validation, then mess with the building).

PR #5035, opened 2026-07-22
Fix Blank ann_graph_topics stubs

jbothma:
Stubs are being exported because Entity.external was always returning False because of the id statement. Let's ignore the id statement when determining externality. It feels a little weird - there may be other things working from a built store that care? but I think it's an ok local fix.

There's still a race condition between enrichers and ann_graph_topics - if an entity gets dropped from the enricher, it could be "blank" with just the topic stub until the next ann_graph_topics run and default export.

---

Screenshot 2026-07-22 at 13 19 50


pudo:
The core `Entity.external` fix + test from this PR is extracted onto current `main` in #5113 (with co-author credit). The other parts have been superseded by #5105 (wikidata inputs) and #5109 (analyzer now uses `has_published_substance` instead of `Entity.external`, so it can also discount its own prior patches). Closing in favour of #5113 — shout if anything got lost in translation.

jbothma:
Just moving to validating so I can confirm it's all sorted after this batch of runs

PR #5059, opened 2026-07-23
Add Cambodia Senate PEP crawler

### LINKED ISSUE #709: Cambodia Senate
### Data URL

https://senate.gov.kh/search-senator-all/

### Publisher

Senate of Cambodia

KH

### Type of data

### Coverage region

region:Asia/Pacific

### Can you tell us more?

This issue covers the upper chamber of the Cambodian Parliament (Senate / ព្រឹទ្ធសភា, 62 seats — 58 indirectly elected by commune councillors and the NA across 8 regions, 2 royal appointees, 2 NA appointees). 6-year term.

**Preferred sources**

1. **[Senate search page (senate.gov.kh)](https://senate.gov.kh/search-senator-all/)** — official human-facing senator directory; renders client-side via jQuery/DataTables.
2. **AJAX backend endpoint** — `POST https://senate.gov.kh/wp-content/themes/senate/ajax/generalsearch.php` with form-data `session=5&keyword=&chck_status=1` and headers `X-Requested-With: XMLHttpRequest` + `Referer: https://senate.gov.kh/search-senator-all/`. Returns a JSON array of exactly 62 senators with `name` (Khmer), `gender`, `dob`, `party`, `phone`, `photo`, `biography` (PDF URL), `start`, `end`, `status`. Helper endpoints expose committees/regions; per-senator detail at `/wp-content/themes/senate/ajax/detail.php?session=5&sid=<str_id>`.

- **[Senate members by region (5th legislature)](https://senate.gov.kh/senator-by-region-5/)** — regional view, same data.
- **[IPU Parline – Cambodia Senate](https://data.ipu.org/parliament/KH/KH-UC01/)** — structural metadata only.
- **Wikipedia "Senate (Cambodia)"** — narrative; no structured list of all 62 for the 5th legislature.
- **[Open Development Cambodia – Senate tag](https://opendevelopmentcambodia.net/tag/senate/)** — aggregates news/maps, no roster.

**Scope**

5th Senate (`session_id=5`), inaugurated 3 April 2024, 6-year term. Filter `session=5` for current; `session=1..4` for historical. Names are Khmer-only with embedded honorifics (e.g. `សម្តេចអគ្គមហាសេនាបតីតេជោ`).

herrhellana:
Adds a PEP crawler for the **Senate of Cambodia** (upper chamber, 5th legislature, 62 seats).

## Source
Official senator directory AJAX endpoint (`senate.gov.kh/.../generalsearch.php`, `session=5`), returning all 62 senators as JSON.

## Output
- Position: *Member of the Senate of Cambodia* (`Q21295127`)
- 62 senators emitted as PEPs with Khmer name, gender, date of birth, party and term start; biography PDF as sourceUrl.
- Guards against a source data-entry error where a few records' start date is the birth year (ignored so the senator stays current).
- Citizenship `kh` per Constitution Art. 34 (New) (see code comment).

Closes opensanctions/crawler-planning#709

### DIFF
--- a/datasets/kh/senate/crawler.py
+++ b/datasets/kh/senate/crawler.py
@@ -1,0 +1,86 @@
+from typing import Any
+
+from zavod import Context
+from zavod import helpers as h
+from zavod.stateful.positions import categorise
+
+# The senator directory is populated by this AJAX endpoint; session 5 is the current
+# (5th) legislature.
+SESSION = "5"
+HEADERS = {
+    "X-Requested-With": "XMLHttpRequest",
+    "Referer": "https://senate.gov.kh/search-senator-all/",
+}
+
+
+def crawl(context: Context) -> None:
+    position = h.make_position(
+        context,
+        name="Member of the Senate of Cambodia",
+        country="kh",
+        wikidata_id="Q21295127",
+    )
+    categorisation = categorise(context, position, default_is_pep=True)
+    if not categorisation.is_pep:
+        return
+    context.emit(position)
+
+    records: list[dict[str, Any]] = context.fetch_json(
+        context.data_url,
+        method="POST",
+        headers=HEADERS,
+        data={"session": SESSION, "keyword": "", "chck_status": "1"},
+        cache_days=1,
+    )
+    if not records:
+        raise ValueError("Senate AJAX endpoint returned no senators")
+
+    for record in records:
+        str_id = record.pop("str_id")
+        name = record.pop("name")
+        assert name, f"Empty name for senator {str_id}"
+
+        person = context.make("Person")
+        person.id = context.make_slug(str_id)
+        person.add("name", name, lang="khm")
+        person.add("gender", record.pop("gender", None))
+        h.apply_date(person, "birthDate", record.pop("dob", None))
+        person.add("political", record.pop("party", None), lang="khm")
+        person.add("sourceUrl", record.pop("biography", None))
+        # Senators must be Khmer citizens (Constitution of Cambodia, Article 34 (New)).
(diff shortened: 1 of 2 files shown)


datasets/kh/senate/crawler.py
```diff
@@ -0,0 +1,225 @@
...
+        wikidata_id="Q21295127",
+        inception_date=[INCEPTION_DATE],
+        lang="eng",
+    )
+    categorisation = categorise(context, position)
+    if not categorisation.is_pep:
+        return
+    context.emit(position)
+
+    cutoff = h.earliest_term_start(TOPICS)
```
brrttwrks: I think this whole cutoff check is redundant. Doesn't [make_occupancy](https://zavod.opensanctions.org/helpers/#zavod.helpers.make_occupancy) already do this under the hood?

brrttwrks (commented):
I think the start/end dates maybe need a double check. The checking for cutoff outside make_occupancy will likely cause issues down the road - something to avoid. I think the issue with the start date (Occupancy) often being the birth date (source data dirtiness) will correctly be handled in make_occupancy - i.e. ignore in these cases. Otherwise, maybe make start date None in these cases?

herrhellana:
what do you mean by `make_occupancy()` being able to handle DOBs passed as startDates? we have logic for determining Occupancy status based on end dates, but not for how plausible the source value is. i think we need this extra sanity check here, although it does just the bare minimum (sufficient for current state of the source data). 

that being said, we can also toss in more preventative sanity checks like compare mandate dates against `CONVENING_DATES`, since we have them hardcoded here anyway, or check start <= end. we don't typically do it in crawlers, but maybe we want to implement some checks as such not just at the level of this crawler. thinking out loud.

PR #5060, opened 2026-07-23
Add Singapore Parliament PEP crawler

### LINKED ISSUE #718: Singapore Parliament
### Data URL

https://www.parliament.gov.sg/mps/list-of-current-mps

### Publisher

Parliament of Singapore

SG

### Type of data

### Coverage region

region:Asia/Pacific

### Can you tell us more?

Singapore has a unicameral Parliament. Composition mixes elected MPs (from GRCs and SMCs), Non-Constituency Members (NCMP, from runners-up), and Nominated Members (NMP, appointed by the President). The 15th Parliament since the 2025 general election.

**Preferred sources**

1. **[Parliament of Singapore – List of Current MPs](https://www.parliament.gov.sg/mps/list-of-current-mps)** — official site. Server-rendered HTML, single page (no pagination). 108 MP entries with stable selectors (`class="mp-last-name"`, `class="mp-sort-name"`). Each card has title (Mr/Ms/Dr), full name, profile slug, portrait, and constituency string (e.g. "Sengkang GRC"). Per-MP detail at `/mps/list-of-current-mps/mp/details/{slug}`.

- **[data.gov.sg – Members of Parliament (Annual)](https://data.gov.sg/datasets/d_20f58ffbb830ae29e574715f02fa4526/view)** — aggregate gender/category counts only, no names. Unsuitable.
- **[SingStat Table M891031](https://tablebuilder.singstat.gov.sg/table/TS/M891031)** — same aggregate data.
- **IPU Parline / Wikipedia** — secondary; the official site is the authoritative complete list.

**Scope**

15th Parliament (since 2025 GE). Constituency field encodes the GRC/SMC system (e.g. "Sengkang GRC", "Bukit Panjang SMC"); NCMP/NMP designations should appear in the same field on the source page. Profile detail pages carry party, biography, and member-type details for enrichment (108 extra requests if needed). No machine-readable feed — change detection relies on full re-crawl diff.

herrhellana:
Adds a PEP crawler for the **Parliament of Singapore** (unicameral).

## Source
Official list of current MPs (`parliament.gov.sg/mps/list-of-current-mps`), single server-rendered page with stable selectors (`mp-last-name`, `mp-sort-name`, `constituency`).

## Output
- Position: *Member of the Parliament of Singapore* (`Q21294917`)
- 107 members emitted as PEPs with constituency (GRC/SMC) or NCMP/NMP designation, keyed on the per-member detail slug.
- Citizenship `sg` per Constitution Art. 44(2)(a) (see code comment).

Closes opensanctions/crawler-planning#718


nvmbrasserie (approved):

nvmbrasserie:
Refactor pass over `crawler.py` — 273 → 199 lines, output verified identical at every step except the deliberate additions.

- **Fetch the flight document directly.** 32 lines of `self.__next_f.push(...)` chunk reassembly replaced by one `fetch_text` with an `RSC` header — the server serves that document on request. Needed `_rsc` as well (our HTTP cache keys on URL alone, so the response would have collided with the page's HTML) and `encoding="utf-8"` (`text/x-component` carries no charset, so HTTP says Latin-1).
- **One `crawl_member` instead of three functions.** The sitting-member roster was crawled separately, rebuilding the same person and the same undated occupancy just to hang three more fields on them. It's now loaded into a dict and joined on the name, so `make_member` and `crawl_sitting_member` both disappear. The join is asserted complete.
- **Added `audit_data`**
- **`parse_terms`:** a `Term` per branch rather than pre-setting `start`/`end` to `None` and overwriting; `NamedTuple` over a frozen dataclass; dropped `squash_spaces`, a no-op on all 15 values.
- **Two hard failures downgraded to warnings** where the cost is one record rather than wrong data. The rest still raise: a short roster would silently drop members, and a Parliament with no date range would show its members as still sitting.
- **Added** `name_in_chinese` / `name_in_tamil` (tagged `zho` / `tam`) and `government_directory_link` as a second `sourceUrl` — +321 statements. Names no longer go through the review framework.

nvmbrasserie:
Very solid work overall! @herrhellana would you mind going through this once more when you have a moment, and flagging anything I overlooked?

PR #5065, opened 2026-07-23
Add Indonesia People's Representative Council (DPR) PEP crawler

### LINKED ISSUE #711: Indonesia Dewan Perwakilan Rakyat (DPR)
### Data URL

https://www.dpr.go.id/anggota

### Publisher

Dewan Perwakilan Rakyat Republik Indonesia

ID

### Type of data

### Coverage region

region:Asia/Pacific

### Can you tell us more?

Indonesia's bicameral legislature has the DPR (Dewan Perwakilan Rakyat, 580 seats since 2024) as the lower house and the DPD as the upper house. This issue covers the DPR. MPs are elected from 84 electoral districts (dapil) across 38 provinces.

**Preferred sources**

1. **[DPR member directory (dpr.go.id)](https://www.dpr.go.id/anggota)** — official portal. 580 MPs organized across 84 dapil; filterable index at `/anggota/index/dapil/{1..84}`, member detail at `/anggota/detail/id/{id}`. **Akamai-protected** — plain curl/WebFetch returns HTTP 403; production crawler requires Zyte or headless browser (Playwright) with browser fingerprint + pacing.

- **[SATU DATA DPR RI](https://data.dpr.go.id/)** — same Akamai gate as the main site; no openly callable API. Apiary docs exist at `https://dprgoid.docs.apiary.io/` but the underlying service is not public.
- **Wikipedia "[Daftar anggota DPR 2024-2029](https://id.wikipedia.org/wiki/Daftar_anggota_Dewan_Perwakilan_Rakyat_Republik_Indonesia_2024%E2%80%932029)"** — structured tables organized by province/dapil, marked "sementara" (provisional) and citing KPU. Useful as secondary cross-check / fallback if Akamai blocks too aggressively.
- **[OpenParliament.id](https://openparliament.id/daftar-anggota-dpr/)** — civic-tech aggregator; mixes terms, incomplete HTML, no API.
- **[BPS – DPR by sex](https://www.bps.go.id/en/statistics-table/2/MjQ4MCMy/anggota-dewan-perwakilan-rakyat--dpr--menurut-jenis-kelamin.html)** — aggregate gender stats only.
- **[KPU infopemilu DCT](https://infopemilu.kpu.go.id/Pemilu/Dct_dpr)** — candidate lists (DCT) pre-election, not elected MPs.
- **IPU Parline – Indonesia** — 403/404 to programmatic fetch; no bulk export.

**Scope**

2024–2029 term, 580 seats (up from 575), 84 dapil across 38 provinces, 8 post-threshold parliamentary parties. Honorifics (`H.`, `Hj.`, `Ir.`, `S.E.`, `Dr.`, religious titles) heavily embedded in names — strip with rigour helpers.

herrhellana:
Adds a PEP crawler for the **People's Representative Council (DPR)**, the lower chamber of Indonesia's parliament (580 seats, 84 electoral districts).

## Source
Official member directory (`dpr.go.id/anggota`), iterating the 84 dapil index pages.

⚠️ **Unverified.** The site is **Akamai-protected and client-rendered** (plain requests return HTTP 403; even archived pages are JS shells), so it is fetched via the **Zyte API** (browser rendering, Indonesian exit) and could not be run locally, nor could the rendered DOM be inspected. Members are anchored on the documented `/anggota/detail/id/<id>` link pattern (robust to CSS changes) with the link text as the name. **Please sanity-check the first production run** (~580 members across 84 dapil); selectors may need adjustment. The Wikipedia "Daftar anggota DPR 2024–2029" table is a documented fallback.

## Output
- Position: *Member of the People's Representative Council of Indonesia* (`Q21328632`)
- Members keyed on detail id, with dapil constituency.
- Citizenship `id` per Law No. 7/2017 Art. 240(1) (see code comment).

Closes opensanctions/crawler-planning#711


herrhellana:
the site migrated to: https://www.dpr.go.id/en/tentang-dpr/informasi-anggota-dewan

herrhellana:
note that there are also member detail pages featuring dob and place of birth, but its structure varies across members, so the decision was not to crawl it (+ it'll add extra GraphQL request funk). however, we're collecting those pages' URLs -- seem to be stable and live even across historical legislatures.

herrhellana:
Name-cleaning audit against the fresh historical crawl (`2,084` Persons):

- There are `187` one-word cleaned Person names. `15` were already one-word names in the source; the other `172` become one-word only after removing recognizable Indonesian honorifics and/or academic post-nominals. I found no case where a personal-name component was removed. Indonesian mononyms are expected here.
- I traced the audit through `data/datasets/id_dpr/statements.pack`, comparing each cleaned `Person:name` with its preserved `original_value`, rather than inferring from `names.txt`.
- The broad bare suffix variants need special care, but in this output they fire only in six unambiguous credential chains: bare `MA` 3 times, `S.E.` twice, and `SE`, `SH`, `S.Ds.`, and `S.M.` once each (some occur on the same person). Examples are `WALMAN SAIHAAN SH, SE, MM.MBA` and `EVA KUSUMA SUNDARI MA, MDE`; none is an isolated possible surname/initial.
- Specifically, `Suparlan, SH` is correctly cleaned to `Suparlan`: DPR records use both `Suparlan, SH`/`SUPARLAN, SH.` and plain `Suparlan`, so `SH` is the undotted `S.H.` (*Sarjana Hukum*), not personal initials.
- `Kol. (Purn). Drs. JAFAR NAINGGOLAN, M.M.` is also a title/credential construction: `Kol.` = Colonel, `(Purn)` = retired, and `Drs.` = *Doktorandus*. DPR attendance records show `KOL (PURN) DRS. JAFAR NAINGGOLAN, M.M.` and other DPR records use `JAFAR NAINGGOLAN, M.M.`, confirming the core name is `JAFAR NAINGGOLAN`. The YAML still needs an exact `Kol. (Purn).` prefix variant if we want this source spelling stripped by the exact-affix cleaner.

Conclusion: the one-word outputs are supported by the raw values and do not indicate over-stripping in this run. The remaining visible issue is under-stripping of a few military-rank variants, not excessive stripping.

herrhellana:
the total N count is also consistent with what the source exposes + historical totals we expect. 

should be ready to go.

datasets/id/dpr/crawler.py
```diff
@@ -0,0 +1,320 @@
...
+PERIODE_ID = 6
+LEGISLATURES: dict[int, tuple[str, str]] = {
+    # Period 1 (1999-2004) is outside PEP coverage and the API only exposes a
+    # handful of incomplete records for it, so it is intentionally omitted.
+    2: ("2004-10-01", "2009-09-30"),
+    3: ("2009-10-01", "2014-09-30"),
+    4: ("2014-10-01", "2019-09-30"),
+    5: ("2019-10-01", "2024-09-30"),
+    6: ("2024-10-01", "2029-09-30"),
+}
```
nvmbrasserie, line 32: The historical terms are a good addition. The one thing I'd like changed is the term structure living in the file: `PERIODE_ID`, `LEGISLATURES`, `check_for_new_legislature()` and the assert at :280 are all the same decision. Since `in_office_only` (:309) and `no_end_implies_current` (:267) key off the constant, a stale `PERIODE_ID` would mark an old legislature as current, and the probe raises after 2029 so we'd emit nothing.

The source does expose this: `wherePeriode` is a `…WhereHasConditions`, so `periode` is a relation we can select from. Could we introspect it, take the dates from there, and keep `h.earliest_term_start` as the cutoff? All four pieces then go away. `cl/chamber_deputies/crawler.py:167-191` is close to this shape, and `be/chamber/crawler.py:25-32` has the per-term dataclass, which also tidies up `crawl_member`'s arg list.

datasets/id/dpr/crawler.py
```diff
@@ -0,0 +1,320 @@
...
+# This is above the chamber's 580 seats while pagination remains supported in
+# case a legislature contains enough mid-term replacements to exceed it.
+PAGE_SIZE = 1000
+
+# The frontend signs every GraphQL request with a static token and a hardcoded HMAC
+# secret, both read from the site's axios interceptor. If the endpoint starts
+# returning 401/403, they have likely rotated: re-extract by grepping the site's JS
+# bundles (/_next/static/chunks/*.js) for "x-api-signature".
+API_TOKEN = "48a07687-2a14-4647-9d42-23d7f8ebfa45"
+SIGNING_SECRET = "LfmqpWYMaEuQA42LcDvmgbBgG4NDmZp73yr8G8pZ"
```
nvmbrasserie, line 43: Could these move to an env var plus 1Password? `os.environ.get("OPENSANCTIONS_ID_DPR_API_TOKEN")`, the way `us/congress/crawler.py:11` does it. We don't keep credentials in git, even ones the site ships in its own JS.

Worth a quick discussion on the approach too, since this is the only crawler in the tree that signs requests. 7d53a0cc8 avoided the secrets by letting the page sign for itself, and that interceptor was already rewriting `variables`, so it could rewrite `wherePeriode` as well — roughly five renders, and you'd keep the historical coverage you added. What do you think? As it stands, a key rotation leaves us with a 401 and a comment pointing at the JS bundles.

datasets/id/dpr/id_dpr.yml
```diff
@@ -0,0 +1,491 @@
...
+  # GraphQL endpoint used by the client-rendered member listing
+  url: https://www.dpr.go.id/gql
+  format: JSON
+  lang: ind
+
+tags:
+  - list.pep
+
+names:
+  prefixes_strip:
```
nvmbrasserie, line 38: The audit you ran was thorough, and it's part of what makes me think the table isn't the right home for this — 396 suffixes, around 400 of the file's 491 lines. Values like `", SE.M, SC"` and `", B, Eng"` are per-record noise rather than titles, and `name_titles.md` routes that to the review system.

The part I'd want to fix: with no `names.schema_rules` and no `is_name_irregular`, the first degree we haven't listed goes into a `Person:name` and nothing surfaces it. Could we keep the closed honorific set deterministic (`Dr.`, `H.`, `Hj.`, `KH.`, `Prof.`, `Ir.`) and pass the rest to `h.apply_reviewed_name_string(..., llm_cleaning=True)` with `reject_chars`/`reject_strings`? Best to leave `default_accepted` off so the reviews still get looked at.

datasets/id/dpr/crawler.py
```diff
@@ -0,0 +1,320 @@
...
+            context.log.info(
+                "Skipping legislature outside PEP coverage",
+                period_id=period_id,
+                period_end=period_end,
+            )
+            continue
+        members = fetch_roster(
+            context,
+            period_id,
+            in_office_only=period_id == PERIODE_ID,
```
nvmbrasserie, line 309: `in_office_only` drops current-term members who were replaced mid-term, even though we emit the equivalent people for 2004-2019. Filtering in the query also means we can't warn about them. The status column is what the filter matches on — could we select it and set the occupancy status per row instead?

On DOB/POB: `priorities.md` treats them as essential where the source has them, and with 187 one-word names a birth date does a lot of the matching work. Varying structure seems like a case for warning on the rows we can't parse rather than skipping the field. Worth checking whether `anggota` exposes `tglLahir`/`tempatLahir` — if it does, it's close to free.

datasets/id/dpr/crawler.py
```diff
@@ -0,0 +1,320 @@
...
+    orderBy: $orderBy
+  ) {
+    paginatorInfo { total currentPage lastPage hasMorePages perPage }
+    data {
+      idPeriode
+      noAnggota
+      idAnggota
+      dapil { id dapil }
+      anggota { nama id photofileLink }
+      riwayatFraksi { idFraksi fraksi { id fraksi singkatan } }
```
nvmbrasserie, line 70: No objection to keeping the query itself, a GraphQL POST needs it. But we request `noAnggota`, `photofileLink`, `dapil { id }`, `idFraksi` and `fraksi { id singkatan }` without using any of them, while the fields we do want (status, period dates, possibly DOB) aren't selected. One introspection call would let us select exactly what we map.

There's also no `context.audit_data` — acceptable with GraphQL as long as the selection stays minimal, which is the same argument. Two nits: `singkatan` is fetched and dropped, and `political` at :256 takes the full `"Fraksi Partai …"` string where we only want the party.

PR #5067, opened 2026-07-23
Add Georgia Parliament PEP crawler

### LINKED ISSUE #695: Georgia Parliament
- Data URL: https://web-api.parliament.ge/api/colleagues/deputies
- Publisher: https://parliament.ge/
- Publisher country/territory code: ge
- Type of data: PEPs (Politically Exposed Persons)
- Coverage region: region:Europe

Official Parliament backend JSON API. **Requires both** `Accept: application/json` AND `Accept-Language: en` request headers (default returns HTTP 500). Returns 101 seated MPs (statutory 150; ~49 opposition seats currently vacant due to post-2024-election boycott). Per-MP profile at `https://parliament.ge/en/parliament-members/{id}/biography`. `data.gov.ge` unreachable.

herrhellana:
Adds a PEP crawler for the **Parliament of Georgia** (unicameral).

## Source
Official member backend JSON API (`web-api.parliament.ge/api/colleagues/deputies`). Requires a browser `User-Agent` (else 403) and both `Accept` + `Accept-Language` headers (else HTTP 500); `Accept-Language: en` returns Latin-transliterated names.

## Output
- Position: *Member of the Parliament of Georgia* (`Q21290878`)
- 101 seated members emitted as PEPs (of 150 statutory; ~49 vacant after the post-2024 opposition boycott) with name, gender and date of birth.
- Citizenship `ge` per Constitution Art. 37(4) (see code comment).

Closes opensanctions/crawler-planning#695

### DIFF
--- a/datasets/ge/parliament/crawler.py
+++ b/datasets/ge/parliament/crawler.py
@@ -12,8 +12,8 @@ def crawl_member(
     position: Entity,
     categorisation: PositionCategorisation,
 ) -> None:
-    first_name = member.pop("firstName").strip().title()
-    family_name = member.pop("familyName").strip().title()
+    first_name = member.pop("firstName").strip()
+    family_name = member.pop("familyName").strip()
 
     person = context.make("Person")
     person.id = context.make_slug(str(member.pop("id")))
--- a/datasets/ge/parliament/crawler.py
+++ b/datasets/ge/parliament/crawler.py
@@ -1,15 +1,15 @@
 from typing import Any
 
-from zavod import Context
+from zavod import Context, Entity
 from zavod import helpers as h
-from zavod.stateful.positions import categorise
+from zavod.stateful.positions import PositionCategorisation, categorise
 
 # The API needs a browser User-Agent (else 403) and both an Accept and Accept-Language
 # header (else HTTP 500); Accept-Language "en" yields Latin-script (transliterated) names.
 HEADERS = {
     "User-Agent": (
         "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 "
-        "(KHTML, like Gecko) Chrome/124.0.0.0 Safari/537.36"
+        "(KHTML, like Gecko) Chrome/124.0.0.0 Safari/537.36 (zavod; opensanctions.org)"
     ),
     "Accept": "application/json",
     "Accept-Language": "en",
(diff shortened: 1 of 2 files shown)


herrhellana:
Pushed a refactor + metadata pass (production-readiness):

- Structured into a per-record `crawl_member` helper.
- **Strict-interpretation:** guard against non-MP rows (the endpoint is a generic "colleagues" feed — today all `mp=True`); fail loudly on an unknown `gender.id` instead of silently dropping it; added `context.audit_data(..., ignore=[...])` to catch new API fields. `contactInfo` (phone numbers) is deliberately ignored.
- Identify the client in the User-Agent.
- Title → **Georgia Members of Parliament**; `coverage.frequency` → **monthly**.

Output unchanged: 203 entities / 1225 statements, clean issues.log.

Left alone: the `.title()` name casing (name cleaning is out of scope for a refactor — flagging for your call). Patronymic (`fatherName`, present on ~12 rows) is available if you want to capture it as `Person:fatherName` later.

herrhellana:
this one should be ready

datasets/ge/parliament/crawler.py
```diff
@@ -0,0 +1,98 @@
...
+    biography = member.pop("biography")
+    if biography:
+        doc = html.fromstring(f"<div>{biography}</div>")
+        paragraphs = [
+            h.element_text(el) for el in h.xpath_elements(doc, ".//p | .//li")
+        ]
+        person.add("biography", "\n".join(p for p in paragraphs if p))
+
+    contact_info = member.pop("contactInfo")
+    for email in contact_info["emails"]:
```
leonhandreke, line 38: We generally try to avoid adding phone numbers for politicians. Emails also? Not sure.

datasets/ge/parliament/crawler.py
```diff
@@ -0,0 +1,98 @@
...
+    family_name = member.pop("familyName").strip().title()
+
+    person = context.make("Person")
+    person.id = context.make_slug(str(member.pop("id")))
+    h.apply_name(person, first_name=first_name, last_name=family_name, lang="eng")
+    person.add("gender", str(member.pop("gender")["id"]))
+
+    h.apply_date(person, "birthDate", member.pop("dob"))
+    person.add("birthPlace", member.pop("birthPlace"), lang="eng")
+    person.add("notes", member.pop("declaration"))  # link to declaration pdf
```
leonhandreke, line 25: Not sure we need this?

leonhandreke (commented):
Added some commits, among them filling `fatherName`. Would you agree?

datasets/ge/parliament/crawler.py
herrhellana: yeah i wasnt sure myself, but those are just nice URLs. albeit in georgian. maybe in indulged a bit into my tendency to preserve data

datasets/ge/parliament/crawler.py
herrhellana: i think we leave emails, but yes true for phone numbers. i forgot.

herrhellana:
i don't think we need `fatherName` because it's mostly `None`, and Georgia doesn't really use patronymic names in official documents the way Russia does. I think it's safe to omit it.

PR #5073, opened 2026-07-23
[cd_national_assembly] Add DR Congo National Assembly PEP crawler

herrhellana:
## Summary
Adds a PEP crawler for the **National Assembly of the DR Congo** (lower house), the clean, buildable-now half of the parliament.

- Source: the National Assembly's official **WordPress REST API** (`assembleenationale.cd/wp-json/wp/v2/deputes`, `_embed`ded taxonomies), current legislature `2024 - 2028` (`legislature=192`).
- Scoped to **currently-serving** deputies (`mandats = "En cours"`; substitutes who have taken up a seat also show this). Suspended (40) and ended (21) mandates and 7 status-less records are counted and logged, not emitted.
- Emits one `Position` (Member of the National Assembly of the DRC, `Q21295979`, `gov.national`/`gov.legislative`) and a current `Occupancy` per deputy with name, electoral constituency + province, `citizenship=cd`.

## Notes / scoping
- The API exposes **no per-deputy party/parliamentary-group link** (`acf` empty; `partis-politiques`/`groupes` are unreferenced separate post types), so political affiliation is not captured. The `role-groupe-parlementaire` taxonomy is an intra-group *role* (Membre/Président/…), not a party, so it is intentionally not used.
- **Senate not included:** `senat.cd` is currently in maintenance mode (last-good roster only via Wayback). Tracked in the planning issue; a `cd_senate` crawler can follow once the site is back.

## Citizenship basis
Deputies must be Congolese nationals (Constitution Art. 102(1): "être Congolais"). Cited in a code comment.

## Validation
- `zavod crawl` → **971 entities** (485 Person + 485 Occupancy + 1 Position), **0 issues**.
- `zavod validate` passes; assertions 400–560 Person.
- Integrity checks: no dangling occupancy holder/post refs; all `role.pep` persons hold an occupancy and carry citizenship.
- `mypy --strict` clean.

Part of opensanctions/crawler-planning#836

### DIFF
--- a/datasets/cd/national_assembly/cd_national_assembly.yml
+++ b/datasets/cd/national_assembly/cd_national_assembly.yml
@@ -1,0 +1,47 @@
+title: DR Congo Members of the National Assembly
+entry_point: crawler.py
+prefix: cd-na
+coverage:
+  frequency: monthly
+  start: 2026-07-23
+load_statements: true
+ci_test: false # Live external government website, not available in CI
+summary: >-
+  Deputies of the National Assembly of the Democratic Republic of the Congo, the lower
+  house of Parliament
+description: |
+  Deputies of the National Assembly (Assemblée nationale) of the Democratic Republic of
+  the Congo, the lower house of the bicameral Parliament. It has 500 deputies elected for
+  five-year terms in multi-member constituencies.
+
+  This dataset records each deputy of the current legislature with their name, political
+  group, electoral constituency and province, sourced from the National Assembly's
+  official website.
+url: https://assembleenationale.cd/tous-les-deputes/
+publisher:
+  name: Assemblée nationale de la République démocratique du Congo
+  description: >
+    The National Assembly is the lower house of the bicameral Parliament of the
+    Democratic Republic of the Congo, alongside the Senate.
+  url: https://assembleenationale.cd/
+  country: cd
+  official: true
+data:
+  url: https://assembleenationale.cd/wp-json/wp/v2/deputes
+  format: JSON
+  lang: fra
+
+tags:
+  - list.pep
+
+assertions:
+  min:
+    schema_entities:
+      Person: 400
+      Position: 1
+    country_entities:
+      cd: 400
+  max:
+    schema_entities:
+      Person: 560
+      Position: 1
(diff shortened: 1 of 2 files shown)


herrhellana:
Checked whether this can be extended to cover past legislatures. It can't — evidence:

`/wp-json/wp/v2/legislature` lists 14 terms, but querying `deputes?legislature=<id>` per term and reading `X-WP-Total` returns deputies for **2024–2028 only**; every older term returns 0. The non-zero `count` on the older terms comes from other post types sharing the `legislature` taxonomy (`anciens-membres`, `historique`, `groupes`, …), not from deputy records.

So instead of multi-term coverage, this branch now:

- walks the `legislature` taxonomy bounded by `h.earliest_term_start(...)` rather than hardcoding term id `192`;
- drops the current-only filter, keeping ended and suspended mandates — **485 → 553 deputies** — with `mandats` mapped through a `mandate_status` lookup that raises on an unknown label.

`anciens-membres` (56 records) is deliberately *not* crawled: it is former **Bureau** leadership rather than membership, has no per-person dates, and the pre-2006 bodies (Conseil législatif, HCR-PT) were constitutionally distinct institutions — naming a position from the legislature term would assert something the source does not say. Recorded in the YAML as a possible follow-up.

Two calls worth a second opinion:

1. `no_end_implies_current=False` is set literally rather than as `term.period_end is None`, because this source retains departed deputies (marked `Terminé`) instead of removing them.
2. 17 `deputes` records carry no `legislature` term at all and remain excluded, as before — emitting them would mean inventing a legislature assignment.

herrhellana:
should be ready

nvmbrasserie:
Really solid work on this one — the structure was already in good shape, so most of what I've pushed on top is cosmetic, plus a bit of unused logic taken out.

The one substantive change: it no longer walks the legislature taxonomy. The taxonomy lists every term back to 1960, but the source only publishes members against the sitting one. Not much point parsing historical terms when there's no historical data behind them — if they ever backfill, we can put the walk back then.

Two smaller things: deputy records now go through `audit_data`, so a new field or taxonomy warns instead of disappearing quietly, and paging moved from `page` to `offset` — the API answers a page past the last one with a 400, which any total that's an exact multiple of 100 would have run into.

Output is unchanged either way: 553 deputies, 7,058 statements, empty `issues.log`.

nvmbrasserie (approved):

PR #5079, opened 2026-07-23
Daily issues 23 07

nvmbrasserie:


datasets/au/dfat_sanctions/au_dfat_sanctions.yml
```diff
@@ -284,6 +284,16 @@ lookups:
...
+      - match: "approximately: between 1959 and 1965"
+        values:
+          - 1959
+          - 1960
+          - 1961
+          - 1962
+          - 1963
+          - 1964
+          - 1965
+
```
jbothma, line 296: no need to change, this, but for next time, I've heard we tend to just do the start and end for ranges like this.

datasets/us/cms_npi/crawler.py
```diff
@@ -132,11 +134,6 @@ def crawl_othernames(context: Context, fh: TextIO) -> None:
         if name is None:
             continue
         entity.add("alias", name)
-        # `alias` is the only property on these othername records. `clean_name`
-        # lets some non-names through (e.g. "NONE") that the `type.name` lookup
-        # then drops, leaving an empty entity. Skip emitting those.
-        if not entity.has_statements:
-            continue
```
jbothma, line 139: out of interest, why are we dropping this?

jbothma (approved):

datasets/us/cms_npi/crawler.py
nvmbrasserie: We're dropping it because it was a workaround we no longer need.

That check was there to skip "empty" records — entries where the name turned out to be garbage like "NONE". The problem was our name-cleaning function wasn't catching those, so the junk slipped through, and got thrown away later. The `has_statements` check was us catching those hollow records at the end and quietly skipping them.

In this PR we fixed the actual name-cleaning function so it now recognizes `"NONE"` and similar junk up front. Once that's fixed, those garbage names never make it through in the first place, so the empty records can't happen anymore

Issue #4605, opened 2026-06-17
[ext_eu_esma_firds ] isin-... property issuer references missing id NK-...


jbothma:
This warning is now showing as a consequence of fixing the fact that [non-risk-tagged entities are no longer included by the enricher used by ext_eu_esma_firds](https://github.com/opensanctions/operations/issues/2583).

The `sanction.linked` topic has to date not propaged from sanctioned Security entities to their issuer. [This is planned to be fixed soon.](https://github.com/opensanctions/opensanctions/issues/4538)

In the meantime, it looks like at least the sample of issuers I've checked are included in `default` and `sanctions` via `eu_esma_saris`. This warning applies to the ext_eu_esma_firds dataset in isolation.

jbothma:
ok, currently this is only showing up for `isin-US3682872078 property issuer references missing id lei-213800R54EFFINMY1P02`

isin-US3682872078 is included in eu_esma_saris with reg.warn.

isin-US3682872078 gets sanction.linked because NK-N7dfJjNtqB9hibmDaoviPf is sanctioned.

There are two funky things going on here:

1. a security references its issuer directly. sanction.linked doesn't propagate to the issuer so we can end up with an issuer that isn't included via enrichment if the security is included and tagged sanction.linked. We could propagate sanction.linked to a security's issuer from a sanction.linked security.
2. in eu_esma_firds, isin-US3682872078 references lei-213800R54EFFINMY1P02 as its issuer, but it sounds like the same field in the firds xml might also be its trading venue, which might not deserve the sanction.linked, and maybe that field doesn't necessarily mean issuer.

(1) isn't a case that would come up normally with our current set of datasets so I don't think we need to introduce an additional rule in ann_graph_topics for sanction issuers at this time. The dangling reference warning will pop up again if we forget to do so when we add such a dataset.

(2) might be fixable by looking at additional data:

> In the FULINS XML, no. The Issr field is a single, unqualified LEI. RTS 23 field 5 is literally defined as "Issuer or trading venue operator identifier", and no flag in the record tells you which meaning applies. Formally the venue-operator meaning is reserved for venue-created instruments without an issuer (derivatives, etc.), so anything with a real issuer should carry the issuer's LEI — but the Gazprom ADR disproves that in practice: its CFI is EDSXFR (a depositary receipt, unambiguously an issued instrument), yet all 56 records carry the venue operator's LEI. So the CFI code alone can't tell you whether to trust the LEI.
>
[...]

PR #5080, opened 2026-07-23
[zavod] Enricher: drop dangling references to unpublishable entities

jbothma:
Rube Goldberg machine

*Visual representation of this fix*

Based on #5027. Addresses the dangling-reference warnings tracked in #4605 (e.g. `isin-US3682872078 property issuer references missing id lei-213800R54EFFINMY1P02` in `ext_eu_esma_firds`).

## Problem

In topic-gated enrichers, a promoted (internal) entity can reference an expansion neighbour that doesn't pass the topic gate — e.g. a security whose FIRDS-reported issuer carries no risk topic. The neighbour is emitted external-only, so the published dataset contains a reference to an entity that isn't there.

## Change

1. **Prune the reference from the published entity.** Before an internal emit, entity-type property values pointing at unpublishable targets are removed from non-edge entities, with an info log. Edges are untouched: `should_promote` already only publishes them when all endpoints are publishable.

2. **Keep the relationship visible externally.** Dropping the reference entirely would blind `ann_graph_topics`, which reads the external view to discover adjacency and emit topic patches (the convergence loop from #4875) — the referenced entity could then never become publishable. So the pruned references are re-emitted in an external stub entity (same id/schema, only the pruned props): the analyzer sees the relationship, the exporter and validators (internal view) don't. Once the analyzer tags the target and it passes the gate on a later run, the reference is published for real.

Tested at both levels: a unit test for the prune helper, and an end-to-end topic-gated run asserting the internal view has no `issuer` value while the external view retains it, alongside the external-only issuer.

### DIFF
--- a/zavod/zavod/runner/util.py
+++ b/zavod/zavod/runner/util.py
@@ -5,6 +5,7 @@ from followthemoney import registry
 from followthemoney.schema import Schema
 
 from zavod.constants import ANALYZER_DATASETS
+from zavod.context import Context
 from zavod.entity import Entity
 from zavod.store import View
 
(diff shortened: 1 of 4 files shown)


jbothma:
Screenshot 2026-07-23 at 16 15 22

zavod/zavod/runner/util.py
```diff
@@ -80,13 +86,63 @@ def check_publishability(
...
+    analyzer (which reads the external view and may tag the referenced entity,
+    making it publishable on a later run) without the exporter seeing it."""
+    pruned: list[tuple[Property, str]] = []
+    if entity.schema.edge:
+        return pruned
+    for prop in list(entity.iterprops()):
+        if prop.type != registry.entity:
+            continue
+        for other_id in entity.get(prop):
+            if not publishable.get(other_id, False):
```
pudo: Why is this a dict rather than just a set? More curious than anything :)

pudo (commented):
yeah it's getting complicated. But it's consistent, so let's do it.

zavod/zavod/runner/util.py
jbothma: Can't see any reason why - I've simplified now - thanks for spotting!

Issue #5104, opened 2026-07-24
Wikidata RCA family members no longer published: topic gate can't see ann_graph_topics role.rca tags


pudo:
## Symptom

PEP family members imported from Wikidata are no longer published. Example: [Emmanuel Macron (Q3052772)](https://www.opensanctions.org/entities/Q3052772/) shows no family at all — no parents, no spouse, no sibling — even though Wikidata has all four claims (P22 → Q28838206, P25 → Q28838264, P26 → Q21511883, P3373 → Q29860422) and the enricher follows them (`PROPS_FAMILY` maps P22/P25 to "parent").

The relatives still exist in `default`, but only as nameless shells: Q28838206's canonical entity has exactly **one** internal statement — the `role.rca` topic patch from `ann_graph_topics` (`first_seen 2026-07-02`). All Wikidata statements about him (name, Family edge, DoB) are external, so the site renders a bare "Legal entity" profile with a `role.rca` chip and nothing else, and Macron's profile has no `familyPerson` edges.

## Root cause

The topic-gating feedback loop between the wikidata enricher and `ann_graph_topics` is broken at the dataset-inputs level:

1. Since 25b7b1449 ("Gate enrichment graph context on risk topics", 2026-06-30), `zavod.runner.enrich.save_match` only publishes expanded graph context when `should_promote` finds a risk topic on the entity **in the subject view** — the store built from the enrich dataset's `inputs`.
2. `role.rca` is assigned by `ann_graph_topics`, which is **not** in `wikidata.yml`'s `inputs` (peps, sanctions, wanted, enforcement, wd_*, …). So the gate looks up e.g. Q28838206 in a scope where no dataset knows him, finds no topics, and the Person + Family edge are emitted `external=True` → dropped from `default`.
3. The analyzer side keeps working: it reads `default` with `external=True`, sees the external Family edges next to the PEP, and keeps emitting `role.rca` patches — into a dataset the wikidata run never reads. The promotion loop the analyzer docstring promises ("affected entities become … eligible for further enrichment") never closes.

Timing fits: gating deployed June 30; the first wikidata run after that flipped the family entities to external; by July 2 the analyzer patch is the only internal statement left.

[...]

PR #5113, opened 2026-07-24
Fix Entity.external to ignore the synthetic BASE_ID statement

pudo:
`Entity.statements` appends a synthetic BASE_ID checksum statement that is always `external=False`, so `Entity.external` returned `False` for every assembled entity with an id — this was the root cause behind the analyzer publishing nameless topic-only stubs (#5104). Compute the property over the entity's property statements (`_iter_stmt()`) instead.

This is the `zavod/zavod/entity.py` fix and test from #5035, found by @jbothma, extracted onto current `main`. The rest of that PR has since been superseded:

- the `wikidata.yml` `ann_graph_topics` input landed in #5105,
- the analyzer no longer consumes `Entity.external`: #5109 replaced the call site with `has_published_substance`, which needs deliberately different semantics (it also discounts internal statements from `ANALYZER_DATASETS`, so a prior internal patch can't keep an otherwise-unpublished entity internal forever — see `test_patch_external_despite_prior_own_patch`),
- the patch-externality test scenarios are covered by the tests added in #5109.

That leaves `Entity.external` with no production callers, but it is public zavod API and shouldn't stay silently broken for the next caller.

Closes #5035.


jbothma:
I don't think anything else is using Entity.external. I think this can be closed n favour of https://github.com/opensanctions/opensanctions/pull/5114

pudo:
Awesome, closing.

PR #5120, opened 2026-07-24
[us_nk_jointventures] Fix mypy --strict type errors

leonhandreke:
Narrow `context.dataset.url` (`str | None`) before `fetch_html`, and swap raw `.xpath()` for the typed `h.xpath_element` / `h.xpath_strings` helpers.


leonhandreke:
@claude just assert directly, no need for an intermediate var

leonhandreke:
@claude And use xpath_string to save the second assert

PR #5125, opened 2026-07-24
Fixes 24 07 26

nvmbrasserie:


datasets/dk/pep/crawler.py
```diff
@@ -137,14 +137,18 @@ def crawl_old_pep_item(
...
         context,
         entity,
         position,
-        True,
-        end_date=removal_date.strip(),
+        no_end_implies_current=False,
+        end_date=removal_date.strip() if removal_date is not None else None,
+        status=OccupancyStatus.ENDED if removal_date is None else None,
         categorisation=categorise(context, position, default_is_pep=True),
     )
```
jbothma, line 153: This dataset is PEPs by definition, so I feel opting out of endDate-based logic when endDate is missing is fair here.

I agree with this fix.

jbothma (approved):

Issue #4945, opened 2026-07-17
helpers: apply_number's _float_str zeroes small values (%.2f) and formats the same number two ways


pudo:
Found in a systematic audit of the zavod helper library (2026-07-17). Verified by repro against `zavod/zavod/helpers/numbers.py:13-17,58`.

`_float_str` formats non-integral floats with a hardcoded `%.2f`, and the `Decimal` branch always emits 2 decimals:

- Any nonzero value < 0.005 is stored as `"0.00"` — a nonzero amount published as zero. `apply_number(wallet, "amountUsd", 0.004)` → `"0.00"`; via the string-parse path `"0.005"` → `"0.01"`.
- The same number gets two representations depending on input type: `Decimal("5")` → `"5.00"` but `float 5.0` → `"5"` — inconsistent values for identical quantities, which matters for statement dedup.

The rounding also silently discards precision on legitimate high-precision values (crypto amounts, coordinates-adjacent measurements) with no warning and no way to opt out.

## Blast radius

The 5 crawlers using `apply_number` (abuja/tokyo/black_sea MOU tonnage, `_global/interpol` height/weight, `_global/ransomwhere` balances). `ransomwhere`'s `balanceUSD` float amounts are the realistic hit — sub-cent balances become `"0.00"`.

## Fix direction

Preserve significant digits instead of fixed 2-dp (e.g. `repr`-style shortest round-trip formatting, or a per-call precision parameter); make float and Decimal paths produce identical strings for equal values. Related: the upstream parse defects feeding this formatter are tracked in opensanctions/followthemoney#331.

Issue #4941, opened 2026-07-17
helpers: make_security mints global isin-* entity IDs from unvalidated input


pudo:
Found in a systematic audit of the zavod helper library (2026-07-17).

`make_security` (`zavod/zavod/helpers/securities.py:8-17`) only calls `.upper()` on the incoming string — no strip, no `rigour.ids.ISIN.is_valid`/`normalize`, even though ISIN is a checksummed identifier and rigour has the validator. The ID is minted directly from the raw string into a **global** namespace shared across datasets (`Security` entities merge on it):

- `make_security(context, "N/A")` → id `isin-N/A` — collides globally with every other dataset's "N/A" row.
- `make_security(context, "US0378331005 ")` (trailing space) → id `isin-US0378331005 `, which never merges with the clean form.
- For an invalid ISIN, FtM's `format: isin` on `Security:isin` silently rejects the value at `entity.add` — so the emitted Security has a garbage global ID and **no `isin` property at all**.
- `cc = isin[:2]` infers a country from the first two characters of whatever arrived, stamping phantom countries on mangled input.

## Blast radius

3 callers, all passing raw source strings without validation: `datasets/ru/nsd_isin/crawler.py:59` (scraped page values), `datasets/eu/esma_saris/crawler.py:68` (CSV field), `datasets/_global/gleif/crawler.py:241`.

## Fix direction

Gate on `rigour.ids.ISIN.normalize` (strip + checksum) and return `Optional[Entity]` or raise, mirroring the `make_address` contract; only infer country from a validated ISIN. **Note: rejecting/normalizing changes some existing entity IDs — needs coordination.**

tamimbinhakim:
Fix in #5546. On the coordination note: I checked the published `eu_esma_saris` (2,931 securities) and `ru_nsd_isin` (17,987) data and neither contains an invalid or non-normalized ISIN id, so normalizing changes no existing entity IDs there. mypy also turned up a fourth caller, `zavod/shed/firds.py`, which already pre-filters with `ISIN.is_valid`; it just needed the `None` guard.

PR #5126, opened 2026-07-24
Keep the decimal part of fractional Excel cells

### LINKED ISSUE #4944: helpers: convert_excel_cell truncates fractional numeric cells with int()

`convert_excel_cell` (`zavod/zavod/helpers/excel.py:31-32`) converts every xls numeric cell with:

```python
if cell.ctype == XL_CELL_NUMBER:
    return str(int(cell.value))
```

The intent is to strip xlrd's `.0` float artifacts from integer IDs, but `int()` truncates toward zero, so any genuinely fractional cell is silently corrupted: `12.75` → `"12"`, `3.7` → `"3"`, `0.75` → `"0"`. Its sibling `parse_xls_sheet` (`excel.py:98`) handles the same cells correctly via `stringify` — the two paths disagree about the same data.

## Blast radius

2 crawlers, both sanctions lists, and both apply the helper to **every** cell of the sheet: `datasets/ae/local_terrorists/crawler.py:152` and `datasets/jp/mof_sanctions/crawler.py:347`. Corruption manifests as soon as those sheets contain any fractional numeric cell (amounts, measurements).

## Fix direction

Only strip the decimal part when the value is integral (`cell.value.is_integer()`), otherwise stringify the full value — matching `parse_xls_sheet`'s behavior.

DMZ22:
Closes #4944.

`convert_excel_cell` converts every numeric cell with `str(int(cell.value))`. The `int()` is there for a good reason — xlrd hands back all numbers as floats, so an integer id arrives as `1234.0` — but it truncates toward zero, so any genuinely fractional cell is silently corrupted.

Driving the current numeric branch with real `xlrd` cells:

| cell value | current | expected |
|---|---|---|
| `1.0` | `"1"` | `"1"` |
| `1234.0` | `"1234"` | `"1234"` |
| `0.0` | `"0"` | `"0"` |
| `12.75` | **`"12"`** | `"12.75"` |
| `3.7` | **`"3"`** | `"3.7"` |
| `0.75` | **`"0"`** | `"0.75"` |
| `-2.5` | **`"-2"`** | `"-2.5"` |
| `0.004` | **`"0"`** | `"0.004"` |
| `99.999` | **`"99"`** | `"99.999"` |

Six of nine wrong, and `0.75 → "0"` is the shape that matters: a nonzero amount published as zero.

The fix strips the artifact only when `cell.value.is_integer()`, and stringifies otherwise — which is exactly what the sibling `parse_xls_sheet` already does with the same cells, so the two paths now agree instead of disagreeing about identical data. Integer ids are unaffected.

Both current callers apply this to every cell of the sheet (`datasets/ae/local_terrorists/crawler.py`, `datasets/jp/mof_sanctions/crawler.py`), so any fractional numeric in those sheets is affected today.

### On verification

I should be straight about what I could and could not run. **zavod does not install on Windows** — `pyicu` has no wheel and `plyvel` needs a C toolchain — so I could not execute your test suite locally.

What I did instead: mirrored the numeric branch exactly and drove it with real `xlrd.sheet.Cell` objects to produce the table above, then confirmed each of the six assertions in the new test passes against the patched logic and fails against the current one.

The test I added is a real repo test and runs in your CI. It constructs cells directly rather than touching the `book.xls` fixture, so nothing binary changes:

```python
assert convert_excel_cell(book, Cell(XL_CELL_NUMBER, 1.0)) == "1"
assert convert_excel_cell(book, Cell(XL_CELL_NUMBER, 12.75)) == "12.75"
```

Worth flagging: I deliberately left the two sibling issues alone. Both #4941 and #4945 note that a fix changes already-published entity ids or values, so they need a call from you rather than a drive-by PR. This one has no such consequence — it only stops data being destroyed.

### DIFF
--- a/zavod/zavod/helpers/excel.py
+++ b/zavod/zavod/helpers/excel.py
@@ -29,7 +29,15 @@ def convert_excel_cell(book: Book, cell: Cell) -> str | None:
     """
     # https://xlrd.readthedocs.io/en/latest/api.html#xlrd.sheet.Cell
     if cell.ctype == XL_CELL_NUMBER:
-        return str(int(cell.value))
+        # xlrd reports every number as a float, so integer identifiers arrive
+        # as e.g. 1234.0. Drop that artifact only when the value really is
+        # integral: int() truncates toward zero, which would silently turn a
+        # genuine 12.75 into "12". This matches parse_xls_sheet, which
+        # stringifies the same cells.
+        assert isinstance(cell.value, float)
+        if cell.value.is_integer():
+            return str(int(cell.value))
+        return stringify(cell.value)
     elif cell.ctype in (XL_CELL_EMPTY, XL_CELL_ERROR, XL_CELL_BLANK):
         return None
     if cell.ctype == XL_CELL_DATE:
--- a/zavod/zavod/tests/helpers/test_excel.py
+++ b/zavod/zavod/tests/helpers/test_excel.py
@@ -1,4 +1,6 @@
 import xlrd  # type: ignore
+from xlrd import XL_CELL_NUMBER  # type: ignore
+from xlrd.sheet import Cell  # type: ignore
 from openpyxl import load_workbook
 from zavod.context import Context
 from zavod.helpers.excel import (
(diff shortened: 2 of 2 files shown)


leonhandreke:
Hi, thanks for your interest in zavod an looking into this! I looked into this a bit myself and went with a more simple one-liner because this is quite an edge case in our codebase. I merged it as part of https://github.com/opensanctions/opensanctions/pull/5154. I hope you understand.

PR #5137, opened 2026-07-27
[us_bis_mieu] Add BIS military-intelligence end users (EAR § 744.22)

pudo:
## What

New dataset `us_bis_mieu` covering the 13 military intelligence organizations named in [15 CFR 744.22(f)(2)](https://www.ecfr.gov/current/title-15/subtitle-B/chapter-VII/subchapter-C/part-744/section-744.22) — BIS's military-intelligence end-user (MIEU) restriction. Exports to these end users require a BIS license reviewed with a presumption of denial, and US persons are separately restricted from supporting them (§ 744.6(b)(5)).

## Why

These entities are named in the regulation text itself, not on any BIS-published list (Entity List, MEU List, DPL, UVL), so they never flow into the Trade Consolidated Screening List — verified against the CSL feed, which has 12 source lists and no MIEU section. Only 4 of the 13 (GRU, RGB, IRGC-IO, DGCIM) are covered by other US federal lists, via independent OFAC SDN designations; 7 were absent from the OpenSanctions corpus entirely, including China's Intelligence Bureau of the Joint Staff Department, both Cuban military intelligence services, and Belarus's military intelligence directorate.

## Design

Follows the `cn_sanctions` pattern:

- `mieu.csv` (git-controlled, force-added past the global `*.csv` ignore like `datasets/cn/sanctions/sanctions.csv`) is the source of truth: names verbatim from the regulation, acronyms as aliases, `Date` = effective date of the Federal Register rule that added each entity, that rule's URL as `Source URL`.
- The crawler doubles as a change monitor: it polls the eCFR versioner API (not bot-blocked, unlike the eCFR website) and warns once per amendment date to § 744.22 that is missing from `config.discovery.reviewed_versions`, with a runbook comment in the yml telling the issues agent how to resolve the warning by updating the CSV. It also warns if the section is removed or renamed — which would catch the pending July 2024 proposed rule (89 FR 60985) redesignating it as § 744.24.
- `manual_check` (90d) covers what the monitor can't see: finalization of that proposed rule and informal BIS guidance.
- New program `meta/programs/US-BIS-MIEU.yml` (issuer `us_bis`, measure: Export control).

Note: for the named countries the (f)(2) list is explicitly non-exhaustive; for all other countries it is exhaustive — stated in the dataset description.

## Verification

- Crawl emits 13 Organizations + 13 Sanctions, empty `issues.log`; `programId` resolves to `US-BIS-MIEU` on every sanction.
- Negative test: removing `2022-09-16` from `reviewed_versions` produces exactly one "Unreviewed amendment" warning with the section and full-text API URLs.
- `mypy --strict` clean; FR citations and per-rule effective dates verified against the Federal Register API; the amendment-comment claims verified by diffing eCFR section versions.

Not included: release into a collection (separate step).


nvmbrasserie (approved):
Looks good to me! I don't have any suggestion for this one

PR #5143, opened 2026-07-27
[gb_nca_most_wanted] Fix mypy --strict type errors

leonhandreke:
All 10 errors were `None`-attribute access on lxml `.find()` results (`_Element | None`) and `.text` (`str | None`). Added `assert ... is not None` narrowing where the code already assumed the element/text was present (it would crash on `None` either way), rather than swapping to `h.element_text` which would change direct-text to full text-content semantics.


leonhandreke:
@claude drop the superflous `, variable` after each assert. Also, can we get away with only one `assert value.text is not None`?

PR #5148, opened 2026-07-27
[lt_pep_declarations] Fix mypy --strict type errors

leonhandreke:
Type the JSON objects as `dict[str, Any]`, correct `parse_affiliations`'s return type to `list[tuple[Entity, Entity]]` (fixes the unpack in the caller too), and add `assert ... is not None` narrowing on the occupancy/spouse entities. `make_id` gets a `# type: ignore[arg-type]` (its arg is hashed into published IDs, so leaving it untouched avoids re-keying).


leonhandreke:
@claude in the make_id, can't you str() to make the signature pass but keep the ids intact?

leonhandreke:
Will be gone soon

PR #5155, opened 2026-07-27
[us_fincen_special_measures] Skip non-representable class-of-transactions rows

nvmbrasserie:
Stacked on #5142. Follow-up addressing a review point: two overview-table rows are *classes of transactions*, not real-world entities, and were being emitted as bogus `Company` entities.

## What

- **Convertible Virtual Currency Mixing** and **Mexican Gambling Establishments** are Section 311 *class-of-transactions* measures — there is no OpenSanctions schema for the class itself, only for the parties it names. They are now listed in a `skip_entity` lookup: the main entity and its sanction are not emitted, only the members mined into `details.csv` (the ten casinos, the eight mixers).
- Detail-row entity IDs are now coded on the raw measure name (`make_id(measure, name)`) rather than the main entity's already-hashed ID.
- A `subsidiary`/`owner`/`related` party under a skipped measure has no main entity to link to, so it is emitted on its own (this affects one row — the ChipMixer operator, a non-designated related person).

## Not included

Country/jurisdiction rows (Burma, DPRK, Iran, Nauru, Ukraine) remain `PublicBody`. Skipping them is possible with the same mechanism, but their findings anchor ~82 `related` mentions and 2 `self` facts to the country entity, so how to represent jurisdiction-level 311 findings is a separate, larger decision left for follow-up.

## Verification

- `zavod crawl` completes with zero warnings; empty `issues.log`. Entity count 926 → 921 (2 bogus umbrella entities + their 2 sanctions removed; the operator's link to the removed umbrella no longer emitted).
- Both class umbrellas gone; all 10 casinos and 8 mixers retained.
- `ruff` clean; `mypy --strict` clean.

### DIFF
--- a/datasets/us/fincen_special_measures/README.md
+++ b/datasets/us/fincen_special_measures/README.md
@@ -27,6 +27,13 @@ that warning is in the `config.discovery` section of the dataset `.yml`.
     from the main entity.
   - `related` — creates the entity plus an UnknownLink; for parties the
     documents connect to the target without designating them.
+
+  When a table row names a *class of transactions* rather than an entity we can
+  represent (e.g. "Convertible Virtual Currency Mixing", "Mexican Gambling
+  Establishments"), it is listed in the `skip_entity` lookup in the dataset
+  `.yml`. The class itself is not emitted; only its `target` members are. A
+  `subsidiary`/`owner`/`related` row under such a measure has no main entity to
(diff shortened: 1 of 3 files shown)


pudo (approved):
Let's do this for now and figure out cleaner modeling as we go.

PR #5165, opened 2026-07-28
Clean up position names and add missing QIDs across PEP crawlers

leonhandreke:
Went through all `make_position` calls that pass a `wikidata_id` (116 of them) plus the ones that arguably should, and fixed three things.

## Tag the position name as English

56 files.

- 61 calls had an English title but no `lang` at all, so the name got emitted untagged — added `lang="eng"`.
- `de/bundesrat`: "Mitglied des Bundesrates" (`lang="deu"`) → "Member of the Bundesrat of Germany", the English label on https://www.wikidata.org/wiki/Q15835370.
- `ie/parliament`: "Teachta Dála" → "Member of the Dáil of Ireland".
- Special cases:
  - Checked no call passes a non-English `summary`/`description`/`subnational_area`, since `make_position` applies the same `lang` to those. Only `nl/house_of_representatives` and the `au/*` + `pg` subnational areas have any, all English.
  - `es/parliament` threads `lang=lang` through a helper — both QID-passing callers already pass `"eng"`, the `"spa"` one passes `wikidata_id=None`. Left alone.

## Add the QID where one exists

15 positions that were `make_id`-keyed even though Wikidata has an item for them. Keying on the QID lets them line up with the same position from other sources.

- `bg/parliament`, `ee/riigikogu`, `fr/assemblee`, `fr/senat`, `hk/legco` (member + president), `kg/jogorku_kenesh`, `lv/saeima`, `mn/parliament`, `mx/deputies`, `mx/senators`, `tm/mejlis`, `tr/parliament`, `uz/legislative_chamber`.
- `cn/peoples_congress_wikipedia` used a typographic apostrophe and no QID, so it emitted a *second* Position for the job `cn/npc` already covers via https://www.wikidata.org/wiki/Q10891456. Both now emit the same entity.
- Verified each item is `instance of: position`, subclasses the right legislator type, and its jurisdiction matches the dataset's country.
- Special cases:
  - Every one of these re-keys its Position from a hash to the QID, so expect churn in the next run's diff.
  - Four had no country in the name ("deputy of Saeima", Riigikogu, State Great Khural, Grand National Assembly) — renamed in the same commit, since the id changes either way.
  - `ci/senate` looks like it belongs here but no Wikidata position item seems to exist for the Ivorian Senate.

## Name the country in the position name

25 names, 19 datasets.

Real collisions, where the bare name was ambiguous:

- "Member of the Senate" was emitted identically by `it/senate`, `nl/senate` and `pl/senate` — now "…of the Italian Republic" / "…of the Netherlands" / "…of the Republic of Poland".
- `nl/house_of_representatives` "Member of the House of Representatives", while 7 other datasets already said "…of {Cyprus, Japan, Malta, …}".
- `gb/commons` "Member of the House of Commons" vs `ca/commons` "…of Canada".
- Also `it/deputies` and `bh/nuwab` (Iraq's chamber shares the English name).

Bare but not yet ambiguous, normalised for consistency:

- `il` ("Knesset Member" → "Member of the Knesset of Israel" — also had the word order backwards), `is`, `tw`, `in` (Lok Sabha + Rajya Sabha), `lt`, `pl/sejm`, `ru`, `my` (2 in the crawler + 4 lookup entries in the yml), `cu`, `cn`.

Deliberately excluded:

- `eu/meps` and `se/riksdag` MEP — supranational, no country belongs there.
- Subnational positions disambiguated by region: `au/*` (8 datasets), `be/flemish_parliament`, `be/walloon_parliament`.
- Names already carrying a nationality adjective: `hr` ("Croatian"), `gr` ("Hellenic"), `de/bundestag` ("German"), `ch` ("Swiss"), `ar` ("Argentine"), `us/congress` ("United States senator").
- `_wikidata/categories` has its own "Member of the Sejm" spec, but those positions are `make_id`-keyed by design, so renaming would re-key them. Out of scope.

---

Nothing enforces any of this, so it'll drift again. Possible followup: have `make_position` assert that `wikidata_id is not None` implies `lang == "eng"`.

### DIFF
--- a/datasets/ad/consell_general/crawler.py
+++ b/datasets/ad/consell_general/crawler.py
@@ -73,6 +73,7 @@ def crawl(context: Context) -> None:
         country="ad",
         topics=["gov.national", "gov.legislative"],
         wikidata_id="Q20177831",
+        lang="eng",
     )
     context.emit(position)
     categorisation = categorise(context, position, default_is_pep=True)
--- a/datasets/al/kuvendi/crawler.py
+++ b/datasets/al/kuvendi/crawler.py
@@ -89,6 +89,7 @@ def crawl(context: Context) -> None:
         country="al",
         topics=["gov.national", "gov.legislative"],
         wikidata_id="Q20177772",
+        lang="eng",
     )
     context.emit(position)
 
--- a/datasets/au/act_parliament/crawler.py
+++ b/datasets/au/act_parliament/crawler.py
@@ -90,6 +90,7 @@ def crawl(context: Context) -> None:
         country="au",
         subnational_area="Australian Capital Territory",
         wikidata_id="Q6814365",
+        lang="eng",
     )
     categorisation = categorise(context, position, default_is_pep=True)
     context.emit(position)
--- a/datasets/au/nsw_parliament/crawler.py
+++ b/datasets/au/nsw_parliament/crawler.py
@@ -161,6 +161,7 @@ def crawl(context: Context) -> None:
             country="au",
             wikidata_id=config["wikidata_id"],
             topics=["gov.state", "gov.legislative"],
+            lang="eng",
         )
         categorisation = categorise(context, position)
         context.emit(position)
--- a/datasets/au/nt_parliament/crawler.py
+++ b/datasets/au/nt_parliament/crawler.py
@@ -79,6 +79,7 @@ def crawl(context: Context) -> None:
         country="au",
         subnational_area="Northern Territory",
         wikidata_id="Q26998278",
+        lang="eng",
     )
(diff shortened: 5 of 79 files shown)


leonhandreke:
@herrhellana @pudo do you agree with putting the country name in the position title? This was inconsistent before (i.e. some where "Member of the Saeima" and some "Member of the Saeima of Latvia"). I feel putting the country name in the caption is just pragmatic, even if it's strictly duplicated information. WDYT?

herrhellana:
this is great, thank you! i'd say we should keep the country names in position names as the positions' genealogical lineage varies across countries?

PR #5177, opened 2026-07-29
Guard tabular parsers against silent column misattribution

### LINKED ISSUE #4933: helpers: tabular parsers silently clobber duplicate-header columns; parse_html_table iterates nested tables; parse_pdf_table repeats page headers as data
Found in a systematic audit of the zavod helper library (2026-07-17). Three related defects in the tabular parsing helpers — all silent-misattribution class, which these parsers otherwise guard against loudly (`assert len(headers) == len(cells)`).

## 1. Duplicate headers silently drop a column (all four parsers)

Headers are slugified and rows built as `dict(zip(headers, cells))` — when two headers collide after slugification (literal duplicates, case variants, punctuation variants like "Date-of-Birth" vs "Date of Birth"), the later cell silently overwrites the earlier. Verified: headers `Name | Name | DOB` with row `original script | latin script | 1970` → `{'name': 'latin script', 'dob': '1970'}` — a whole column gone, invisible even to the `row.pop()` + `context.audit_data()` pattern.

Affected: `helpers/html.py:88-96,110` (`parse_html_table`, 44 caller files), `helpers/pdf.py:120` (`parse_pdf_table` — two *empty* header cells both slugify to `""` and clobber), `helpers/excel.py:123-124,176-180` (both sheet parsers; `parse_xlsx_sheet` has ~59 references). A source *adding* a duplicate column is exactly the schema drift crawlers should fail loudly on.

Fix: `assert len(set(headers)) == len(headers)` after slugification, in all four.

## 2. `parse_html_table` treats nested tables' rows as outer-table rows

`table.findall(".//tr")` (`helpers/html.py:83`) is a descendant search. Verified: an outer table `Name | Info` with a nested table inside the Info cell yields a spurious extra row `{'name': 'inner1', 'info': 'inner2'}` — inner cell text emitted as an entity name — while the legitimate outer row's cell is polluted (`'inner1inner2'`, no separator). Silent whenever the inner column count matches; nested layout tables are common in government CMS HTML.

Fix: iterate only rows whose nearest `<table>` ancestor is the target table.

## 3. `parse_pdf_table` emits repeated page headers as data rows by default

With default `headers_per_page=False` (`helpers/pdf.py:100-120`), headers are captured once on the first page; on later pages every row — including a repeated header row — is yielded as data. The helper's own test asserts the junk (`test_pdf.py:82-83`: `rows[3]["forenames"] == "Forenames"`). 5 of 12 callers don't pass `headers_per_page` (us/wy|ne|az med_exclusions, jp/meti_eul, li/posting_sanctions) — safe only while their PDFs never repeat headers.

Fix: detect row == headers and skip (or warn), independent of the flag.

DMZ22:
Closes #4933.

Three fixes, following the suggestions in the issue:

1. **Duplicate headers**: all four parsers (`parse_html_table`, `parse_pdf_table`, `parse_xls_sheet`, `parse_xlsx_sheet`) now assert that headers are unique after slugification. Rows are built with `dict(zip(headers, cells))`, so a duplicate header silently dropped the earlier column's cell. This also covers the two-empty-PDF-header-cells case, since both slugify to `""`.

2. **Nested tables**: `parse_html_table` used `table.findall(".//tr")`, a descendant search that also matched rows of tables nested inside a cell and emitted them as outer-table rows. It now only parses rows whose nearest `<table>` ancestor is the target table.

3. **Repeated page headers**: `parse_pdf_table` with the default `headers_per_page=False` emitted repeated per-page header rows as data. A row whose slugified cells equal the headers is now skipped and logged, independent of the flag.

A few notes:

- `test_parse_pdf_table_multiple_pages` previously asserted the repeated header row (`rows[3]["forenames"] == "Forenames"`); it now expects 5 rows with that row skipped. The page-2 `"First\nName"` row still comes through — it slugifies to `first_name`, not `forenames`, so it is a different header row rather than a repeat, and detecting that generically seemed out of scope here.
- `parse_xls_sheet` validates headers at the first data row rather than at header build, because `join_header_rows` can extend them across several rows.
- The xls duplicate-header test uses a small stub sheet with real `xlrd` cells, since xlrd cannot write files to produce a fixture.
- A crawler whose source legitimately serves duplicate headers will now fail loudly instead of dropping a column — as I read the issue, that is the intended behaviour, but happy to downgrade the assert to a warning if you'd rather roll this out gently.

Dev note: zavod doesn't install on Windows (pyicu/plyvel), so I couldn't run the test suite locally. I verified the changes by driving the three helper modules directly against the repo fixtures (`book.xls`, `book.xlsx`, `test_pdf.pdf`) with zavod internals stubbed: all existing expectations are unchanged before/after, and the updated PDF expectations above come from running the fixture PDF through pdfplumber. CI should be treated as authoritative.

### DIFF
--- a/zavod/zavod/helpers/excel.py
+++ b/zavod/zavod/helpers/excel.py
@@ -82,6 +82,7 @@ def parse_xls_sheet(
     Cells with links are included as keys with _url appended to the original key.
     """
     headers: list[str] | None = None
+    headers_validated = False
     for row_ix, row in enumerate(sheet):
         if row_ix < skiprows:
             continue
@@ -119,6 +120,15 @@ def parse_xls_sheet(
                         cell = f"column_{idx}"
                     headers.append(slugify_text(cell, "_") or "")
             continue
+
+        if not headers_validated:
+            # Headers are final once the first data row is reached (they may
+            # span several rows via join_header_rows). Records are built by
+            # zipping headers with cells, so a duplicate header would silently
+            # drop the earlier column's cell.
+            duplicates = {hdr for hdr in headers if headers.count(hdr) > 1}
+            assert not duplicates, f"Duplicate headers: {sorted(duplicates)}"
+            headers_validated = True
 
         for header, value in zip(headers, cells):
             record[header] = stringify(value)
@@ -170,6 +180,10 @@ def parse_xlsx_sheet(
                 if header_slug is None and header is not None:
                     header_slug = f"column_{idx}"
                 headers.append(header_slug)
+            duplicates = {hdr for hdr in headers if headers.count(hdr) > 1}
+            # Records are built by zipping headers with cells, so a duplicate
+            # header would silently drop the earlier column's cell.
+            assert not duplicates, f"Duplicate headers: {sorted(duplicates)}"
             continue
 
         record: dict[str, str | None] = {}
--- a/zavod/zavod/helpers/html.py
+++ b/zavod/zavod/helpers/html.py
@@ -81,7 +81,15 @@ def parse_html_table(
       - `zavod.helpers.links_to_dict`
     """
     headers = None
-    for rownum, row in enumerate(table.findall(".//tr")):
+    rows = [
+        row
+        for row in table.findall(".//tr")
+        # A descendant search also matches rows of tables nested inside a
+        # cell; only rows whose nearest <table> ancestor is the target table
+        # belong to it.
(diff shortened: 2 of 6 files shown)


jbothma:
Thanks for this! 

I merged in https://github.com/opensanctions/opensanctions/pull/5323 where I clarified the fixtures a bit. I also changed the HTML nested table check into a warning just so we can see if anything seriously depends on the current behaviour before dropping those rows.

PR #5181, opened 2026-07-29
Add list.pep.bulk tag and load_all_datasets() catalog scan

pudo:
Groundwork for selecting xref scopes via followthemoney dataset queries (e.g. `#list.pep - #list.pep.bulk`) instead of hand-maintained dataset name lists. Two parts:

**New tag `list.pep.bulk`** — applied to the 26 `list.pep` datasets that are not suitable for broad PEP cross-referencing: declaration registries, sub-national officeholder lists, and `un_ga_protocol`, which is matched in a dedicated scope instead. Since tag matching is exact-string (no prefix semantics), these datasets keep their `list.pep` tag alongside the sub-tag; the convention is documented in the metadata docs. Three parliament datasets (`al_kuvendi`, `bg_parliament`, `nz_parliament`) were missing the plain `list.pep` tag and get it here.

**`zavod.meta.load_directory_catalog()`** — builds a catalog from every dataset YAML under a datasets directory, defaulting to `settings.DATASETS_PATH` (`ZAVOD_DATASETS_PATH`, or the `datasets/` folder of the repository checkout containing the zavod installation). Dataset queries with `#tag` selectors need the full dataset universe — including unreleased datasets and ones outside the `default` collection — while the lazy archive-backed catalog only knows explicitly requested datasets, so tag selectors would silently under-match against it.

The function returns a **fresh catalog instance** rather than populating the `get_catalog()` singleton: the two have different metadata sources (repository specs vs published `index.json` artifacts), and mixing them in one instance makes the result depend on load order. To let callers choose which catalog to resolve against, **`get_multi_dataset()` now takes the catalog as an explicit first argument** — a breaking signature change; all existing call sites (enrichers, CLI, the two `_analysis` analyzers, matcher training) pass `get_catalog()` and keep their current behavior.

The scan loads with the new `traverse_children=False` option on `load_yaml()`: child resolution via `get()` can backfill `index.json` artifacts from the archive over the network, which a full directory scan neither needs (`add()` links parents and children in any load order) nor wants.

Verified with `kombinat check-scopes`: the query `#list.pep - #list.pep.bulk` plus the three explicit non-PEP scope members resolves to the previous 157-dataset goodpeps xref scope plus seven datasets promoted out of the bulk tag during review (164 total; an eighth promoted dataset, `lt_pep_declarations`, is newly marked disabled since the source is dead, and disabled datasets are excluded from scope resolution), and `#list.debarment & #sector.devbank` resolves to exactly the five MDB debarment lists. The consumer change (kombinat scope resolution) lands separately in the operations repo and requires this image to be published first.

### DIFF
--- a/datasets/lt/pep_declarations/lt_pep_declarations.yml
+++ b/datasets/lt/pep_declarations/lt_pep_declarations.yml
@@ -1,6 +1,7 @@
 title: Lithuania PEPs from the register of private interests
 entry_point: crawler.py
 prefix: lt-pinreg
+disabled: true
 coverage:
   frequency: monthly
   start: "2024-08-16"
--- a/datasets/lt/pep_declarations/lt_pep_declarations.yml
+++ b/datasets/lt/pep_declarations/lt_pep_declarations.yml
@@ -19,7 +19,7 @@ description: |
 deprecation: |
   As of April 2025, the Lithuanian Chief Official Ethics Commission (VTEK) has required
   API authentication for access to the declarations search page, which prevents
-  automated crawling of the data. We've [reached out to VTEK to request access](https://github.com/opensanctions/opensanctions/issues/2212), but 
+  automated crawling of the data. We've [reached out to VTEK to request access](https://github.com/opensanctions/opensanctions/issues/2212), but
   not received a response yet. The dataset is therefore marked as deprecated and eventually
   removed in favor of other sources of PEP data for Lithuania.
 publisher:
--- a/zavod/zavod/meta/__init__.py
+++ b/zavod/zavod/meta/__init__.py
@@ -1,8 +1,13 @@
 from pathlib import Path
 from banal import hash_data
 from functools import cache
+from followthemoney.dataset import DataCatalog
 from followthemoney.exc import MetadataException
+from pydantic import ValidationError
+from yaml import YAMLError
 
+from zavod import settings
+from zavod.exc import ConfigurationException
 from zavod.logs import get_logger
 from zavod.meta.dataset import Dataset
 from zavod.meta.catalog import ArchiveBackedCatalog
@@ -22,10 +27,40 @@ def load_dataset_from_path(path: Path) -> Dataset | None:
     return get_catalog().load_yaml(path)
 
 
-def get_multi_dataset(names: list[str]) -> Dataset:
+def load_directory_catalog(datasets_path: Path | None = None) -> ArchiveBackedCatalog:
+    """Build a catalog from every dataset YAML found under the datasets directory.
+
+    Use this to get the full universe of datasets — including unreleased ones and
+    those outside the `default` collection — for evaluating dataset queries: the
+    process-wide catalog is lazy and only knows datasets that were explicitly
+    requested, so `#tag` selectors would silently match against an incomplete
+    catalog. Returns a fresh catalog on every call; get_catalog() is unaffected.
+
+    Args:
+        datasets_path: Directory tree to scan; defaults to settings.DATASETS_PATH
+            (the ZAVOD_DATASETS_PATH environment variable).
+    """
+    path = datasets_path or settings.DATASETS_PATH
+    if not path.is_dir():
+        raise ConfigurationException(
+            f"Datasets path does not exist: {path} (set ZAVOD_DATASETS_PATH)"
+        )
+    catalog = ArchiveBackedCatalog()
+    for yml_path in sorted(path.glob("**/*.y*ml")):
+        try:
+            catalog.load_yaml(yml_path, traverse_children=False)
+        except (YAMLError, ValidationError, MetadataException, TypeError) as exc:
+            log.warning(
+                "Skipping invalid dataset YAML",
+                path=yml_path.as_posix(),
+                error=str(exc),
+            )
+    return catalog
+
+
+def get_multi_dataset(catalog: DataCatalog[Dataset], names: list[str]) -> Dataset:
     """The scopes of a dataset is the set of other datasets on which analysis or
     enrichment should be performed by the runner."""
-    catalog = get_catalog()
     inputs: list[Dataset] = []
     for input_name in names:
         try:
@@ -52,6 +87,5 @@ def get_multi_dataset(names: list[str]) -> Dataset:
             "summary": "Synthetic, ad-hoc virtual collection for multiple input datasets",
             "hidden": True,
         }
-        scope = catalog.make_dataset(data)
-        catalog.add(scope)
+        catalog.make_dataset(data)
     return catalog.require(name)
--- a/zavod/zavod/meta/catalog.py
+++ b/zavod/zavod/meta/catalog.py
@@ -10,7 +10,7 @@ class ArchiveBackedCatalog(DataCatalog[Dataset]):
     def __init__(self) -> None:
         super().__init__(Dataset, {})
 
-    def load_yaml(self, path: Path) -> Dataset | None:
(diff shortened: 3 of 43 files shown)


datasets/am/hetq/am_hetq_officials.yml
```diff
@@ -22,6 +22,7 @@ description: |
   The data has been archived from the data.hetq.am API on 2024-04-22.
 tags:
   - list.pep
+  - list.pep.bulk
```
pudo: am_hetq shouldn't be bulk

datasets/ca/foreign_reps/ca_foreign_reps.yml
```diff
@@ -14,6 +14,7 @@ description: |
 url: https://w05.international.gc.ca/protocol-protocole/Heads-Chefs.aspx
 tags:
   - list.pep
+  - list.pep.bulk
```
pudo: ca_foreign reps shouldn't be bulk

datasets/de/abgeordnetenwatch/de_abgeordnetenwatch.yml
```diff
@@ -18,6 +18,7 @@ description: |
   Data released under CC0 1.0 license.
 tags:
   - list.pep
+  - list.pep.bulk
```
pudo: de_aw shouldn't be bulk

datasets/dk/pep/dk_pep.yml
```diff
@@ -16,6 +16,7 @@ description: |
   PEPs associated with them.
 tags:
   - list.pep
+  - list.pep.bulk
```
pudo: this has DOB, so it should be good to match. not bulk

datasets/lt/pep_declarations/lt_pep_declarations.yml
```diff
@@ -33,6 +33,7 @@ publisher:
     and the persons working in them ethically serve the people.
 tags:
   - list.pep
+  - list.pep.bulk
```
pudo: source is dead, don't add here.

datasets/mx/governors/mx_governors.yml
```diff
@@ -34,6 +34,7 @@ publisher:
   official: false
 tags:
   - list.pep
+  - list.pep.bulk
```
pudo: mx_gov also not bulk

datasets/si/zvezoskop/si_zvezoskop.yml
```diff
@@ -35,6 +35,7 @@ description: |
   and networks.
 tags:
   - list.pep
+  - list.pep.bulk
```
pudo: not bulk

datasets/us/cia_world_leaders/us_cia_world_leaders.yml
```diff
@@ -30,6 +30,7 @@ description: |
 url: https://www.cia.gov/resources/world-leaders/foreign-governments/
 tags:
   - list.pep
+  - list.pep.bulk
```
pudo: not bulk

zavod/zavod/meta/__init__.py
```diff
@@ -22,10 +27,40 @@ def load_dataset_from_path(path: Path) -> Dataset | None:
...
+    those outside the `default` collection — for evaluating dataset queries: the
+    process-wide catalog is lazy and only knows datasets that were explicitly
+    requested, so `#tag` selectors would silently match against an incomplete
+    catalog. Returns a fresh catalog on every call; get_catalog() is unaffected.
+
+    Args:
+        datasets_path: Directory tree to scan; defaults to settings.DATASETS_PATH
+            (the ZAVOD_DATASETS_PATH environment variable).
+    """
+    path = datasets_path or settings.DATASETS_PATH
```
pudo, line 43: no datasets_path argument here, always `settings`.

zavod/zavod/meta/catalog.py
```diff
@@ -10,16 +10,20 @@ class ArchiveBackedCatalog(DataCatalog[Dataset]):
     def __init__(self) -> None:
         super().__init__(Dataset, {})
 
-    def load_yaml(self, path: Path) -> Dataset | None:
+    def load_yaml(self, path: Path, fetch_children: bool = True) -> Dataset | None:
```
pudo: instead of `fetch_children` let's call it `traverse_children`

pudo (commented):
Notes

datasets/am/hetq/am_hetq_officials.yml
pudo: Done — bulk tag removed; am_hetq_officials now resolves into goodpeps.

datasets/ca/foreign_reps/ca_foreign_reps.yml
pudo: Done — bulk tag removed; ca_foreign_reps now resolves into goodpeps.

datasets/de/abgeordnetenwatch/de_abgeordnetenwatch.yml
pudo: Done — bulk tag removed; de_abgeordnetenwatch now resolves into goodpeps.

datasets/dk/pep/dk_pep.yml
pudo: Done — bulk tag removed; dk_pep now resolves into goodpeps.

datasets/lt/pep_declarations/lt_pep_declarations.yml
pudo: Done — bulk tag removed. One consequence to confirm: lt_pep_declarations has a coverage end date but is not marked disabled, so without the bulk tag it now resolves into goodpeps via #list.pep. If a dead source shouldn't be matched there, it would need disabled: true (scope resolution drops disabled datasets).

datasets/mx/governors/mx_governors.yml
pudo: Done — bulk tag removed; mx_governors now resolves into goodpeps.

datasets/si/zvezoskop/si_zvezoskop.yml
pudo: Done — bulk tag removed; si_zvezoskop now resolves into goodpeps.

datasets/us/cia_world_leaders/us_cia_world_leaders.yml
pudo: Done — bulk tag removed; us_cia_world_leaders now resolves into goodpeps.

zavod/zavod/meta/__init__.py
pudo: Done — the parameter is gone; load_directory_catalog() always reads settings.DATASETS_PATH, and tests monkeypatch that setting.

zavod/zavod/meta/catalog.py
pudo: Done — renamed to traverse_children.

datasets/lt/pep_declarations/lt_pep_declarations.yml
pudo: Follow-up: lt_pep_declarations is now marked disabled: true (commit 08dd18f83), so it stays out of all xref scopes despite losing the bulk tag.

zavod/zavod/meta/__init__.py
pudo: Update: restored the datasets_path argument (defaulting to settings.DATASETS_PATH) — dropping it forced tests to monkeypatch the settings module, which is the uglier dependency.

PR #5182, opened 2026-07-29
[au_nsw_parliament] Rewrite for the redesigned parliament website

pudo:
## Why

The NSW parliament site was redesigned. `members/Pages/all-members.aspx` now 301s to `/members-and-electorates/members-and-ministers` and no longer contains the `prlMembers` table, so the crawler has failed on every run since 2026-07-29:

```
Runner failed: Expected 1 elements, got 0 for xpath "//table[@id='prlMembers']"
```

## What changed

The new listing is rendered client-side from a Funnelback search index, so the member list now comes from that JSON endpoint instead of scraping the page. Term start dates and biographies aren't in the index and are read from each member's profile page. The member's numeric ID is recoverable from the index URL, which keeps person entity IDs on the same `member-<pk>` slug as before.

`data.url` moves to the search endpoint and `dates.formats` becomes `%a %d %B %Y` to match how the new site renders dates.

## Verification

Crawled locally and diffed against the last good production run (`20260628235901-omc`):

| | old | new |
|---|---|---|
| Person | 135 | 135 |
| Position | 2 | 2 |
| Occupancy | 135 | 135 |
| statements | 2106 | 2118 |

- **Person entity IDs: all 135 preserved.** Position IDs: both preserved.
- **Zero issues logged** (`issues.log` is empty).
- Only property-level change: `Person:biography` 111 → 123 (+12). Every other property count is identical.

Occupancy IDs all change. `make_occupancy` hashes the raw pre-parse date string, and the redesigned site renders term starts as `Sat 25 October 2014` where the old one rendered `25 Oct 2014`. The parsed `startDate` (`2014-10-25`) and every other occupancy property — `holder`, `post`, `status`, `constituency` — are unchanged.

## Provenance

The issues agent worked this out three separate times (runs [30421338793](https://github.com/opensanctions/opensanctions/actions/runs/30421338793), [30425685500](https://github.com/opensanctions/opensanctions/actions/runs/30425685500), [30429614382](https://github.com/opensanctions/opensanctions/actions/runs/30429614382)) and hit its 60-turn limit before committing each time, so all three attempts were discarded. This code is recovered from the third run's log, then independently crawl-verified and linted locally as above — the numbers in this PR are from my own run, not the agent's.

A rewrite of this size doesn't fit the autofix job's budget, so the dataset should be kept out of that loop; opening this PR also stops the twice-daily retries, since `issues_agent` skips datasets with an open PR.

### DIFF
--- a/datasets/au/nsw_parliament/au_nsw_parliament.yml
+++ b/datasets/au/nsw_parliament/au_nsw_parliament.yml
@@ -27,12 +27,13 @@ publisher:
   country: au
   official: true
 data:
-  url: https://www.parliament.nsw.gov.au/members/Pages/all-members.aspx
-  format: HTML
+  # Search index backing https://www.parliament.nsw.gov.au/members-and-electorates/members-and-ministers/all-members
+  url: https://search.parliament.nsw.gov.au/s/search.json
+  format: JSON
   lang: eng
 
 dates:
-  formats: ["%d %b %Y"]
+  formats: ["%a %d %B %Y"]
 tags:
   - list.pep
 
--- a/datasets/au/nsw_parliament/crawler.py
+++ b/datasets/au/nsw_parliament/crawler.py
@@ -1,3 +1,6 @@
+import re
+from typing import Any
+
 from zavod import Context
 from zavod import helpers as h
 from zavod.entity import Entity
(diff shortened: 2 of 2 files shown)


leonhandreke (approved):
Wonderful, grazie mille!

PR #5183, opened 2026-07-29
exporters: don't publish delta_url when the delta exporter is disabled

### LINKED ISSUE #5140: delta_url is published even when the delta exporter is disabled, pointing to a stale manifest
## Problem

`ext_us_ofac_press_releases` publishes a `delta_url` in its index.json even though the delta exporter is not enabled for this dataset (`exports:` in `datasets/_externals/ext_us_ofac_press_releases.yml` only lists `statistics.json`). The linked delta manifest's newest entry is from 2025-10-29, while the dataset itself is actively updating (`last_change: 2026-07-24`, `last_export: 2026-07-27` as of writing). This looks to a consumer like delta/incremental updates are available and current, when in fact they stopped being produced roughly 9 months ago.

```
$ curl -s https://data.opensanctions.org/datasets/latest/ext_us_ofac_press_releases/index.json | jq '.last_change, .last_export, .delta_url'
"2026-07-24T11:11:25"
"2026-07-27T10:03:59"
"https://data.opensanctions.org/artifacts/ext_us_ofac_press_releases/20260727100359-hrb/delta.json"

$ curl -s https://data.opensanctions.org/artifacts/ext_us_ofac_press_releases/20260727100359-hrb/delta.json | jq '.versions | keys | sort | last'
"20251029160524-iqx"
```

## Root cause

In `zavod/zavod/exporters/metadata/__init__.py`, `write_dataset_index()` sets `delta_url` by checking whether a `delta.json` manifest file exists on disk / in the archive — it does not check whether the delta exporter is actually part of the current run's `exports:` config:

```python
delta_index_path = dataset_resource_path(dataset.name, DELTA_INDEX_FILE)
if delta_index_path.is_file():
    meta["delta_url"] = make_artifact_url(dataset.name, version.id, DELTA_INDEX_FILE)
else:
    for version in iter_dataset_versions(dataset.name):
        object = get_artifact_object(dataset.name, DELTA_EXPORT_FILE, version=version.id)
        if object is not None:
            meta["delta_url"] = make_artifact_url(dataset.name, version.id, DELTA_INDEX_FILE)
            break
```

`write_delta_index(dataset)` and `write_dataset_index(...)` are called unconditionally in `zavod/zavod/exporters/__init__.py` regardless of whether `DeltaExporter` is in the exporter set built from `context.dataset.model.exports`. `write_delta_index()` falls back to scanning up to 100 past versions via `iter_dataset_versions()` / `get_artifact_object()` for any historical `entities.delta.json`, and if it finds one, rewrites the dataset-level `delta.json` manifest to include it. This means once a dataset has ever produced a delta export, `delta_url` keeps getting published indefinitely — even after the delta exporter is removed from `exports:` — with the manifest silently going stale.

## Suggested fix (pinch of salt, these are vibes)

`write_dataset_index` / `write_delta_index` should only publish `delta_url` when the delta exporter is currently enabled for the dataset (i.e. check `dataset.model.exports` / whether `DeltaExporter.FILE_NAME` is configured), rather than relying solely on the presence of historical artifacts.

## Affected dataset

`ext_us_ofac_press_releases` — either re-enable the delta exporter (add `entities.delta.json` to `exports:`) or stop publishing `delta_url` for it, depending on which is intended.

NickyGuants:
Fixes #5140

`write_delta_index()` rebuilds `delta.json` on every export by scanning *archived* versions for `entities.delta.json`, with no reference to whether the dataset still runs the delta exporter. `write_dataset_index()` then saw that file and published a `delta_url` (and failing that, scanned the archive for old delta exports itself and published one anyway):

```python
# export_dataset(), before
write_delta_index(dataset)                                 # unconditional
write_dataset_index(dataset, DatasetVersionResult.SUCCESS) # publishes delta_url if delta.json exists
```

So `ext_us_ofac_press_releases`, whose `exports:` lists only `statistics.json`, kept advertising a delta feed on every daily run — the newest entry frozen at 2025-10-29, the last run before the exporter came off.

## Changes

- `write_dataset_index()` takes `delta_enabled` and only emits `delta_url` when set. Both the local-`delta.json` branch and the archive-backfill branch are gated; the backfill branch is what was producing the stale URL in production.
- `export_dataset()` skips `write_delta_index()` entirely when the exporter is off, so no stale `delta.json` is written or archived.
- `get_exporter_names(dataset)` lifts the four lines `export_data()` already had, so the "empty `exports:` means defaults, statistics always on" rule has one home. `export_dataset()` and `publish.archive_failure()` derive `delta_enabled` from it.

`delta_enabled` is required rather than defaulted deliberately. A default of `True` would have kept the three existing `write_dataset_index()` test call sites untouched, but it also silently reinstates this bug for any future caller who forgets to pass it — the whole defect is a delta decision made by omission. Making it required means the type checker asks every caller, and there are only two in production code.

An earlier draft put the enabled-exporter lookup in a new `exporters/registry.py`, because `metadata` importing `zavod.exporters` for it would be circular. Passing the flag down instead inverts the dependency — `metadata` stays a leaf and gains no new imports — so that module is gone and `DEFAULT_EXPORTERS`/`EXPORTERS` stay exactly where they were.

## Open question

This stops *advertising* the delta feed; it doesn't touch the `entities.delta.json` artifacts already in `/artifacts/{dataset}/{version}/`. They stay reachable by direct URL for anyone who bookmarked one, and if a dataset ever re-enables the exporter, `write_delta_index()` will pick those pre-gap versions back up and list them alongside new ones — a feed with a silent hole in the middle. Options are to leave them (current behaviour), or have the delta index stop at the first gap. Happy to follow up if you'd prefer the latter, but it seemed out of scope for the fix.

## Manual cleanup

The stale `delta.json` already published for `ext_us_ofac_press_releases` won't be removed by this — the next run just stops linking it, and the copy under `datasets/latest/ext_us_ofac_press_releases/` stays until someone deletes it from the bucket by hand. That matches the existing `_warn_about_stale_latest_files()` policy of warning rather than auto-deleting.

## Verification

- New regression test `test_delta_url_not_published_when_delta_exporter_disabled` reproduces the production scenario: export with the exporter on, archive it, switch `exports` to statistics-only, clear the working dir as a fresh run would, export again. Confirmed red on a pre-fix worktree of main — `index for version …-bbb advertises …/…-bbb/delta.json, but the delta feed only carries ['…-aaa']` — and green here.
- `test_get_exporter_names_config_shapes` pins the exporter resolution across all four config shapes (no `exports:`, explicit without statistics, explicit with statistics, statistics-only), since nothing covered it before. Verified to fail under three separate mutations of the resolution logic.
- `make lint` — clean.
- `pytest zavod/tests/` — 289 passed (287 pre-existing + 2 new).
- `mypy --strict --exclude zavod/tests zavod/` — clean, 124 files.

### DIFF
--- a/zavod/zavod/exporters/__init__.py
+++ b/zavod/zavod/exporters/__init__.py
@@ -43,14 +43,25 @@ EXPORTERS: dict[str, type[Exporter]] = {
     DeltaExporter.FILE_NAME: DeltaExporter,
 }
 
-__all__ = ["export_dataset", "write_dataset_index"]
+__all__ = ["export_dataset", "write_dataset_index", "get_exporter_names"]
+
+
+def get_exporter_names(dataset: Dataset) -> set[str]:
+    """The file names of the exporters enabled for the given dataset.
+
+    A dataset which doesn't configure `exports:` runs the default set. The
+    statistics exporter always runs, because the dataset metadata is derived
+    from its output.
+    """
+    names = set(dataset.model.exports)
+    if not len(names):
+        names.update(DEFAULT_EXPORTERS)
+    names.add(StatisticsExporter.FILE_NAME)
+    return names
 
 
 def export_data(context: Context, view: View) -> None:
-    exporter_names = set(context.dataset.model.exports)
-    if not len(exporter_names):
-        exporter_names.update(DEFAULT_EXPORTERS)
-    exporter_names.add(StatisticsExporter.FILE_NAME)
+    exporter_names = get_exporter_names(context.dataset)
     exporters: list[Exporter] = []
     for name in exporter_names:
         clazz = EXPORTERS.get(name)
@@ -94,7 +105,13 @@ def export_dataset(dataset: Dataset, view: View) -> None:
         context.close()
 
     # Export metadata and issues (after the context is closed & flushed)
-    write_delta_index(dataset)
-    write_dataset_index(dataset, DatasetVersionResult.SUCCESS)
+    # The delta index is built from archived delta exports, so it must only be
+    # written for datasets that still produce them - otherwise it would list
+    # versions from before the exporter was turned off
+    # (https://github.com/opensanctions/opensanctions/issues/5140).
+    delta_enabled = DeltaExporter.FILE_NAME in get_exporter_names(dataset)
+    if delta_enabled:
+        write_delta_index(dataset)
+    write_dataset_index(dataset, DatasetVersionResult.SUCCESS, delta_enabled)
     write_catalog(dataset)
     log.info(f"Exported dataset: {dataset.name}")
--- a/zavod/zavod/exporters/metadata/__init__.py
+++ b/zavod/zavod/exporters/metadata/__init__.py
@@ -84,8 +84,15 @@ def get_base_dataset_metadata(dataset: Dataset) -> dict[str, Any]:
     return meta
 
 
-def write_dataset_index(dataset: Dataset, result: DatasetVersionResult) -> None:
-    """Export dataset metadata to index.json."""
+def write_dataset_index(
+    dataset: Dataset, result: DatasetVersionResult, delta_enabled: bool
+) -> None:
+    """Export dataset metadata to index.json.
+
+    `delta_enabled` states whether the dataset runs the delta exporter, which
+    the caller knows from the exporters it set up. Only then does the index
+    advertise a `delta_url`.
+    """
     catalog = get_catalog()
     version = get_latest(dataset.name, backfill=True)
     if version is None:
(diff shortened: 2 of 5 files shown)


zavod/zavod/exporters/metadata/__init__.py
```diff
@@ -124,24 +131,30 @@ def write_dataset_index(dataset: Dataset, result: DatasetVersionResult) -> None:
...
+            for version in iter_dataset_versions(dataset.name):
+                object = get_artifact_object(
+                    dataset.name, DELTA_EXPORT_FILE, version=version.id
                 )
-                break
+                if object is not None:
+                    meta["delta_url"] = make_artifact_url(
+                        dataset.name, version.id, DELTA_INDEX_FILE
+                    )
+                    break
```
pudo, line 157: What's the logic behind this?

zavod/zavod/exporters/__init__.py
```diff
@@ -94,7 +105,13 @@ def export_dataset(dataset: Dataset, view: View) -> None:
         context.close()
 
     # Export metadata and issues (after the context is closed & flushed)
-    write_delta_index(dataset)
-    write_dataset_index(dataset, DatasetVersionResult.SUCCESS)
+    # The delta index is built from archived delta exports, so it must only be
+    # written for datasets that still produce them - otherwise it would list
+    # versions from before the exporter was turned off
+    # (https://github.com/opensanctions/opensanctions/issues/5140).
```
pudo, line 111: That would crash any downstream consumers assuming that there is a delta file. Would it not be much less disruptive to create the delta file and just keep it empty?

pudo (changes requested):
This is a large infrastructural refactor that would break a lot of assumptions about which files are available for each dataset. It does not just remove `delta_url` from the metadata, which the PR title implies. 

What is the gain here? How does this improve zavod?

PR #5189, opened 2026-07-29
Document position topics, name recognizability and QID verification

leonhandreke:
Follow-up to the position cleanup in https://github.com/opensanctions/opensanctions/pull/5165 and https://github.com/opensanctions/opensanctions/pull/5186 — the skill and docs didn't say to do any of what those PRs fixed.

- `topics`: pass them for positions the crawler names itself, omit for source-supplied ones where the UI decides. Added to `peps.md`, `SKILL.md`, the `make_position` docstring, and three of the four `examples.md` calls (the source-supplied one gets a comment saying why it has none). Vocabulary links to https://www.opensanctions.org/docs/pep/methodology/ rather than restating it.
- Name should be recognizable as belonging to its country when read on its own — either construction works, `Member of the Swedish Riksdag` or `Member of the Senate of the Italian Republic`.
- QIDs: verify `instance of (P31): position` and `applies to jurisdiction (P1001)` before use. Which caught a live bug — `examples.md` keyed the Seanad on Q1396622, a Mexican football league. Fixed to Q18043391, and the two Irish names now match the merged crawler.
- Docstring also gained the missing `id_hash_prefix` entry.

### DIFF
--- a/.claude/skills/crawler-pep/SKILL.md
+++ b/.claude/skills/crawler-pep/SKILL.md
@@ -34,7 +34,7 @@ corpus, are authoritative here.
 
 In addition to the general checks (fields, date formats, language, record count):
 
-- Is there a Wikidata ID for the position(s)? (See `zavod/docs/peps.md`; skip QIDs for per-municipality / per-region positions.)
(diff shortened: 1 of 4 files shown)


herrhellana (approved):
this looks good to me! i'll vibe a few crawlers with the updated skill, but ultimately i think we need to trim it down more and outsource this compartmentalized job to a few subagents invoked per run. the skill is still a tad too lengthy to be manageable for an agent.

i'd also point to omitting `default_pep=True` in `make_position()` in the docs. 

and i updated what to do with the lang arg in the docs and the skill last week here: https://github.com/opensanctions/opensanctions/pull/5124

PR #5191, opened 2026-07-29
Meta/program status field

### LINKED ISSUE #4982: [meta] Add a status field to the Program model
The US program-metadata review (#4978) surfaced several programs that are no longer, or not fully, in force — with no structured way to say so:

- **US-SYR** — revoked outright (EO 14312, July 2025; Caesar Act repealed, 31 CFR 542 removed), but residual SDN entries still carry its tags.
- **US-CORRUPT-353** — the Section 353 designation authority sunset in December 2023; the Engel List is frozen.
- **US-HONGKONG** — the EO 13936 emergency expired July 14, 2026; the program continues only through its HKAA half.
- **US-FSE** — legally in force but the list has been empty since December 2025.

Today the only place to express any of this is summary prose, which neither the UI nor downstream filtering can act on.

### Proposal

Add an optional `status` field to the `Program` model (`zavod/zavod/stateful/programs.py`), e.g. a small controlled vocabulary:

- `active` (default when omitted)
- `wound-down` — designation authority ended (revoked/expired/sunset), but designations made under it may persist in data
- `historical` — no designations under this program remain in any source

Plus authoring guidance in `meta/README.md` (when to set it, and that the summary should still state what happened in prose). The UI could badge non-active programs, and screening users get a machine-readable signal that a hit references a lapsed regime.

Naming and granularity of the vocabulary are open — `revoked` vs `expired` vs `suspended` distinctions exist (EOs get revoked, statutory authorities sunset, emergencies lapse), but a single "no longer designating" bucket may serve users better than legal taxonomy.

nvmbrasserie:


pudo (approved):
I like it, I guess the distinction between `legacy` and `ended` is going to evolve a bit

PR #5198, opened 2026-07-29
[sg_mas_investor_alert] Handle new source fields and multi-value parsing

### LINKED ISSUE #4910: [sg_mas_investor_alert] release issue
cf. https://www.opensanctions.org/issues/sg_mas_investor_alert/

nvmbrasserie:
## Source fields and multi-value parsing

The API now emits a display-string copy of every date and multi-valued field, plus `agency_custom_categories`, `url` and `last_updated_at`. Each was checked redundant against the `_dt`/`_t` originals across all records before going into the `audit_data` ignore list. That was the bulk of the warnings in the last run.

Every multi-valued field separates values with newlines, which none of them split on:

- `website` values were concatenating into nonsense that passed URL cleaning silently, e.g. `https://arqcap.comhttps://gtrmamtechtrading.com`. 60 implausible values, now 0.
- Emails containing a newline failed cleaning and were dropped without a warning. 237 → 364 values.
- Phone lists merged into single values. 536 → 979. Fragments carrying no digits are dropped, removing 630 Tel/Fax and country labels — the phone cleaner rejects nothing, so it has to happen here.
- Addresses merged up to 9 per value. Distinct addresses are separated by a blank line; single newlines are breaks within one address. 762 → 893.

`check_num_found` warned when `numFound` exceeded a magic 999 tied to `rows=1000` in the data URL. A truncated crawl under-collects, so it now asserts `numFound == len(docs)` and will fail rather than quietly drop the tail. The list is at 917 of 1000 rows.

Also fixes `emit_ownership` taking a parameter shadowed by its own loop variable, an email loop that discarded the cleaned value, and warnings that interpolated values into the message text, which made the issues page aggregate distinct problems as one pattern.

## Names

The name field holds websites, aliases, acronyms and MAS's own annotations as well as names, so values like `Amitell Capital Pte Ltd (Impersonated)` and `Telegram Group: "Pictet Official Channel" (t.me/Pictet_Group)` were reaching the `name` property verbatim. The old approach routed anything URL-shaped into a hand-written `names` lookup, where 49 of 59 entries only restated the URL — so every new listing carrying a URL warned until someone added another no-op entry.

That lookup is gone. Names now split two ways:

- **Relationship prose** → the existing `ownership` lookup, which yields the entity's own name plus the counterparty. Keywords keep their preposition, since bare `managed` also matches the 74 `Asset Management` names here. Matching is case-insensitive now, which picks up the `(Operated by …)` records it was missing.
- **Everything else needing cleaning** → a data review, following `_global/adb_sanctions`: an `EntityData` model carrying `website` alongside the five name properties, proposed by `gpt-5.4` and applied only once an analyst accepts. Putting `website` in the model means a reviewer moves a value between `name` and `website` in one pass, rather than the crawler guessing with a regex.

Which values need cleaning is declared in `names.schema_rules` rather than computed in the crawler. `.` in `reject_chars` catches bare domains, at the cost of reviewing ordinary `Pte. Ltd.` names, which are accepted unchanged.

## Rollout

- `ci_test: false` — the crawler needs `OPENAI_API_KEY` and raises without it.
- 228 records start with an open review and are emitted without names until it is accepted. `assert_all_accepted` warns with the outstanding count on every run.


datasets/sg/mas_investor_alert/crawler.py
```diff
@@ -1,46 +1,109 @@
...
+  website      - URLs, bare domains and handles, even where listed as a name
+
+Rules:
+- Preserve original text exactly (no spelling corrections, no title-casing).
+- Do not invent or expand names not present in the input.
+- Split a value holding several names, separated by a slash, a comma or "and".
+- Keep parentheses belonging to the legal name: "Quantum Securities (Singapore) Pte. Ltd"
+  is one name.
+- Discard an "(Impersonated)" marker, and a disclaimer such as "(not affiliated with
+  "MariBank Singapore Private Limited")" together with the institution it names.
```
jbothma: I actually think we need to be a bit careful here - the way I'm reading it, the (Impersonated) marker is saying someone is impersonating this entity.

Image

I think this means this means someone is impersonating 8F Asset Management Pte. Ltd. from the phone numbers +60 13-973 2093, +60 16-329 0757, +60 16-856 0211 

This is the real 8F Asset Management Pte. Ltd

Image

We're about to slap a reg.warn on them

Image

I'm not sure what the solution is. One idea: the name with `(Impersonated)` is emitted as an entity with poi, and an UnknownLink to one other entity who gets the description and in this case, no name, only those phone numbers.

Image

This is a bit of a shitshow - according to el goog, Weatherbie Advisors and Weatherbie Capital are two distict but legitimate businesses. I think MAS is conflating them. I think the suggestion above still works:

- LegalEntity Weatherbie Advisors 
  - topics: poi
- LegalEntity Equilibrium Strategy Academy
  - topics: reg.warn
  - description: This entity impersonates a financial institution that is regulated by the US SEC, Weatherbie Capital LLC and is not regulated by the MAS. Members of the public should take care in ensuring that they are dealing with the actual regulated entity.
- UnknownLink between the two

The description could go on a Sanction too, I'm not too fussed - it's just nice on the naughty entity because it's talking about them as "this entity", but the sanction does point to the entity in question.

jbothma: maybe no topic at all for the benign entity

nvmbrasserie:
8 records carry the marker, in two shapes:

- **6 with only the marked name** (`8F Asset Management`, `Amitell Capital`, `Cygnus Investment`, `Go Asset Management`, `Catalyst Funds Management`, `Financial AI`) — the impersonator is anonymous, and the record's phones, sites and emails are theirs.
- **2 with other names alongside** — `Weatherbie Advisors Ltd` + `Equilibrium Strategy Academy`, and `Point Hope Pte Ltd` + four other `Point Hope` names.

The marker is now routed in the crawler rather than the prompt, so the model never sees one:

```
Weatherbie Advisors Ltd          ← name only, no topics, no sanction
  ↑ UnknownLink  role: "Impersonated by"
Equilibrium Strategy Academy     ← fin, reg.warn, notes, phones, website, sanction
```

The listed entity keeps everything else the record says and is simply left unnamed in the 6 anonymous cases.

Two small deviations from your sketch, both easy to flip:

- I kept the text in `notes`.
- The impersonated party is keyed on its name, matching how `emit_ownership` keys owners, so repeat impersonations of the same business would converge on one entity. None currently do.

Scope: only these 8. The 27 records whose notes describe impersonation without a marker (the `Pentagreen Pte Ltd` pattern) list the impersonator themselves, so `reg.warn` on them is correct and they're unchanged.

datasets/sg/mas_investor_alert/sg_mas_investor_alert.yml
```diff
@@ -44,222 +45,27 @@ assertions:
...
+    LegalEntity:
+      # MAS annotates its names rather than publishing them plain, and uses the name
+      # field for websites too. Punctuation is what marks a value as needing review:
+      # brackets and quotes around annotations and short forms, ":" after a platform
+      # name, "@" before a handle, "." inside a domain. The dot also catches ordinary
+      # "Pte. Ltd." names, which a reviewer accepts unchanged.
+      reject_chars: "()[]{}<>:\"“”@."
+      reject_strings:
+        - " and "
+        - impersonated
```
jbothma: should now not be needed

datasets/sg/mas_investor_alert/crawler.py
```diff
@@ -52,20 +119,40 @@ def emit_ownership(
...
+    impersonated = context.make("LegalEntity")
+    impersonated.id = context.make_id("named", impersonated_name)
+    impersonated.add("name", impersonated_name)
+    context.emit(impersonated)
+
+    link = context.make("UnknownLink")
+    link.id = context.make_id(entity.id, "impersonated", impersonated.id)
+    link.add("subject", impersonated)
+    link.add("object", entity)
+    link.add("role", "Impersonated by")
```
jbothma: nit: I know it's a bit trivial, but maybe worth dropping the "by" just for the principle of using source text verbatim.

```suggestion
    link.add("role", "Impersonated")
```

datasets/sg/mas_investor_alert/crawler.py
```diff
@@ -75,23 +162,77 @@ def emit_relationship(
...
+        # later doesn't reset every one of them.
+        source_data: dict[str, Any] = {
+            prop: sorted(vals) for prop, vals in source.model_dump().items() if vals
+        }
+        source_value = JSONSourceValue(
+            key_parts=[
+                part
+                for prop in sorted(source_data)
+                for part in (prop, *source_data[prop])
+            ],
```
jbothma: sheesh, this is a bit clever for me - I wonder if this could just be like a list and a loop. It might end up nice and simple like

```python
parts = []
for key, values in sorted(source_data.items()):
    parts.append(key)
    parts.append(sorted(values))
```

datasets/sg/mas_investor_alert/crawler.py
```diff
@@ -75,23 +162,77 @@ def emit_relationship(
...
+            emit_ownership(context, entity, name)
+        else:
+            names.append(name)
+    aliases = h.multi_split(alternative_names, NAME_SPLITTERS)
+    previous_names = h.multi_split(former_names, NAME_SPLITTERS)
+
+    source = EntityData(name=names, alias=aliases, previousName=previous_names)
+    if any(
+        h.is_name_irregular(entity, name) for name in names + aliases + previous_names
+    ):
```
jbothma: maybe this can be broken into like three distinct lines, each doing one thing at a time. Or at least two.

datasets/sg/mas_investor_alert/crawler.py
```diff
@@ -101,27 +242,40 @@ def crawl_item(context: Context, item: dict[str, Any]) -> CrawlItemResult:
...
+        if any(char.isdigit() for char in phone):
+            entity.add("phone", phone)
+    # Older records don't use the blank-line separator consistently, so where a
+    # type.address lookup splits a value by hand, leave it to.
+    address_s = item.pop("address_s")
+    if context.lookup("type.address", address_s) is not None:
+        entity.add("address", address_s)
+    else:
+        for address in ADDRESS_SPLITTER.split(address_s):
+            entity.add("address", address)
```
jbothma: doesn't it work to just do this?
```python
for address in ADDRESS_SPLITTER.split(address_s):
            entity.add("address", address)
```
I believe the lookups should take effect on the single string passed in when the splitter ends up not splitting anything.

datasets/sg/mas_investor_alert/crawler.py
```diff
@@ -101,27 +242,40 @@ def crawl_item(context: Context, item: dict[str, Any]) -> CrawlItemResult:
...
     h.apply_date(entity, "modifiedAt", item.pop("modifieddate_dt"))
-    for email in h.multi_split(item.pop("email_s"), [";", " and "]):
+    # None of these can occur inside an address, and the field also carries country
+    # labels ("Singapore: a@b.com") and space-separated lists.
+    for email in h.multi_split(item.pop("email_s"), [";", ",", ":", "/", " ", "\n"]):
+        # Splitting that aggressively leaves fragments of the surrounding free text.
         email_clean = registry.email.clean(email)
         if email_clean is not None:
-            entity.add("email", email)
+            entity.add("email", email_clean)
```
jbothma: could we just to this
```python
for email in h.multi_split(item.pop("email_s"), [";", ",", ":", "/", " ", "\n"]):
    entity.add("email", email)
```
and exclude the cruft via lookups like we do in every other crawler?

jbothma (approved):
cool - I think my comments are almost all cosmetic - it'd be nice to make the code just a little bit more simpleton since the crawler itself is already a fair bit of logic.

PR #5206, opened 2026-07-30
[us_ofac_cons] Map programs by legal authority, not list membership

nvmbrasserie:
NS-MBS and CAPTA bundle designations made under several unrelated legal authorities, so keying them by list name put the same authority under two different programs depending on which list delivered it. Key those two lists by their per-entry program tag instead, and retire US-NS-MBS and US-CAPTA.

The other consolidated lists stay keyed by list name: their entries carry the same tags as SDN entries (UKRAINE-EO13662 and RUSSIA-EO14024 appear on both SSI and NS-MBS), so the tag alone cannot tell them apart, and keying them by tag would dissolve US-SSI and US-NS-CMIC as collateral.

Also, for consolidated entries, free-text `program` now keeps OFAC's own tag and the list name moves to `summary`, since FtM has no property for list membership. This drops the HKAA special case, which falls out of the general rule. 561-Related maps to US-IRAN: the entry is listed under 31 CFR 561.201(c), which that program already covers.

Verified against a test crawl: exactly 13 entities change programId (12 from US-NS-MBS, 1 from US-CAPTA), no entity loses one, and us_ofac_sdn is unaffected. Adds a programId fill-rate assertion to both datasets, since the existing `program` assertion stays satisfied even if no tag resolves to a key.

Refs #4980

### DIFF
--- a/datasets/us/ofac/ofac_advanced.py
+++ b/datasets/us/ofac/ofac_advanced.py
@@ -83,6 +83,19 @@ SANCTION_FEATURES = {
     "Listing Date (CMIC)": "listingDate",
 }
 IMO_RE = re.compile(r"IMO \d{6,9}")
+# Lists whose entries are keyed by their own program tag, because the list bundles
+# designations made under several unrelated legal authorities. `programId` names the
+# authority, so for these the tag is the key and list membership is only a delivery
+# mechanism. The other consolidated lists each correspond to a single authority, so
+# their list name already identifies it - and their entries carry the same tags as
+# SDN entries (e.g. UKRAINE-EO13662 appears on both SSI and NS-MBS), so the tag
+# alone cannot tell them apart. Each dataset maps the keys it can encounter in its
+# own sanction.program lookup. cf. #4980
+AUTHORITY_KEYED_LISTS = {
+    "SDN List",
+    "Non-SDN Menu-Based Sanctions List",
+    "CAPTA List",
+}
 
 
 def get_relation_schema(party_schema: Schema | None, range: Schema | None) -> Schema:
@@ -541,7 +554,14 @@ def parse_id_reg_document(
 
 
 def extract_sanctions_measure_name(entry: Element, refs: Element) -> str | None:
-    """Extract the source program tag from a sanctions entry."""
+    """Extract the source program tag from a sanctions entry.
+
+    An entry can carry several Program-type measures, i.e. name several legal
+    authorities at once. We take the first one, so the others survive only in
+    `provisions`. Which tag comes first is decided by OFAC's element order, so
+    the choice is arbitrary where the tags resolve to different programs.
+    Making `programId` multi-valued is tracked separately, cf. #4980.
+    """
     for measure in entry.findall("./SanctionsMeasure"):
         sanctions_type_id = measure.get("SanctionsTypeID")
         sanctions_type = get_ref_text(refs, "SanctionsType", sanctions_type_id)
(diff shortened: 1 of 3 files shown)


nvmbrasserie:
### Deprecation needs its own field, not `status`

This PR retires `US-NS-MBS` and `US-CAPTA` with a `# DEPRECATED` header comment, following the `GB-SAMLA` precedent. I looked at whether the new `status` field from #5191 could carry that instead. It can't, and I think the gap is worth filling separately.

`status` records the legal status of the regime. Per `meta/README.md`, it "says nothing about whether the source data currently carries entries", and `ended` is explicitly reserved for regimes genuinely no longer law, "not ones merely dormant, unused, or narrowed while still in force".

Both `legacy` and `ended` share a precondition: the regime is no longer fully in force. They differ only on whether designations still bite. Neither of these two keys passes that precondition:

| list | designation events | earliest | latest |
|---|---:|---|---|
| Non-SDN Menu-Based Sanctions | 54 | 2020-12-14 | **2026-07-17** |
| CAPTA | 1 | 2019-03-14 | 2019-03-14 |

NS-MBS took a new designation ten days before the snapshot I tested against, under authorities (CAATSA, HKAA, EO 14024, EO 14014, EO 14059) that are all in force. CAPTA is dormant, but its authorities (NKSR, IFSR, HIFPAA, the Russia provisions) remain in force and it can receive entries at any time. Marking either as `legacy` or `ended` would assert something false about the law.

What we are actually recording is different: these keys described *lists* rather than legal authorities, so we retired the keys while the underlying regimes carry on. That is a modelling lifecycle, orthogonal to legal status. `US-HONGKONG: legacy` shows the field used correctly by contrast, EO 13936's emergency expired while HKAA designations persist.

So the vocabulary has no value meaning "key retired, regime still in force", and neither `meta/README.md` nor `zavod/docs/programs.md` documents any retirement mechanism. `GB-SAMLA` hit the same wall and also resorted to a comment.

Proposal for a follow-up change:

- A separate field on the `Program` model, e.g. `deprecated: true`, optionally with `superseded_by: [<keys>]` to record where the designations went.
- Policy prose in `meta/README.md` stating that deprecation and `status` are independent: a deprecated key can wrap a fully active regime.
- A unit test, plus converting the `GB-SAMLA`, `US-NS-MBS` and `US-CAPTA` comments over to it.

The benefit over a comment is that consumers and our own tooling could then tell "do not use this key" from "this regime ended", which are currently indistinguishable in the metadata. Happy to pick this up if it sounds right.

pudo:
Can we maybe do this in two steps: one that establishes the programs - then we wait a day or two, and then the actual crawler updates? This way we don't end up referencing unpublished programs for key US lists.

nvmbrasserie:
All six programs the change newly references (US-RUSHAR, US-UKRRUS-REL, US-CAATSA, US-BURMA, US-NARCO, US-IRAN) already exist on main and are already referenced by the live `us_ofac_sdn`. So no new programs are created here

nvmbrasserie:
Follow-up commit 216d98d3, metadata only — completes the program side of this change.

- **US-CAATSA**: `dataset` → `us_ofac_sdn` (that dataset ingests the designations; `wd_oligarchs` emits no `programId`), added the `CAATSA - IRAN`/`CAATSA - RUSSIA` aliases and `Asset freeze`. Boilerplate summary replaced — supersedes the US-CAATSA hunk in #4981.
- All 13 remapped entries turn out to carry **non-blocking** directive or menu measures, so the receiving programs now record the union of operative measures per `meta/README.md`: `+Financial restrictions` on **US-UKRRUS-REL** (EO 13662 Sectoral Directives 1/2) and **US-NARCO** (EO 14059 loan/credit prohibition, per [jy2542](https://home.treasury.gov/news/press-releases/jy2542)). US-RUSHAR, US-BURMA and US-IRAN already covered theirs.
- Variation clause added to all five receiving summaries; the US-NS-MBS and US-CAPTA deprecation headers now record that per-entry severity lives only in `Sanction:provisions` until a normalized `Sanction:measures` lands.

Outstanding: the Section 235 menu items imposed on SSB could not be verified (State and OFAC pages both unavailable), so US-CAATSA's `measures` is a floor rather than the full union.

nvmbrasserie:
Split the program metadata out into #5272, which this PR is now based on, since program data takes a while to propagate to data.opensanctions.org and should be live before the attributions move.

- **#5272** (base, targets `main`) — `meta/programs/` only, 8 files, no entity data moves.
- **this PR** — `ofac_advanced.py` and the two dataset configs, 3 files. Unchanged in substance; the branch was rebuilt on #5272 rather than reworked.

Merge #5272 first, then this one.

PR #5166, opened 2026-07-28
Add /dataset-metadata skill and rework metadata docs

pudo:
Reworks `zavod/docs/metadata.md` into the authoritative reference for dataset metadata conventions and adds a terse `/dataset-metadata` skill that applies it.

## Docs (`zavod/docs/metadata.md`)

- `title`, `summary`, `description` promoted to their own sections with concrete conventions:
  - titles start with the issuing country's short English name (or the international issuer's name/acronym), never possessive; established acronyms kept;
  - summaries are one plain fragment just above 50 chars, no title repetition, no humor;
  - descriptions cover scope and institutional context only — no per-record field lists, ETL mechanics, or sourcing narration.
- New **Maintainer notes (YAML comments)** section: crawler-level notes (failure modes, source quirks, runbooks) go in a top-of-file `#` block, and each non-type lookup gets a short purpose comment. Comments record evidence, never speculation.
- Frequency house defaults by dataset type (sanctions `daily`, PEP `monthly`, registries `weekly`, frozen dumps `never` + schedule), and the actual schedule resolution order (`coverage.schedule` → `deploy.schedule` → frequency mapping, minute randomized per dataset — verified against the deploy tooling).
- Curated documentation for previously undocumented fields: `hidden`, `disabled`, `data.lang`, `resolve`, `full_dataset`, `manual_check`, and the `deploy.*` keys with their deployed defaults.

## Skills

- New `.claude/skills/dataset-metadata/SKILL.md`: procedure-only checklist deferring every rule to the doc, with a no-speculation guard for comments and a metadata-load verify step.
- `.claude/skills/legislature-metadata/SKILL.md` slimmed to the legislature-specific title/description/frequency pattern; per-member attribute enumeration dropped from descriptions (derivable from data statistics, drifts with the crawler). All other fields delegate to `/dataset-metadata`; the two skills cross-reference so either entry point works.

## Trial application

`[de_bundestag]` metadata updated with the new conventions as a worked example (second commit): scope-focused description, plain summary, publisher `name`/`name_en` split.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5233, opened 2026-07-31
skill and docs updates

herrhellana:


herrhellana:
@goose-life i made some changes to the docs -- do you agree? i also replaced the non-existing module path `zavod.logic.pep.categorise` with `zavod.stateful.positions.categorise` in position categorization, the rest is minor.

pudo:
Did you see the metadata skill I asked for feedback on?

herrhellana:
reading it now -- not related to the current PR though?

pudo:
Edit the same files so there's a conflict.

herrhellana:
hm then i'm looking at the [wrong PR](https://github.com/opensanctions/opensanctions/pull/5008/changes#diff-3e1646376007ad477983b97581dc362485fea03c05e58d8af5b31555ef67093a)?

pudo:
https://github.com/opensanctions/opensanctions/pull/5166

herrhellana:
nice, thank you, i'll look at the diff!

.claude/skills/crawler-pep/examples.md
```diff
@@ -61,8 +62,8 @@ def crawl(context: Context) -> None:
         wikidata_id="Q...",
         lang="eng",  # crawler-supplied names are always English
     )
-    categorisation = categorise(context, position, default_is_pep=True)
-    # Gate even with default_is_pep=True: the position may have been un-flagged
+    categorisation = categorise(context, position)
+    # Gate even though default_is_pep is True: the position may have been un-flagged
```
nvmbrasserie, line 66: `default_is_pep=True` is helpful for visibility and for understanding the nature of the source. I don't think we should remove it

herrhellana: @pudo suggested (a while back) that we do since it's `True` by default, and it's been a convention in a lot of merged legislative crawlers in the past months. it sounds more like a taste preference?

.claude/skills/crawler-pep/examples.md
```diff
@@ -157,9 +158,12 @@ lookups:
...
 ```
 
-## Pattern C: Multi-position crawler with `default_is_pep=True`
+## Pattern C: Multi-position crawler
 
-For sources that list officials across known, enumerable position types:
+For sources that list officials across known, enumerable position types. `POSITIONS`
+here is an enumeration-as-data table, which is a legitimate module-level constant —
+unlike a per-record `dict` passed between functions, which reviewers reject: parse each
+record straight into its entity rather than into an intermediate `dict`.
```
nvmbrasserie: ideally, positions would not be hard-coded in the code and would always be sent for categorisation review. I wouldn't add this bit either

nvmbrasserie (changes requested):

.claude/skills/crawler-pep/examples.md
herrhellana: the use of `.yml` lookups in`us_congress` for multiple positions looks like a nice practice? i rewrote the multi-position guideline (see the latest commit).

nvmbrasserie:
Deep review done, focused on the changes to `zavod/docs/peps.md` and `zavod/zavod/helpers/positions.py`. No action points — good to merge as is.

Verified rather than assumed:

- `default_is_pep` really does default to `True` (`bool | None = True`), so a plain `categorise(context, position)` call is equivalent and dropping the argument from the example changes nothing.
- The `zavod.logic.pep` → `zavod.stateful.positions.categorise` docstring fix points at the right module.
- The lookup mechanics in "Where a position's definition belongs" behave as documented. Checked against a real `Dataset` carrying that exact lookup: a matched option exposes `.name`, `.topics` and `.wikidata_id`; an option with `value: null` and no `name` yields `res.name is None` rather than raising; an unmatched label yields `None`. `us_congress` uses the same three keys in production.
- Both cross-references resolve: `datapatch_lookups.md#mapping-to-richer-concepts` and `strict_interpretation.md#key-categorical-fields-need-complete-coverage`.

#5294 is closed — nothing left in it.

nvmbrasserie (approved):

PR #5243, opened 2026-08-02
Add zavodlake, a parquet statement lake workbench

pudo:
Prototyping workbench in `contrib/zavodlake` exploring parquet statement artifacts as the substrate for export, xref, analysis and enrichment processes.

**What's here:**
- `build` command: backfills `statements.pack` + `index.json` for all leaves of a collection (default: `default`) and converts each pack into a deduplicated, typed `data/lake/<dataset>/statements.parquet` via duckdb — dedupe on statement id (richest-row representative), sorted by `entity_id`, zstd.
- `PLAN.md`: the experiment agenda and agreed long-term layout (versioned artifacts, per-dataset pointer files, event-driven builds).
- `perf.md`: research notes and benchmark results — point-lookup numbers on the full 127M-statement lake, the (negative) LevelDB-build experiment, and validation of the nomenklatura `DuckDBStore` (companion PR in nomenklatura).

Experiment harnesses (`exp*_*.py`) are deliberately not committed.

### DIFF
--- a/contrib/zavodlake/PLAN.md
+++ b/contrib/zavodlake/PLAN.md
@@ -1,0 +1,34 @@
+---
+description: Experiment agenda for the zavodlake parquet statement lake
+date: 2026-08-02
+tags: [zavodlake, parquet, duckdb, prototyping]
+---
+
+# zavodlake: parquet statement lake experiments
+
+The `build` command in this package backfills `statements.pack` and metadata for every
+dataset in a collection and converts each pack into a deduplicated, typed
+`data/lake/<dataset>/statements.parquet`.
+
+## Idea
+
+The fundamental idea is that these parquet files might be able to shift burden away from
+the export, xref, analysis and enrichment processes — which today each fetch and load the
+full statement set — towards re-usable, queryable artifacts.
+
+## Experiments
+
+1. **Parquet-backed nomenklatura Store.** Can the lake drive a `Store` implementation that
+   is slower but runs off re-usable artifacts rather than fetching all statements? This
+   will likely not work in export (graph traversal requires a lot of queries), but might
+   work in xref, which only reads individual entities in its second phase.
+2. **Faster LevelDB store builds.** Can a LevelDB store be built more quickly by feeding
+   it from a parquet query instead of the pack files?
+3. **Stateless xref decision tool.** Can the lake provide a backend store for an
+   interactive web-based xref decision tool running as a stateless Cloud Run job?
+4. **Inverted-index parquet.** Can we generate a second set of parquet files containing a
+   tokenized inverted index à la `nomenklatura.index`, usable to drive xref and enrichment
+   straight from bucket artifacts?
+
+None of these necessarily intersect with the resolver — the entity queries involved can
+just be `entity_id IN (...)` lookups on source ids.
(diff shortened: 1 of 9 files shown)


jbothma:
Follow-up on `perf.md` in this PR, from benchmarking the shipped batch store against a production-scale resolver. I think the linker-mapping load in `DuckDBBatchView._build` is worth optimising before the store switch is decided.

`_build` (nomenklatura 4.14.0, `store/duckdb_batch.py`) loads the `entity_id -> canonical_id` mapping with `conn.executemany("INSERT INTO ... VALUES (?, ?)", batch)` in 50k batches. Row-wise inserts into DuckDB run at roughly **155 µs/row**, flat as the table grows, so this dominates a view build once the resolver is real.

Measured against a resolver with 1,284,727 live POSITIVE edges (1,998,490 mapping rows), swapping **only** the mapping load for a `csv.writer` + `read_csv`:

Isolated, same pre-materialized rows handed to both paths:

| rows | `executemany` | `csv` + `read_csv` | speedup |
|---|---|---|---|
| 100,000 | 15.74 s | 0.09 s | 174x |
| 400,000 | 61.95 s | 0.25 s | 252x |

End-to-end `store.view(scope)` over a parquet lake built from 21 local dataset packs (6,914,390 statements; 1,399,384 non-external):

| scope | statements in relation | as shipped | with the swap | speedup |
|---|---|---|---|---|
| 5 smallest datasets | 1,410 | **325.4 s** | 1.6 s | 207x |
| all 21 datasets | 6,914,390 | **325.7 s** | 2.4 s | 137x |

The pair of end-to-end rows is the part I'd flag: **the build takes the same ~325 s whether the view covers 1,410 statements or 6.9M**, because the mapping is loaded in full regardless of scope. The actual statement work is the 1.6 -> 2.4 s difference. Resulting `stmt_table` and `edge_table` are md5-identical between the two paths.

That also means the cost lands on small scopes hardest, and it is paid **per view** — `_build` runs eagerly in `DuckDBBatchView.__init__` and views aren't reused, so e.g. a local-enricher run that builds a target view and a subject view pays it twice.

Worth raising here specifically because exp1b's "store init 0.12 s" measured the earlier `duckdb_.py` draft, which built its mapping from `iter_pairs()` once per *store*. nomenklatura#348 replaced that with the materialized `_build`, which drains `linker.mappings()` (one row per identifier per cluster, not per edge) once per *view* — so that line no longer describes the merged code. `tests/store/test_duckdb_batch.py` uses `Linker({})` or a two-pair fixture, so nothing in CI catches it either.

I can't account for how the 0.1 s sync / 115 s export figure on #5516 avoided this — my guess is a much smaller resolver in that environment, but I'd rather someone check than assume the number is wrong. Worth re-running with a full linker before the LevelDB -> lake switch is called.

A few notes for whoever picks this up:

- `nomenklatura/blocker/index.py` (`load_entities`) already does exactly the right thing — batch to a `csv.writer`, then `INSERT INTO ... SELECT * FROM read_csv(...)`. It's a near drop-in template, and `duckdb_batch.py` is the only remaining `executemany`-into-DuckDB in the repo.
- `duck.connect()` sets `python_enable_replacements=False`, and neither `pyarrow` nor `pandas` is a dependency, so CSV (or an explicit `conn.register()`) rather than handing DuckDB a dataframe.
- The mapping table is joined twice (statement CTAS and edge CTAS), so it needs to stay a real table — it can't be inlined into a single query.

Normalising for indentation and docstrings, the change is +10/-8 lines in one hunk, with nothing outside the mapping-load block touched.

pudo:
Nice research. The store now writes a CSV. 

I think we can close this PR - it was just research, and we have an actual implementation now :)

PR #5267, opened 2026-08-03
[lt_pep_declarations] deprecation

brrttwrks:
In preparation for deprecation on Monday, 17 August 2026, creating PR:
- remove from collections (peps)
- update metadata file


datasets/lt/pep_declarations/lt_pep_declarations.yml
jbothma: could add deprecation date here

brrttwrks: I went ahead and added the date in the description.

PR #5270, opened 2026-08-03
[cz_national_sanctions.yml] temporary outage fixes

### LINKED ISSUE #5122: [cz_national_sanctions.yml] temporary outage
The ministry stopped publishing the XLSX version of the national sanctions list on the source page, so the crawler has been failing since 2026-07-24 (last successful run 2026-07-23). The `.xlsx` anchor it matches is no longer present on the page:

> Expected 1 elements, got 0 for xpath `//a[contains(@href, ".xlsx")]/span[contains(text(), "Vnitrostátní sankční seznam")]/../@href`

**What we tried:** repointing the source to the CSV published in the ministry's open-data section (`.../o_ministerstvu/otevrena_data/index_5.html`), fetched through Zyte browser rendering since the attachment link is only rendered client-side. It didn't work.

The XLSX has been up and down over the past little while, so this is most likely just one more temporary outage of the source rather than a permanent change. Suggest we give it a couple of days to come back before committing to the CSV rewrite.

Source page: https://mzv.gov.cz/jnp/cz/zahranicni_vztahy/sankcni_politika/sankcni_seznam_cr/vnitrostatni_sankcni_seznam.html

nvmbrasserie:


datasets/cz/national_sanctions/crawler.py
```diff
@@ -1,52 +1,46 @@
...
 from rigour.names import pick_name
 
 from zavod import Context, helpers as h
 
 # detect anything more complex than word/word/word (and handle question mark woopsie)
 REGEX_LAST_NAME = re.compile(r"^[\w\?]+( ?/\s*[\w\?]+)*$")
+# The note on a cancelled or amended entry states the effective date(s) in prose,
+# e.g. "Zápis byl zrušen ke dni 24. února 2025 v souladu s § 7 ...", sometimes
+# with a numeric month. The formats are handled by the dataset `dates` config.
+REGEX_STATUS_DATE = re.compile(r"ke dni\s+(\d{1,2}\.\s*\w+\.?\s*\d{4})")
```
jbothma: could we include `Zápis byl zrušen` in here to be sure we're capturing the right date?

datasets/cz/national_sanctions/crawler.py
```diff
@@ -1,52 +1,46 @@
...
-    # -> Provision of the European Union regulation, the factual basis of which the subject fulfilled by action
-    provision = row.pop(
-        "ustanoveni_predpisu_evropske_unie_jehoz_skutkovou_podstatu_subjekt_jednanim_naplnil"
-    )
+    status_note = row.pop("entry_status_note").strip()
+    # One note states the wrong year, so the dates pass through a lookup.
+    status_dates = [
+        context.lookup_value("status_note_dates", text, text)
+        for text in REGEX_STATUS_DATE.findall(status_note)
+    ]
```
jbothma: instead of passing through a regex and lookup, what do you think of just passing the entire status field through a lookup which returns a dict with `endDate` and `notes` keys? Then there's no risk of the regex missing end dates.

jbothma: and maybe a separate key for modification dates. It looks like currently we'd stuff all dates available in the status field in the endDate prop which wouldn't be right.

datasets/cz/national_sanctions/crawler.py
```diff
@@ -82,32 +72,35 @@ def crawl_item(context: Context, row: dict[str, str | None]) -> None:
...
+            context.log.warning(
+                "No cancellation date in status note", note=status_note, name=name_field
+            )
+        h.apply_dates(sanction, "endDate", status_dates)
+    elif status_key == "amended":
+        # An amended entry is superseded by a later, valid entry for the same
+        # subject, so the dates state when it was amended, not when it ended.
+        h.apply_dates(sanction, "date", status_dates)
+    else:
+        assert status_key == "valid", (status, status_key)
```
jbothma: I guess it might be nice if we just warned if it wasn't known, which we could do using the lookup and the warn arg there. Then a new status or typo in status doesn't abort the run and miss new data.

datasets/cz/national_sanctions/crawler.py
```diff
@@ -82,32 +72,35 @@ def crawl_item(context: Context, row: dict[str, str | None]) -> None:
...
-    # Omezující opatření -> Restrictive measures
-    sanction.add("provisions", row.pop("omezujici_opatreni"), lang="ces")
-
-    # Číslo usnesení vlády
-    # -> Government resolution number
-    sanction.add("recordId", row.pop("cislo_usneseni_vlady"), lang="ces")
+    sanction.add("reason", row.pop("conduct_description"), lang="ces")
+    sanction.add("provisions", row.pop("restrictive_measures"), lang="ces")
+    # The legal instrument of the entry, i.e. a government resolution.
+    sanction.add("recordId", row.pop("entry_legal_instrument"), lang="ces")
```
jbothma: `Právní předpis zápisu` (Legal regulation of registration) doesn't sound right for this prop

datasets/cz/national_sanctions/crawler.py
```diff
@@ -117,24 +110,29 @@ def crawl_item(context: Context, row: dict[str, str | None]) -> None:
...
+
+    columns = context.get_lookup("columns")
+    with open(path, encoding="utf-8") as fh:
+        for record in csv.DictReader(fh):
+            row = dict()
+            for header, value in record.items():
+                # An unmapped heading is passed through under its source name, so
+                # the pops in crawl_item fail or audit_data warns about it. A row
+                # with fewer values than headings has None values.
+                column = columns.get_value(header, header)
```
jbothma, line 164: nice

datasets/cz/national_sanctions/cz_national_sanctions.yml
```diff
@@ -27,11 +27,13 @@ tags:
   - issuer.west
 url: "https://mzv.gov.cz/jnp/cz/zahranicni_vztahy/sankcni_politika/sankcni_seznam_cr/vnitrostatni_sankcni_seznam.html"
 data:
-  url: https://mzv.gov.cz/jnp/cz/zahranicni_vztahy/sankcni_politika/sankcni_seznam_cr/vnitrostatni_sankcni_seznam.html
+  # The sanctions list page only publishes a PDF; the machine-readable CSV is
+  # linked from the ministry's open data section.
+  url: https://mzv.gov.cz/jnp/cz/o_ministerstvu/otevrena_data/index_5.html
```
jbothma, line 32: it might be worth checking hashes on the page listing the PDF and the CSV for a couple of months - for a long time the page linking to the CSV was stated to be the open data version, yet wasn't updated for a year or two. So it could just be worth checking they update these two pages in lockstep going forward.

datasets/cz/national_sanctions/crawler.py
jbothma: or better, see manual extraction comment below

datasets/cz/national_sanctions/crawler.py
jbothma: maybe check the key/prop/translation mappings for the rest too - did claude fall over its feet a bit?

jbothma (commented):
glad this is coming back

nvmbrasserie:
Addressed the review in a9af92eb2:

- The whole status note now goes through a `status_note` lookup returning `end_dates` and `amendment_dates` per note; the regex and the `status_note_dates` lookup are gone. Unmatched notes and cancelled entries without an end date warn.
- Amendment dates go to `modifiedAt` instead of `date` (`date` counts as `temporalStart`, so it was shifting the sanction interval).
- No more assert on the entry status — the lookup warns and the run continues.
- `Právní předpis zápisu` moved off `recordId` (the values are full sentences, not IDs) to `description`. Happy to extract just the resolution number into `recordId` instead if you prefer.
- Re-checked the rest of the column keys: `nationality_or_registered_office`, `entry_or_amendment_date`, `entry_legal_regulation`. Dropped the Czech month date config, which only existed for the prose dates.
- Instead of watching the two pages by hand, the crawler compares the PDF and CSV filenames (both published as `Vnitrostatni_sankcni_seznam_<date>`, currently `2026_07_23`) and warns if the open data CSV falls behind.

Full crawl is clean: 52 entities, no issues.

jbothma:
> Právní předpis zápisu moved off recordId (the values are full sentences, not IDs) to description. Happy to extract just the resolution number into recordId instead if you prefer.

was I wrong here about the legislation? is it legislation in the sense that one of these individual references is the document where that person gets sanctioned? Or are they references to broad legislation like an act under which a person could be sanctioned? The robot response isn't really telling me I'm wrong or right, it's solving a different one to what I thought the issue is.

Has some human checked these column mappings? Can I get proof of life? :P

nvmbrasserie:
you were right that it didn't belong to `recordId`, sorry, I should have addressed it properly. I double checked the property mappings and was happy with what I saw. I think description is a better property for it. It is the government resolution that created the entry

jbothma:
lovely stuff, thanks

PR #5277, opened 2026-08-03
[us_cuba_sanctions] Watch the current-list pages for change detection (#4979)

### LINKED ISSUE #4979: [us_cuba_sanctions] Change detection watches the frozen initial-publication page
The crawler's entity data comes from a hand-maintained Google Sheet, with `assert_dom_hash` change detection on the State Department pages meant to warn us when the source lists change.

For the Cuba Prohibited Accommodations list, the watched page is the **initial publication** (September 2020, frozen by definition):
https://www.state.gov/cuba-prohibited-accommodations-list-initial-publication/

State maintains a separate current-list page, updated as recently as July 14, 2025:
https://www.state.gov/cuba-sanctions/cuba-prohibited-accommodations-list

Since the watched page never changes, list updates will not trigger the "Check for data updates" warning and the Google Sheet can silently go stale.

Suggested fix:
- Point the accommodations change detection at the current-list page and refresh the DOM hash.
- While in there, verify the Cuba Restricted List URL too — the crawler watches `https://www.state.gov/division-for-counter-threat-finance-and-sanctions/cuba-restricted-list`, and `https://www.state.gov/cuba-sanctions/cuba-restricted-list` appears to be the canonical current-list page ("As of July 14, 2025").
- Note for whoever picks this up: state.gov serves an HTTP-200 "Technical Difficulties" page to non-browser user agents, so verify page content with a realistic UA.

Found during the program-metadata review (#4978).

dennismdejong:
Fixes #4979

## Change
- Accommodations change detection now watches the live current-list page `https://www.state.gov/cuba-sanctions/cuba-prohibited-accommodations-list` (was the frozen initial-publication page).
- Restricted-list change detection now watches `https://www.state.gov/cuba-sanctions/cuba-restricted-list` (canonical URL; 301-redirects to the page previously watched).
- yml `description` links updated to match.

## Change-detection hashes
- Accommodations: `cde71ba89a1b51b1b1d8d780d4902f72ca1d384c` — verified by maintainer against Zyte (correct).
- Restricted list: reverted to `a146ff14f0a283a4a80afaaf0f46637574aa78c2` — the new URL 301-redirects to the same page already watched, so the existing Zyte-computed hash stays valid. The earlier `60164794…` was a curl-only hash (breadcrumb widget differs under a browser) and warned on every run; it has been removed.

## Per maintainer review
- The accommodations page contains 8 properties effective 2025-07-14 missing from the hand-maintained Google Sheet (incl. Meliá Trinidad Península, Meliá Costa Rey). Sheet update is being handled by OpenSanctions; the accommodations hash bump is sequenced with it.
- The unrelated `.gitignore` commit (superpowers plans / worktrees) has been dropped from this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,9 @@ apicache-py3
 # Emacs
 *~
 task-prompts/
+
+# Superpowers implementation plans
+docs/superpowers/plans/
+
+# Git worktrees
+.worktrees/
--- a/datasets/us/cuba_sanctions/crawler.py
+++ b/datasets/us/cuba_sanctions/crawler.py
@@ -10,7 +10,7 @@ ORIGINAL_ACCOMMODATIONS_URL = (
     "https://www.state.gov/cuba-sanctions/cuba-prohibited-accommodations-list"
 )
 ACCOMMODATIONS_URL = "https://docs.google.com/spreadsheets/d/e/2PACX-1vQMquWjNWZ09dm9_mu9NKrxR33c6pe4hpiGFeheFT4tDZXwpelLudcYdCdME820aKJJo8TfMKbtoXTh/pub?gid=1890354374&single=true&output=csv"
-ORIGINAL_RESTRICTED_ENTITIES_URL = "https://www.state.gov/division-for-counter-threat-finance-and-sanctions/cuba-restricted-list"
+ORIGINAL_RESTRICTED_ENTITIES_URL = "https://www.state.gov/cuba-sanctions/cuba-restricted-list"
 RESTRICTED_ENTITIES_URL = "https://docs.google.com/spreadsheets/d/e/2PACX-1vQMquWjNWZ09dm9_mu9NKrxR33c6pe4hpiGFeheFT4tDZXwpelLudcYdCdME820aKJJo8TfMKbtoXTh/pub?gid=0&single=true&output=csv"
 CONTENT_XPATH = ".//div[@class='entry-content']"
 ACTIONS = [
@@ -63,6 +63,9 @@ def crawl_restricted_entities(context: Context) -> None:
     node = doc.find(CONTENT_XPATH)
     # Chrome save HTML only
     # xmllint --format --html --encode UTF-8
+    # TODO: recompute this hash from the new URL
+    # https://www.state.gov/cuba-sanctions/cuba-restricted-list
+    # via Zyte fetch (state.gov serves a Technical Difficulties page to curl).
     if not h.assert_dom_hash(node, "a146ff14f0a283a4a80afaaf0f46637574aa78c2"):
         context.log.warning("Restricted List content changed. Check for data updates")
 
--- a/datasets/us/cuba_sanctions/crawler.py
+++ b/datasets/us/cuba_sanctions/crawler.py
@@ -32,10 +32,7 @@ def crawl_accommodations(context: Context) -> None:
     node = doc.find(CONTENT_XPATH)
     # Chrome save HTML only
     # xmllint --format --html --encode UTF-8
-    # TODO: recompute this hash from the new URL
-    # https://www.state.gov/cuba-sanctions/cuba-prohibited-accommodations-list
-    # via Zyte fetch (state.gov serves a Technical Difficulties page to curl).
-    if not h.assert_dom_hash(node, "6dc9087e0ccb2e13fc2389ba4176ab114996ad32"):
+    if not h.assert_dom_hash(node, "cde71ba89a1b51b1b1d8d780d4902f72ca1d384c"):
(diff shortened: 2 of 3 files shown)


pudo:
Thanks for picking this up! I ran the branch end-to-end against Zyte and reproduced each hash independently. The accommodations half is correct; the restricted-list half needs to come out; and there's a data-reconciliation step that has to happen before the accommodations hash can be bumped.

First, a method check: fetching the landing page through the crawler's own path (`fetch_html` → Zyte `browserHtml`) gives `0355144d…`, matching the committed hash — so my numbers are comparable to what the crawler sees in production.

### ✅ Accommodations URL + hash — correct

`https://www.state.gov/cuba-sanctions/cuba-prohibited-accommodations-list` fetches cleanly with `waitForSelector`, and `cde71ba89a1b…` matches the Zyte DOM exactly. Verified twice (standalone fetch, and a full `zavod crawl` where no warning is logged).

### ❌ Restricted-list hash — wrong, warns on every run

`https://www.state.gov/cuba-sanctions/cuba-restricted-list` **301-redirects to the URL already in the crawler** (`division-for-counter-threat-finance-and-sanctions/cuba-restricted-list`). It's the same page: Zyte returns an identical DOM for both, hashing to `a146ff14…` — the value already committed.

`60164794…` is the hash of the *curl* response. The cause is specific to this page: its `entry-content` contains a breadcrumb widget whose inline script runs `bcEllipse.style.display = "none"`. A headless browser applies that, so `browserHtml` carries `<span class="bc_ellipse" style="display: none;">`; a raw HTTP fetch does not. That single attribute is the whole hash difference.

This is why the curl-vs-Zyte validation gave a false green: it holds for the landing page and the accommodations page (identical raw vs. rendered) and breaks precisely here. Worth remembering generally — this crawler hashes `browserHtml`, so hashes have to be computed through `fetch_html`, not curl.

Running this branch:

```
warning  DOM hash changed: expected=60164794c60c… actual=a146ff14f0a2…
warning  Restricted List content changed. Check for data updates
```

That would fire on every weekly run and mask the next real change — strictly worse than today. Please revert this hash to `a146ff14…`. The URL swap itself is cosmetic (it just adds a redirect hop); fine to keep or drop, your call.

For confidence that this list really is current: `fe8ddb65b [us_cuba_sanctions] update hash - no listed entity changes` (2026-06-19), and the source sheet already carries 8 restricted-list rows with `EffectiveDate 2025-07-14`.

### ⚠️ The accommodations change is real, and bumping the hash discards it

This is exactly the failure #4979 predicted, so the fix works — but the current page ("As of July 14, 2025") contains 8 properties effective 2025-07-14 that are **missing from the hand-maintained Google Sheet**:

- Grand Aston La Habana
- Grand Aston Varadero Beach Resort
- Iberostar Selection La Habana (aka "Torre K")
- INNSiDE Habana Catedral
- Varadero Sol Caribe
- Meliá Trinidad Península
- Meliá Costa Rey
- Sol Turquesa Beach

Six exist as entities via the restricted list but without the Prohibited Accommodations sanction; **Meliá Trinidad Península and Meliá Costa Rey are absent from our output entirely.**

Also new on the page: three alias additions (Hotel Sevilla Habana Affiliated by Meliá; Roc Casa del Mar; Lagunas del Mar) and several renames/corrections — América/Barcelona/Louvre/Real → "Hotel …", Hotel Sevilla → Hotel Sevilla Affiliated by Melia, `Plaza (also: Hotel Islazul Plaza Camaguey)` → `Hotel Plaza` with postcode 70 100 → 72 100, and Gran Hotel Bristol Kempinski's address now "485 Av. Bélgica". No delistings — every name in the sheet still appears on the page.

So the hash bump is the last step of this fix, not the first: the sheet gets updated, then the hash is set to match. We (OpenSanctions) maintain that sheet, so that part is on us — I'll take care of it. Happy to land the URL change ahead of it if you'd rather not wait; we'd just keep the accommodations hash mismatching in the meantime so the reminder keeps firing.

### Minor

- The `.gitignore` commit (`docs/superpowers/plans/`, `.worktrees/`) is unrelated to this fix — that kind of local tooling is better in a personal global gitignore. Could you drop it from the PR?
- The sheet's `SourceURL` column still points at the old URLs, so emitted `sourceUrl` values keep the initial-publication and division pages even after this change. Both are live and 200, so nothing is broken — just noting that the sheet, not the crawler constants, is what reaches the output here.

To summarise what I'd like to see before merge: revert the restricted-list hash to `a146ff14…`, drop the `.gitignore` commit. The accommodations URL change is good to go, and we'll sequence its hash with the sheet update.

pudo:
https://github.com/opensanctions/opensanctions/pull/5290

PR #5287, opened 2026-08-04
[zavod] validator: warn when Ownership.asset is not an Asset (#2550)

### LINKED ISSUE #2550: Create validator that validates whether Ownership.asset is an Asset
https://www.opensanctions.org/entities/lei-549300DTNTPKNKGJ7P43/ is owned by https://www.opensanctions.org/entities/NK-f7H2E3mdvYsFdeFfxxTvWw/, but it's an Organization

dennismdejong:
Fixes #2550

## Change
- New `OwnershipAssetValidator`: resolves each `Ownership.asset` id via the view and warns (no abort) when the target's schema is not `Asset`.
- Registered in `VALIDATORS`.

## Tests
- `test_ownership_asset_validator` added (Organization as a non-Asset target, matching the issue's case).
- `pytest zavod/tests/`: 294 passed; mypy --strict clean.

Note: `Company` is a subtype of `Asset` in the FTM model, so Companies are correctly not flagged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,9 @@ apicache-py3
 # Emacs
 *~
 task-prompts/
+
+# Superpowers implementation plans
+docs/superpowers/plans/
+
+# Git worktrees
+.worktrees/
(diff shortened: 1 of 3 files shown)


pudo:
see https://github.com/opensanctions/opensanctions/pull/5288

PR #5294, opened 2026-08-04
Keep PEP status out of the position lookup

nvmbrasserie:
Stacked on #5233 — base is `agent-skill-upd`, so merge that one first.

Two commits. Net change to `zavod/docs`: one reflowed bullet in `peps.md`, plus two reverts. Everything else is skill files and one docstring.

### 1. Review corrections to the PEP guidance

**The one real defect.** Pattern C's `position` lookup carried a `value: null` option for `Territorial delegation`. Territorial delegates are sworn House members and their holders are PEPs, so the example dropped a real position in the YAML — before `categorise`, so it never reaches the review database and nobody can rule on it in the UI. It was also wrong for the source it was modelled on: the Congress API's `chamber` only yields the two enumerated values. Dropped the option, and noted in the skill that a doubtful position goes in with `default_is_pep=None` rather than being skipped — the shape [`cl_info_probidad`](https://github.com/opensanctions/opensanctions/blob/main/datasets/cl/info_probidad/crawler.py#L89) uses, where the lookup supplies `default_is_pep` per label (including `null`) and the UI still decides.

The rest are corrections, not new guidance:

- The bullet on lookups carrying `topics` read against the rule above it, that topics are omitted for positions read out of the source data. Now says a hand-enumerated label is the crawler naming the position.
- The multi-term example called `crawl_term`, which it never defined.
- That example claimed ISO strings "compare correctly here, including year-only term bounds". They don't — a year-only bound sorts below any full date in the same year. But the consequence is negligible: `earliest_term_start(["gov.national","gov.legislative"])` returns 1996-08-11, ten years past the 20-year after-office threshold, so only a term ending in the cutoff year differs and `make_occupancy` rejects its holders anyway. Dropped the claim; left the comparison alone.
- `ci_test: false` cited a "2-minute CI budget". No per-dataset timeout is configured anywhere in `.github/workflows/`; replaced with the grounds `metadata.md` gives.
- `make_occupancy` did not document `period_start`, `period_end`, `election_date`, `categorisation` or `key_prefix`.
- Registered `/crawler-constants-to-yml` in CLAUDE.md's task-guide list.

### 2. Revert the total-count-guard anti-pattern — scope reduction

#5233 added a new prohibition to `strict_interpretation.md`: no hand-rolled guard on the crawl's total record count, because "the metadata `assertions` block owns totals". Reverted, along with the skill bullet enforcing it, for a maintainer to decide separately.

The objection is substantive as well as procedural. Metadata assertions are checked on export (`zavod run`, or `zavod validate` while developing), not by `zavod crawl`. Deferring the total to them means a crawl that emits nothing completes quietly — the outcome that doc's own opening line exists to prevent.

### Deliberately not changed

Earlier drafts of this PR overreached into the docs and were withdrawn:

- **How `value: null` behaves.** #5233 documents a convention (`value: null` + no `name` marks a label that isn't a position). A draft here added "PEP status is decided by `categorise`, never by the lookup" — but `cl_info_probidad` shows a lookup legitimately carrying `is_pep: true|false|null` as the `default_is_pep`, and `value: null` mechanically just clears `value`, it doesn't mean "not a position". The absolute was wrong, so nothing is stated about those mechanics. The fix is confined to the bad example.
- **When a lookup miss should crash vs skip.** `strict_interpretation.md`'s "increasing loudness" ladder already sanctions `warn_unmatched=True` that "warns and skips the record". Drafts here required failing on a miss, in prose and in the example; both contradicted that ladder and are gone.
- The pre-existing topics rule in "Creating positions" is untouched; the qualifier lives in the new section that needed it.
- `.claude/skills/crawler-constants-to-yml/SKILL.md` is untouched.

`mypy --strict` is clean on `positions.py`; the change there is docstring text. `pytest` isn't installed in the local venv, so CI's test run is the check.


nvmbrasserie:
Withdrawn — the review of #5233 found nothing needing a change in `zavod/docs/peps.md` or `zavod/zavod/helpers/positions.py`. The branch is reset to base, so this PR carries no diff (the title reflects an earlier, abandoned draft). Findings are summarised on #5233.

PR #5300, opened 2026-08-04
[wd_categories] Cover Lithuanian PEP offices from lt.wikipedia

nvmbrasserie:
lt_pep_declarations died when VTEK put its register of private interests behind API authentication. Add lt.wikipedia categories for the Lithuanian offices that are PEP positions by definition - president, prime minister and deputies, ministers and vice-ministers, Seimas members and its speaker, LT members of the European Parliament, the courts of final instance, the prosecutor general and deputies, the central bank governor, the chief of defence and ambassadors - and assert positions wherever the category is specific to a single office.

### DIFF
--- a/datasets/_wikidata/categories/wd_categories.yml
+++ b/datasets/_wikidata/categories/wd_categories.yml
@@ -2065,13 +2065,6 @@ config:
       language: fi
 
     # country: Lithuania
-    #
-    # Replaces the coverage of lt_pep_declarations, whose source (the VTEK
-    # register of private interests) started requiring API authentication in
-    # April 2025. lt.wikipedia categorises Lithuanian office-holders by office
-    # much more completely than en.wikipedia. A position is only asserted where
-    # the category is specific to one office; ambassadors and mayors vary by
-    # posting or municipality, so those only get the topic.
     - category: Lietuvos_politikai
       language: lt
       country: lt
--- a/datasets/_wikidata/categories/wd_categories.yml
+++ b/datasets/_wikidata/categories/wd_categories.yml
@@ -2064,17 +2064,206 @@ config:
     - category: Suomen_hallinto
       language: fi
 
+    # country: Lithuania
+    #
+    # Replaces the coverage of lt_pep_declarations, whose source (the VTEK
+    # register of private interests) started requiring API authentication in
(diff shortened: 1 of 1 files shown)


datasets/_wikidata/categories/wd_categories.yml
```diff
@@ -2064,17 +2064,206 @@ config:
...
+      language: lt
+      country: lt
+      topic: role.pep
+      position:
+        name: Member of the European Parliament
+        wikidata_id: Q27169
+        country: eu
+        topics:
+          - gov.igo
+          - gov.legislative
```
pudo: For these we don't need to make the position (see `eu_meps`)

datasets/_wikidata/categories/wd_categories.yml
```diff
@@ -2064,17 +2064,206 @@ config:
...
+    - category: Lietuvos_vicepremjerai
+      language: lt
+      country: lt
+      topic: role.pep
+      position:
+        name: Deputy Prime Minister of Lithuania
+        country: lt
+        topics:
+          - gov.executive
+          - gov.national
```
pudo: I would also do w/o position

pudo (commented):
Thank you very much for thinking of this!

nvmbrasserie:
Gone: the positions on the deputy PM, ministers, vice-ministers and MEP categories — `wd_peps` models the LT ministerial offices per ministry and `eu_meps` owns the EP seat, so these were only a vaguer second position per holder. All four still mark their members as PEPs. The `Lietuvos_susisiekimo_viceministrai` negcat went with them; it only existed to keep wikipedia's miscategorised vice-ministers out of that ministers position.

Kept: the QID-backed offices (president, PM, Seimas member and speaker, prosecutor general, bank governor, chief of defence), where Wikidata's P39 coverage is patchy — only 3 of 10 bank heads and 3 of 17 chiefs of defence carry the claim — and the judiciary plus deputy prosecutor general, where no Wikidata position item exists at all, so `wd_peps` can never supply one (none of the 31 constitutional or 72 supreme court judges have a P39 for their court). Glad to drop those too if you'd rather.

jbothma (approved):
pretty cool! did claude do a lot of the research here? Looks pretty effective!

PR #5319, opened 2026-08-05
Pass a two-digit year base where sources publish two-digit years

### LINKED ISSUE #4930: helpers: no guardrail against %y date formats parsing birth dates into the future

### LINKED ISSUE #4940: [be_fod_sanctions] Two-digit birth years pivot into the future — 5 sanctioned persons published with DOBs like 2068-07-16

### LINKED ISSUE #4974: 29 datasets configure %y date formats — two-digit years can pivot into the wrong century
Follow-up to #4930. `strptime`'s `%y` pivot maps two-digit years 00–68 to 20xx and 69–99 to 19xx, so any pre-1969 date parses a century too late ("16-07-68" with `%d-%m-%y` → 2068-07-16). The birthDate guard from #4930 rejects the future-birth-date symptom at cleaning time, but every other date property (sanction start/end dates, listing dates, etc.) parsed through a `%y` format remains exposed, and the wrongly-pivoted value is stored as fully valid.

A repo-wide scan (2026-07-17) found 29 datasets configuring `%y` in `dates.formats`:

- `datasets/_global/afdb_sanctions` — `%d-%b-%y`
- `datasets/_global/c4ads_xinjiang` — `%m/%d/%y`
- `datasets/_global/c4ads_xinjiang_mining` — `%m/%d/%y`
- `datasets/_global/icij_offshoreleaks` — `%d/%m/%y`
- `datasets/_global/un_1718_vessels` — `%d-%b-%y`
- `datasets/_global/un_ga_protocol` — `%d-%b-%y`
- `datasets/au/dfat_sanctions` — `%d/%m/%y`
- `datasets/be/fod_sanctions` — `%d-%m-%y` (actively corrupting birth dates, tracked in #4940)
- `datasets/ca/dfatd` (ca_dfatd_sema_sanctions) — `%b-%y`
- `datasets/cn/sanctions` — `%d.%m.%y`
- `datasets/cu/inventario_legislatura` — `%d/%m/%y`
- `datasets/il/mod_crypto` — `%m.%d.%y`
- `datasets/il/wmd_sanctions` — `%d %m %y`
- `datasets/in/nse_debarred` — `REVOKED ON %d.%m.%y` etc.
- `datasets/lv/fiu_sanctions` — `%d/%m/%y`
- `datasets/tr/fcib` — `%m/%d/%y`
- `datasets/us/al/med_exclusions` — `%m/%d/%y`
- `datasets/us/bis_antiboycott` — `%m/%d/%y`
- `datasets/us/bis_oac` — `%b-%y`
- `datasets/us/de/med_exclusions` — `%m/%d/%y`
- `datasets/us/fdic_failed_banks` — `%d-%b-%y`
- `datasets/us/hi/med_exclusions` — `%m/%d/%y`
- `datasets/us/mo/med_exclusions` — `%m/%d/%y`
- `datasets/us/ms/med_exclusions` — `%m.%d.%y`, `%m/%d/%y`
- `datasets/us/nd/med_exclusions` — `%m/%d/%y`
- `datasets/us/oh/med_exclusions` — `%m/%d/%y`
- `datasets/us/sc/med_exclusions` — `%m/%d/%y`
- `datasets/us/wv/med_exclusions` — `%m/%d/%y`
- `datasets/ve/asamblea_nacional` — `%d/%m/%y`, `%d – %m – %y`

## Candidate approaches (undecided)

Whichever pivot scheme is chosen, a value that is wrong but lands *inside* the accepted range is undetectable — that is intrinsic to two-digit years. The framework options below replace strptime's arbitrary 1969 pivot with a declared or derived one; the cleaning-time birthDate guard from #4930 stays as a backstop in all variants.

Mechanically, all variants hook into `extract_date`: `prefixdate.parse_formats` is a trivial first-match loop, so zavod can run the loop itself, know which format matched, and rewrite the year when that format contains `%y`.

### Option A: `base_century` in the dataset dates config

```yaml
dates:
  formats: ["%d-%m-%y"]
  base_century: 1900
```

Two-digit years are mapped into the 100-year window starting at `base_century` (`1900` → 1900–1999; `2000` → 2000–2099; intermediate floors like `1930` → 1930–2029 handle sources whose values span centuries). Could be made **required** whenever a configured format contains `%y`, failing dataset validation otherwise — no dataset could silently inherit the 1969 pivot, and this issue becomes a mechanical config rollout over the 29 datasets above.

- Explicit and deterministic: the maintainer looks at the source once and declares intent; the same input always parses the same way regardless of when the crawler runs.
- Works for future-leaning fields too (`base_century: 2000` for debarment end dates).
- One dataset-wide setting: a dataset mixing 19xx birth dates with 20xx end dates in `%y` formats needs a window floor that covers both (e.g. 1930), or crawler-level handling.

### Option B: `assume_past` — pivot at the current date

Declared per dataset (`dates: assume_past: true`) or inferred per property (`birthDate`/`deathDate` are always past): when a `%y` format produces a date in the future, subtract 100 years, so two-digit years map into the trailing 100-year window ending today.

- Zero judgment required for always-past fields: the most recent non-future reading is the only valid one, so this is arguably correct-by-construction for birth dates.
- Config-free in the per-property variant; fixes every `%y` dataset's birth dates at once.
- Not deterministic across time: the same source string parses differently depending on the run date (slow reparse drift, e.g. "27" flips from 1927 to 2027 when the year 2027 arrives).
- Unusable for fields that can legitimately be in the future (sanction/debarment end dates), so the dataset-wide variant is wrong for mixed datasets — several med-exclusion lists above have exactly that shape.

### Option C: status quo — per-dataset crawler fixes

Keep the framework as-is; each dataset replaces its `%y` format with explicit century handling in the crawler where exposure is real, or documents why the field can never predate 1969 (e.g. a listing date on a list founded in the 2000s).

Many of the exposed fields are exactly the med-exclusion / debarment lists where a wrongly-pivoted end date would also interact with `h.is_active` (#4929).

leonhandreke:
`strptime` reads two-digit years 00-68 as 20xx and 69-99 as 19xx, so a date outside that window parses into the wrong century and is stored as fully valid. `be_fod_sanctions` publishes birth dates like 2068-07-16 today (https://github.com/opensanctions/opensanctions/issues/4940).

- `h.extract_date`, `h.apply_date` and `h.apply_dates` take a `two_digit_year_base`, the earliest year a two-digit year may denote.
- Parsing a `%y` format without a base year raises a dataset issue.
- `h.TWO_DIGIT_BIRTH_YEAR_BASE` is 100 years ago, because birth dates are always before now. Case dates use the earliest possible event year.

The century logic itself lives in prefixdate: https://github.com/pudo/prefixdate/pull/3. **CI fails until that is released** — this branch pins `prefixdate == 0.6.0`, which does not exist yet.

## Which datasets need it

29 datasets configure a `%y` format. For each one I streamed the latest published `statements.pack`, re-ran the dataset's own format list over every date value, and recorded which format actually matched.

**Found** is measured from that pass: counts are published values matching a `%y` format, and "never" means no published value of that property matches one.

**Change** separates the two kinds of claim. *Measured* is the earliest published value of that property parsed by a **four-digit** year format — the only values that can evidence how far back the property reaches, since a two-digit value already carries the wrong century. *Assumed* is my reasoning for the base year, which is domain knowledge and not derived from the data; it is what a reviewer needs to check.

### Base year passed (17)

| Dataset | Found | Change |
|---|---|---|
| `c4ads_xinjiang` | 2707 of 2707 `incorporationDate` are `%m/%d/%y` | `incorporationDate` after 1950. No four-digit value exists to measure against. Assumed: the XPCC was founded in 1954, so no company of its is older. |
| `c4ads_xinjiang_mining` | 7031 of 7031 `endDate` are `%m/%d/%y` | `endDate` after 1990. No four-digit value exists to measure against. Assumed: no mining permit is older, and permits expire in the future. |
| `us_fdic_failed_banks` | 575 of 575 `dissolutionDate` are `%d-%b-%y` | `dissolutionDate` after 2000. No four-digit value exists to measure against. Assumed: the FDIC list covers bank failures since October 2000. |
| `un_ga_protocol` | 521 of 521 appointment dates are `%d-%b-%y` | occupancy `start_date` after 1945. No four-digit value exists to measure against. Assumed: the UN was founded in 1945, so no office-holder was appointed earlier. |
| `us_bis_antiboycott` | 100 of 100 `listingDate` are `%m/%d/%y` | `listingDate` after 1977. No four-digit value exists to measure against. Assumed: the U.S. antiboycott regulations came into force in 1977. |
| `un_1718_vessels` | 57 of 59 `startDate` are `%d-%b-%y` | `startDate` after 2000. No four-digit value exists to measure against. Assumed: the 1718 Committee was established in 2006. |
| `us_hi_med_exclusions` | 206 of 206 `startDate` and 12 of 15 `endDate` are `%m/%d/%y` | both after 1977. Measured: earliest four-digit `endDate` is 2019-06-06, and `startDate` has none. Assumed: exclusions cannot predate the 1977 Medicare-Medicaid Anti-Fraud and Abuse Amendments, and can end in the future. |
| `be_fod_sanctions` | 137 `birthDate` and 137 `listingDate` are `%d-%m-%y`; the rest ISO | `birthDate` after 115 years ago (`h.TWO_DIGIT_BIRTH_YEAR_BASE - 15`), `listingDate` after 1990. Measured: four-digit `birthDate` values run 1925-01-25 to 2004-05-20, so the plain 100-year window starts too late; earliest `listingDate` is 2002-05-29. Assumed: a sanctioned person is old enough to have acted, so 15 years of extra room is safe, and the oldest measures implement UNSCR 1267 (1999). |
| `cn_sanctions` | 156 of 557 `startDate` and 1 of 3 `endDate` are `%d.%m.%y` | both after 2010. Measured: earliest four-digit `startDate` is 2019-12-02, earliest `endDate` 2026-04-24. Assumed: China started to publish counter-sanctions in 2019. |
| `in_nse_debarred` | 330 of 6318 `endDate` match 9 different `REVOKED … %y` formats; `date` never does | `endDate` only, after 1990. Measured: earliest four-digit `endDate` is 2012-04-09. Assumed: no order is older, and a debarment can end in the future. |
| `il_wmd_sanctions` | 121 of 276 `modifiedAt` are `%d %m %y`; `birthDate` and `startDate` never | `modifiedAt` only, after 2000. Measured: earliest four-digit `modifiedAt` is 2018-03-12. Assumed: no designation on the list is older. |
| `us_mo_med_exclusions` | 14 of 263 `startDate` are `%m/%d/%y`; `date` never | `startDate` only, after 1977. Measured: earliest four-digit `startDate` is 2011-02-03. Assumed: the exclusion cannot predate the 1977 Amendments. |
| `us_wv_med_exclusions` | 3 of 165 `startDate` are `%m/%d/%y` | `startDate` only, after 1977. Measured: earliest four-digit `startDate` is 2010-04-20. Assumed: the exclusion cannot predate the 1977 Amendments. |
| `icij_offshoreleaks` | 566 relationship `startDate` and 14 `endDate` are `%d/%m/%y`; incorporation and dissolution dates never, over all 34.2M statements | relationship dates only, after 1950. Not measured — I skipped the second 34.2M-statement pass. Assumed: the oldest officer relationships in the leaked registers are from the 1950s. |
| `ca_dfatd_sema_sanctions` | 1 `birthDate` is `%b-%y` (`Sep-99`); `listingDate` and `buildDate` never | `birthDate` only, after 115 years ago (`h.TWO_DIGIT_BIRTH_YEAR_BASE - 15`). Measured: four-digit `birthDate` values run 1924-02-21 to 2004-08-31, so the plain 100-year window starts too late. Assumed: a sanctioned person is old enough to have acted, so 15 years of extra room is safe. |
| `il_mod_crypto` | 1 `birthDate` is `%m.%d.%y` (`01.02.92`); sanction dates never | `birthDate` only, after 100 years ago (`h.TWO_DIGIT_BIRTH_YEAR_BASE`). Measured: four-digit `birthDate` values run 1952-11-08 to 2003-07-26, inside the window. Assumed: a birth date is always before now. |
| `us_de_med_exclusions` | 1 `startDate` is `%m/%d/%y`; `endDate` never | `startDate` only, after 1977. Measured: earliest four-digit `startDate` is 2007-01-19. Assumed: the exclusion cannot predate the 1977 Amendments. |

### No change (12)

The `%y` format never matches a published value in these, so passing a base year would be speculative. Their `%y` config is dead and worth removing separately.

| Dataset | Found |
|---|---|
| `us_sc_med_exclusions` | all 1345 `startDate` are ISO. Inferred, not measured: the XLSX cells reach the crawler as datetimes, which is why |
| `us_al_med_exclusions` | 2079 dates are ISO or `%m/%d/%Y` |
| `us_ms_med_exclusions` | 312 dates across 3 properties are ISO |
| `us_nd_med_exclusions` | 193 `startDate` are ISO or `%m/%d/%Y` |
| `us_oh_med_exclusions` | 3858 `birthDate` and `startDate` are ISO or `%m/%d/%Y` |
| `afdb_sanctions` | 2543 `startDate` and `endDate` are ISO |
| `au_dfat_sanctions` | 2957 `birthDate` match 8 other formats; 7714 listing dates match none, their raw values being `2026-04-14T00:00:00` (inferred, not measured: they reach the crawler as datetimes) |
| `tr_fcib` | 3732 dates match `%d.%m.%Y`, ISO or `%d %m %Y` |
| `lv_fiu_sanctions` | 186 dates are ISO or `%d.%m.%Y.` |
| `cu_inventario_legislatura` | 470 `birthDate` are ISO |
| `us_bis_oac` | all 211 `listingDate` are `%B %Y`; the configured `%b-%y` never matches |
| `ve_asamblea_nacional` | 6 `birthDate` match `%d-%m-%Y`, `%d/%m/%Y` or `%d %b %Y` |

### The birth-date window

`h.TWO_DIGIT_BIRTH_YEAR_BASE` is `RUN_TIME.year - 100`, a window of 1926-2025 today. Measured against the published data that is too late at the old end for two sanctions lists: four-digit birth dates run 1925-01-25 to 2004-05-20 in `be_fod_sanctions` and 1924-02-21 to 2004-08-31 in `ca_dfatd_sema_sanctions`. A two-digit `24` or `25` there would read as 2024 or 2025.

Both now pass `h.TWO_DIGIT_BIRTH_YEAR_BASE - 15`, a window of 1911-2010. A person on a sanctions list is old enough to have acted, so giving up the last 15 years costs nothing and covers the measured range with room at both ends. `il_mod_crypto` keeps the plain constant, its measured range being 1952-2003.

This changes no published value today: the 137 two-digit birth dates on the Belgian list run 1964 to 2006 and resolve identically under either window. It does move the risk to the other end: a two-digit birth year after 2010 on these two lists would now read as 1911 onwards. That is the trade — you cannot be on a sanctions list at that age — but it is where the window now runs out.

Closes https://github.com/opensanctions/opensanctions/issues/4930
Closes https://github.com/opensanctions/opensanctions/issues/4940
Closes https://github.com/opensanctions/opensanctions/issues/4974

### DIFF
--- a/datasets/_global/un_ga_protocol/crawler.py
+++ b/datasets/_global/un_ga_protocol/crawler.py
@@ -72,15 +72,13 @@ def crawl(context: Context) -> None:
                 topics=["gov.national"],
             )
             start_date = start_date if len(start_date) < 18 else None
-            if start_date is not None:
-                start_date = h.extract_date(
-                    context.dataset,
-                    start_date,
-                    # The UN was founded in 1945, so no appointment is older.
-                    two_digit_year_base=1945,
-                )[0]
             occupancy = h.make_occupancy(
-                context, entity, position, start_date=start_date
+                context,
+                entity,
+                position,
+                start_date=start_date,
+                # The UN was founded in 1945, so no appointment is older.
+                two_digit_year_base=1945,
             )
 
             # entity.add("date_of_appointment", )
--- a/datasets/_global/un_ga_protocol/crawler.py
+++ b/datasets/_global/un_ga_protocol/crawler.py
@@ -1,5 +1,4 @@
 from urllib.parse import urljoin
-
 from rigour.mime.types import PDF
 from rigour.names import remove_person_prefixes
 
--- a/datasets/_global/un_ga_protocol/crawler.py
+++ b/datasets/_global/un_ga_protocol/crawler.py
@@ -1,4 +1,5 @@
 from urllib.parse import urljoin
+
 from rigour.mime.types import PDF
 from rigour.names import remove_person_prefixes
 
@@ -72,6 +73,13 @@ def crawl(context: Context) -> None:
                 topics=["gov.national"],
             )
             start_date = start_date if len(start_date) < 18 else None
+            if start_date is not None:
+                start_date = h.extract_date(
+                    context.dataset,
+                    start_date,
+                    # The UN was founded in 1945, so no appointment is older.
+                    two_digit_year_base=1945,
+                )[0]
             occupancy = h.make_occupancy(
                 context, entity, position, start_date=start_date
             )
--- a/datasets/_global/un_ga_protocol/crawler.py
+++ b/datasets/_global/un_ga_protocol/crawler.py
@@ -77,8 +77,9 @@ def crawl(context: Context) -> None:
                 entity,
                 position,
                 start_date=start_date,
-                # The UN was founded in 1945, so no appointment is older.
-                two_digit_year_base=1945,
+                # As of 2026, longest current head started tenure 1982 (Teodoro Mbasogo).
+                # https://www.guinnessworldrecords.com/world-records/65343-longest-serving-president-current
+                two_digit_year_base=1980,
             )
 
             # entity.add("date_of_appointment", )
--- a/zavod/zavod/helpers/__init__.py
+++ b/zavod/zavod/helpers/__init__.py
@@ -50,6 +50,7 @@ from zavod.helpers.change import (
 )
 from zavod.helpers.crypto import extract_cryptos
 from zavod.helpers.dates import (
+    TWO_DIGIT_BIRTH_YEAR_BASE,
     apply_date,
     apply_dates,
     backdate,
@@ -132,6 +133,7 @@ __all__ = [
     "replace_months",
     "backdate",
     "within_max_age",
+    "TWO_DIGIT_BIRTH_YEAR_BASE",
     "apply_number",
     "convert_excel_cell",
     "convert_excel_date",
--- a/datasets/_global/c4ads_xinjiang/crawler.py
+++ b/datasets/_global/c4ads_xinjiang/crawler.py
@@ -1,4 +1,5 @@
 import csv
+
 from zavod import Context
 from zavod import helpers as h
 from zavod.shed.internal_data import fetch_internal_data
--- a/datasets/_global/c4ads_xinjiang/crawler.py
+++ b/datasets/_global/c4ads_xinjiang/crawler.py
@@ -1,5 +1,4 @@
 import csv
-
 from zavod import Context
 from zavod import helpers as h
 from zavod.shed.internal_data import fetch_internal_data
@@ -18,7 +17,13 @@ def crawl(context: Context) -> None:
             entity.add("name", name_zho, lang="zho")
             entity.add("name", row.pop("Company_Name_English"), lang="eng")
             entity.add("country", "cn")
-            h.apply_date(entity, "incorporationDate", row.pop("Date_of_Establishment"))
+            h.apply_date(
+                entity,
+                "incorporationDate",
+                row.pop("Date_of_Establishment"),
+                # The XPCC was founded in 1954, so no company is older.
+                two_digit_year_base=1950,
+            )
             entity.add("sector", row.pop("Industry"), lang="zho")
             entity.add("address", addr_zho, lang="zho")
             entity.add("topics", "export.risk")
--- a/datasets/_global/c4ads_xinjiang_mining/crawler.py
+++ b/datasets/_global/c4ads_xinjiang_mining/crawler.py
@@ -27,7 +27,13 @@ def crawl_row(context: Context, row: dict[str, str]) -> None:
     own.add("owner", owner.id)
     own.add("asset", entity.id)
     own.add("recordId", row.pop("Permit_Number"))
-    h.apply_date(own, "endDate", row.pop("Permit_Expiration_Date"))
+    h.apply_date(
+        own,
+        "endDate",
+        row.pop("Permit_Expiration_Date"),
+        # No permit on the list predates 1990, and permits expire in the future.
+        two_digit_year_base=1990,
+    )
 
     context.emit(entity)
     context.emit(owner)
--- a/datasets/_global/icij_offshoreleaks/crawler.py
+++ b/datasets/_global/icij_offshoreleaks/crawler.py
@@ -9,6 +9,8 @@ from followthemoney import model
 from zavod import Context, Entity
 from zavod import helpers as h
 
+# The oldest officer relationships in the leaked registers are from the 1950s.
+TWO_DIGIT_RELATIONSHIP_YEAR_BASE = 1950
 SCHEMATA: dict[str, str] = {}
 ADDRESSES_FULL: dict[str, list[str | None]] = {}
 ADDRESSES_COUNTRIES: dict[str, list[str] | None] = {}
@@ -215,8 +217,18 @@ def make_row_relationship(context: Context, row: dict[str, str | None]) -> None:
     if res.schema is not None:
         rel = context.make(res.schema)
         rel.id = context.make_slug(_start, _end, link)
-        h.apply_date(rel, "startDate", row.pop("start_date"))
-        h.apply_date(rel, "endDate", row.pop("end_date"))
+        h.apply_date(
+            rel,
+            "startDate",
+            row.pop("start_date"),
+            two_digit_year_base=TWO_DIGIT_RELATIONSHIP_YEAR_BASE,
+        )
+        h.apply_date(
+            rel,
+            "endDate",
+            row.pop("end_date"),
+            two_digit_year_base=TWO_DIGIT_RELATIONSHIP_YEAR_BASE,
+        )
         rel.add(res.status, row.pop("status"))
         rel.add(res.link, link)
         rel.add("publisher", source_id)
--- a/datasets/_global/un_1718_vessels/crawler.py
+++ b/datasets/_global/un_1718_vessels/crawler.py
@@ -58,7 +58,13 @@ def crawl(context: Context) -> None:
             vessel.add("type", row.pop("type"))
             # Create the sanction
             sanction = h.make_sanction(context, vessel, program_key=PROGRAM_KEY)
-            h.apply_date(sanction, "startDate", row.pop("date_of_designation"))
+            h.apply_date(
+                sanction,
+                "startDate",
+                row.pop("date_of_designation"),
+                # The 1718 Committee was established in 2006.
+                two_digit_year_base=2000,
+            )
             for program in PROGRAMS:
                 value = row.pop(program).strip()
                 if value and value.lower() != "na":
--- a/datasets/be/fod_sanctions/crawler.py
+++ b/datasets/be/fod_sanctions/crawler.py
@@ -1,5 +1,4 @@
 from csv import DictReader
-
 from rigour.mime.types import CSV
 
 from zavod import Context
@@ -70,7 +69,15 @@ def crawl_row(
         entity.add("gender", row.pop("Gender"))
         entity.add("birthPlace", row.pop("Birth place"))
         entity.add("country", row.pop("Birth country"))
-        h.apply_dates(entity, "birthDate", row.pop("Birth date"))
+        h.apply_dates(
+            entity,
+            "birthDate",
+            row.pop("Birth date"),
+            # The list holds birth dates from 1925, before the default 100-year
+            # window. A sanctioned person is old enough to have acted, so their
+            # birth date can be another 15 years back.
+            two_digit_year_base=h.TWO_DIGIT_BIRTH_YEAR_BASE - 15,
+        )
         entity.add("position", row.pop("Function"))
     else:
         # There's an organization with the following Firstname
@@ -81,7 +88,13 @@ def crawl_row(
 
     sanction = h.make_sanction(context, entity)
     for listing_date in row.pop("Publication date"):
-        h.apply_date(sanction, "listingDate", listing_date)
+        h.apply_date(
+            sanction,
+            "listingDate",
+            listing_date,
+            # The oldest measures implement UN Security Council Resolution 1267 (1999).
+            two_digit_year_base=1990,
+        )
     for embargo in row.pop("Embargos"):
         program_id = h.lookup_sanction_program_key(context, embargo)
         sanction.add("programId", program_id)
--- a/datasets/be/fod_sanctions/crawler.py
+++ b/datasets/be/fod_sanctions/crawler.py
@@ -1,4 +1,5 @@
 from csv import DictReader
+
 from rigour.mime.types import CSV
 
 from zavod import Context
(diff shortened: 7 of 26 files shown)


pudo:
`un_ga_protocol` pre-parses the date before handing it to `h.make_occupancy`, but the helper hashes the raw `start_date` string into the occupancy ID (`zavod/zavod/helpers/positions.py:185-197`):

```python
    parts = [
        person.id,
        position.id,
        "started",
        start_date or "unknown",
        ...
    ]
    occupancy.id = context.make_id(*parts, hash_prefix=key_prefix)
```

So the ID changes for every occupancy whose date parses — per the table above that is all 521 appointment dates, i.e. effectively the whole dataset: `started 16-Jul-68` becomes `started 1968-07-16`.

It also costs the source form of the value. `make_occupancy` calls `apply_date` itself, so the `original_value` it records is now the ISO string rather than what the PDF said.

Both go away if the helper takes the argument and the crawler keeps passing the raw string:

```python
def make_occupancy(
    ...
    two_digit_year_base: int | None = None,
) -> Entity | None:
    ...
    h.apply_date(
        occupancy, "startDate", start_date, two_digit_year_base=two_digit_year_base
    )
```

```python
occupancy = h.make_occupancy(
    context,
    entity,
    position,
    start_date=start_date,
    # The UN was founded in 1945, so no appointment is older.
    two_digit_year_base=1945,
)
```

More generally this is the one place the new argument turns out to be viral: any helper that wraps `apply_date` needs it threaded through to be usable. `make_occupancy` is the only such caller in this PR's path, but it is worth knowing that is the shape of the cost.

Unrelated, for the record: the body still says CI fails until prefixdate is released. 0.6.0 is on PyPI and all checks are green on `c2ea3ec`.

zavod/zavod/helpers/__init__.py
```diff
@@ -132,6 +133,7 @@ def crawl(context: Context) -> None:
     "replace_months",
     "backdate",
     "within_max_age",
+    "TWO_DIGIT_BIRTH_YEAR_BASE",
```
jbothma, line 136: I'm toying with the idea of

```python
# zavod.constants exist, but isn't used like this yet
from zavod import constants as c

...

h.apply_date(entity, prop, date, base_year=c.TWO_DIGIT_BIRTH_YEAR_BASE)
```

But perhaps accessing via `h.` leads you more closely to its context - `zavod.helpers.dates` - so I think what you have is best.

datasets/_global/un_ga_protocol/crawler.py
```diff
@@ -73,7 +73,12 @@ def crawl(context: Context) -> None:
             )
             start_date = start_date if len(start_date) < 18 else None
             occupancy = h.make_occupancy(
-                context, entity, position, start_date=start_date
+                context,
+                entity,
+                position,
+                start_date=start_date,
+                # The UN was founded in 1945, so no appointment is older.
+                two_digit_year_base=1945,
```
jbothma: nit: I think the logic here is a bit wrong - I think this dataset is merely when someone took the head of state/gov appointment - it's not a UN role. So I think this is the date and logic this is looking for https://www.guinnessworldrecords.com/world-records/65343-longest-serving-president-current?

jbothma (approved):
fun!

datasets/_global/un_ga_protocol/crawler.py
pudo: So we put 1980?

zavod/zavod/helpers/__init__.py
pudo: I like `from zavod import constants as c`. That's also easy to refactor en masse if we choose to.

pudo:
@jbothma Would you perhaps apply the changes so we can merge? I'm just worried that otherwise it goes out of sync and becomes an evergreen :)

zavod/zavod/helpers/__init__.py
jbothma: I'm not gonna do the constants thing now, since TWO_DIGIT_BIRTH_YEAR_BASE depends on settings.RUN_TIME, and it's a bit weird for constants to depend on settings, rather than the other way around. There's definitely some room for improvement in these run-time variables though

PR #5371, opened 2026-08-07
fixes 07 08

nvmbrasserie:


datasets/_wikidata/categories/wd_categories.yml
```diff
@@ -172,6 +172,32 @@ lookups:
         value: 1951-06
       - match: "1942-02-29"
         value: 1942-02
+      - match: "1585-02-30"
+        value: 1585-02
+      - match: "1858-06-31"
```
nvmbrasserie: do we expect these dates? we haven't had these before

herrhellana: it seems to be a `nomenklatura` issue

herrhellana: proposed fix: https://github.com/opensanctions/opensanctions/pull/5373/changes

Issue #5400, opened 2026-08-10
[us_ar_med_exclusions] review / fix name and property value validity


brrttwrks:
Names appear to be incorrect - person name == "American". 

Review source data and crawler to assess and fix name validity.

PR #5406, opened 2026-08-11
[us_ar_med_exclusions] fix misclassifying legal entities as persons

brrttwrks:
The provider-name lookup table was only ever consulted when the value failed a word-shape heuristic, so entries like "American, All" (a mangled form of "All American Medical") passed the heuristic and were forced into Person regardless of a matching lookup entry. Check the lookup first so it can override the heuristic, and add entries for the other business names found misclassified the same way.

This closes ticket #5400


nvmbrasserie (approved):

nvmbrasserie:
`changed=127 dataset=us_ar_med_exclusions entities=6631 statements=38705`
not disruptive

PR #5409, opened 2026-08-11
Always use the last successful version when backfilling without specified version

jbothma:
Fixes https://github.com/opensanctions/operations/issues/2762

When fetching an artifact from the archive without specifying a version, the last successful version is used, rather than merely the latest one with the same resource name.

Moves `set_last_successful_version` from `cli.etl.run` to `publish.publish_dataset` so that `zavod run` and `zavod publish` both update `last_successful`.

Doesn't require a specified version to be successful. That could be future work.

This fixes the issue where the catalog shows metadata for the latest dataset even if that was a failed run with no resources.

In the end, this felt like an important part that should be included in https://github.com/opensanctions/opensanctions/pull/4873 rather than a band-aid that wouldn't be needed if we did #4873. So let's do the small simple thing and then update and finish the bigger one.

### DIFF
--- a/zavod/zavod/archive/__init__.py
+++ b/zavod/zavod/archive/__init__.py
@@ -210,23 +210,38 @@ def get_artifact_object(
     dataset_name: str, resource: str, version: str | None = None
 ) -> ArchiveObject | None:
     backend = get_archive_backend()
+    if version is None:
+        versions_data = get_versions_data(dataset_name)
+        if versions_data is not None:
+            history = VersionHistory.from_json(versions_data)
+            version = history.last_successful.id if history.last_successful else None
+
     if version is not None:
         name = f"{ARTIFACTS}/{dataset_name}/{version}/{resource}"
         object = backend.get_object(name)
         if object.exists():
             return object
-    else:
-        for v in iter_dataset_versions(dataset_name):
-            name = f"{ARTIFACTS}/{dataset_name}/{v.id}/{resource}"
-            object = backend.get_object(name)
-            if object.exists():
-                return object
+
+    # Temporary fallback for datasets that haven't succeeded since last_successful
+    # TODO: Remove after the stragglers have succeeded or been removed.
+    # stragglers: lt_pep_declarations, hu_national_assembly, ps_local_freezing
+    for v in iter_dataset_versions(dataset_name):
+        # We could check if the index.json has a resources array here, but only
+        # lt_pep_declarations needs it, and it's being removed 2026-08-17
+        # so let's nto bother
+        name = f"{ARTIFACTS}/{dataset_name}/{v.id}/{resource}"
+        object = backend.get_object(name)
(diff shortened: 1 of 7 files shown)


jbothma:
Testing:

Against prod, the datasets without resources (the symptom the user described) are
  - datasets exporting nothing but maybe stats and deltas (mostly enrichers)
  - datasets whose last run was failed

```
$ curl -s https://data.opensanctions.org/datasets/latest/index.json   | jq '[.datasets[] | select((.resources // []) | length == 0) | .name]'
[
  "lt_pep_declarations",
  "ext_icij_offshoreleaks",
  "ext_us_cms_npi",
  "ext_cy_companies",
  "ua_rada",
  "ext_ua_edr",
```

With this fix, it's down to just the the datasets with restricted exporters.

```
$ cat data/catalogs/opensanctions.json  | jq '[.datasets[] | select((.resources // []) | length == 0) | .name]'
[
  "ext_gb_coh_psc",
  "ext_lv_business_register",
  "ext_ua_edr",
```

zavod/zavod/archive/__init__.py
```diff
@@ -210,23 +210,35 @@ def get_artifact_object(
...
+        version = "20250403160648-kkg"
+        name = f"{ARTIFACTS}/{dataset_name}/{version}/{resource}"
+        object = backend.get_object(name)
+        if object.exists():
+            return object
 
     # FIXME: legacy fallback option of using the latest release
     # REMOVE THIS AFTER MIGRATION
     name = f"{DATASETS}/{LATEST}/{dataset_name}/{resource}"
     object = backend.get_object(name)
```
pudo: this is very long since ready to go.

zavod/zavod/publish.py
```diff
@@ -63,12 +63,18 @@ def _archive_artifacts(dataset: Dataset, extra_artifacts: list[str] = []) -> Non
...
+
     Every file we persist about this run is uploaded to /artifacts/{dataset}/{version}/
 
     Listed resources plus index and collection catalog are copied to
     /datasets/{RELEASE}/{dataset}/ backward compatibility and
     /datasets/{LATEST}/{dataset}/ for discovery without the full catalog.
     """
+    version = get_latest(dataset.name, backfill=False)
+    if version is None:
+        raise ValueError(f"No working version found for dataset: {dataset.name}")
```
pudo, line 76: I feel really strongly that we should make this go away in the other (prefix path) PR again. version is passed, not discovered.

pudo (approved):
Nice! This does the right direction. I wonder if we at some point we want a dataclass a la `DatasetVersions` with `working`, `successful`, `recent` that we hand around to always be explicit?

zavod/zavod/publish.py
jbothma: agree - just moving it up here, but its days are numbered.

PR #5415, opened 2026-08-12
[us_ofac] Emit the Company schema upgrade for owned organizations

### LINKED ISSUE #5416: [us_ofac_sdn] Ownership:asset should reference Asset, but 3178 references point at Organization
We've added validation for entities being emitted with a schema that isn't compatible with the relations the entity is referenced in. This is highlighting an existing issue rather than a new bug.

### LINKED ISSUE #5417: [us_ofac_cons] Ownership:asset should reference Asset, but 277 references point at Organization
We've added validation for entities being emitted with a schema that isn't compatible with the relations the entity is referenced in. This is highlighting an existing issue rather than a new bug.

jbothma:
Emit sanctioned entities only once they have both properties (no empty emits) and the schema's been upgraded to company for those with an owner (Owned organisation is a company).

Increase in memory consumption is within normal fluctuation:
entities: Maximum resident set size (kbytes): 1,761,536
schemata: Maximum resident set size (kbytes): 1,763,360

potentially disruptive - let's discuss - 3000 Organizations get a schema change and appear as MOD in delta. No other properties change.

### DIFF
--- a/datasets/us/ofac/ofac_advanced.py
+++ b/datasets/us/ofac/ofac_advanced.py
@@ -235,7 +235,7 @@ def parse_location(context: Context, refs: Element, location: Element) -> Featur
 
 
 def parse_relation(
-    context: Context, refs: Element, el: Element, parties: dict[str, Schema]
+    context: Context, refs: Element, el: Element, parties: dict[str, Entity]
 ) -> None:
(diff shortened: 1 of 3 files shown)


jbothma:
dupe of https://github.com/opensanctions/opensanctions/pull/5289

PR #5427, opened 2026-08-13
[in_nse_debarred] fix splitting issue

herrhellana:
ref issue found in https://github.com/opensanctions/opensanctions/pull/5412

### DIFF
--- a/datasets/in/nse_debarred/crawler.py
+++ b/datasets/in/nse_debarred/crawler.py
@@ -1,3 +1,4 @@
+import re
 import shutil
 import zipfile
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


datasets/in/nse_debarred/crawler.py
```diff
@@ -18,6 +19,100 @@
...
+    r"^(?:(?:mr|mrs|ms|dr)\.(?=\s*\w)|(?:mr|mrs|ms|dr|shri|smt)\.?\s+)",
+    re.IGNORECASE,
+)
+BUSINESS_WORD = re.compile(
+    r"\b(?:advisory|advisor|academy|capital|company|consultancy|consultant|"
+    r"enterprise|financial|firm|gainer|infotech|institute|investment|market|"
+    r"money|organisation|research|securities|services|solutions|system|tips|"
+    r"trader|trading)\b",
+    re.IGNORECASE,
+)
```
jbothma, line 31: this feels like a job for an LLM and Data Reviews, just at a glance, seeing this many regexes

jbothma: comment was from when there were about 7 regexes

datasets/in/nse_debarred/crawler.py
```diff
@@ -6,18 +7,101 @@
...
+    name: str,
+    source_url: str | None,
+) -> tuple[OwnershipExtraction, str | None] | None:
+    """Request review of an ownership name and return only accepted extraction."""
+    if OWNERSHIP_MARKER.search(name) is None:
+        return None
+
+    source_value = TextSourceValue(
+        key_parts=name,
+        label="Entity / Individual Name ownership extraction",
```
jbothma, line 83:
```suggestion
        label="Entity / Individual Name column",
```

or just `Entity / Individual Name`

jbothma (commented):
This is looking great so far.

In this case, am I right in thinking it's going to mark the proprietor with the warning and not the company, while it's actually the company with the warning?

Image

PR #5437, opened 2026-08-14
[eu_journal_sanctions] Add consolidated framework-act parsers and snapshots

pudo:
## What

- **`scripts/parse_*.py`** — one self-contained parser per framework act, 15 in total: Russia 833/2014, Libya 2016/44, Belarus 765/2006, terrorist list 2580/2001, al-Qaida 881/2002, Bosnia Decision 2011/173, Afghanistan 753/2011, Myanmar 401/2013, CAR 224/2014, Burundi 2015/1755, ISIL/Al-Qaida 2016/1686, Mali 2017/1770, Human Rights 2020/1998, Russia internal repression 2024/1485, and Russian destabilising activities 2024/2642. Each fetches its pinned date-suffixed consolidated CELEX from CELLAR (via `zavod.shed.ojeu`, cached under gitignored `scripts/out/`) and extracts every target-bearing annex into the reviewed CSV contract. Where a regime has no regulation, the CFSP decision itself is the framework act.
- **`data/consolidated/*.csv`** — the extracted snapshots, from header-only files for placeholder annexes up to 1,746 rows for 833/2014.
- **`data/FORMAT.md`** — the CSV contract, including the fidelity boundary: cells hold the source's printed wording; transcription does structural extraction only (label→column mapping, splits dictated by printed labels or list structure). `startDate` and the entity date columns stay verbatim, country wording is preserved, name categorisation follows printed labels only, and `notes` holds bare prose with labels stripped. Semantic normalization (date parsing, country codes, identifier canonicalization, name review) is deferred to the consuming crawler via lookups, `dates` config, and the names review system.
- **`validate_csv.py`** — offline structural validator for every reviewed CSV: header/contract, CELEX provenance, program/measure vocabulary, FtM schema compatibility, bare-date shapes, multi-value cells, row uniqueness.
- **`scripts/common.py`** — flat leaf utilities shared across parsers (cell codec, date-shape recognizers, annex block access) under the independence-first ground rules in `scripts/CLAUDE.md`.
- **`eu_journal_sanctions.yml` / `Makefile`** — a `consolidation:` lookup pinning which consolidated version each snapshot reflects, one parse line per framework; `make` re-runs every parser and validates all CSVs.

## Design

Each parser runs inside an agentic fix-it loop: it fails closed on any structure it has not been taught — unknown annexes, changed headers, unrecognized labels, markers, date formats, or row shapes — with one-line errors naming the annex and the problem, and no rule is wider than the formats observed in its document. Closed annex registries classify every annex as target / expected-empty / non-target, so a newly populated annex becomes a visible failure rather than silently ignored data.

CSV filenames and the `celex` column are keyed on the framework act, so the git diff between consolidation versions is the review artifact and entity IDs stay stable across updates.

## Verification

- `make` runs offline end to end and reproduces every snapshot byte-identical; `validate_csv.py` passes all 17 reviewed CSVs.
- Each contract widening was checked with a column-aware diff against the previous snapshots: existing cell values unchanged except where a rule deliberately moved them (e.g. UN listing-basis prose from `reason` to `notes` in the Libya file).
- `mypy --strict` and `ruff` clean on all parsers, `common.py`, and the validator.


pudo:
## Verification: non-external statements are unchanged, except one

I checked the claim that this PR leaves the set of **non-external** statements identical. It essentially holds — one statement changes.

The new emission itself is clean: every `context.emit` added in `crawler.py` passes `external=True`, and `crawl_unconsolidated_row` / `crawl_context_row` are untouched. The risk is elsewhere — the PR changes **dataset-wide config that the old path also consumes**: `dates.formats` (+11 formats), `type.date` (7 → 262 options, `lowercase: true`), `type.country`, `type.email`, `type.identifier`, plus new `type.address`, `type.gender`, `type.url`. All of those fire in `value_clean` on every `entity.add()`, including the pre-existing rows.

So I replayed the untouched property-assignment code verbatim under both `eu_journal_sanctions.yml` versions over the live source sheets (703 unconsolidated + 5,281 context rows) and diffed every property value.

### One divergence in 5,984 rows

`context.csv` row 2519 — Hossein SHAMKHANI (List ID 1951), `DOB = "1984 or 1985"`:

```
birthDate   OLD ['1984']   NEW ['1984', '1985']
```

A newly added `type.date` option splits the value; the old config fell back to extracting the single year. One added non-external statement. Nothing else differs — worth a deliberate ack rather than a silent ride-along.

### Everything else verified clean

- **`sanction.program`** (47 → 77 options): nothing removed or modified, additions only. All 699 sanctions in the current production run (`20260816122501-kbs`) already carry a `programId`, so no journal row currently falls through to a newly-added code — `programId` / `program` / `programUrl` unchanged.
- The four `type.date` options that *look* deleted in the diff (`Approximately 1977`, `Between Aug. and Sep. 1977`, `Between 1960 and 1975`, `Between 1959 and 1965`) are only case-normalized; `lowercase: true` keeps them matching identically (confirmed by direct lookup comparison).
- **`names:` config** (`schema_rules`, `suggest_weak_alias_person_single_token`) is read only by `check_names_regularity` / `apply_reviewed_names` / `clean_names`, none of which the old path calls — it is not part of `value_clean`.
- **No entity-ID collisions**: 16,368 new IDs (8,184 entities plus their sanctions) against every ID the old path can produce — 806 unconsolidated entity/related/UnknownLink/wallet IDs, 4,973 context IDs, and 62,868 enumerated old sanction IDs across all program codes — **zero overlap**. No existing statement can have its `external` flag flipped.
- `zavod/zavod/shed/ojeu/cellar.py` is purely additive; no existing function touched.
- Deleted `extract_tables.py` / `extract/` / `validate_csv.py` are not imported by `crawler.py` (`entry_point: crawler.py`).
- `assertions` are unchanged and `min`-only, so the ~8k new external entities can't trip them.

### Caveats

- No full crawl was run — `extract_program_code` and the CELLAR discovery need network and a warm cache. The program-key conclusion rests on the production artifact showing zero sanctions without a `programId`.
- `http: timeout: 300` now also applies to the old path's EUR-Lex fetches. No statement impact, but it changes failure timing.
- Operational, not statement-level: the CSV emission runs *after* the old path and calls `apply_reviewed_names` + `assert_all_accepted`, which need the review database. A failure there aborts the run and publishes nothing — a new failure mode for the existing data.
- `zavod/docs` still references the now-deleted `extract_tables.py`.

PR #5443, opened 2026-08-14
[hu_national_assembly] xml guard

nvmbrasserie:


datasets/hu/national_assembly/crawler.py
```diff
@@ -34,13 +50,45 @@ def fetch_xml(context: Context, endpoint: str, params: dict[str, str] = {}) -> E
             "access_token": ACCESS_TOKEN,
         },
     )
-    _, _, _, text = zyte_api.fetch_text(
-        context,
-        url,
-        geolocation="HU",
-        cache_days=1,
```
jbothma, line 41: Yeah a local run succeeded for me too, so I ran it in prod now and it also succeeded.

I think as a first step, we can close it - maybe it'll actually be fairly reliable. But the PR is correct in realising the broken value was cached.

Next time it fails, I'd suggest the simpler fix of just doing 

```
result = zyte_api.fetch()
assert status_code == 200, status_code
try ...
    context.cache.set()
except Exception as e:
    result.invalidate()
    raise
context.cache.set()
...extract...
```

similarly to https://github.com/opensanctions/opensanctions/blob/main/datasets/ch/parlament/crawler.py#L159 (except this one forgets to set the cache)

so we'd rely on the crawler retrying. But at least the cache would be invalidated, and it might just be enough to be robust. Before we implement aaaaaall the logic here.

The way we test this amount of logic is... with automated tests, which we have for fetch_html which does aaaaaall this logic. So I think the long term fix is to make this pattern easier to use generally and not just for html.

Issue #5434, opened 2026-08-14
us_dod_chinese_milcorps: move source data from Google Sheet to repo CSV


nvmbrasserie:
Follow `bg_mft_national` / `cn_sanctions`: manually-maintained data lives as `sanctions.csv` in the dataset directory, served from `raw.githubusercontent.com`.

- Export the published Google Sheet to `datasets/us/dod_chinese_milcorps/sanctions.csv`; point `data.url` at the raw GitHub URL
- Document row conventions + update procedure in the yml `config:` comment

**Open question:** can the list itself be automated? DoD publishes each 1260H update as a PDF attached to a defense.gov release; the crawler already detects new releases via `assert_dom_hash`. Assess whether the PDF is parseable enough to emit directly (or at least to generate CSV rows for review), which would remove the manual sheet entirely.

PR #5458, opened 2026-08-17
[us_dod_chinese_milcorps] Move source data from Google Sheet to repo CSV

fabio-rovai:
This addresses #5434.

## Change

The manually maintained source list moves from the published Google Sheet into the repository as `datasets/us/dod_chinese_milcorps/sanctions.csv`, following the pattern used by `bg_mft_national` and `cn_sanctions`. The dataset's `data.url` now points at the raw.githubusercontent.com URL for that file. The row conventions and the update procedure for new 1260H releases are documented in a top-of-file comment in the dataset yml, which is where the repo's dataset-metadata guidance places maintainer runbooks. The CSV is force-added because `*.csv` is gitignored globally, matching how the precedent files are tracked.

The issue's open question about parsing the 1260H PDF directly is not addressed here. This PR only performs the mechanical move.

## Verification

An end-to-end `zavod crawl` needs a Zyte API key for the release-page hash check, and the dataset is marked `ci_test: false` for the same reason, so verification here is at the data-contract level. The committed CSV is byte-identical to what the crawler currently fetches from the published sheet, with 397 data rows. Its header contains exactly the nine columns that `crawler.py` pops, so `context.audit_data` passes with no leftovers. All Start date and End date values are ISO formatted. The data yields 246 unique companies including parents, inside the yml assertion band of 120 to 280 Company entities. The edited yml re-parses cleanly with all expected top-level keys.

## Context

We run open register-integrity audits (https://github.com/fabio-rovai/register-integrity-index) and are glad to contribute identifier-conformance checks if useful.


pudo:
duplicate. Thanks, though!

PR #5459, opened 2026-08-17
[be_chamber] Fetch without browser rendering, skip unfetchable profiles

nvmbrasserie:
Runs have been failing with UnblockFailedException on a member profile page, a different member each run. The source is fine: every page the crawler parses is server-rendered HTML with no JavaScript, and all ten term list pages plus profiles from the newest and oldest terms parse from a plain GET.

fetch_html defaults to browserHtml, which the crawler never overrode, so each run drove 1457 full browser page loads (eight terms are inside earliest_term_start) at a site that throttles sustained clients and answers with a table-less IIS 403 page when it refuses. Ask for httpResponseBody instead; the Content-Type header declares iso-8859-1, so accented names decode correctly.

A single 403 also aborted the whole run, discarding all 645 members. Warn and skip that member instead. Skipping rather than emitting a partial person is deliberate: the person ID includes the birth date, which only the profile page carries, so emitting without it would mint a second entity for the same member. The Person minimum in the assertions catches it if this stops being occasional.

### DIFF
--- a/datasets/be/chamber/crawler.py
+++ b/datasets/be/chamber/crawler.py
@@ -7,12 +7,13 @@ from zavod import Context, helpers as h
 from zavod.extract import zyte_api
 from zavod.stateful.positions import categorise
 
-
 UNBLOCK_VALIDATOR = "//table[@width='100%']"
 POSITION_TOPICS = ["gov.legislative", "gov.national"]
 # Every page we parse is server-rendered ColdFusion HTML with no JavaScript involved,
 # so the fetches ask Zyte for httpResponseBody rather than its default browserHtml.
 HTML_SOURCE = "httpResponseBody"
+# Retry patiently: when throttled, the site answers with a table-less IIS 403 page.
+RETRIES = 7
 
 # Matches DOB in plain-text bios:
 #   French: "Née à Tournai le 22 mai 1963."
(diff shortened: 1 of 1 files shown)


datasets/be/chamber/crawler.py
```diff
@@ -53,13 +56,28 @@ def crawl_person(
...
-        unblock_validator="//table",
-        absolute_links=True,
-        cache_days=30,
-    )
+    try:
+        pep_doc = zyte_api.fetch_html(
+            context,
+            profile_url,
+            unblock_validator="//table",
+            html_source=HTML_SOURCE,
```
jbothma: I don't think we gain anything from making HTML_SOURCE a constant at the top of the module. The API right now is to pass in a string value, so we can just pass the explicit value "httpResponseBody" here and the code would be more readable.

A better API would be to define that string as a constant in zyte_api, then we could pass `zyte_api.HTTP_RESPONSE_BODY` or something, and/or allow only specific strings in the type defintion of `fetch_html`.

datasets/be/chamber/crawler.py
```diff
@@ -53,13 +56,28 @@ def crawl_person(
...
+        # includes their birth date, which only the profile page carries, so emitting
+        # them without it would mint a second entity for the same member. Skip them
+        # this run instead of losing the whole term - the Person count assertion
+        # catches it if this ever stops being occasional.
+        context.log.warning(
+            "Could not fetch profile page",
+            name=squash_spaces(name),
+            profile_url=profile_url,
+        )
+        return
```
jbothma: it runs monthly, failed today, but succeeded when it last ran a month ago. So it feels premature to drop entities for a month at a time. We really only allow a crawler to drop entities in extreme cases. Let's get rid of the exception and allow more retries. That's more commonly the solution.

PR #5468, opened 2026-08-18
Fixes 18 08 26

nvmbrasserie:


datasets/br/ceis/br_ceis.yml
```diff
@@ -57,3 +57,7 @@ lookups:
         # From the registry: https://casadosdados.com.br/solucao/cnpj/kasvi-importacao-e-distribuicao-de-produtos-para-laboratorios-ltda-13324282000396
         # matched based on the tax number: 13324282000396
         values: "KASVI IMPORTACAO E DISTRIBUICAO DE PRODUTOS PARA LABORATORIOS LTDA"
+      - match: "12.441.527/0001-30"
+        # From the registry: https://casadosdados.com.br/solucao/cnpj/associacao-de-educacao-e-cultura-12-441-527-0001-30-12441527000130
+        # matched based on the tax number: 12441527000130
+        values: "W J C ARTISTIC PROMOTIONS LTDA"
```
jbothma:
```suggestion
      - match: "12.441.527/0001-30"
        # From a second entry with the same company registration number in the source dataset.
        # matched based on the tax number: 12441527000130
        values: "WJC PROMOÇÕES ARTÍSTICAS LTDA - ME"
```

I think we should be very wary of bringing in names that are not in the source dataset. And in this case we don't even need to pull a name from one of the other three name columns - we can pull a name from the same column :D 

Image
Image

datasets/eu/fsf/eu_fsf.yml
```diff
@@ -254,6 +254,9 @@ lookups:
         value: Vessel
       - match:
           - eu-fsf-eu-12692-74
+          # A research arm of the AEOI, published by the EU with
+          # subjectType code="person".
+          - eu-fsf-eu-13894-78 # Nuclear Research Center for Agriculture and Medicine
```
jbothma, line 259: at least it's not like this is serious business `</sarcasm>`

jbothma (commented):
other than br_ceis, good to go.

PR #5481, opened 2026-08-19
PEPs: Moldova MPs

### LINKED ISSUE #627: Moldova parliament members
**Data URL**
https://parlament.md/member-of-the-parliament.nspx

** Publisher**
Parliament of the Republic of Moldova.

**Publisher country/territory code**
MD

**Type of data**
PEPs

**Coverage region**
region:Europe

**Can you tell us more?**

herrhellana:
closes https://github.com/opensanctions/crawler-planning/issues/627

### DIFF
--- a/datasets/md/parliament/crawler.py
+++ b/datasets/md/parliament/crawler.py
@@ -1,0 +1,208 @@
+import json
+import re
+from datetime import datetime, time
+from typing import Any
+from zoneinfo import ZoneInfo
+
+from zavod.entity import Entity
+from zavod.extract.zyte_api import fetch_html
+from zavod.stateful.positions import PositionCategorisation, categorise
+
+from zavod import Context
+from zavod import helpers as h
+
+# In-page requests that retrieve and slim all available rosters.
+BROWSER_SCRIPT = """
+(async () => {
+  const emit = (value) => {
+    const pre = document.createElement("pre");
+    pre.id = "payload";
+    pre.textContent = JSON.stringify(value);
+    document.body.appendChild(pre);
+  };
+  const controller = new AbortController();
+  const timer = setTimeout(() => controller.abort(), 40000);
+  const post = async (body) => {
+    const response = await fetch("/Routing/MakePostRequest", {
+      method: "POST",
+      headers: {"Content-Type": "application/json; charset=utf-8"},
+      body: JSON.stringify(body),
+      signal: controller.signal,
+    });
+    if (!response.ok) {
+      throw new Error(`POST ${body.methodName} returned HTTP ${response.status}`);
+    }
+    return JSON.parse(await response.text());
+  };
+
+  try {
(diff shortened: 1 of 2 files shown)


nvmbrasserie:
Tidy-up on md_parliament, no change to the emitted data.

- In-page script cut back to what must run in the page; shape checks moved to Python, where a failure is legible instead of costing four retried renders.
- .NET term dates parsed without a regex.
- `audit_data` on both records — new source fields were being dropped silently.
- Metadata fixes: `data.url` was the site root; `zyte_timeout: 90` lowered the budget below the 300s default; the `ci_test` note blamed Cloudflare rather than the in-page-only API.
- Title/summary/description standardised per `zavod/docs/metadata.md`, with the crawler's failure modes recorded at the top of the yml.

Runs clean against Zyte.

nvmbrasserie (approved):

datasets/md/parliament/crawler.py
```diff
@@ -0,0 +1,199 @@
...
+# The endpoint only answers calls made from within the rendered page; an identical
+# external POST returns an empty 200.
+BROWSER_SCRIPT = """
+(async () => {
+  const emit = (value) => {
+    const pre = document.createElement("pre");
+    pre.id = "payload";
+    pre.textContent = JSON.stringify(value);
+    document.body.appendChild(pre);
+  };
```
nvmbrasserie, line 26: This is a very long browser script, and I'm on the fence here. I tried every other way to get to the same result, but none of them succeeded. It's our only way there. @jbothma maybe you could take one final look before we merge it?

nvmbrasserie:
Why the in-page script — everything else is blocked:

- **External POST replay** (JSON content-type, `Referer`/`Origin`/`X-Requested-With`): endpoint answers 200 with an empty body.
- **Same replay with harvested session cookies:** Zyte returns 520 "Website Ban" for *any* attached cookie — even on a plain GET that succeeds without them. Identical for session-cookie-only and with `cf_clearance` dropped.
- **The anti-forgery token is a red herring:** `services.js` reads `RequestVerificationToken` from an element that isn't on the page, so the site itself sends `undefined`. Our request body already matches the site's own call.
- **DOM scraping:** the site's Angular app never populates the roster in Zyte's browser — a 15s `waitForSelector` on a member card timed out over four renders.
- **`networkCapture`:** nothing to capture, since that same app never fires the XHR.

datasets/md/parliament/crawler.py
brrttwrks: @nvmbrasserie , given its JS, I think this seems OK. We could omit the error checking, but those seem minimal here and useful. I would merge this.

I did check for other approaches as well, including accessibility features, but the DOM has to load first anyway for that and it is accessed via Cloudflare. This seems like the only way, but it does feel a little fragile.

PR #5487, opened 2026-08-19
us_nv / us_nc med_exclusions: provider schema and controlling-interest names

nvmbrasserie:
Splits the `us_nv_med_exclusions` and `us_nc_med_exclusions` work out of #5484 so it can be reviewed on its own.

Both lists pair each excluded provider with the people who hold a controlling interest in it, and both were tripping `Ownership:asset should reference Asset`. Satisfying that by calling the provider a `Company` is a claim the source never makes — which is how Nevada ended up emitting a practitioner as both a `Company` and a `Person`, giving `No common schema: Company and Person` in the dataset run and in the `default` collection. Each commit here decides the schema from what the columns actually say.

### us_nv_med_exclusions

- **Don't make a practitioner a Company.** The self-ownership check compared names within a row, so Roderick Hill slipped through: he co-owns Zion Safe Haven with Joyce M. Hill, so his own row names her, not him, while the Zion row names him as a holder. The check now reads the controlling-interest column across the whole list — a provider the list names as a holder anywhere is a person, so no schema upgrade and no ownership edge. Of 470 rows with a holder, 8 providers appear in the column; only Roderick Hill's behaviour changes.
- **One holder per controlling-interest name.** Cells listing several holders arrived as one person carrying every name ("Jini Muhammad / Dominick Hurd", "Michael Hobbs / Regina Hobbs" — ten of them, each also short an ownership edge). They now split per line, 473 edges against 462. Two cells where the line break isn't a separator are spelled out in a lookup. Two cells name a business rather than a person ("PHC Hospice LLC", "OneCare Health Services, LLC"), so the holder's schema comes from `rigour`'s `extract_org_types` instead of the column heading.
- **Don't read a credential as a legal form.** `extract_org_types` reads a generational suffix as a legal form — "Claude Arthur Verbal II" comes back as `('II', 'ii')` — and `OD` and `PT` go the same way, both credentials a holder on a medical list plausibly carries. Filtered out; output unchanged on the current list.

### us_nc_med_exclusions

- **Give the provider the schema its row implies.** Both warnings were `Ownership:asset` pointing at a non-Asset, 88 references at a `Person` and 85 at a `LegalEntity`, with a different cause each. 91 of the 178 rows name the provider as their own owner (ALFRED KWASI FOLUKE owned by ALFRED KWASI FOLUKE); those are practitioners, which is what resolution was working out by merging each pair into a `Person`. The provider now carries `Person` directly and there's no ownership edge — one record per practitioner instead of two. The other 87 providers are unambiguously businesses (1ST CHOICE HEALTHCARE SERVICES LLC, ANGELS HOME HEALTH, MARCIA L REMENTER DMD PA), so they get `Company`, which is also an `Asset`. Both `LegalEntity` buckets are empty now, so the bounds move to the two schemata that hold providers plus `entity_count`.

### Verification

Crawl, validate and export clean on both, `issues.log` empty, `mypy --strict` and `ruff` clean.

| | before | after |
|---|---|---|
| us_nv | 6 assemble errors, 469 range warnings | none; 1758 things (LegalEntity 913, Person 470, Company 375) |
| us_nc | 2 range warnings over 173 refs | none; 233 things (Person 164, Company 69) |

Nevada was verified with the two Zyte calls substituted for direct fetches, since there's no API key locally.

### Known limits

The Nevada guard is exact-match evidence and sees only this dataset, so it narrows the `Company`/`Person` conflict rather than closing it: a holder spelled differently from their provider row still slips through (this very PDF has "Monteray Sellers" vs "Monteray Sellars"), and cross-dataset resolution can't be anticipated from within the file. Closing the class means dropping `Ownership` for `UnknownLink`, which trades away the ownership semantics for the ~460 rows that really are businesses — worth deciding once across the `us_*_med_exclusions` family rather than per state.

### DIFF
--- a/datasets/us/nc/med_exclusions/us_nc_med_exclusions.yml
+++ b/datasets/us/nc/med_exclusions/us_nc_med_exclusions.yml
@@ -31,18 +31,25 @@ data:
 
 dates:
   formats: ["%m/%d/%Y"]
-# Individual providers arrive twice, as the excluded LegalEntity and as the Person who
-# owns it, and dedupe merges the pair into a Person - so LegalEntity drains into Person.
+# A provider given as their own owner is a Person and the rest are Companies, so every
+# provider carries one of the two schemata and nothing is left in the LegalEntity bucket.
 assertions:
   min:
+    entity_count: 180
     schema_entities:
-      LegalEntity: 50
-      Person: 100
+      Person: 120
+      Company: 40
   max:
+    entity_count: 350
     schema_entities:
-      LegalEntity: 300
       Person: 250
+      Company: 150
 lookups:
+  type.name:
+    options:
+      # The ownership column names the person, and repeats what they trade as
+      - match: 'ROSALIND CURRENCE BLAND d\b\a ABUNDANT LOVE HOME CARE SERVICES'
+        value: ROSALIND CURRENCE BLAND
   type.identifier:
     options:
       - match: # NPI/ATYPICAL ID in the source
--- a/datasets/us/nv/med_exclusions/crawler.py
+++ b/datasets/us/nv/med_exclusions/crawler.py
@@ -8,6 +8,14 @@ from zavod import Context, helpers as h
 from zavod.extract import zyte_api
 
 CONTROLLING_INTEREST = "persons_with_controlling_interest_of_5_or_more"
+# rigour reads each of these as a legal form, but on a medical provider they are a
+# generational suffix or a credential: "Claude Arthur Verbal II", "Gary Ridenour OD".
+PERSON_SUFFIXES = {"ii", "iii", "iv", "od", "pt"}
+
+
+def is_business(name: str) -> bool:
+    """Whether a name carries a legal form, which a person's name doesn't."""
+    return any(norm not in PERSON_SUFFIXES for _, norm in extract_org_types(name))
 
 
 def controlling_interest_holders(context: Context, value: str | None) -> list[str]:
@@ -68,7 +76,7 @@ def crawl_item(
 
         for holder in holders:
             # The column says persons, but a couple of cells name a business.
-            owner = context.make("Company" if extract_org_types(holder) else "Person")
+            owner = context.make("Company" if is_business(holder) else "Person")
             owner.id = context.make_id(holder, entity.id)
             owner.add("name", holder)
             owner.add("country", "us")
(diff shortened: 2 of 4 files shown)


datasets/us/nv/med_exclusions/crawler.py
```diff
@@ -1,12 +1,40 @@
...
+
+    A cell names one holder per line, except where a line break falls inside a name
+    or where two names run together with no separator at all. The lookup spells those
+    out, since nothing in the cell distinguishes them.
+    """
+    if not value:
+        return []
+    result = context.lookup("controlling_interest", value)
+    names = result.values if result is not None else value.split("\n")
+    return [name.lstrip("&").strip() for name in names if name.strip("& ")]
```
jbothma, line 32: wht do you think of making the lookup warn if len(h.multisplit(value, "\n ")) >= 4 but no lookup matches? That should have discovered the current cases, right?

datasets/us/nc/med_exclusions/us_nc_med_exclusions.yml
```diff
@@ -31,18 +31,25 @@ data:
 
 dates:
   formats: ["%m/%d/%Y"]
-# Individual providers arrive twice, as the excluded LegalEntity and as the Person who
-# owns it, and dedupe merges the pair into a Person - so LegalEntity drains into Person.
+# A provider given as their own owner is a Person and the rest are Companies, so every
+# provider carries one of the two schemata and nothing is left in the LegalEntity bucket.
```
jbothma, line 35:
```suggestion
# A provider given as their own owner is a Person and the rest are Companies, so every
# provider carries one of those two schemata.
```

datasets/us/nc/med_exclusions/us_nc_med_exclusions.yml
```diff
@@ -31,18 +31,25 @@ data:
...
     schema_entities:
-      LegalEntity: 300
       Person: 250
+      Company: 150
 lookups:
+  type.name:
+    options:
+      # The ownership column names the person, and repeats what they trade as
+      - match: 'ROSALIND CURRENCE BLAND d\b\a ABUNDANT LOVE HOME CARE SERVICES'
+        value: ROSALIND CURRENCE BLAND
```
jbothma, line 52: how about using name cleaning for this? there's also `DR. FIDELIS SIMPLUS AKAGHA/BRADLELY HOMES` but the question is how we'll detect new ones

jbothma (approved):
perhaps nv can get a little company asserstion

pudo:
Would this be a good testing ground for `Control`?

PR #5488, opened 2026-08-19
tm_mejlis: account for every field on the deputy profile

### LINKED ISSUE #5489: [tm_mejlis] mejlis.gov.tm unreachable from our network path (Zyte 421)
`mejlis.gov.tm` is unreachable from our network path, and has been intermittently since June. The site itself is up.

## Current failure

Failing for 3 consecutive runs since 2026-08-19 11:33. Last success `20260729094248-gkz` (2026-07-29 09:42), first of this batch `20260819113314-onm`, latest `20260819113440-ift`.

```
Runner failed with HTTPError on https://api.zyte.com/v1/extract
response_code=421
response_text={"type":"/website/connection-error","title":"Website Connection Error","status":421,
  "detail":"Failed to establish a network connection to the website. This could be due to the
  server being down or network connectivity issues. Verify the website is operational and accessible."}
```

## What I probed

- **Direct fetch, from a laptop in Europe, just now**: `https://mejlis.gov.tm/convocation?lang=en` → `200`, 101,083 bytes, 2.9s. A deputy profile → `200`, 52,030 bytes, 0.6s. So the server is serving content and the outage is in the path, not the host.
- **DNS**: resolves to `216.250.11.241`, MX `mail.sanly.tm`. No NXDOMAIN, nothing stale.
- **Zyte**: can't be probed without the API key, but the 421 is Zyte's own report that *it* failed to open a connection to the host — not a block page, not a challenge, not a 403.

## History — the same symptom on both transports

Before we moved to Zyte, the production runner failed the same way directly:

```
20260629135214-cdj, 20260721160321-buf:
Runner failed with ConnectionError
HTTPSConnectionPool(host='mejlis.gov.tm', port=443): Max retries exceeded with url:
/convocation?lang=en (Caused by NewConnectionError(...))
```

That's what prompted #4909 ("[tm_mejlis] Try running via zyte"), closed 2026-07-23 when 7eb67fec7 switched the crawler to the Zyte API. It bought two successful runs on 23 July and one on 29 July — and that 29 July success came after five failures the same hour, so it was already flaky rather than fixed.

So: our runner couldn't open a TCP connection in June and July, Zyte's exit can't open one now, and the host answers a European consumer connection fine. The reading that fits all three is a host that is intermittently unreachable from commercial/hosting networks — and one that recovers on its own, since it has done so twice.

## Why this is an outage rather than a fix

No crawler change makes a TCP connection succeed. Retrying is what has worked historically, and this dataset is on a monthly schedule, so there's little cost to waiting.

One code-side hypothesis is open as #5488: both fetches ask Zyte for `geolocation: TM`, on the theory that a Turkmenistan exit reaches a host that drops foreign traffic. That's untested — no API key locally — so it's a candidate, not a fix, and it shouldn't hold up marking this as an outage.

nvmbrasserie:
Splits the `tm_mejlis` refactor out of #5484 so it can be reviewed on its own. No output change — same 249 entity ids, no property-value differences.

### Strict interpretation

The profile's five `label:` rows were read one at a time by a helper that returned `None` when a label was missing, and only `Year of Birth` was ever asked for, so the other four went out silently. They now go into a `dict[str, Element]` keyed by label, each field is popped where it's used, and `context.audit_data(fields, ignore=["Position", "Committee"])` closes the record — a renamed or added label warns instead of vanishing. Values keep their markup, which lets the constituency and the biography come out of their own popped block rather than a second top-level selection.

Three more places went from silence to a signal:

- the birth year was filtered through `^\d{4}$` and dropped without a word, where `h.apply_date` warns and keeps `original_value`;
- the constituency was `if len(district) > 0`, now `h.xpath_element`, which raises;
- the id regex on each roster href was `if match is not None`, now an `assert`, since the selector already required `single-deputy/`.

All 124 profiles carry exactly one name, one district and five labelled rows today, so none of these fire.

### Structure and HTTP

`parse_deputy_links` returns the href the roster publishes instead of rebuilding it from a hardcoded host, and dedupes with `setdefault`. Both fetches now ask Zyte for `httpResponseBody` from `geolocation: TM`: the raw body carries all 124 links, so the roster was paying for a browser render it doesn't need, and the location is the open question behind the `421 /website/connection-error` that has failed the last three production runs. Types move to `zavod.util.Element`, which is what the `h.xpath_*` helpers and Zyte's `fetch_html` speak.

### Left unmapped, deliberately

`Position:` gives a committee role for 52 of the 124 deputies ("Member of the Committee on social policy of the Mejlis of Turkmenistan") and an outside job for the rest ("Therapist at Bayramaly etrap hospital"), with deputy 39 carrying both spliced into one string — one field, two meanings, so it stays in the `ignore` list with that written down. The dataset description claimed the dataset records the committee, which it never did, so that claim is gone. The `http.user_agent` block goes too: `zyte_api` sends its own agent, and the site serves a plain request with no agent at all, so the comment about it rejecting the default was wrong as well.

### Verification

Crawl, validate and export clean, `issues.log` empty, `ruff` and `mypy --strict` clean. Person 124, Position 1, `tm` 125 — all inside the assertions.

**Untested:** the Zyte transport. Every local run substituted `zyte_api.fetch_html` with a direct fetch, since there's no API key here, so `geolocation="TM"` against the live 421 wants one manual `zavod crawl` with the key before merge.

Follow-up worth having: each biography opens with `Date and place of birth: 7 January 1988, city of Ashgabat`, so the full birth date and birth place are there for the taking — both "essential when the source provides them" per the data priorities, against the year-only field used today.

### DIFF
--- a/datasets/tm/mejlis/crawler.py
+++ b/datasets/tm/mejlis/crawler.py
@@ -1,84 +1,97 @@
 import re
-from lxml.html import HtmlElement
 
 from zavod import Context, helpers as h
 from zavod.entity import Entity
-from zavod.stateful.positions import PositionCategorisation, categorise
 from zavod.extract.zyte_api import fetch_html
+from zavod.stateful.positions import PositionCategorisation, categorise
+from zavod.util import Element
 
+DEPUTY_LINK_XPATH = "//a[contains(@href, 'single-deputy/')]"
+# Deputy profile links look like ``single-deputy/168?lang=en``; the number is the
+# stable source id we key the person entity on.
 DEPUTY_RE = re.compile(r"single-deputy/(\d+)")
-YEAR_RE = re.compile(r"^\d{4}$")
 
 
-def parse_deputy_ids(doc: HtmlElement) -> list[str]:
-    """Collect the deputy ids linked from the convocation roster page, in order."""
-    ids: list[str] = []
-    for href in h.xpath_strings(doc, "//a[contains(@href, 'single-deputy/')]/@href"):
+def parse_deputy_links(doc: Element) -> dict[str, str]:
+    """Map the id of each deputy linked from the convocation roster to their profile."""
+    links: dict[str, str] = {}
+    for href in h.xpath_strings(doc, f"{DEPUTY_LINK_XPATH}/@href"):
         match = DEPUTY_RE.search(href)
-        if match is not None and match.group(1) not in ids:
-            ids.append(match.group(1))
-    return ids
+        assert match is not None, href
+        links.setdefault(match.group(1), href)
+    return links
 
 
-def field_value(right_block: HtmlElement, label: str) -> str | None:
-    """Return the text of the profile's `label:` row value, if present."""
+def parse_profile_fields(right_block: Element) -> dict[str, Element]:
+    """Read the profile's ``label:`` rows, keyed by label without the colon.
+
+    The profile is a vertical label/value list rather than a column-oriented table,
+    so `h.parse_html_table()` doesn't fit. The values keep their markup: both the
+    constituency and the biography are structured inside their own value block.
+    """
+    fields: dict[str, Element] = {}
     for row in h.xpath_elements(
         right_block, ".//div[contains(@class, 'key_value_wrapper')]"
     ):
-        key = h.element_text(h.xpath_element(row, ".//span[@class='key']")).rstrip(":")
-        if key == label:
-            return h.element_text(
-                h.xpath_element(row, ".//*[contains(@class, 'value')]")
-            )
-    return None
+        label = h.element_text(h.xpath_element(row, ".//span[@class='key']"))
+        value = h.xpath_element(row, ".//*[contains(@class, 'value')]")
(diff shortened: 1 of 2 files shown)


nvmbrasserie:
Blocked on verification against the live site: `mejlis.gov.tm` is unreachable from our network path right now — see #5489, filed as an outage.

That issue is also the evidence behind the `geolocation: TM` in this PR. The production runner failed with `ConnectionError`/`NewConnectionError` on direct fetches in June and July, Zyte's exit now fails with `421 /website/connection-error`, and the host answers a European consumer connection fine (`200`, 101 KB, checked just now). So `geolocation: TM` is a hypothesis about which exit can reach the host, not a verified fix — it wants one `zavod crawl` with `OPENSANCTIONS_ZYTE_API_KEY` set, once the path is back up, before this merges.

The parsing changes here are independent of that and are verified: crawl, validate and export clean with the two Zyte calls substituted for direct fetches, and no property-value differences against the pre-refactor baseline.

nvmbrasserie:
closing due to the deprecation

PR #5490, opened 2026-08-19
PEPs: Dominican Republic National Congress

### LINKED ISSUE #635: Dominican Republic National Congress - Chamber of Deputies and Senate
### Data URL

https://camaradediputados.gob.do/
https://senadord.gob.do/

### Publisher

Cámara de Diputados / Senado de la República Dominicana

DO

### Type of data

### Coverage region

region:Latin America

### Can you tell us more?

Bicameral parliament: 190 deputies + 32 senators.

**Chamber of Deputies** (`camaradediputados.gob.do`):
- WordPress-based site with the REST API enabled at `/wp-json/`
- Member data appears to be stored as posts and is queryable via the WordPress content API
- Not a dedicated legislative API, but members are reachable via `/wp-json/wp/v2/posts?search=diputado` or similar
- Requires exploring the post types and taxonomy to find the right endpoint

**Senate** (`senadord.gob.do` — note redirect from old `senado.gob.do`):
- Also WordPress; no dedicated member API exposed
- An OAI-PMH endpoint was previously listed at `senado.gob.do/senado/oai.aspx` — this is now defunct after the domain migration

**Approach:** The Chamber WordPress API is the most tractable entry point. Crawling the member post type should yield name, party, and constituency for all 190 deputies. Senate may require HTML scraping of the member directory.

- [x] I understand.

herrhellana:
closes https://github.com/opensanctions/crawler-planning/issues/635

### DIFF
--- a/datasets/do/congress/crawler.py
+++ b/datasets/do/congress/crawler.py
@@ -1,0 +1,114 @@
+from itertools import count
+
+from zavod.entity import Entity
+from zavod.stateful.positions import PositionCategorisation, categorise
+
+from zavod import Context
+from zavod import helpers as h
+
+
+def discover_legislator_ids(context: Context) -> set[int]:
+    ids: set[int] = set()
+    # Search is required to run at all (an empty keyword returns zero results), so the
+    # full roster is discovered by taking the union of searches for each vowel. Spanish
+    # full names (given names + two surnames) reliably contain every vowel at least
+    # once, confirmed empirically: each of the 5 searches independently already
+    # returns the same complete set of legislators.
+    for keyword in "aeiou":
+        for page in count(1):
+            data = context.fetch_json(
+                context.data_url + "legisladores",
+                params={"page": page, "keyword": keyword},
+                cache_days=1,
+            )
+            for result in data.pop("results"):
+                ids.add(result.pop("legisladorId"))
+
+            if page * data.pop("pageSize") >= data.pop("total"):
+                break
+    return ids
+
+
+def crawl_legislator(
+    context: Context,
+    positions: dict[str, tuple[Entity, PositionCategorisation]],
+    legislador_id: int,
+) -> None:
+    data = context.fetch_json(
+        context.data_url + f"legislador/{legislador_id}", cache_days=1
+    )
+
+    person = context.make("Person")
+    person.id = context.make_slug("person", str(legislador_id))
+    h.apply_name(
+        person,
+        first_name=data.pop("nombres").strip(),
+        last_name=data.pop("apellidos").strip(),
+        lang="spa",
+    )
+    # Both deputies and senators must be Dominican citizens (Constitution of the
+    # Dominican Republic, Art. 79 for senators, Art. 82 for deputies which
+    # cross-references Art. 79):
+    # https://drlawyer.com/espanol/leyes/constitucion-de-la-republica-dominicana/
+    person.add("citizenship", "do")
+    person.add("profession", data.pop("profesion", None))
+    person.add("email", data.pop("correoInstitucional", None))
+
+    party = data.pop("partido")
+    if party is not None:
+        person.add("political", party.pop("nombre"))
+
+    representation = data.pop("representacion")
+    district = representation.pop("circunscripcion")
+    role = context.lookup("role", representation.pop("funcion"))
+    if role is None or role.value is None:
+        return
+    position, categorisation = positions[role.value]
+    if not categorisation.is_pep:
+        return
+
+    constituency = representation.pop("provincia")
+    if district != "No aplica":
+        constituency = f"{constituency}, {district}"
+
+    occupancy = h.make_occupancy(
+        context,
+        person,
+        position,
+        no_end_implies_current=False,
+        start_date=representation.pop("inicio"),
+        end_date=representation.pop("fin"),
+        categorisation=categorisation,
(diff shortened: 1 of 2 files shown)


datasets/do/congress/crawler.py
```diff
@@ -0,0 +1,114 @@
...
+    if district != "No aplica":
+        constituency = f"{constituency}, {district}"
+
+    occupancy = h.make_occupancy(
+        context,
+        person,
+        position,
+        no_end_implies_current=False,
+        start_date=representation.pop("inicio"),
+        end_date=representation.pop("fin"),
```
nvmbrasserie, line 80: `inicio`/`fin` are term boundaries shared by everyone in the term (`periodo: "2024-2028"`), so these want to be `period_start`/`period_end`. As written, a future `endDate` reads as CURRENT, so every sitting member stays current until Aug 2028 even if they leave, and `no_end_implies_current=False` never fires because `end_date` is always set. `ejercicio` can't rescue us — a `2020-2024` record still says "En Curso". Same choice as `cl/chamber_deputies/crawler.py:131`.

datasets/do/congress/crawler.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def crawl_legislator(
+    context: Context,
+    positions: dict[str, tuple[Entity, PositionCategorisation]],
+    legislador_id: int,
+) -> None:
+    data = context.fetch_json(
+        context.data_url + f"legislador/{legislador_id}", cache_days=1
+    )
```
nvmbrasserie, line 39: Could we add `context.audit_data` here and in `discover_legislator_ids`? Without it the pops aren't doing any work, and we're reading about six of the record's 27 keys with no way to notice new ones.

datasets/do/congress/crawler.py
```diff
@@ -0,0 +1,114 @@
...
+            data = context.fetch_json(
+                context.data_url + "legisladores",
+                params={"page": page, "keyword": keyword},
+                cache_days=1,
+            )
+            for result in data.pop("results"):
+                ids.add(result.pop("legisladorId"))
+
+            if page * data.pop("pageSize") >= data.pop("total"):
+                break
```
nvmbrasserie, line 28: The keyword really is mandatory, so search-based discovery is fair. But it matches across several fields, not just names (`PRM` → 167, `Diputado` → 118), which is why each vowel reports `total: 224` — `a` and `e` each return all 224 ids alone. So four of the five passes are redundant, and `pageSize` is pinned at 10 server-side: ~115 requests where ~23 would do. Main thing: every response carries `total` and we never check our ids against it. That's the guard against a truncated roster.

datasets/do/congress/crawler.py
```diff
@@ -0,0 +1,114 @@
...
+        no_end_implies_current=False,
+        start_date=representation.pop("inicio"),
+        end_date=representation.pop("fin"),
+        categorisation=categorisation,
+    )
+    if occupancy is None:
+        return
+    occupancy.add("constituency", constituency)
+    context.emit(occupancy)
+    context.emit(position)
```
nvmbrasserie, line 87: (nit) This emits each position once per legislator, ~224 times. Better in `crawl()` after `categorise`.

datasets/do/congress/do_congress.yml
```diff
@@ -0,0 +1,71 @@
...
+  min:
+    schema_entities:
+      Person: 190
+      Position: 2
+    country_entities:
+      do: 190
+  max:
+    schema_entities:
+      Person: 450
+      Position: 2
```
nvmbrasserie, line 71: (nit) min 190 / max 450 against an actual ~224 (190 deputies + 32 senators) — this would pass with every senator missing. Something like 210-260 would catch drift.

datasets/do/congress/crawler.py
```diff
@@ -0,0 +1,114 @@
...
+    role = context.lookup("role", representation.pop("funcion"))
+    if role is None or role.value is None:
+        return
+    position, categorisation = positions[role.value]
+    if not categorisation.is_pep:
+        return
+
+    constituency = representation.pop("provincia")
+    if district != "No aplica":
+        constituency = f"{constituency}, {district}"
```
nvmbrasserie, line 72: (nit) Source string in code — worth a named constant or a lookup, so it's easy to spot if the source changes the wording.

datasets/do/congress/crawler.py
```diff
@@ -0,0 +1,114 @@
...
+    data = context.fetch_json(
+        context.data_url + f"legislador/{legislador_id}", cache_days=1
+    )
+
+    person = context.make("Person")
+    person.id = context.make_slug("person", str(legislador_id))
+    h.apply_name(
+        person,
+        first_name=data.pop("nombres").strip(),
+        last_name=data.pop("apellidos").strip(),
```
nvmbrasserie, line 46: (nit) `.strip()` covers the trailing spaces the source has (`'Alcibíades '`, `'Ángel '`), but `squash_spaces` would also catch the internal double spaces I saw in its concatenated names.

datasets/do/congress/do_congress.yml
herrhellana: good call yes

herrhellana:
@nvmbrasserie i think i've addressed all the comments, could you please check? thank you!

PR #5505, opened 2026-08-21
[tt_parliament] PEPs

### LINKED ISSUE #818: Trinidad and Tobago Members of Parliament
### Data URL

https://www.ttparliament.org/members/

### Publisher

https://www.ttparliament.org/

tt

### Type of data

### Coverage region

region:Latin America

### Can you tell us more?

Covers: House of Representatives and Senate. Format: html. Per-member fields: full name (with honorifics), party (UNC/PNM), constituency, photo, profile URL, biography, parliamentary career, committee membership. ~42 members; languages: en. Access notes: List is paginated (5 pages of 10 at /house/members/page/N/). WordPress REST API (wp-json/wp/v2) is blocked by the Solid Security plugin (401), so no JSON endpoint. Site's TLS certificate chain is incomplete, which broke WebFetch; curl with a browser User-Agent works without issue. No Cloudflare, captcha, or JS rendering required.

- [x] I understand.

herrhellana:
closes https://github.com/opensanctions/crawler-planning/issues/818

### DIFF
--- a/datasets/tt/parliament/crawler.py
+++ b/datasets/tt/parliament/crawler.py
@@ -1,0 +1,177 @@
+import re
+from dataclasses import dataclass
+from urllib.parse import urlencode
+
+from zavod.entity import Entity
+from zavod.stateful.positions import PositionCategorisation, categorise
+from zavod.util import Element
+
+from zavod import Context
+from zavod import helpers as h
+
+# The parliament selector labels each term like "12th Republican Parliament
+# (28 Aug 2020 - 18 Mar 2025)"; the sitting term is labelled "Current - ...".
+TERM_RE = re.compile(r"\((\d{1,2} \w+ \d{4}) - (\d{1,2} \w+ \d{4})\)")
+
+
+@dataclass
+class Post:
+    """A position together with its PEP categorisation, so both travel as a unit."""
+
+    position: Entity
+    categorisation: PositionCategorisation
+
+
+def make_post(context: Context, name: str, wikidata_id: str | None = None) -> Post:
+    position = h.make_position(
+        context,
+        name=name,
+        country="tt",
+        topics=["gov.national", "gov.legislative"],
+        wikidata_id=wikidata_id,
+        lang="eng",
+    )
+    return Post(position, categorise(context, position))
+
+
+def parse_term(label: str) -> tuple[str | None, str | None]:
+    """Extract the term's raw start and end date strings from a selector label.
+
+    Returns the source date strings as-is (parsing is left to the date helpers),
+    or `(None, None)` for the sitting parliament, which the source labels
+    "Current" and leaves undated.
+    """
+    if label.startswith("Current"):
+        return None, None
+    match = TERM_RE.search(label)
+    if match is None:
+        raise ValueError(f"Cannot parse parliament term from label: {label!r}")
+    return match.group(1), match.group(2)
+
+
+def crawl_member(
+    context: Context,
+    row: Element,
+    member: Post,
+    period: tuple[str | None, str | None],
+    speaker: Post | None = None,
+) -> None:
+    profile_url = h.xpath_string(row, "./@href")
+    slug = profile_url.rstrip("/").rsplit("/", 1)[-1]
+
+    raw_name = h.xpath_string(row, './/div[contains(@class, "member_name")]/text()')
+    name = h.strip_name_titles(context, raw_name)
+    assert name is not None
+
+    person = context.make("Person")
+    person.id = context.make_id(slug)
+    person.add("name", name, original_value=raw_name if name != raw_name else None)
+
+    constituencies = h.xpath_strings(
+        row, './/div[contains(@class, "member_constituency")]/text()'
+    )
+    constituency = constituencies[0] if constituencies else None
+
+    is_speaker = constituency == "Speaker of the House"
+    if is_speaker:
+        if speaker is None:
+            raise ValueError(f"Unexpected Speaker row outside the House: {profile_url}")
+        post = speaker
+    else:
+        post = member
+
+    if not post.categorisation.is_pep:
+        return
+
+    parties = h.xpath_strings(
+        row, './/div[starts-with(@class, "member_affiliation")]/text()'
+    )
+    if parties:
+        party_res = context.lookup("party", parties[0])
+        person.add("political", party_res.value if party_res else parties[0])
+    # Constitution of the Republic of Trinidad and Tobago, ss. 41 (Senate) and 47
+    # (House of Representatives), both requiring Trinidad and Tobago citizenship:
+    # https://www.constituteproject.org/constitution/Trinidad_and_Tobago_2007
+    person.add("citizenship", "tt")
+    person.add("sourceUrl", profile_url)
+
+    period_start, period_end = period
+    occupancy = h.make_occupancy(
+        context,
+        person,
+        post.position,
+        period_start=period_start,
+        period_end=period_end,
+        categorisation=post.categorisation,
+    )
+    if occupancy is None:
+        return
+    if not is_speaker and constituency is not None:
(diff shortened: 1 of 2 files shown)


datasets/tt/parliament/crawler.py
```diff
@@ -0,0 +1,177 @@
...
+        pid = h.xpath_string(option, "./@value")
+        if pid == "All":
+            continue
+        period = parse_term(h.xpath_string(option, "./text()"))
+        period_end = period[1]
+        if period_end is not None:
+            end = h.extract_date(
+                context.dataset, period_end, fallback_to_original=False
+            )[0]
+            if end < h.earliest_term_start(house.position.get("topics")):
```
brrttwrks, line 169: I think we want to avoid doing this here. This is handled in the make_occupancy method, if I'm not mistaken.

brrttwrks (commented):
This looks good. Just check the one thing - redundant inclusion check.

datasets/tt/parliament/crawler.py
herrhellana: this check is not redundant, its purpose is to skip processing entire legislative terms outside the PEP coverage window. `make_occupancy()` only drops individual occupancies outside of the coverage window, but it doesn't prevent us from first processing terms, where each position is known to be outside of the coverage window. these two operate at different levels (roaster vs. individual positions).

so the `earliest_term_start()` helper saves processing times and avoids paginated fetches of old legislatures we don't need.

nvmbrasserie (commented):
Nice first pass — the typed HTML helpers, `strip_name_titles` with `original_value`, and the `earliest_term_start` gating all look good. A few things below, mostly about keeping position decisions out of the crawler.

datasets/tt/parliament/crawler.py
```diff
@@ -0,0 +1,177 @@
...
+    )
+    senate = make_post(
+        context, "Member of the Senate of Trinidad and Tobago", "Q19319420"
+    )
+    # No Wikidata QID: the closest candidate, Q109538297, is `instance of` (P31)
+    # "public office" rather than "position", and carries no `applies to
+    # jurisdiction` (P1001) statement, so it fails the check for supplying a QID.
+    speaker = make_post(
+        context, "Speaker of the House of Representatives of Trinidad and Tobago"
+    )
```
nvmbrasserie, line 156: Let's keep just one position here — `Member of the Parliament of Trinidad and Tobago`, with its topics — created once in `crawl()`. Hardcoding a table of positions with looked-up QIDs puts decisions in the code that belong in the categorisation UI.

This also lets `Post`/`make_post` and the Speaker handling go away.

datasets/tt/parliament/crawler.py
```diff
@@ -0,0 +1,177 @@
...
+from zavod.entity import Entity
+from zavod.stateful.positions import PositionCategorisation, categorise
+from zavod.util import Element
+
+from zavod import Context
+from zavod import helpers as h
+
+# The parliament selector labels each term like "12th Republican Parliament
+# (28 Aug 2020 - 18 Mar 2025)"; the sitting term is labelled "Current - ...".
+TERM_RE = re.compile(r"\((\d{1,2} \w+ \d{4}) - (\d{1,2} \w+ \d{4})\)")
```
nvmbrasserie, line 14: Could we split this rather than match it? `partition("(")`, `rstrip(")")`, `partition(" - ")` gets both halves, and the date helpers already know the format from `dates.formats`. The regex reads like validation, but `\w+` accepts any word as a month. We treat regexes as a last resort in parsing generally.

Keep the two halves raw when you pass them on — `make_occupancy` keys the occupancy ID on the date strings as given.

datasets/tt/parliament/crawler.py
```diff
@@ -0,0 +1,177 @@
...
+def make_post(context: Context, name: str, wikidata_id: str | None = None) -> Post:
+    position = h.make_position(
+        context,
+        name=name,
+        country="tt",
+        topics=["gov.national", "gov.legislative"],
+        wikidata_id=wikidata_id,
+        lang="eng",
+    )
+    return Post(position, categorise(context, position))
```
nvmbrasserie, line 34: I'd skip the wrapper type and pass `position` and `categorisation` as plain arguments — `datasets/ke/national_assembly/crawler.py` does it that way, and it's easier to follow at the call sites. With a single position there's nothing left to bundle.

datasets/tt/parliament/crawler.py
```diff
@@ -0,0 +1,177 @@
...
+
+def crawl_member(
+    context: Context,
+    row: Element,
+    member: Post,
+    period: tuple[str | None, str | None],
+    speaker: Post | None = None,
+) -> None:
+    profile_url = h.xpath_string(row, "./@href")
+    slug = profile_url.rstrip("/").rsplit("/", 1)[-1]
```
nvmbrasserie, line 60: Let's key on the raw value: `context.make_id(raw_name)`. We always use the source value as-is for IDs rather than something derived from it — see the [entity ID guide](https://github.com/opensanctions/opensanctions/blob/main/zavod/docs/best_practices/entity_id.md).

datasets/tt/parliament/crawler.py
```diff
@@ -0,0 +1,177 @@
...
+    )
+    constituency = constituencies[0] if constituencies else None
+
+    is_speaker = constituency == "Speaker of the House"
+    if is_speaker:
+        if speaker is None:
+            raise ValueError(f"Unexpected Speaker row outside the House: {profile_url}")
+        post = speaker
+    else:
+        post = member
```
nvmbrasserie, line 81: Giving the Speaker its own position means passing a `speaker` argument down through `crawl_chamber` into `crawl_member` to route one row per term, plus a `ValueError` for when it turns up in the Senate. With a single position none of that is needed — the only thing left is that the House roster puts `Speaker of the House` where other members have their constituency, so we just shouldn't write it to `Occupancy:constituency`.

datasets/tt/parliament/crawler.py
```diff
@@ -0,0 +1,177 @@
...
+    query = urlencode(
+        {"members_search": "1", "keywords": "", "pid": pid, "house": house}
+    )
+    url: str | None = f"{context.data_url}?{query}"
+    seen: set[str] = set()
+    while url is not None:
+        if url in seen:
+            raise ValueError(f"Pagination loop detected at {url}")
+        seen.add(url)
+        doc = context.fetch_html(url, cache_days=1)
```
nvmbrasserie, line 133: `cache_days=1` on the paginated listing is worth another look. Paginated pages are the case `caching.md` specifically warns about: when a member is added, everyone after them shifts across page boundaries, so a stale page combined with a fresh one can drop or duplicate someone.

Concluded terms are the opposite situation — those rosters are settled and don't need refetching every run. No caching for the sitting term and a longer window for the rest would give us fresher current data and a cheaper crawl at the same time.

datasets/tt/parliament/crawler.py
```diff
@@ -0,0 +1,177 @@
...
+
+    period_start, period_end = period
+    occupancy = h.make_occupancy(
+        context,
+        person,
+        post.position,
+        period_start=period_start,
+        period_end=period_end,
+        categorisation=post.categorisation,
+    )
```
nvmbrasserie, line 106: A question rather than a change request. Every row carries a `current_status` div — `Current Member` or `Former Member` — which we don't read. For dated terms it likely doesn't matter, since `make_occupancy` derives status from `period_end`. But the sitting parliament has no dates, so every row comes out as `status: current`, including the 22 of 95 RP13 rows the source marks Former Member.

Worth establishing: is the flag about the person or about that specific term? Past terms show both values, which points to person-level. And is `ended` right with no end date to back it, or is `unknown` more honest?

nvmbrasserie (commented):
Two follow-ups: one on caching, one question about a field we're not reading.

PR #5506, opened 2026-08-21
[il_mod_crypto] switch to up to date source

nvmbrasserie:
Switch `il_mod_crypto` to the Hebrew source page, which is a strict superset of the
English one: 35 vs 34 orders, 490 vs 448 wallet addresses.
The English page is currently missing ASO 18/26 (37 wallets, published 01.07.2026)
entirely, plus 5 ASO 1/25 wallets that have been up in Hebrew since Feb 2025.

- refresh the `releases.csv` / `wallets.csv` diff snapshots and the DOM hash
- add `maintenance.md` documenting how the manually maintained Google Sheet is
  filled in (schema, start/end dates, platform, order PDF link)

The 42 wallets are already in the sheet, so the next run picks them up.


brrttwrks (approved):
This looks good to me. Thanks for assessing the diff between the two pages.

PR #5515, opened 2026-08-23
Version-scoped run artifacts, explicit version threading, and run manifests

### LINKED ISSUE #3786: Stop backfilling `statistics.json` for failed runs
The easy way is to just disable backfill here:

https://github.com/opensanctions/opensanctions/blob/7ffef77115d19e6cb41fe878d129203005ab4d4f/zavod/zavod/exporters/metadata.py#L44-L52

But the more correct way would be to set the version to backfill. Which, usually, is the latest local version.

But then I see this beauty:

https://github.com/opensanctions/opensanctions/blob/7ffef77115d19e6cb41fe878d129203005ab4d4f/zavod/zavod/archive/__init__.py#L147-L155

So calling `get_artifact_object` with a `version` set will always backfill something, even if it's not that version. But that's not what we want.

So let's figure out a path here.

As part of this bug, we should also get rid of this comment here: https://github.com/opensanctions/opensanctions/blob/7ffef77115d19e6cb41fe878d129203005ab4d4f/zavod/zavod/publish.py#L81-L86

Spun out from https://github.com/opensanctions/opensanctions/issues/2541

### LINKED ISSUE #4835: Collections should export a manifest.json
When we export a collection, the system will automatically backfill and then include the latest successful run of each dataset in the collection. Unfortunately, that means that the precise composition of the collection in terms of dataset versions becomes a bit of an ambiguous thing. For example, a dataset crawler could have started its work at T+0, then the collection export starts at T+1, and the crawler uploads its result at T+3, but with the timestamp of T+0 (the `settings.RUN_TIME`). That means we can't really track back which versions are in the collection, because the "dataset time is smaller than T+1" heuristic fails. 

I'd suggest we make a standard artifact called manifest.json. It could contain: 

```
{
  "datasets": {"us_ofac_sdn": "<version>", "eu_fsf": "<version>"}
}
```

(Just realized that we can divine it from the `last_seen` statement metadata at the moment, but that's horrible.)

pudo:
Fixes opensanctions/operations#2675, fixes #4835, fixes #3786. Replaces #4873.

Every stage of a run now operates on an explicit, verifiable version. Local run artifacts move from a flat `data/datasets/<name>/` into `data/datasets/<name>/_artifacts/<version>/`, mirroring `artifacts/<name>/<version>/` in the archive, and the version travels as a `followthemoney.dataset.Version` from the CLI boundary down.

## Run manifests (#4835)

A `Manifest` (`zavod.runtime.manifest`) records which version of each leaf dataset a run consumes, resolved **once** when the run starts instead of lazily inside the store builder:

- The crawl creates it for source datasets; the export creates it for collections. `manifest.json` is archived with the other run artifacts, so the composition of any published collection version can be inspected.
- Resolution per leaf: the run's own version → newest local artifact dir with statements → last successful archive version → omitted with a warning (visible in the artifact).
- The store syncs against the manifest digest (datasets mapping only): matching pins reuse the LevelDB, changed pins auto-rebuild. `--keep-store` is now exact instead of hopeful.
- Analytical commands (`xref`, `dedupe`, `dedupe-edges`, `wikidata-reconcile`) persist a transient manifest under `_state/`, so consecutive runs share a stable scope; `-r` re-resolves it.
- All statement access goes through `manifest.statements()`, which reads pinned versions locally or streams them from the archive and **errors on absence** — replacing `iter_dataset_statements` and its silent fallback to stale archive data. `load-db` and `dump-file` iterate the pinned set directly, so a `zavod run` can no longer export one version set and load-db another.

## Pipeline changes

- `zavod crawl/export/publish/run` take a version (`-v` / positional); the process default is a fresh version, and no environment variable is consulted — `ZAVOD_VERSION` and `ZAVOD_RELEASE` are no longer read (the legacy date-stamped CDN purge derives from the version). Stages running in separate processes must agree on a version explicitly. Standalone `zavod export` of a collection now works (export is the producer for collections).
- Producer/consumer split: the producer of a run (`Manifest.create`) resets the version dir; consumers load the manifest and hard-fail if it's absent rather than fabricating input.
- The `versions.json` snapshot is written when a run concludes (`publish_dataset` on success, `archive_failure` on failure), not at run start — so a run that fails before its artifact directory was created still registers its version in the history. `publish_dataset` refuses a version that has no run manifest.
- Delta backfill and `entities.hash` are keyed on the last successful version; the hash file lands in the version dir so publish archives it again.
- A failed run's `index.json` no longer carries backfilled statistics: `statistics.json` is read strictly from the run's own artifact directory, so failure indexes have no stale entity counts (#3786; pinned by `test_metadata_failure_no_statistics_no_warning`).
- The timestamp index is cached at `_state/timestamps/<last-successful-version>/` with a completion marker, so it is rebuilt only when a new version is published. Caveat: `zavod crawl/run` default to `--clear-data`, which wipes `_state/` including this cache — it only pays off with `--keep-data` for now.
- Disabled datasets in `zavod run` exit 0 without registering a version or a failure index: a disabled dataset has no run to record, and its last published version stays authoritative.
- Dry-run mode is removed. It existed to keep debug crawls from corrupting the shared output directory (now structurally impossible: version dirs are isolated and nothing leaves the machine before `publish`) and to skip the timestamp index build (now cached). A plain crawl into its unpublished version dir is the debug tool, with full statements/issues/resources output.

## Testing

The test suite is fully ported to the new semantics: explicit `Version` threading instead of `settings` monkeypatching, assertions on version-scoped artifact paths and `Manifest.statements()`, and shared run primitives in `zavod.tests.util` (`get_manifest`, `make_context`, `get_test_view`, `run_dataset`). Full suite and `mypy --strict` green. The contrib scripts (`egrul`, the wikidata debug tools) are updated for the new `Context` contract.

## Must land before merge / known gaps

- **kombinat coordination**: ported in opensanctions/operations#2813 (catalog generation onto `get_best_version`/`backfill_artifact`, xref scopes and PEP statistics onto transient manifests, release date derived from the run time) — merge that in lockstep with the first image built from this PR.
- Crawler-emitted resources still write to the flat path with a two-location fallback in publish; unifying is a follow-up.
- No GC yet for accumulated `_artifacts/<version>/` and `_state/timestamps/<version>/` dirs — to be solved deliberately, not per-cache.
- `ZAVOD_ARCHIVE_BACKFILL_STATEMENTS` and `ZAVOD_BACKFILL_RELEASE` are deleted (no remaining consumers); re-introduce on the manifest path if a use case returns.


zavod/zavod/archive/__init__.py
```diff
@@ -100,19 +97,15 @@
     STATISTICS_FILE,
     DELTA_EXPORT_FILE,
 ]
-# Files we persist about a run other than DatasetResources.
-EXTRA_ARTIFACTS = [
+# The only files a failed run publishes. Half-generated data files stay local.
+FAILURE_ARTIFACTS = [
```
jbothma: > Half-generated data files stay local.

what do we mean by this? in one sense, it could be nice if we only upload what's been registered in the manifest. on the other hand, I'm not sure how we'd half-generate a file without crashing.

zavod/zavod/archive/__init__.py
```diff
@@ -140,125 +133,140 @@ def clear_data_path(dataset_name: str) -> None:
...
+
+def get_best_version(dataset_name: str) -> Version | None:
+    """Get the best version of a dataset, ie. the last successful one if available,
+    otherwise the latest one."""
+    history = get_version_history(dataset_name)
+    return history.last_successful or history.latest
 
 
 def iter_dataset_versions(dataset_name: str) -> Generator[Version, None, None]:
     """Iterate over all versions of a given dataset."""
```
jbothma, line 180: Am I right in thinking deltas is the only remaining use case where we're ok with stepping back through versions (until we find a version where there was a delta). if so, a nice change might be to make iter_dataset_versions really just a private part of deltas.

an alternative could be introducing `VersionHistory.last_delta: Version` [which might be nulled when delta exporter is removed ](https://github.com/opensanctions/opensanctions/issues/5140) (out of scope here).

zavod/zavod/runtime/manifest.py
```diff
@@ -0,0 +1,152 @@
...
+                leaf resolves to its best available version.
+
+        Each leaf resolves to the newest local artifact directory holding a
+        statements file, then to the last successful version in the archive."""
+        datasets: dict[str, Version] = {}
+        for leaf in dataset.leaves:
+            if version is not None and leaf.name == dataset.name:
+                datasets[leaf.name] = version
+                continue
+            local = latest_local_artifact_version(leaf.name, STATEMENTS_FILE)
```
jbothma, line 83: is this currently only to support the dev use case where we're figuring something out at the collection level and we want to use a local run of one of its leaves? If so, it could be nice to document that (maybe both here and at the function definition). It's handy functionality but worth pointing out it's freaky and not intended for prod behaviour

zavod/zavod/exporters/metadata/__init__.py
```diff
@@ -168,25 +168,28 @@ def get_catalog_dataset(dataset: Dataset) -> dict[str, Any]:
...
-    # Get a barebones metadata object, only relevant before the first export
-    meta = get_base_dataset_metadata(dataset)
-
+    meta = {}
     # Use the latest published index file, if available.
-    path = get_dataset_artifact(dataset.name, INDEX_FILE)
-    if path.is_file():
-        with open(path) as fh:
-            meta.update(json.load(fh))
+    version = get_best_version(dataset.name)
```
jbothma, line 173: when would we want a failed version when there's never been a successful run? a premature release? Shoudl it rather be a warning and skip?

zavod/zavod/exporters/metadata/__init__.py
```diff
@@ -70,11 +77,11 @@ def get_base_dataset_metadata(dataset: Dataset) -> dict[str, Any]:
             # so an emptied dataset's last change is whatever the last successful
             # run reported.
             if last_change is None:
-                last_change = settings.RUN_TIME_ISO
+                last_change = viso
```
jbothma, line 80: How about spelling the meaning out a bit more

```suggestion
                last_change = version_date_iso
```

zavod/zavod/tests/exporters/test_metadata.py
```diff
@@ -42,8 +43,14 @@ def test_metadata_collection_export(
         # The written index conforms to the output contract zavod validates against.
         CatalogDatasetModel.model_validate(index)
 
-    collection_path = settings.DATA_PATH / "datasets" / collection.name
-    export_dataset(collection, view)
+    # In a production run, sources are published before a collection exports;
+    # the catalog takes each dataset's run information from its last published
+    # version.
+    publish_dataset(testdataset1, version)
```
jbothma, line 49: I would love these tests running at the `zavod.cli.etl.run` level so they're a bit more representative and stay in sync with that `run` but just how to do that hasn't quickly revealed itself to me yet. Maybe in one of the next PRs. test.util.run_dataset is probably the solution though - nice.

zavod/zavod/archive/__init__.py
jbothma, line 101: lol an allow list is a lot simpler than deleting a bunch of files!

zavod/zavod/publish.py
```diff
@@ -1,51 +1,104 @@
...
 
 log = get_logger(__name__)
 
 
-def _archive_artifacts(dataset: Dataset, extra_artifacts: list[str] = []) -> None:
-    """
-    Upload every file we persist about a run to /artifacts/{dataset}/{version}/.
-
-    Also publishes the version history to the dataset's stable version history location.
+def _ensure_versions_file(
```
jbothma, line 23: would it be nice if we explicitly created the version file in (hopefully a handful of) clear points in the pipeline, and just assert it exists like we're asserting the part exists, on publish/archive_failure?

it feels like something that'd be nice not to be defensive on but I can't visualise all the those points right now

jbothma (approved):
nice to be up to date with these changes

zavod/zavod/exporters/metadata/__init__.py
leonhandreke: I personally think we should warn and skip if no last successful version exists. It's a very edge-y case anyway.

leonhandreke:
Question - what do do about `catalog.json`, now that we have the concept of a manifest? Currently, it resolves to the last successful version of the scope itself, all sub-collections and all leaves. That is true both for the root-level kombinat and the dataset-level `catalog.json`

For the root, we want to keep this behavior.

For the dataset-level `catalog.json`, it's less clear what to do:
1. get rid of dataset-level `catalog.json` entirely (probably not)?
2. read leaves and scope version as defined in `Manifest`
  a. drop sub-collections from `catalog.json`
  b. use last-successful for sub-collections? (meh)

I feel 2a or 1. But I have no idea what uses dataset-level `catalog.json` and for what, and whether it's at all useful now that `manifest.json` exists.

jbothma:
how about 2c - keep sub-collections but drop the metadata that doesn't make sense as a super-collection member?

Is the point of the catalog so that yente has dataset metadata? when yente makes a dataset from its datasets listing, they're a much more limited version than what we publish, aren't they? Can we perhaps define a pydantic model which get_catalog_dataset's result meets, so that e.g. collections don't need a version? But they can have a title and description.

pudo:
We can't get rid of `catalog.json` - it's yente-readable and many of our peeps use it (eg. `kyb/catalog.json`). 

I think having a datasets `catalog.json` based on and consistent with `manifest.json` would be really awesome :)

zavod/zavod/exporters/metadata/__init__.py
leonhandreke: I did it this way.

zavod/zavod/exporters/metadata/__init__.py
leonhandreke: Done.

zavod/zavod/runtime/manifest.py
leonhandreke: I added a warning

zavod/zavod/archive/__init__.py
leonhandreke: The vibes tell me:

```
iter_dataset_versions has two production uses:

  - metadata/init.py:138 (zavod/zavod/exporters/metadata/__init__.py:138): write_dataset_index finds the newest earlier delta export when the current run has no delta index.
  - metadata/init.py:259 (zavod/zavod/exporters/metadata/__init__.py:259): write_delta_index collects up to 100 earlier delta exports.
```

I think both could probably work the same way that `versions.json` works, i.e. by appending/overwriting the previous value, but only if something changed. But I think for now, this is fine and this PR is already large enough.

leonhandreke:
Sometimes you gotta pull the trigger.

PR #5535, opened 2026-08-25
[eu_meps] Switch to the European Parliament Open Data API

leonhandreke:
## Why
The MEP list XML endpoint (`.../meps/en/full-list/xml`) is behind an AWS WAF challenge. The Zyte browser workaround stopped passing it, so `eu_meps` has failed every run since 2026-08-22.

## Background and timeline
This crawler has read the MEP list XML endpoint since 2018 (https://github.com/opensanctions/opensanctions/commit/f49f97788), and stayed on it in its current Zyte/WAF form. The European Parliament only launched its Open Data Portal and API in January 2023, years after the crawler started, which is why we never used it before. The old XML feed also lists current members only and carries no dates.

## What
Fetch from the European Parliament Open Data API v2 (https://data.europarl.europa.eu/api/v2, licensed CC BY 4.0) instead. No WAF bypass, no Zyte. The API is a clearly better source:

- Covers current and former MEPs (all terms since 1979); we keep the PEP-relevance window via `h.earliest_term_start`, not only sitting members.
- Emits one dated `Occupancy` per parliamentary term, with real start and end dates and status — the XML feed had none.
- Emits the presidency, vice-presidency and quaestorship a member holds alongside the mandate as their own dated Positions, by Wikidata QID.
- Adds gender, citizenship, birth date and birth place.
- Records the national party on `Person.political` and the parliamentary group on `Occupancy.politicalGroup`, following the house convention for legislature crawlers.

## Rate limiting
The API allows 500 requests per 5 minutes, and a full run makes ~4000 calls. A small leaky-bucket rate limiter paces requests, gated on the cache so a cached rerun is not slowed.

## Identifiers
Persons keep their `eu-meps-{id}` ids — the API MEP id equals the old XML `<id>`, so no person re-keying.

## Final counts
A full crawl emits 7,320 entities:

| schema | count | why |
|---|---|---|
| Person | 2,711 | of the 3,702 MEPs in terms 4-10, those with a mandate inside the ~20-year PEP window; ~1,000 pre-2006-only members drop out |
| Occupancy | 4,605 | ~1.7 per person: several terms, plus the presidency periods |
| Position | 4 | Member, President, Vice-President and Quaestor of the European Parliament |

Party and group are text on `Person.political` and `Occupancy.politicalGroup`, so there are no Organization or Membership entities. Largest contingents by country match the big member states plus historical UK MEPs: Italy 307, France 290, Germany 278, Spain 197, UK 185.

## Notes
- `ci_test` stays false: ~4000 API calls per run is too many for CI.
- Assertions are set from this crawl: Person 2300-5000, Position 1-5.

🤖 Generated with Claude Code using Opus 4.8

### DIFF
--- a/datasets/eu/meps/crawler.py
+++ b/datasets/eu/meps/crawler.py
@@ -15,6 +15,70 @@ from zavod.stateful.positions import PositionCategorisation, categorise
 # One term roster fits in a single response; this bounds it and flags overflow.
 ROSTER_LIMIT = 10000
 
+# Fields each record type carries that the crawler deliberately leaves unmapped.
+# Every object the API returns has `id` (the JSON-LD node IRI) and `type` (the
+# JSON-LD @type tag).
+
+# A PeriodOfTime, as carried by `temporal` and `memberDuring`.
+PERIOD_IGNORE = ["id", "type"]
+
+# A parliamentary term, i.e. the `org/ep-{n}` corporate body.
+TERM_IGNORE = [
+    "id",  # JSON-LD node IRI
+    "type",  # JSON-LD @type tag
+    "identifier",  # term number, already the loop index
+    "label",  # term number as a label
+    "altLabel",  # always "European Parliament"
+    "prefLabel",  # always "European Parliament"
+    "classification",  # always EU_INSTITUTION
+    "represents",  # not set on the institution body
+    "source",  # EP provenance flag
+    "linkedTo",  # refs to related bodies
+    "isVersionOf",  # parent body ref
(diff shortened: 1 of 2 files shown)


leonhandreke:
@nvmbrasserie would appreciate if you could take this over next week - I think it's already in an okay place. If you don't get to it, just leave it for me when I'm back.

nvmbrasserie (commented):
I've added a few suggestions here: https://github.com/opensanctions/opensanctions/pull/5573. Output looks good

PR #5552, opened 2026-08-27
[tr_kap] Add Türkiye KAP listed company ownership and boards crawler

### LINKED ISSUE #856: Türkiye KAP (Public Disclosure Platform) — listed company ownership and boards
### Data URL

https://www.kap.org.tr/tr/sirketler/ALL

### Publisher

Merkezi Kayıt Kuruluşu A.Ş. (MKK) — Central Securities Depository of Türkiye

TR

### Type of data

Companies (National registers of companies, organizations, etc)

### Coverage region

region:Asia/Pacific

### Can you tell us more?

KAP (Kamuyu Aydınlatma Platformu / Public Disclosure Platform) is Türkiye's statutory
disclosure channel for capital markets. It is defined in the Capital Markets Law No. 6362,
art. 3(1)(k), and the CMB's Public Disclosure Platform Communiqué requires that all
information subject to public disclosure be filed there with electronic signature. Operated
by MKK since March 2014 (previously run within Borsa İstanbul from June 2009), under CMB
regulation. Publisher should be marked official — same shape as FINRA, which we already
treat as official: statutory mandate, non-governmental legal form.

**Why it's worth doing:** it carries real, structured ownership for Turkish listed
companies, which is otherwise very hard to get. The national trade registry gazette
(ticaretsicil.gov.tr) publishes far richer ownership data for all ~1.9M Turkish companies,
but as non-searchable 300 DPI scans with no enumeration surface, so it is not crawlable —
TOBB sells the parsed equivalent as a paid subscription. KAP is the one Turkish corporate
ownership source that is free, public and structurally clean. Small universe, but it is the
public face of MKK's ownership data, so there is no separate MKK source to chase.

**Universe:** ~1,110 KAP member entities, of which ~746 are BIST-traded.

**Access:** unauthenticated, no CAPTCHA, no rate limiting observed, no geoblocking (checked
from ~40 vantage points incl. Frankfurt, all 200). Next.js SSR with the full JSON payload
embedded in the HTML. Company detail at `/tr/sirket-bilgileri/genel/{id}-{slug}`, summary
at `/tr/sirket-bilgileri/ozet/{id}-{slug}`. Full crawl estimated at ~466 MB and ~25 minutes.

**Fields available** (verified on live company pages):

- Ownership: `shareholdingStructureGrid` — `shareholder`, `shareInCapital`, `ratioInCapital`,
  `votingRightRatio`. Covers direct and indirect holders above 5%.
- Boards: `boardMembersGrid` — name, title, first-elected date, independence flag,
  executive/non-executive.
- Identifiers: `tradeRegNo`, `tradeRegOffice`, `tradeRegDate`, `taxNo`, `taxOffice`.
- Capital: `paidCapital`, `authorizedCapital`, `nominalValueOfShares`,
  `nominalValuePerShare`, `registeredOrBearerShare`, `shareGroup`, `typeOfPrivilege`.
- Market: `stockCode`, `market`, `marketOrStockExchange`, `relevantSubMarketOrExchange`,
  `bistIndices`.
- Contact/company: address, city, website, sector, scope of activities.

**Two implementation gotchas, both verified:**

1. **The field named `isin` is not an ISIN.** It holds the BIST ticker — the column header
   binds `"columnName":"Borsa Kodu"` to `"columnKey":"isin"`, and the value is e.g. `ADEL`
   for Adel Kalemcilik. Four characters, not twelve. Mapping it to `Company:isinCode` would
   silently fill the property with tickers. A search of company and listing pages found no
   12-character ISIN-shaped token anywhere. Real Turkish ISINs are issued by MKK as the
   National Numbering Agency but are not surfaced through KAP company profiles.
2. **KAP emits national ID numbers in its HTML.** `tcknYkn` / `tcknYknVkn` keys are present
   in the payload, carrying base64-encoded 11-digit TCKNs that are not rendered in the page.
   Any KAP-derived dataset must strip these deliberately rather than by accident.

**Caveat on data status:** KAP is an official *disclosure* platform, so content is
issuer-declared under legal obligation, not independently verified by MKK or the CMB. Same
epistemic status as SEC EDGAR filings. Shareholding percentages are what the company says
they are.

**Not checked:** whether debt-instrument or fund pages carry real ISINs — guessed paths
404'd. Worth closing that gap if ISIN linkage matters. Also could not confirm LEI on company
pages (zero occurrences found), despite secondary reporting suggesting subsidiaries carry
one.

Related: #234 (Turkey Companies umbrella ticket) — KAP is listed there as one of several
candidate sources under "Istanbul stock exchange", but has never been scoped or split out.

- [X] I understand.

tamimbinhakim:
Adds a crawler for **KAP, Türkiye's Public Disclosure Platform**: all KAP member companies with their shareholders, boards and subsidiaries.

## Source
KAP is a Next.js site; the data is in the server-rendered flight payload, which the crawler parses (no API found). The list page gives 1,114 members and their permalinks; each general information page carries identifiers, shareholders, board and subsidiaries. Two things differ from the ticket:
- KAP does rate-limit: after a few hundred pages it answers 429, and before that it starts serving 200 pages without the company payload. Detail pages go through Zyte per `http_operations.md`, so `ci_test: false`. The list page is one request and is fetched directly.
- 73 members (types IGMS and DG) have no general page at all; they are emitted with name and city only.

## Output
- `Company` per member (ID from KAP's numeric code): name, alias, ticker, tax and trade registry numbers, incorporation date, capital, sector, status, address, website, e-mail. `corp.public` for traded companies.
- `Ownership` per shareholder row (direct and indirect, `ownershipType`), with percentage and share value. Placeholder rows (Diğer, Halka Açık, Toplam, Yok) are skipped by lookup. When a shareholder's name is exactly another KAP member's title, the owner is that member; otherwise a per-company `LegalEntity`.
- `Directorship` per board member: `Person` with gender and profession, or `LegalEntity` where a representative acts for a legal-person member; role, first election date, independence/executive flags.
- `Ownership` per subsidiary: `lei-` IDs when KAP gives an LEI (joins GLEIF), else per-company; percentage, share value, relation.
- The `isin` column is the BIST ticker (as the ticket notes) and maps to `ticker`. National ID numbers (TCKN) present in the page data, including 11-digit values in subsidiary `taxNo` fields, are never emitted.

## Validation
- `zavod crawl` (Zyte): 27,162 entities: 4,694 Company, 5,726 Person, 3,495 LegalEntity, 7,137 Ownership (2,891 direct, 662 indirect, 3,584 subsidiary), 5,882 Directorship. All 1,041 members with a page yielded data. 7 warnings, all `Invalid LEI` where a filer put a registry number in the LEI column; happy to downgrade those to info if you prefer quiet runs.
- `zavod export` with assertions passes (bands at roughly 80% / 200% of these counts).
- Spot check: Koç Holding's shareholders (43.75 / 7.29 / 6.15), 40 subsidiaries and 11 board members match the live page.
- No 11-digit identifiers in the output (checked every property).
- ruff, `mypy --strict`, yamllint clean.

Not included: senior managers and portfolio managers (separate tables), share classes and listing details, and the activity description (HTML legalese). Happy to add any of these if useful.

Closes opensanctions/crawler-planning#856


pudo (commented):
Thanks for this — the entity model and the handling of the identity numbers look right. Before we go further on the code style, I want to change the data path: the site has a plain JSON API behind the Next.js frontend, so the flight-payload scraping (`flight_payload`, `parse_json_at`, `json_after_key`, `parse_items`) can go entirely.

What I found on the live site (via Zyte, our network can't reach kap.org.tr directly):

**1. The company list is a JSON endpoint.** One call returns all 1,126 members, including the header fields the crawler currently pulls off each detail page: tax number, trade registry number and date, paid capital, ticker, city, and the numeric `companyCode` we use for the entity ID.

```
GET https://www.kap.org.tr/tr/api/company/items/ALL/A
Accept-Language: tr
```

This replaces the list-page payload parsing, the permalink map, and the `rfind("{")`-before-`kapMemberTitle` header extraction.

**2. Detail pages can be fetched as a pure RSC stream.** Request the same page URL with the header `RSC: 1` and you get `text/x-component`: newline-delimited `id:json` records, each parseable with `json.loads` as-is. Walk the parsed tree for dicts carrying `itemObject` and you have the same `itemKey -> value` map. Verified on Koç Holding: 29 items, identical to what the current parser yields. No regex over `self.__next_f.push`, no bracket counting.

The `RSC: 1` header has to go through Zyte, so `zyte_api.fetch_text` needs a `headers` parameter (the docstring already mentions one, the signature doesn't have it). Small zavod change, happy to take that or leave it to you.

**3. There is also a per-item history API**, `GET /tr/api/company-detail/get-history/<mkkMemberOid>/<itemKey>/N`. For the board and subsidiary tables its first entry equals the current page value, but for the direct-shareholder key the entries came back unordered with reformatted numbers, and it's one request per item. I would not build on it; noting it in case it's useful later for change tracking.

I probed for a single "all items of one company" endpoint and grepped the endpoint table in all 20 JS bundles — there isn't one, so it stays at one request per company, just a much simpler one.

**Other things for the rework**

- `clean`, `keyed_text`, `keyed_key`, `clean_percent` re-implement what lookups and the `h.` helpers should cover. Placeholder values like `-`, `---`, `YOK` belong in a lookup on the property, not in a generic string cleaner; percentages should go through the framework's number handling.
- The member types without a page: the list API shows 56 IGMS + 14 DG = 70, versus 73 in the PR description. Worth re-checking once the list source changes.
- The header dict from the list API has `taxOffice`, `tradeRegOffice`, `kayitliSermayeTavani`, `relatedMemberTitle` (the auditor) etc. — decide explicitly per field and run them through `audit_data`.

tamimbinhakim:
Thanks, that was a good find. Switched it over:

- The list comes from `/tr/api/company/items/ALL/A` now. The flight payload parsing, the permalinks and the header extraction are all gone. It turns out the detail page also takes `mkkMemberOid` directly, so no permalinks are needed there either.
- Detail pages are fetched with `RSC: 1`. I added `headers` to `zyte_api.fetch_text` as its own `zavod:` commit, with a test.

Two things about the RSC stream you might not have hit:

- Not every row is one line of JSON. Longer text comes as `id:T<hex byte length>,<text>` and can contain newlines, and the item value then just says `"$43"`. So the parser reads rows using that length and resolves the references. Anything it doesn't recognise raises. All 1,126 pages parse.
- KAP's cache ignores `Vary: rsc`. If someone opened the normal page in the last half hour, you get cached HTML back even with the header. Adding `?_rsc=1` like the Next.js client does fixes that, and the fetch asserts `text/x-component` in case it comes back.

On your other points:

- `clean`, `keyed_text`, `keyed_key` and `clean_percent` are gone. Placeholders are type lookups now, and so are the messy website and e-mail values and the registry numbers people typed into the LEI column. That was also the source of the 7 warnings, so the run is clean now. Numbers go through `numbers:` on the dataset. That means `h.apply_number` for capital, and `registry.number.parse` with the same settings for percentage and shares value, since those are string props. The one helper left reads the select fields (`{"key": ..., "text": ...}`), because some filers typed free text into them.
- Members without a page: 56 IGMS + 14 DG, plus 7 DDK (Borsa İstanbul, SPK, MKK etc.) that also don't have one, so 77. The 73 was from the old HTML list, which only had 1,114 members.
- List fields: the ones in use are title, code, tax no, trade reg no and date, paid capital, city, stock code, trading status, member type and state. The rest are in `audit_data` ignore with a reason: tax/registry office, capital ceiling, the auditor, and KAP's internal type flags. Three `KAP TEST` accounts and the BISTECH circuit breaker entry are skipped.
- Agreed on the history API, left it out.

Two fixes I found along the way:

- `stockCode` can be `GARAN, TGB`, and banks and brokers carry their 3-letter exchange member code in there too. So only the longer codes become `ticker`, and `corp.public` comes from `payIslemDurumu` instead of "has a code".
- Board members: foreigners are sometimes listed as their own representative, which made 45 of them `LegalEntity`. A seat is now only a legal person if the representative is someone else, and that person is linked with `Representation` instead of going into the description.

Full uncached run: 27,084 entities (4,701 Company, 5,904 Person, 3,328 LegalEntity, 7,133 Ownership, 5,903 Directorship, 112 Representation), no warnings. Export with assertions passes, Koç still matches the live page, and there are no 11-digit IDs in the output. Rebased on main.

Also, when you have a moment, #5545 (`us_ori_admin_actions`, new crawler) and #5546 (ISIN validation in `make_security`, for #4941) are still waiting on a first look. Both merge cleanly. No rush, happy to change either one.

PR #5579, opened 2026-09-01
daily issues 9-1

herrhellana:


datasets/au/dfat_sanctions/au_dfat_sanctions.yml
```diff
@@ -303,6 +303,8 @@ lookups:
         value: 9422457
       - match: 9434980
         value: 9434890
+      - match: 947115
+        value: null
```
nvmbrasserie, line 307: IMO numbers are crucial for vessel identification, so it’s always good to preserve them whenever possible. in this case, it appears to be a typo in the source. the correct number should be: 9417115

let’s correct it. I believe this is the vessel in question: https://www.marinetraffic.com/en/ais/details/ships/shipid:5344879/mmsi:0/imo:9417115/vessel:KOTI

herrhellana: this is wonderful, thank you! i'll commit an edit.

PR #5581, opened 2026-09-01
issues 9-1 part II

herrhellana:


datasets/il/mod_crypto/wallets.csv
```diff
@@ -249,9 +248,7 @@ MOHAMMED A S ALQIBTI,5062738,,22/05/1989,970599929075,
 ,,,,TMznnujRXPoKUL5RU2bkhT2qo7Qrfx2gc2,
 ,,,,TNAGa4B4YRAS58fH8eaAsX4qRxtSmFArTz,
 ,,,,TNEKEuzfg9BiaQtQKBGSJoH9CVvJphzoLP,
-,,,,TP5cEKbcKknQsbcNwfyS7WLZgPJCrWg4wg,
 ,,,,TPV2rmU42Q7vgteRNnc9rKkmXT5cEgGTST,
-,,,,TQQeghVLiFYNByzumZPXCu5PwXxK1X3WHB,
```
nvmbrasserie, line 254: I've never seen them dropping wallets without a published notice, it's weird. Let's park it for a couple of days and see if they come back. Otherwise, we'll update the hash and remove them accrodingly

datasets/au/dfat_sanctions/au_dfat_sanctions.yml
```diff
@@ -304,7 +304,7 @@ lookups:
       - match: 9434980
         value: 9434890
       - match: 947115
-        value: null
+        value: 9417115
```
nvmbrasserie: perfect, thanks! we can also add a brief comment about the source typo. could you maybe also email them asking to fix it? they are usually quite responsive

herrhellana: sounds good! i can link the source you mentioned to the comment in .yml?

datasets/il/mod_crypto/wallets.csv
herrhellana: yeah the diff was only on the wallets.csv. I checked those entries on Drive against what's in the Seizure Order PDFs, and these wallets were missing in the linked sources. removed them from the Drive sheet, too, and updated the diff

datasets/au/dfat_sanctions/au_dfat_sanctions.yml
nvmbrasserie: yes, I think we did it this way before: https://github.com/opensanctions/opensanctions/blob/22239e3b2b278796abb40de2e6911285e80a70d5/datasets/gb/fcdo_sanctions/gb_fcdo_sanctions.yml#L213

herrhellana: emailed the DFAT team: they're correcting the mistyped IMO number in the source.

datasets/il/mod_crypto/wallets.csv
nvmbrasserie: okay, could you update the google sheet (remove these rows from there)? we can merge it then

herrhellana: i did it on monday

datasets/au/dfat_sanctions/au_dfat_sanctions.yml
herrhellana: they will fix it, but the IMO number is still invalid as of today. keeping the datapatch for now.

PR #5593, opened 2026-09-02
[us_finra_actions] migrate to XLSX source

### LINKED ISSUE #5591: [us_finra_actions] site migrated from a paginated HTML table to a web-component portal
zyte unblock issue: https://www.opensanctions.org/issues/us_finra_actions/

The root cause is that the FINRA site has migrated from a Drupal table to an Angular web-component portal, but now it also features a XLSX bulk export (+not behind Cloudflare) -- maybe it's better if we switch to crawling it vs rewriting the pagination and cache-busting logic. The switch would imply data reviews in the UI (annoying, but better maintenance in the long-term & aligned with the new repo practices).

herrhellana:
closes: https://github.com/opensanctions/opensanctions/issues/5591

source switch from the Zyte-scraped HTML portal (which is down) to XLSX bulk export. 

Name handling:
- deterministic cleaning/splitting first, then UI reviews (every comma is corrupted into ";" which is used to split entities, aliases, titles, and what not)
- name lookups are retired in favor of the new name cleaning framework 

Tried preserving the old crawler IDs where possible to avoid a huge rekey.


herrhellana:
## Entity-ID scheme and rekey impact

This dataset keys entity IDs on `make_slug(<raw name part>)`. That contradicts the entity-ID guide (`zavod/docs/best_practices/entity_id.md`: never `make_slug` personal names or free text) — it is kept **deliberately** as the pre-existing published scheme, because the alternative is a full rekey of ~15k matchable entities. The crawler documents the invariant: the slug is computed from the raw split part *before* any lookup resolution, which also pins the legacy numeric IDs like `us-finra-142866` (rows where the name column holds a bare CRD number and the name is restored via the `type.name` lookup).

**Measured against the last successful production run (2026-08-15):** production had 14,971 LegalEntities carrying 15,016 raw crawler IDs; the new crawl emits 14,937. **14,880 raw IDs survive (99.1%). 93 production entities (0.6%) are left with no surviving raw ID**, and 57 new IDs appear. No `context.rekey` is used — at this scale the resolver churn is acceptable versus the rekey choreography.

Breakdown of the 93:

* **~60 are junk entities the old comma/alias splitting fabricated**, now correctly gone: alias fragments (`us-finra-f-k-a-fahnestock-co-inc`, `us-finra-aka-alexandria-montgomery`, `us-finra-as-successor-to-j-p-morgan-clearing-corp`), parenthetical debris (`us-finra-respondent-or-ahmed`, `us-finra-cssu-or-the-firm`, `us-finra-iii`, `us-finra-ph-d`), and single-token shards of comma firm names (`us-finra-goldman`, `us-finra-merrill-lynch`, `us-finra-bear`, `us-finra-corp`, `us-finra-fenner-smith-incorporated`). Losing these is a correction, not a regression.
* **The rest are real entities whose ID absorbs an alias tail** now that the full source string is kept intact for name review (`us-finra-daniel-b-irving` → `us-finra-daniel-b-irving-irving`, `us-finra-chung-dinh-tran` → `us-finra-chung-dinh-tran-also-known-as-tony-tran`, `us-finra-axos-clearing-llc` → `us-finra-axos-clearing-llc-1`), **or whose XLSX spelling differs from the old HTML listing** (`us-finra-edward-joseph-rudiger-jr` → `us-finra-edward-j-rudiger-jr`, `us-finra-trautman-wasserman-and-company` → `…-company-inc`, `us-finra-malcolm-d-pryor` → `us-finra-malcolmn-d-pryor` — the old lookup fixed that source typo; review now fixes the name while the raw form keys the ID).

## Review volume

The first production run will create **327 name reviews in the UI** (326 distinct irregular name strings — a/k/a, f/k/a, `("the Firm")` junk, BD numbers, trailing debris like `Axos Clearing LLC;1`). In production the LLM cleaning pre-fills each review with a suggested split/categorisation, so acceptance is mostly one-click. Regular names (~14.6k entities) apply deterministically and never touch the queue; new reviews only appear when FINRA publishes new irregular strings.

Also of note: the 3-step name-migration protocol wasn't applicable here — the old HTML source is dead, and review keys derive from the raw strings, which changed with the source anyway. Until a review is accepted, the entity carries its raw string as `name`, so nothing is lost in the interim.

## Known source gap: 8 rows without names

8 rows in the export have **both** name columns empty and are skipped with a `Row has no individual or firm name` warning — the respondent is only recoverable from unstructured text, which we don't transcribe:

* 3 recent OHO default decisions (2020–2021) where the name + CRD appear only in the PDF filename in the *Title* column, with an empty *Summary*: Patrick Matlock (CRD 5760255, case `2020067731001`), Jordan Palmer John (CRD 6924720, `2021071672302`), David Hixon (CRD 4707468, `2021070867702`).
* 5 older rows (2006–2008) where the name appears only inside the *Summary* document text: Rebecca Amy Reichman (`2008012096001`, twice — NAC + OHO decisions on the same case), Brenden Christopher Lilly (`2008012102601`), Michelle Holton (`2006004691101`), and one OCR-mangled "CORB1N T. JONES" (`2006004969704`).

These 8 warnings are the crawler's entire steady-state warning volume.

datasets/us/finra_actions/crawler.py
```diff
@@ -1,172 +1,215 @@
...
+
+
+def split_individual_names(
+    context: Context,
+    raw_name: str,
+    raw_crd: str | None,
+    suffix_tokens: set[str],
+    case_id: str,
+) -> list[tuple[str, str | None]]:
+    """Split names and re-attach suffix or alias continuation parts."""
```
jbothma, line 68: it'd be nice to have some examples in the docstring of what this is actually solving - what's a continuation part? what sort of suffix needs re-attachment?

datasets/us/finra_actions/us_finra_actions.yml
```diff
@@ -22,20 +22,44 @@ publisher:
...
+    - ", Sr."
+    - "Sr."
+    - ", Sr"
+    - "Sr"
+    - ", II"
+    - "II"
+    - ", III"
+    - "III"
+    - ", IV"
+    - "IV"
```
jbothma, line 56: this is often an important part of someone's name - John Smith is a different person from John Smith Jr and John Smith III

datasets/us/finra_actions/us_finra_actions.yml
```diff
@@ -48,9 +72,36 @@ assertions:
 dates:
   formats: ["%m/%d/%Y"]
 lookups:
+  crd:
+    options:
+      # Swapped/duplicated column values: the CRD column holds a name whose
```
jbothma, line 77: what's a CRD?

datasets/us/finra_actions/crawler.py
```diff
@@ -1,172 +1,215 @@
...
+                    "Individual name continuation has no preceding name",
+                    continuation=part,
+                    individual_name=raw_name,
+                )
+                continue
+            names[-1] = f"{names[-1]}{separator}{part}"
+            continue
+        names.append(part)
+
+    if raw_crd is None:
```
jbothma, line 88:
```suggestion
    # Central Registration Depository (an identifier for this)
    # https://www.finra.org/registration-exams-ce/classic-crd
    if raw_crd is None:
```

jbothma:
It's nice to put this up front so we have an idea what the implications of this change are

```
Delta export complete          [us_finra_actions] added=4707 dataset=us_finra_actions deleted=5132 metric=delta_counts modified=28796
```

datasets/us/finra_actions/crawler.py
```diff
@@ -1,172 +1,215 @@
...
-        table = response.find(".//table")
-        assert table is not None, "Validated FINRA page did not contain a table"
+    path = context.fetch_resource("source.xlsx", context.data_url)
+    context.export_resource(path, XLSX, title=context.SOURCE_TITLE)
+
+    wb = load_workbook(path)
+    ws = wb.worksheets[0]
+    suffix_tokens = {
+        token.casefold()
+        for suffix in context.dataset.names.suffixes_strip
```
jbothma, line 208: this config key is intended for h.strip_name_titles and nothing else. Using it for this custom purpose can lead to bad suprises (things being stripped that shouldn't) if someone starts using the helper

datasets/us/finra_actions/crawler.py
jbothma: "reattach" sounds complex.

In the past we've often done something like this:

```python
# the bare minimum cleaning before you can split safely in 99% of cases
NOT_SPLITS = [
    "Esq",
    "Jr",
]
for not_split in NOT_SPLITS:
    # replace the semicolon version with a comma version
raw_ish_names = split(";")
...more name cleaning...

# I like the comparison of num CRDs with num names and warning

# could still pass names through name cleaning framework or at least is_irregular() and warn if it looks irregular to detect more work needed

# and at some point we make an id

# once this has run in prod and we've dealt wih reviews and stuff, we can delete the old_id and context.rekey()
for raw_ish_name in raw_ish_names:
    if h.is_irregular(raw_ish_name):
        context.log.warning((maybe we need more splitting?")
    id_old = context.make_slug(raw_ish_name)
    id = context.make_id(raw_ish_name)
    context.rekey(id_old, id_new)
    entity= context.make(...)
    entity.id = id
    entity.add("name", clean_name)
```

jbothma (changes requested):
let's look at some examples of the 5000 removals and additions - is there a pattern to the change in input to make_slug() which we could use to generate the old key and use context.rekey to gracefully migrate those keys to our new rawish_name strings? but if it's hard to understand the code, it's probably fine to just accept the rekey - I think it's disruptive anyway with 30k american names getting modifications

datasets/us/finra_actions/us_finra_actions.yml
herrhellana: added a comment

PR #5602, opened 2026-09-03
[se_soe] Add the 2025 report and consolidate leadership into one CSV

nvmbrasserie:
The government replaced the annual *Verksamhetsberättelse för bolag med statligt ägande* with a slimmer report, [Årlig information om bolag med statligt ägande 2025](https://www.regeringen.se/rapporter/2026/07/arlig-information-om-bolag-med-statligt-agande-2025/), which no longer publishes board rosters — only chairs and CEOs. `check_updates()` matched the old series title only, so the new report went unnoticed; it now asserts on the set of reports the source page offers, matching both titles.

## Data now lives in one CSV

`leadership.csv` in the repo replaces the Google Sheet, one row per appointment, with `report` naming the editions that attest it and `source_url` carrying the source page plus each edition's PDF. The crawler rejects duplicate `(name, position, company)` keys and a `source_url` that disagrees with `report`.

The 2024 board rosters are kept — the 2025 report is supplementary, not a replacement.

## Name corrections found while merging

- **Maria Håkansson** (2024) and **Maria Hammarskjöld Håkansson** (2025) are one person, previously emitted as two entities. Merged, keeping the former as an alias.
- The 2025 report misspells Svenska Spel's CEO as *Anna Johansson*; [the company spells it Anna Johnson](https://om.svenskaspel.se/kontakt/anna-johnson/). Corrected, with the deviation recorded in `notes`. Deliberately **not** aliased — "Anna Johansson" is a common Swedish name and would cause false matches.
- Chair names wrap inside their column, so they are read from a bounding box rather than page text, which had silently truncated three of them.
- Three CEOs named in the 2025 quote attributions had already left; the names come from the report's own footnotes instead.
- Adds Susanne Norman, Swedavia's acting CEO in the 2025 report, on the same position as the substantive holder.

## Extraction scripts replaced by docs

Every edition has changed layout, so the one-time PDF scripts were not reusable — the 2024 one was useless for the 2025 report. `extraction.md` captures what actually transfers: the traps in these PDFs, the acting-officeholder rule, and the verification steps.

## Result

`Person 286 · Position 114 · Company 38 · Occupancy 324`, clean run, empty `issues.log`, `mypy --strict` and `ruff` pass.

Two company entity IDs change, since IDs derive from the name: `AB Göta` → `AB Göta kanalbolag` (a truncation in the retired sheet, contradicted by the report itself) and `Svenska rymdaktiebolaget` → `SSC Space AB` (a real rename, old name kept as `previousName`).

`last_checked` is deliberately untouched here — date bumps stay on #5586.


datasets/se/soe/extraction.md
```diff
@@ -0,0 +1,86 @@
...
+- Check surnames against the company's own site. The 2025 report gives Svenska Spel's
+  CEO as `Anna Johansson`; the company spells it `Anna Johnson`.
+- Scan for single-token names, stray punctuation, digits and double spaces.
+
+Put a real former name in `alias`. Don't alias a typo — `Anna Johansson` is a common
+Swedish name and would cause false matches. Note it in `notes` instead.
+
+## Merging
+
+The Google Sheet used until September 2026 is frozen history — edit the CSV.
```
jbothma: let's at least mark the sheet as deprecated or disused YYYY-MM-DD in its name when we merge this so we can delete it eventually. And then I think we can drop this line since it'll become stale.

datasets/se/soe/extraction.md
```diff
@@ -0,0 +1,86 @@
...
+Industrifonden and Norrlandsfonden are foundations, not companies with state
+ownership, and get no rows here.
+
+## Verify before committing
+
+- Compare each `(position, company)` slot against the previous edition. A changed
+  occupant is usually a real appointment, but can be one person under two spellings —
+  `Maria Håkansson` and `Maria Hammarskjöld Håkansson` were one person.
+- Check surnames against the company's own site. The 2025 report gives Svenska Spel's
+  CEO as `Anna Johansson`; the company spells it `Anna Johnson`.
```
jbothma: hmmm - perhaps only when there's a discrepancy in similar names for a company from one year to the next? We don't normally check every name against another source

datasets/se/soe/extraction.md
leonhandreke, line 78: Does the thing need to know about the previous sheet?

datasets/se/soe/se_soe.yml
```diff
@@ -10,19 +13,9 @@ manual_check:
...
-    1. Update the 'company_names' lookup table in this YAML with mappings from
-       PDF text snippets to official company names (run once to identify missing entries)
-    2. After extraction, manually review and correct the output CSV - name parsing
-       may split/merge names incorrectly or miss people entirely
-
-    The final dataset requires manual quality control before publishing.
-
-    NOTE: Check if the data has become available in machine-readable format
-    (e.g., structured data, API, or database).
+    Check whether a new report has been published, and extract it if so.
```
leonhandreke, line 16: Could this message make clear what the audience is? Human, LLM? What am I supposed to do as a human processing the ticket?

datasets/se/soe/crawler.py
```diff
@@ -1,86 +1,150 @@
...
+
+
+def crawl_row(context: Context, row: dict[str, str]) -> None:
+    company_name = row.pop("company")
+    pep_name = row.pop("name")
+    position_name = row.pop("position")
+    # Validated in crawl(); source_url is what we publish.
+    row.pop("report")
+    # Maintainer annotation, not the FollowTheMoney property. Not published.
+    row.pop("notes")
```
jbothma:
```suggestion
    row.pop("extraction_notes")
```

or similar, let's make it really clear this is unpublished.

datasets/se/soe/extraction.md
```diff
@@ -0,0 +1,86 @@
...
+
+Crop each name's column bounding box; don't read page text line by line. Names wrap
+onto a second line, and line-based extraction silently truncated three of them in the
+2025 report. Also de-hyphenate names split across lines (`Susanne Anders- son`).
+
+## The CEO is prose, not a field
+
+The chair is a labelled field, but the CEO is named only in a quote attribution:
+`"…" – Caroline Arehult, VD`. Never take a footnoted or `tf` attribution as the
+current CEO — read the footnote and record what it says, with the reason in `notes`.
```
jbothma: remember to update if we rename the column

datasets/se/soe/extraction.md
```diff
@@ -0,0 +1,86 @@
...
+
+- Compare each `(position, company)` slot against the previous edition. A changed
+  occupant is usually a real appointment, but can be one person under two spellings —
+  `Maria Håkansson` and `Maria Hammarskjöld Håkansson` were one person.
+- Check surnames against the company's own site. The 2025 report gives Svenska Spel's
+  CEO as `Anna Johansson`; the company spells it `Anna Johnson`.
+- Scan for single-token names, stray punctuation, digits and double spaces.
+
+Put a real former name in `alias`. Don't alias a typo — `Anna Johansson` is a common
+Swedish name and would cause false matches. Note it in `notes` instead.
```
jbothma: remember to update if we rename the column

datasets/se/soe/extraction.md
```diff
@@ -0,0 +1,86 @@
+# Extracting Swedish state-owned enterprise leadership
+
+How to turn an edition of the government's annual report on the state's company
+portfolio into rows in `leadership.csv`. The report PDF is the source of record.
+
+Every edition so far has changed layout, so expect to write a new throwaway parser
+rather than reuse one.
```
jbothma:
```suggestion
Every edition so far has changed layout, so expect to write a new parser and/or update
extraction instructions based on a review of the previous and current edition. If parser
code was used as part of the extraction of a version, keep it in source control in a module
named `extract_edition_{YYYY}.py` for reference.
```

Let's keep the old script even if this iteration was totally agent-based.

jbothma (approved):
I've been thinking about garbage collecting people from this CSV when they're not a PEP any more and I think it gets a bit too much logic to build into this crawler and isn't needed quite yet. Just referencing https://github.com/opensanctions/opensanctions/issues/5627 here.

I'd really like to see the original CSV and then the new version so that we get a commit that takes us from the version with only 2024 to the version with both 2024 and 2025, showing name changes, new rows, etc. Maybe not worth bothering now - we can filter for rows containing only 2025. But it'd be a nice way to make sure nothing was dropped by the vibes. Just keep in mind such step by step commits for CSV edits going forward. That's all we have when it's vibes updating data.

looks like only a couple of names dropped from the original, and they weren't PEP positions, using

```
csvcut -c1,3,4 datasets/se/soe/leadership.csv > leadership-keys.csv
csvcut -c1,2,3 ~/Downloads/\[se_soe\]\ Sweden\ State\ Owned\ Enterprises\ -\ Sweden\ State-Owned\ Enterprises\ Leadership.csv >2024-keys.csv
comm -3 <(sort 2024-keys.csv ) <(sort leadership-keys.csv )
```

if we drop "https://www.regeringen.se/regeringens-politik/bolag-med-statligt-agande/" which I'm not sure makes sense to add in addition to the other two URLs, then there are only 66 person MODs instead of 280.

PR #5603, opened 2026-09-03
[gem_energy_ownership] August 2026 data, plus entity status and mergers

nvmbrasserie:
Picks up the August 2026 source file and maps three of the five columns it adds.

## Newly mapped

- **`entity_status` → `status`**, recording the 189 dissolved and 117 amalgamated entities we were previously publishing with nothing to show they are defunct. The source's own vocabulary is kept.
- **`merged_into` → `Succession`**, linking a defunct entity to the one it merged into, so a hit on an old name resolves to something live. The source writes these references as floats (`E100001014363.0`), hence `clean_entity_id()`.
- **`entity_status_data_source_url` → `Succession:sourceUrl`**, the evidence for a merger: company-registry pages, GLEIF records, merger announcements. All 83 emitted successions carry one (106 URLs). The source serialises the column as a Python list literal (`"['https://permid.org/']"`), split on the quote-comma separator rather than on commas, because 13 of the URLs contain a comma of their own (`.../Vattenfall+GmbH,+Berlin/...`). Evidence for the 189 `dissolved` entities, which have no succession to hang it on, is not published.

83 of the 117 merger pointers can be linked. The other 34 name entities that appear **nowhere** in the workbook — neither in *All Entities* nor in *Entity Ownership* — so there is nothing to point at and a dangling `successor` would be worse than none. Those are skipped at info level, while a genuinely malformed reference still warns.

## Left ignored, deliberately

- `intermediate_owner_ids` — a derived roll-up of the ownership chain; the individual links are already emitted as direct `Ownership` edges, and its bracketed figures look like effective rather than direct shares. Consistent with `gem_parents_ids`, ignored for the same reason.
- `joint_venture` — boolean, true for 0.8% of rows, no natural property.

## Identifier fixes

Four values were landing on the wrong property and warning every run. `issues.log` is now empty, down from four warnings:

| Value | Was | Now |
|---|---|---|
| `02.315.580/0001-31` | overflowing `permId` | `registrationNumber` (Brazilian CNPJ) |
| `5036181392` | failing LEI validation | `registrationNumber` |
| `549300ED64PCB8PIL726` | overflowing `permId` | `leiCode` |
| `5086807410 Company` | overflowing `permId` | `permId` = `5086807410` |

Plus nine INN codes and a double-spaced `not  found` variant.

## Result

`48,361 entities · 376,580 statements`, empty `issues.log`, `mypy --strict` and `ruff` pass.

`last_checked` is deliberately untouched here — date bumps stay on #5586.


datasets/_global/gem_energy_ownership/crawler.py
```diff
@@ -16,6 +15,9 @@
     "headquarters_subdivision",
     "gem_parents",
     "gem_parents_ids",
+    "intermediate_owner_ids",
+    "joint_venture",
+    "entity_status_data_source_url",
```
jbothma: it could be fun if this was added as sourceUrl do the Succession, for example, if I'm reading this right?

jbothma (approved):

PR #5610, opened 2026-09-04
Issues 9 4

herrhellana:


datasets/_externals/permid.yml
```diff
@@ -79,4 +79,4 @@ assertions:
       Company: 2800
   max:
     schema_entities:
-      Company: 10000
+      Company: 12000
```
nvmbrasserie, line 82: do we know the underlying cause? it's always important to double check the behaviour of the enrichers before editing the assertions to make sure that we understand why it is happening

herrhellana: it just looks like the N has been growing in the source -- they started adding new records in Sept

Issue #5444, opened 2026-08-15
15-digit Chinese 注册号 carries a division-code prefix that cross-checks against the USCC


zbl1998-sdjn:
#4374 re-routed `321000400012591` to `registrationNumber` after it failed USCC validation. That call is right — 15 digits is the pre-2015 business registration number (注册号), not a malformed 18-character USCC. There is one signal the re-route drops that may be worth keeping.

Both identifiers carry an administrative-division code, in different positions:

| identifier | layout |
|---|---|
| 15-digit 注册号 | **6-digit division code of the first registering authority** + 8-digit sequence + 1 check digit |
| 18-char USCC | 1 authority + 1 entity type + **6-digit division code (chars 3–8)** + 9-char entity code + 1 check char |

For the same legal entity these two normally agree. In the case above, `321000` is Yangzhou, Jiangsu under GB/T 2260 — consistent with the entity name Yangzhou Yangjie.

They can legitimately differ if the entity re-registered in a different division after its original filing, so a mismatch is a flag to look at, not an error.

Two places this is usable without querying any registry:

- a record carrying both identifiers should agree on those 6 digits;
- where only the 15-digit value survives, the prefix still names the registering city, which can be checked against the entity's address.

On scope, so I don't overstate it: the 15-digit format does end in a check digit, but I have not implemented or verified that algorithm and am not claiming you can validate the value itself — only the prefix relationship. The 18-character check character is the one specified in GB 32100-2015, which we do implement and tested against 2,149 real codes from two Chinese government disclosure files: https://currawongweb.com/verify/china-usci-checker/

Disclosure: we sell China-side supplier verification, so I have an interest in this area.

piomeer:
I'd like to take this on.

**Planned scope (warning-only, first PR):**
- When a `LegalEntity` has both a 15-digit Chinese 注册号 (`registrationNumber`) and an 18-char USCC (`uscCode`), compare the embedded 6-digit GB/T 2260 division codes (`reg[:6]` vs `uscc[2:8]`).
- On mismatch: emit a structured warning (not a hard reject), since re-registration in another division can legitimately diverge.
- Unit tests for agree / disagree / missing sibling / non-China shapes.

Out of scope for this PR: 15-digit check-digit validation; address vs prefix weak check (can follow up).

Implementation is on branch `fix/5444-china-reg-uscc-division-check` — opening a PR shortly.

leonhandreke:
[...]

Issue #5568, opened 2026-08-30
Adopt the new FtM Control schema (edges, analyzer, crawlers)


pudo:
followthemoney is gaining a `Control` schema (opensanctions/followthemoney#362): a sibling of `Ownership` under `Interest` with a directed `controller` → `controlled` edge, both ranged `LegalEntity`, plus `controlType` (voting, appointment, management, contractual, trust, nominee, state, indirect) and `legalBasis`. It is for one entity controlling another without necessarily owning it, and it removes the need to cast controlled `Organization`s as `Company` just to satisfy `Ownership.asset`'s `Asset` range.

Once a followthemoney release with `Control` is pinned here, the following needs doing.

## 1. Infrastructure that hardcodes `Ownership`

- [ ] `zavod/zavod/integration/edges.py:59` — `PROTECTED_PROPS` has per-schema entries for `Ownership`, `Directorship`, `UnknownLink`, … Add `"Control": ["controlType", "role"]` so control edges with conflicting descriptors are not merged on endpoints+dates alone.
- [ ] `datasets/_analysis/ann_graph_topics/analyzer.py`:
  - `SANCTION_ADJACENCY_EDGES` (line 93) — add `"Control"`.
  - `rule_sanction_control_descent` (line 274) — currently `adjacent.schema.name != "Ownership" or prop.reverse.name != "owner"`; it should also descend `Control` edges via `controller`. The topic it emits is already called `sanction.control`.
  - `rule_export_control_descent` (line 312) — `is_a("Ownership")` check; decide whether control edges should propagate `export.control.linked` too (the docstring warns against adding `Directorship`; `Control` is a different, deliberate assertion of control by the source).
  - Extend `test_ann_graph_topics.py` accordingly.

## 2. Crawlers that currently force control into other schemata

Ordered roughly by how clearly the source distinguishes control from ownership:

- [ ] `zavod/zavod/shed/bods.py:110` — BODS `ownershipOrControlStatement` interests are flattened into `Ownership` with `interests[].type` in `role`. BODS interest types `votingRights`, `appointmentOfBoard`, `influenceOrControl`, `seniorManagingOfficial` should become `Control` (with `controlType` mapped), while `shareholding` stays `Ownership`. Affects `datasets/dk/cvr`.
[...]

PR #5626, opened 2026-09-07
Warn when Chinese registration number and USCC division codes disagree

piomeer:
## Summary
- Implements a warning-only cross-check for #5444: when a LegalEntity carries both a 15-digit Chinese 注册号 and an 18-char USCC, compare the embedded 6-digit administrative division codes.
- Mismatches log a warning (re-registration can legitimately diverge); values are not rejected.
- Adds unit tests for agree/disagree/missing-sibling/non-China shapes and the cleaning-path hook.

## Test plan
- [x] `pytest zavod/zavod/tests/runtime/test_china_ids.py` (10 passed)
- [ ] Reviewer: spot-check an entity that has both IDs (e.g. Yangzhou Yangjie `321000400012591` / `913210007908906337`) — should stay silent
- [ ] Optionally run a China-heavy dataset crawl and confirm mismatch warnings look actionable on the Issues page

Closes nothing automatically; please link #5444 if this fully addresses the first bullet there.

### DIFF
--- a/zavod/zavod/runtime/china_ids.py
+++ b/zavod/zavod/runtime/china_ids.py
@@ -1,0 +1,121 @@
+"""Helpers for Chinese corporate identifier checks.
+
+See https://github.com/opensanctions/opensanctions/issues/5444
+
+Pre-2015 Chinese business registration numbers (注册号) are 15 digits:
+``AAAAAA`` (administrative division of the first registering authority)
++ 8-digit sequence + 1 check digit.
+
+The 18-character Unified Social Credit Code (USCC / 统一社会信用代码) embeds
+the same 6-digit division code in characters 3–8 (1-based).
+
+When a record carries both, the prefixes normally agree. They can
+legitimately differ after a re-registration in another division, so a
+mismatch is a maintenance flag (warning), not a hard reject.
+"""
+
+from __future__ import annotations
+
+import re
+from typing import TYPE_CHECKING
+
+from zavod.logs import get_logger
+
+if TYPE_CHECKING:
+    from zavod.entity import Entity
+
+log = get_logger(__name__)
+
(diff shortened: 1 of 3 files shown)


pudo:
@piomeer Hey! Our team is currently being hit with half a dozen LLM PRs a day. It's creating a significant amount of chaos, and ultimately is not helpful: we can punch tickets into an LLM ourselves. The reason that the issues you're picking up aren't addressed is because they're low priority, or complex to implement. So being forced to review them as PRs forces us away from important work, towards less important work. 

So here's the deal: are you really interested in working on improvements to OS? Then please submit human PRs, and not slop PRs. Please also explain why each PR is useful to you in how you use OS.

If it's just contribution farming, please do it somewhere else.

piomeer:
Thanks @pudo — fair feedback, and I take the point about LLM PR volume.

This PR was too agent-driven. That isn't a useful use of your review time, and I'm sorry for adding to the noise.

**Why I cared about #5444 personally:** I use OpenSanctions-style sanction/PEP data when thinking about counterparty and ownership-chain screening (treasury / due-diligence style workflows). For Chinese legal entities I often see both a pre-2015 15-digit 注册号 and an 18-char USCC on the same profile; when those disagree on the embedded division code it makes me less sure I'm looking at one legal person before I trust any ownership/control links. That's the only reason I picked up that issue — not contribution farming.

**What I'll do next:** please **close this PR**. I won't push more automated patches on it. If I come back to the 注册号/USCC check, it will be a smaller, human-written change I can explain line-by-line, with a clear note on how I use the data — and only after checking it's still wanted.

I'm also staying away from #5568 / Control-schema work after your note there; that clearly needs the team.

Thanks for saying this directly.

PR #5632, opened 2026-09-08
Us nfa enforcements

brrttwrks:
Draft crawler for `us_nfa_enforcement_actions`. Could use some feedback regarding the program, name cleaning, and ID (normalized component).

### DIFF
--- a/datasets/us/nfa_enforcement_actions/crawler.py
+++ b/datasets/us/nfa_enforcement_actions/crawler.py
@@ -14,12 +14,6 @@ from zavod import helpers as h
 FIRST_YEAR = 1981
 # The link params carrying a respondent's key, in the two forms the source has used.
 KEY_PARAMS = ("entityid", "nfaid")
-# A caption naming more than one respondent, e.g. "Acme Futures, et al. (Jane Roe)".
-AMBIGUOUS_CAPTION = re.compile(r"\bet\.?\s*al\b|[()]", re.IGNORECASE)
-# Names often contain an et al and parenthetical names, all post-nominal info that 
-# we can ignore.
-PATTERN_IRREGULAR = r"[()]|et al"
-REGEX_IRREGULAR = re.compile(PATTERN_IRREGULAR, re.IGNORECASE)
 
 
 def listing_url(context: Context) -> str:
--- a/datasets/us/nfa_enforcement_actions/crawler.py
+++ b/datasets/us/nfa_enforcement_actions/crawler.py
@@ -1,0 +1,171 @@
+import json
+import re
+from collections import defaultdict
+from dataclasses import dataclass, field
+from typing import Any
+from urllib.parse import parse_qs, urljoin, urlsplit
+
+from lxml import html
+
+from zavod import Context, settings
+from zavod import helpers as h
+
+# Lower bound of the year slider on the source page; the earliest action is from 1996.
+FIRST_YEAR = 1981
+# The link params carrying a respondent's key, in the two forms the source has used.
+KEY_PARAMS = ("entityid", "nfaid")
+# A caption naming more than one respondent, e.g. "Acme Futures, et al. (Jane Roe)".
+AMBIGUOUS_CAPTION = re.compile(r"\bet\.?\s*al\b|[()]", re.IGNORECASE)
+# Names often contain an et al and parenthetical names, all post-nominal info that 
+# we can ignore.
+PATTERN_IRREGULAR = r"[()]|et al"
+REGEX_IRREGULAR = re.compile(PATTERN_IRREGULAR, re.IGNORECASE)
+
+
+def listing_url(context: Context) -> str:
+    """The page the respondent links are relative to, and the API's expected referer."""
+    url = context.dataset.url
+    assert url is not None, "Dataset metadata is missing `url`"
+    return url
+
+
+def fetch_rows(context: Context) -> list[dict[str, Any]]:
+    """Fetch the full listing from the JSON-RPC endpoint.
+
+    The endpoint only answers requests that carry the session cookie and CSRF token
+    handed out by the listing page, and answers everything else with a redirect to an
+    error page rather than an error status.
+    """
+    page_url = listing_url(context)
+    doc = context.fetch_html(page_url)
+    token = h.xpath_string(doc, ".//meta[@name='csrf-token']/@content")
+    text = context.fetch_text(
+        context.data_url,
+        method="POST",
+        data=json.dumps(
+            {
+                "method": "getEnforcementRegs",
+                "params": [FIRST_YEAR, settings.RUN_TIME.year],
+            }
+        ),
+        headers={
+            "Content-Type": "application/json",
+            "Referer": page_url,
+            "x-csrf-token": token,
+            "x-requested-with": "XMLHttpRequest",
+        },
+    )
+    if text is None or not text.startswith("{"):
+        raise RuntimeError("Listing API returned no JSON; the CSRF handshake failed.")
+    data = json.loads(text)
+    if data.get("Success") is not True:
+        raise RuntimeError(f"Listing API reported failure: {data.get('Message')!r}")
+    rows = data["Result"]
+    assert isinstance(rows, list) and len(rows), data.get("Message")
+    return rows
+
+
+def normalise_key(param: str, value: str) -> str:
+    """Reduce a respondent's link parameter to a single form per respondent.
+
+    Numeric `entityid` values are published both zero-padded to seven digits and
+    unpadded, for about 90 respondents in both forms, so they are padded here. One
+    legacy id is not numeric and is left alone.
+    """
+    value = value.strip()
+    if param == "entityid" and value.isdigit():
+        return value.zfill(7)
+    return value
+
+
+def parse_respondents(
+    context: Context, headline: str, case_id: str
+) -> list[tuple[str, str, str]]:
+    """Return the (respondent key, detail URL, caption) of each link in a headline."""
+    fragment = html.fragment_fromstring(headline, create_parent="div")
+    respondents: list[tuple[str, str, str]] = []
+    for anchor in h.xpath_elements(fragment, ".//a"):
+        caption = h.element_text(anchor)
+        # Some hrefs are written root-relative with a Windows separator, e.g.
+        # "\BasicNet/basic-reg-actions-details.aspx?...".
+        href = (anchor.get("href") or "").replace("\\", "/")
+        query = parse_qs(urlsplit(href).query)
+        keys = [
+            f"{p}:{normalise_key(p, query[p][0])}" for p in KEY_PARAMS if p in query
+        ]
+        if len(keys) != 1:
+            # An unclosed <a/> tag in a 1998 headline leaves a stray, empty anchor.
+            if caption:
+                context.log.warning(
+                    "Cannot identify respondent from link",
+                    case_id=case_id,
+                    caption=caption,
+                    href=href,
+                )
+            continue
+        respondents.append((keys[0], urljoin(listing_url(context), href), caption))
+    return respondents
+
+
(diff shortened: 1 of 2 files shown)


datasets/us/nfa_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,164 @@
...
+        ),
+        headers={
+            "Content-Type": "application/json",
+            "Referer": page_url,
+            "x-csrf-token": token,
+            "x-requested-with": "XMLHttpRequest",
+        },
+    )
+    if text is None or not text.startswith("{"):
+        raise RuntimeError("Listing API returned no JSON; the CSRF handshake failed.")
```
jbothma: We would normally just use context.fetch_json and let things downstream crash if the response wasn't valid JSON or didn't contain the expected keys. That keeps the code a bit cleaner.

Is this common enough that we need to explicitly check for it and have a specific log message? If so, I would log the text response here in case it's something other than the CSRF handshake

datasets/us/nfa_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,164 @@
...
+FIRST_YEAR = 1981
+# The link params carrying a respondent's key, in the two forms the source has used.
+KEY_PARAMS = ("entityid", "nfaid")
+
+
+def listing_url(context: Context) -> str:
+    """The page the respondent links are relative to, and the API's expected referer."""
+    url = context.dataset.url
+    assert url is not None, "Dataset metadata is missing `url`"
+    return url
```
jbothma: this is overkill - you can just stuff context.dataset.url straight into context.fetch_*()

datasets/us/nfa_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,164 @@
...
+def fetch_rows(context: Context) -> list[dict[str, Any]]:
+    """Fetch the full listing from the JSON-RPC endpoint.
+
+    The endpoint only answers requests that carry the session cookie and CSRF token
+    handed out by the listing page, and answers everything else with a redirect to an
+    error page rather than an error status.
+    """
+    page_url = listing_url(context)
+    doc = context.fetch_html(page_url)
+    token = h.xpath_string(doc, ".//meta[@name='csrf-token']/@content")
```
jbothma: if you want to factor things out, this could be come fetch_csrf_token()

datasets/us/nfa_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,164 @@
...
+    text = context.fetch_text(
+        context.data_url,
+        method="POST",
+        data=json.dumps(
+            {
+                "method": "getEnforcementRegs",
+                "params": [FIRST_YEAR, settings.RUN_TIME.year],
```
jbothma: we have a constant `zavod.helpers.dates.MAX_ENFORCEMENT_DAYS` which determines the number of years of enforcements we normally include. Could use that to calculate the starting year.

datasets/us/nfa_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,164 @@
...
+    should be
+    American Futures Group, Inc.
+    """
+    PATTERN_ETAL = r"\s*,?\s*\bet\.?\s+al\b\.?"
+    REGEX_ETAL = re.compile(PATTERN_ETAL, re.IGNORECASE)
+    name_without_etal = REGEX_ETAL.split(name, 1)[0]
+    PATTERN_PARENTHETICAL = r"\(([^()]+)\)"
+    REGEX_PARENTHETICAL = re.compile(PATTERN_PARENTHETICAL, re.IGNORECASE)
+    name_without_parentheticals = re.sub(REGEX_PARENTHETICAL, "", name_without_etal)
+    return name_without_parentheticals
```
jbothma: we would normally just do `clean_name = h.multisplit(raw_name, ["et al."])[0]` if you only want the first name. After a run, look at the names.txt export, see if there are more cases you should be cleaning

for future unseen cases, consider using h.is_name_irregular with strings like "et al" in the namespec to detect and warn on dirty names that need cleaning.

datasets/us/nfa_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def parse_row(context: Context, row) -> None:
```
jbothma: as a bit of a convention, if a function emits based on the thing it's passed, we'd name it crawl_row rather than parse_row

datasets/us/nfa_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,164 @@
...
+    respondents: list[tuple[str, str, str]] = []
+    for anchor in h.xpath_elements(fragment, ".//a"):
+        caption = h.element_text(anchor)
+        # Some hrefs are written root-relative with a Windows separator, e.g.
+        # "\BasicNet/basic-reg-actions-details.aspx?...".
+        href = (anchor.get("href") or "").replace("\\", "/")
+        query = parse_qs(urlsplit(href).query)
+        keys = [
+            f"{p}:{normalise_key(p, query[p][0])}" for p in KEY_PARAMS if p in query
+        ]
```
jbothma: I'd say this more processing on entity ID inputs than we like, from experience.

what happens if we drop key normalization? we won't be emitting any broken references in e.g. an UnknownLink referring to 00000123 when the entity had 123, right? is it just that we might need to dedupe John Smith 00000123 and John Smith 123 at some point? If so, let's drop normalise_key and leave it to dedupe.

and then constructing this id scheme (nfaid or entityid) with the id, isn't really something we do. We're already doing concatenation of parts in context.make_id. So I'd suggest something like a dataclass instead of it becoming a 4-tuple

```python

@dataclass
class Respondent:
    scheme: str
    id: str
    url: str
    name_raw: str

def parse_respondents() -> list(Respondent):
    ...

for respondent in parse_respondents():
    entity.id = context.make_id(respondent.scheme, respondent.id, respondent.name_raw)
```

jbothma (commented):
nice stuff

broadly I think there's cleaning we can simplify or drop - shout if anything isn't clear

datasets/us/nfa_enforcement_actions/crawler.py
brrttwrks: Yes. I've noticed LLMs are very cagy on checking things, which normally might make sense, but here it doesn't as it will fail anyway and that is the desired outcome. I'll keep an eye on these "redundant" checks more. I've also noticed there are times when the LLM proposes to count things when we're already asserting sane counts on our side explicitly, leading to more code and redundancy.

datasets/us/nfa_enforcement_actions/crawler.py
brrttwrks: Uh, yeah!

datasets/us/nfa_enforcement_actions/crawler.py
brrttwrks: Nice. We reduced the number of requests considerably here.

datasets/us/nfa_enforcement_actions/crawler.py
brrttwrks: Got it.

datasets/us/nfa_enforcement_actions/crawler.py
brrttwrks: Agreed. I reworked that considerably.

datasets/us/nfa_enforcement_actions/crawler.py
brrttwrks: I understand what you mean. I've reworked this with your approach, sending irregular names for review, which are quite few after the simple multisplit.

datasets/us/nfa_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,194 @@
...
+
+def fetch_rows(context: Context) -> list[dict[str, Any]]:
+    """Fetch the full listing from the JSON-RPC endpoint.
+
+    The endpoint only answers requests that carry the session cookie and CSRF token
+    handed out by the listing page, and answers everything else with a redirect to an
+    error page rather than an error status.
+    """
+    url = context.dataset.url
+    assert url is not None
```
jbothma, line 67: aaa I was gonna suggest this but I guess it's needed for the type checker because context.datast.url is str | None

```suggestion
```

datasets/us/nfa_enforcement_actions/crawler.py
```diff
@@ -0,0 +1,194 @@
...
+    # ever named in a complaint are therefore not in this dataset. Sanctions imposed —
+    # fines, bars, suspensions, withdrawals — are published only on the per-case detail
+    # pages, which NFA excludes from crawling, so this dataset records that an action was
+    # taken but not what it imposed.
+    if category == "COMPLAINTS":
+        return
+
+    date = row.pop("content_date_sort")
+    respondents = parse_respondents(context, row.pop("HEADLINE_TEXT"))
+    for respondent in respondents:
```
jbothma, line 124: what if we changed the loop to be over the fragments, and make the entity in the function we call from the loop (edits after `if category`):

```suggestion
def crawl_row(context: Context, row: dict[str, str]) -> None:
    case_id = row.pop("CASE_ID")
    category = row.pop("ACTION_CATEGORY_CODE")

    # Complaints are excluded. A complaint is the charging document that opens a disciplinary
    # case and states allegations that have not been adjudicated; respondents who were only
    # ever named in a complaint are therefore not in this dataset. Sanctions imposed —
    # fines, bars, suspensions, withdrawals — are published only on the per-case detail
    # pages, which NFA excludes from crawling, so this dataset records that an action was
    # taken but not what it imposed.
    if category == "COMPLAINTS":
        return

    date = row.pop("content_date_sort")
    fragment = html.fragment_fromstring(headline, create_parent="div")
    for anchor in h.xpath_elements(fragment, ".//a"):
        crawl_respondent(context, case_id, anchor)
```

Then the whole loop body gets factored out to crawl_respondent. We don't even need the Respondent class any more (or the tuple it replaced) . It's named crawl_ because it emits one respondent.

Sorry I just thought of this.

we can also return the entity from crawl_respondent, and link all the respondents in a case by making an Article entity for the link we're currently putting into sourceUrl, and and linking each respondent to it using a Documentation
https://zavod.opensanctions.org/extract/enforcements/#create-article-and-documentation-entities

I completely forgot about that in the last review.

datasets/us/nfa_enforcement_actions/us_nfa_enforcement_actions.yml
```diff
@@ -0,0 +1,81 @@
...
+  max:
+    schema_entities:
+      LegalEntity: 150
+
+lookups:
+  type.name:
+    options:
+      - match:
+          - "Vincent Capital Group LLC ("
+        value: Vincent Capital Group LLC
```
jbothma, line 71: shouldn't need this if all names go through name cleaning

datasets/us/nfa_enforcement_actions/crawler.py
jbothma: I think we can use h.apply_reviewed_names directly actually - rather than h.is_irregular. Let's just make a nice clear comment unpacking what we should do with the cleaning, i.e. the part after et al doesn't become an alias or something.

Hopefully we can make such a notice nice and visible on the review UI soon. this is a good usecase / test case for such a notice.

jbothma (approved):
btw with `ci_test: false` and a comment "openai key not available in ci"  you can probably ship this and refine if needed

datasets/us/nfa_enforcement_actions/crawler.py
brrttwrks: I reworked it to include articles and documentation. It added some complexity, but nothing we weren't already doing really.

PR #5637, opened 2026-09-08
Raise for missing exports, ensure only complete exports are discoverable

jbothma:
Closes https://github.com/opensanctions/operations/issues/2647

- An exporter which finalizes but whose self.path doesn't exist results in a FileNotFoundError - **not caught any more**
- A dataset.exports value which isn't match some crawler results in an error but doesn't fail the run
- index.json is published after other artifacts. versions.json is published after that. So a run is only discoverable once its artifacts are all published.

Adds tests for these.


jbothma:
@leonhandreke @pudo main thing to have a second pair of eyes on here is just not catching an exporter artifact being missing any more. Any concerns/vetos?

This differs from the behaviour of a misconfigured exporter option in the dataset yaml which just logs an error, but I think the one is human error on one dataset, while the other is something misbehaving and potentially affecting many datasets.

Issue #5654, opened 2026-09-10
proposal: source string in eu_journal CSVs


jbothma:
It's a bit painful to review the eu_journal agent PRs.

it could be cool to try requiring the eu journal extraction script to include the raw string used to extract any given entity in a csv cell. 

That chunks the validation to a size that's more manageable for a human. It also enables some fairly nice automated validation:

- python script checks if source string exists verbatim in the source as a whole (does this get broken by consolidation edits?)
- python script checks if each value in the csv is contained in that string, llm explains any differences, e.g. paragraph breaks dropped from reason string
- llm checks if each value in the source is extracted to some column in the csv.

Much simpler implementation of the "is everything in the CSV also in the source" direction is in https://github.com/opensanctions/opensanctions/compare/eu-journal-check-values-in-source?expand=1 - the LLM could explain the discrepancies very nicely. We could instruct the crawler to run it on any modified files and comment on discrepancies in the PR.

It was a bit slow at then looking at the source and checking whether everything's been captured in the CSV.

The suggestion of capturing source string is just to try and do the deterministic stuff deterministically, and keep keep llm jobs small. 

Is this something we want?

pudo:
I'm a bit worried that this would make the PRs even harder to review: you've now got repetititve, wrapped multi-line strings inside the CSV, which would make it entirely impossible to read on GitHub; and in most cases they are technically HTML rows so there would be HTML fragments in there, too? In a way, it's asking the LLM to do even more editorial work than now by identifying the specific context - and because that context is inferred, I'm not sure it's suitable for verification at all.

PR #5655, opened 2026-09-10
Document handling eu journal PRs

jbothma:
Closes https://github.com/opensanctions/operations/issues/2841

Documents most of "reviewer should kinda review whether the maintainer did what it was instructed to do" in the dataset yaml where the maintainer instructions are.

Adds a high level flow chart which I think is at least a bit helpful.

Reviewing is still hard - I think https://github.com/opensanctions/opensanctions/issues/5654 could take the edge off but it's out of scope here. 

A sanity check from either of pudo or leon would be smart, just so I don't instruct anyone totally incorrectly.

localhost-eu_journal_sanctions

### DIFF
--- a/datasets/eu/journal_sanctions/eu_journal_sanctions.yml
+++ b/datasets/eu/journal_sanctions/eu_journal_sanctions.yml
@@ -117,15 +117,29 @@ config:
   # The procedure for each warning is above the key that the procedure
   # changes. A comment block here holds a maximum of 32 lines: the YAML
   # check in CI rejects more before a key.
+  #
+  # The documentation is split by role. Maintainer is usually an LLM agent.
+  # Reviewer is usually an OpenSanctions data engineer.
+  key_just_to_allow_more_documentation: "whatever"
 
   # The consolidated version that each data/consolidated/{framework}.csv
   # snapshot comes from. The Makefile gives the same version to the parser.
+  #
+  # Role: Maintainer
   #
   # These two warnings show an error in this mapping. Correct the mapping.
   # - "Snapshot has no consolidation pin": a file in data/consolidated/ has no
   #   entry here. Add the pin, or delete the file.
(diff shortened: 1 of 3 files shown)


pudo (commented):
Some thoughts: 

1. I wonder if the docs page should make it clearer that there are two, fundamentally independent, processes at work here: the a) transcription of amendments and b) the parsing of consolidated texts. Right now it reads a bit like there is one process. 
2. This PR significantly expands the YAML guidance - which I would see as a quality liability, unless it addresses specific failure modes of the extraction. More prompt means more drift, and we should do everything we can to keep these minimal and un-bloated. A lot of the expansion just reads like "make no mistakes". It's not entirely clear to me which failures each additional comment addresses. Can you clarify? 

In short: I'd want to make really sure that the YAML detail is absolutely worth it.

jbothma:
Thanks!

What I'm trying to address in the yaml is that it isn't clear how to tackle one of these PRs as a reviewer. It isn't clear the comments in the yaml is a prompt for an LLM.

I think for us handling these every other week when we've half-forgotten the detail which might have changed since we last saw it, we can keep it a pure prompt if we say "this is a prompt. Human documentation is in the zavod docs". There's risk is that the LLM still references it. 

Then it makes more sense to document the two stages in the zavod docs, so it isn't duplication, it's a different audience.

jbothma:
I've had a go at addressing both comments. I think eventually the prompt should be in a separate file, and should have its own eval suite (set of example inputs and model results like unit tests for the prompt). But that's definitely not something to do the day before go-live :)

pudo (approved):
Hm - agree: the promptiness of the YAML file is something we'll need to manage at some point. I hope that the `eu_journal` source will remain a bit special in just how weird it is, but those are cursed words. 

Re this version: I actually think that phrased as a review policy, they have a shot to not just help reviewers, but also help with data quality in the thing.

PR #5656, opened 2026-09-10
[pk_proscribed_persons] Fix country, idNumber and placeholder CNICs

nvmbrasserie:
Two data-quality fixes in the NACTA proscribed persons crawler, found while looking into the delta on #5586. Independent of that PR — the numbers below are measured against `main`'s `source.json`, and hold for the updated one too.

## `country` and `idNumber` were only set for valid CNICs

Both were added inside the valid-CNIC branch, so the people whose CNIC fails `stdnum` validation carried neither. Everyone on the 4th Schedule is proscribed by a Pakistani authority, so the country holds either way, and the malformed number is still what the publisher holds for that person — worth keeping searchable.

The exception is a fragment too short to be a CNIC. Two records carry a single digit (`5` and `1`); those are dropped, with a warning so the source's data entry errors stay visible rather than being silently swallowed. Note this turns `issues.log` from empty to two warnings.

## Filler CNICs merged unrelated people

`stdnum` only checks a CNIC's length and its province and gender digits, so filler values pass. Twelve are in the source — an incremented `1111111111111` family plus `1234567891011` — and using them as the entity ID collapses different people into one entity:

    pk-cnic-1234567891011  name='Shahzad Ahmed Bajwa'  fatherName='Mushtaq Ahmed Bajwa'
    pk-cnic-1234567891011  name='Yaqoob Saeed Khan'    fatherName='Saeed Ullah'

They are now listed in a `placeholder_cnic` lookup and treated as missing, so those records fall back to name-based IDs.

## Effect

Measured by diffing statement packs before and after, on the same source file:

    country     present on 5003/5195 persons -> 5195/5195
    idNumber    present on 5003/5195 persons -> 5166/5195
    idNumber values not 13 characters: 0
    entity IDs re-keyed: 24  (12 placeholder persons + their sanctions)

Person count is unchanged. The 29 persons without an `idNumber` are the 15 records with a null CNIC, the 12 placeholder-derived entities, and the 2 fragments — all deliberate.

One consequence worth a reviewer's eye: one placeholder record, "Akhtar Muhammad Khalil" (BANNU, KP), now merges into the pre-existing `pk-proscribed-akhtar-muhammad-khalil-bannu-kp`, which already held the same person under a malformed CNIC. That merge looks right, but it is the name-slug fallback's judgement rather than something the source states.

Four `pk-cnic-*` entities still carry two name variants each (`Imran Khan`/`Imran Lala`, `Doctor Gul Alam`/`Dr. Gul Alam`, `Hanif Pashteen`/`Hanif Ullah`, `Amir Sadiq`/`Ansir`). Those share a real CNIC, so they are left merged.


pudo:
This looks great, want to merge?

nvmbrasserie:
[pk_proscribed_persons] changed=743 dataset=pk_proscribed_persons entities=10406 statements=83296

PR #5666, opened 2026-09-12
Add Hong Kong ICAC wanted persons crawler

thomasgdata:
Adds a crawler for the Hong Kong ICAC wanted persons list, requested in opensanctions/crawler-planning#335.

The index at `/en/rc/wanted/index.html` is server-rendered and lists 57 people as linked thumbnails, split into "Newly Wanted Person(s)" and the main list, with recent additions appearing in both. Each links to a profile page carrying a Name / Alias / Charge(s) table and a "Personal Particular" block.

Emitted as `Person` under the `crime` and `wanted` topics, following `gb_nca_most_wanted` and `de_bka_wanted`.

### Two things worth a look

**Every document number is masked at source.** I checked all 66 identifier values across all 57 profiles: HKIC numbers, and passport numbers for nine different countries, are published partially redacted (`P868xxx(x)`, `BB24XXXXX`, `44052XXXXXXXXXXXXX`). None are emitted as `idNumber` or `passportNumber` — they go to `notes` labelled as masked, so the record still shows which documents the ICAC listed without implying a searchable number.

**The particulars block is free text, and the labels drift.** Across 57 profiles it uses 35 distinct labels, including `HKIC No.` and `HKIC` for the same thing, plus `Complexion` and a `Complex` typo. Rather than enumerate them, the crawler maps the five stable ones (year and place of birth, sex, nationality, occupation) and keeps everything else as labelled notes, so a new label is recorded rather than silently dropped.

Two smaller things: the block is `<br/>`-separated, so it reads the direct text nodes — `element_text()` squashes the line breaks away. And family names come from the `hf_family_name` span on the index, which is the only reliable way to split these names.

### What I checked, and what I could not

Validated against all 57 live profile pages plus the index: 57 entities, no parse failures, all with `birthDate` (four-digit year), `lastName` and `gender`, 54 with `nationality`. Nationality is split on `/` for the dual-national and trailing-slash cases. `ruff check`, `ruff format` and `yamllint` all pass against the configs in this repo.

**I have not run it through zavod itself.** I have no Docker available and `plyvel` and `pyicu` will not build in my environment, so the parsing is verified against saved copies of the real pages rather than by a crawl. Worth a run before merge. Happy to adjust the topics, the assertion bounds, or the decision on the masked numbers.

### DIFF
--- a/datasets/hk/icac_wanted/crawler.py
+++ b/datasets/hk/icac_wanted/crawler.py
@@ -1,0 +1,145 @@
+from normality import collapse_spaces
+
+from zavod import Context
+from zavod import helpers as h
+from zavod.util import Element
+
+# Personal particulars are published as "Label : Value" text nodes separated by
+# <br/>, not as a table, and the labels drift between profiles. Anything not
+# mapped below is kept as a note rather than dropped.
+BIRTH_YEAR = "Year of Birth"
+BIRTH_PLACE = "Place of Birth"
+SEX = "Sex"
+NATIONALITY = "Nationality"
+OCCUPATION = "Occupation"
+
+# Values published as unknown carry no information.
+UNKNOWN_VALUES = {"unknown", "n/a", "-"}
+
+# Every identity card and passport number on this source is partially masked
+# (e.g. "P868xxx(x)", "BB24XXXXX"). That holds for all of them across every
+# profile, so none are emitted as idNumber or passportNumber; they are kept as
+# notes so the record still shows which documents the ICAC listed.
+IDENTIFIER_HINTS = ("passport", "hkic", " id ", "id no", "identity", "permit")
+
+
+def is_identifier_label(label: str) -> bool:
+    padded = f" {label.lower()} "
+    return any(hint in padded for hint in IDENTIFIER_HINTS)
+
+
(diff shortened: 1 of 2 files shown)


nvmbrasserie:
@thomasgdata Thank you for this, and for the write-up! Flagging the masked document numbers and the label drift up front saved me a lot of time. Both were the right calls.

I've cleaned it up and pushed two commits to your branch, keeping your structure.

The main issue was `fetch_html` that defaults to `absolute_links=False`, so the relative profile links raised `MissingSchema` on the first fetch and nothing got emitted.

That run surfaced two parsing bugs:
1. Long values wrap onto an unlabelled `<br/>` line, so occupations were cut off at "Company Director of"; continuation lines are now appended, which also recovers the dual nationalities. 
2. And multiple charges are `<br/>`-separated, so `element_text()` fused them into "AGENTCONSPIRACY TO DEFRAUD".

The details (particulars) now map to real properties through a lookup instead of "Label: value" strings in notes, with the labels enumerated so a new one warns. Masked numbers still aren't emitted. Placeholders moved to `type.*` lookups, clearing 47 warnings. My second commit is just a minor touch to the metadata.

Thanks again for contributing this.

nvmbrasserie (approved):

PR #5669, opened 2026-09-13
[us_dea_fugitives] Discover profiles through the official sitemap

jedediahg:
The DEA HTML listing intermittently returns Akamai challenges or Zyte website-ban errors even with browser rendering. Its official XML sitemap and individual profile pages remain accessible through Zyte HTTP requests.

This change discovers profiles from every partition advertised by the sitemap index, then calls the unchanged detail parser. Discovery completes before any profile is processed. It rejects invalid XML, DTDs, unexpected document shapes, unsafe URLs, duplicate partitions and implausible profile counts; failed responses are not cached. The dataset data URL now identifies the XML sitemap. URL-derived entity IDs, extracted properties and export assertions are unchanged.

Validation on final HEAD 33546e7e4:

- All 59 dataset tests pass, including 29 DEA regression cases.
- Strict mypy passes for all three Python files, with the declared development stubs supplied in isolated test storage. Ruff lint/format and git diff checks pass.
- Offline replay of the observed index and all nine partitions yields an exact match for 519 unique profile URLs.
- Full isolated live crawl and validated export: 519 Person entities, all 519 targets, 8,892 statements, result=success, zero issues; completed in 641.5 seconds.
- The live run started on da8a0e92f. The final commit adjusts only the DTD detection API and adds its test. A complete crawl and validated export on final HEAD then passed in 16.19 seconds using the populated cache with networking disabled. All entity IDs, schemata and property values match the live export (ignoring list ordering), and the exported source URL set exactly matches sitemap discovery.

The public listing also reports 519 results. That agreement is useful corroboration, but not a full listing-membership comparison. The sitemap index observed on September 13 is dated September 1; future sitemap freshness or membership drift should remain a monitoring concern.

GitHub reports the upstream build/lint workflows as action_required pending maintainer approval for this fork contribution. No deployment is included in this PR.

Contributor: codex-a1055e


nvmbrasserie:
hey @jedediahg! Thanks for the contribution. The sitemap approach is the right call.

A couple of notes on the cleanup: we don’t usually unit-test crawlers, so I dropped `test_crawler.py` and the `__init__.py` it needed. I also trimmed the crawler down a bit to fit our house style.

The main thing worth keeping is still there: discovery completes before any profiles are fetched. Appreciate the work on this!

PR #5670, opened 2026-09-13
INTERPOL: skip notices without a usable name

jedediahg:
The INTERPOL API can return a notice with both `forename` and `name` set to null. The crawler currently emits its nameless Person and linked Sanction, which downstream name-based screening cannot use.

Skip a notice when normal name handling leaves no `name`, before creating its sanction or emitting either record. Log the skip at info level. Notices with only a first or last name remain included. This also covers empty strings and the existing name-placeholder cleaning.

Add synthetic regression tests for nameless notices, single-name preservation, continued processing after a skipped notice, and linked sanction integrity. Strict typing also exposed two existing HTTP error handlers dereferencing an optional response; preserve and re-raise response-less HTTP errors, with regression coverage.

Validation: all 38 dataset tests passed, including 8 new INTERPOL cases; strict mypy on both changed files, Ruff lint/format, and diff checks passed. An offline replay of the observed null-name API response emitted neither Person nor Sanction and logged one skip. No real notice identifiers or personal data are included in the tests. The full live crawl was not rerun.

Contributor: codex-a1055e


nvmbrasserie:
@jedediahg thanks for catching this! it's a nice fix.

One note on the test file: we don't generally write unit tests for crawlers. They depend on live source data that changes shape without warning. We rely on running the crawler and reviewing `issues.log` and the emitted entities instead. That's why I've dropped` test_interpol_api.py`.

Thanks again for your contribution!

PR #5674, opened 2026-09-14
Document limitations of last_change

### LINKED ISSUE #5017: last_change is incorrect for deletions which don't empty the dataset

jbothma:
We decided just to document the limitations in the source code for now. Since we're planning improved change monitoring primitives, doing a lot of engineering on last_change as a change-monitoring mechanism doesn't make sense.

Closed https://github.com/opensanctions/opensanctions/issues/5017


pudo:
Nice, thanks :)

PR #5682, opened 2026-09-15
[us_ddtc_debarred] Record ITAR debarments under one program key

nvmbrasserie:
Statutory and administrative debarment are two entry paths into the same ITAR 22 CFR 127.7 regime, so both sheets now map to `US-AECA-DEBARRED`. The source-list distinction moves to the free-text `Sanction:program`.

`US-DDTC-SD` and `US-DDTC-AD` are deprecated with a header comment (the same pattern `US-NS-MBS` uses) and kept only so the keys stay resolvable while published statements age out.

Split out of #4981, which stalled once #5206 landed the OFAC half of the plan in it. cf. #4980

### Not yet verified

#4981 claimed a full crawl gave *1502 statements on the new key, 0 on the old*. I have **not** reproduced that, so it should be confirmed with a fresh `us_ddtc_debarred` crawl before merge — and the `assertions` block in the dataset yml may need a bump.

### Two judgement calls worth a look

- `US-AECA-DEBARRED`'s `dataset:` moves from `us_trade_csl` to `us_ddtc_debarred`. Both emit the key; `meta/README.md` says to pick the primary one, and the DDTC list is the authoritative source rather than the aggregator.
- `measures:` changes from `Export control` to `Arms restrictions`, per the README's boundary rule (ITAR covers military-list items, not dual-use). This also affects how the program displays for `us_trade_csl`.


nvmbrasserie:
added=0 dataset=us_ddtc_debarred deleted=0 metric=delta_counts modified=1499 version=20260921151552-efw

PR #5683, opened 2026-09-15
[wd_oligarchs] Add CAATSA Section 241 report as its own program

nvmbrasserie:
The CAATSA Section 241 report identifies senior Russian political figures, oligarchs and parastatal entities, but it designates nobody and imposes no measures — Treasury states explicitly that inclusion is not a determination that a person meets any designation criteria. Folding it into `US-CAATSA`, which covers OFAC's blocking designations, conflates the two.

This adds the informational `US-CAATSA-241` (no `measures`) and references it from `wd_oligarchs`.

Split out of #4981, which stalled once #5206 landed the OFAC half of the plan in it. cf. #4980

### Open question for review

Is a report that designates nobody a *program* at all? It fits the authority-shaped direction of #4980, but it stretches what a program key has meant so far, and it brushes against the unresolved `authorityId` / `ProfileId` point in that issue. This is the one of the three leftovers carrying a real design call — deliberately split out so it can't block the other two.

`US-CAATSA.yml` itself is untouched: main has since rewritten it independently, and it already points at `us_ofac_sdn`.


nvmbrasserie:
One design question this PR doesn't address, worth settling before merge: it collides with #3936.

`wd_oligarchs` attaches the program as `Thing:programId` on the Person, with no `Sanction` entity:

```python
entity.add("programId", CAATSA_PROGRAM)   # datasets/_wikidata/oligarchs/crawler.py:27
```

#3936 decided the opposite — where an entity carries a program and has no related `Sanction`, create the `Sanction` and put the program there, then "deprecate `Thing:program`" with a validator and audit. This PR adds a new instance of the pattern that issue exists to remove, and it would be the only one of its kind: the other two `programId` writers both attach to a `Sanction` (`datasets/us/ofac/ofac_advanced.py:650`, `datasets/be/fod_sanctions/crawler.py:101` — OFAC's `proxy.add` on the next line is a mirror alongside the `Sanction`, not a substitute).

The house pattern for a measure-less, informational program is to emit a `Sanction` anyway — the UA war-sanctions watchlists do `h.make_sanction(context, vessel, program_key="UA-WS-MARE")` (`datasets/ua/war_sanctions/crawler_zyte.py:385`), and those programs carry no `measures` either.

The argument the other way is specific to this source: Treasury states that inclusion in the § 241 report is **not** a determination that a person meets the designation criteria of any sanctions program. Emitting a `Sanction` entity arguably asserts something the source explicitly disclaims — which is the same reason `US-CAATSA-241` has no `measures`.

So: does an informational, non-designating listing get a `Sanction` entity for consistency with #3936 and the `UA-WS-*` precedent, or is `Thing:programId` the right representation and #3936's rule needs an explicit carve-out for this class? Either answer is fine for this PR; it just shouldn't merge by default without the call being made.

Unrelated to the above, for the record: `issuer: us_treasury` + `dataset: wd_oligarchs` (an OS compilation, `official: false`) is not an inconsistency — it matches `CN-CML`/`CN-UEL`/`CN-TAO-*` (`issuer: cn_npc` -> `dataset: cn_sanctions`) and the five `ICIJ-*` programs.

nvmbrasserie:
Resolved by `9fe2eab8`, using the third option: the listing is now a `Risk`, not a `Sanction` and not a bare `programId`.

`h.make_risk` landed in #5696, so `wd_oligarchs` emits a `Risk` per listed person. That keeps the program attribution on a record — with the authority, source and dates — while stating what the source states: FtM defines `Risk` as flagging heightened exposure or scrutiny, not a legally binding restriction. No `Sanction` is asserted, which is what Treasury's own disclaimer requires. The helper still mirrors `programId` onto the person, so nothing is lost for filtering.

Verified by a local crawl and export: 166 Persons, 93 Risks, export clean.

PR #5687, opened 2026-09-16
Document the version-tracked CSV curation pattern

### LINKED ISSUE #5650: [docs] Rube Goldberg / CSV format best practices
I think it's time to formalize this into `zavod/docs`. It's either one or two pieces of documentation, about the logic of Rube Goldberg crawlers, and about global conventions for version-tracked CSV files. Some notes: 

* All extracted content in the CSV files must either be metadata (schema, topic, etc.) or verbatim from the source. Data cleaning (including name cleaning) can happen in data patches. If this needs to be overridden, let's split into an `_original` and `_cleaned` column.
* Every Rube Goldberg crawler needs a removal mechanism and logic (see `cn_sanctions`). It's as important as detecting additions.
* Sometimes the agent actually writes a crawler/parser program as part of the agent run. Should we govern this? How? Should there be a threshold (like 8-10 rows)?

nvmbrasserie:
Add best_practices/curated_datasets.md for the pattern where a reviewed CSV in git is the record of truth: an agent proposes row changes as pull requests, a human reviews, and a crawler maps the file to FollowTheMoney.

The page covers the file and column conventions (verbatim cells, x_original/x_clean, one row per designation, origin, header naming, the ; delimiter, git tracking), how the file is updated as the source changes (discovery as its own step, removal mechanisms per source type, when an extraction justifies a parser program), and the two-commit migration of a source that currently lives in a Google Sheet.

Cut change_detection.md's "Separate discovery from extraction" section to a sentence and a link, so discovery is documented in one place.

Closes #5650


zavod/docs/best_practices/curated_datasets.md
```diff
@@ -0,0 +1,88 @@
...
+
+Curate when the source publishes designations only as prose — a ministry notice, an annex to a regulation, a PDF edition of a list — and there is no structured document to parse instead.
+
+Because the file is tracked in git, every change to the data carries an author, a date and a diff, and a reviewer sees which rows a pull request touches before it is merged.
+
+## File and column conventions
+
+### Every cell is verbatim or a metadata column
+
+The metadata columns are the ones extraction assigns: `type`, `topics`, the program key and the QID. Every other cell holds the source's own text, unmodified.
```
jbothma:
```suggestion
The metadata columns are the ones extraction assigns: e.g. `type`, `topics`, the program key and the QID. Every other cell holds the source's own text, unmodified.
```

zavod/docs/best_practices/curated_datasets.md
```diff
@@ -0,0 +1,88 @@
...
+
+Cleaning belongs downstream, in the crawler and in [datapatch lookups](datapatch_lookups.md), where one option corrects every row carrying the value and the correction re-runs on the next crawl. A value tidied up in the file applies to that row alone, fixed to whichever model extracted it, and nothing can re-derive it when the cleaning rules change.
```
jbothma:
```suggestion
Cleaning belongs downstream, in the crawler and in [datapatch lookups](datapatch_lookups.md) or name-cleaning Data Reviews, where one option corrects every row carrying the value and the correction re-runs on the next crawl. A value tidied up in the file applies to that row alone, fixed to whichever model extracted it, and nothing can re-derive it when the cleaning rules change.
```

zavod/docs/best_practices/curated_datasets.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Split a corrected column into `x_original` and `x_clean`
+
+When the printed value is not usable, add a second column instead of overwriting the first. `x_original` holds what the source prints and `x_clean` the correction; the crawler reads `_clean` and falls back to `_original`.
```
jbothma: what does "fall back to _original" mean? Do we copy every value to both columns except where _clean differs? 

Should we also have a x_notes column which is blank except where it explains to maintainers what was changed and why? We do have a common notes column with such explanations - maybe a single column is sufficient, instead of a per-field notes column.

zavod/docs/best_practices/curated_datasets.md
```diff
@@ -0,0 +1,88 @@
...
+
+Use the same two suffixes on every column that takes corrections, so a reviewer can tell which values came from the source without reading the crawler.
+
+On the emit side the crawler passes the correction as the value and the source text as `original_value=`, so the statement records both: `entity.add("name", clean_name, original_value=raw_name)`, as in `datasets/us/dod_chinese_milcorps/crawler.py`.
+
+### One row per designation, not per entity
+
+The same person designated twice under different measures, or re-listed after a removal, gets a row for each designation. Entities are merged downstream, while the rows retain the program, the dates and the source of each one.
+
+One row per entity has nowhere to record the date a single designation ended, so end dates and removals cannot be expressed at all.
```
jbothma, line 31: I would rephrase this a little - it isn't so much a hard rule as what turns out to be most practical.

"normally..."

It's easiest to make the crawler emit a single Sanction or Occupancy for a single designation or term of office when the curated CSV contains a single row for each designation or term of office. So even when the source is made up of notices, or "events", see if it's practical to maintain the curated CSV as a row per designation.

What we want to avoid is a row per notice/event, which means the crawler has to try and group all the events/updates using a common key, which can lead to duplicate entities if this doesn't work perfectly. This is also harder to review.

zavod/docs/best_practices/curated_datasets.md
```diff
@@ -0,0 +1,88 @@
...
+Discovery produces a review queue and does not write to the CSV. Deciding who is designated, under which program and from when requires reading the official document, so the write is a pull request. [Change detection](change_detection.md) covers the sentinel helpers to build the polling step from.
+
+### Every dataset needs a removal mechanism
+
+Detecting that a designation has ended is as important as detecting a new one. Which mechanism applies depends on what the source publishes.
+
+- **Consolidated sources** republish the complete current list, so a row vanishing from the source is the signal. Discovery compares the source's membership against the reviewed file; without that comparison the dataset records only additions and grows on every update.
+- **Event sources** publish a stream of notices, so the removal is its own notice, extracted like any other event. Absence from a later notice is not a removal, and neither is a URL that stops resolving.
+
+Until the mechanism exists, set the end date rather than deleting the row. Deleting discards the record of which notice designated the entity and when, which both the next extraction and the next review read.
```
jbothma: until which mechanism exists?

```suggestion
Set the end date rather than deleting the row. Deleting discards the record of which notice designated the entity and when, which both the next extraction and the next review read.
```

zavod/docs/best_practices/curated_datasets.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Write a parser program when the extraction recurs
+
+The default is no program: read the document, write the rows.
+
+Write one when the extraction recurs against the same document shape — each new edition of a report, each new version of a regulation. Recurrence is the trigger, not size: no row count makes a parser necessary, so a one-off transcription of a several-hundred-entry PDF gets none, while a twenty-row annex reissued every few months does.
+
+Keep the program next to the crawler and name it for what it parses, so the next edition is a change to reviewable code rather than a fresh extraction. `datasets/se/soe/extract_edition_2024.py` is one module per report edition.
```
jbothma: > so the next edition is a change to reviewable code rather than a fresh extraction

I'm not sure I understand what this says?

Also, something we could work into this section: The primary reviewed artifact is the diff on the curated CSV. The python code is kept for reuse and provenance, but in day-to-day pull request reviews we scrutinise the CSV, not the python code.

```suggestion
Keep the program next to the crawler and name it for what it parses, so the next edition is a change to reviewable code rather than a fresh extraction. `datasets/se/soe/extract_edition_2024.py` is one module per annual report edition.
```

i think it's worth mentioning annual, as in, it's not like we're making monthly python files.

jbothma (commented):
Generally looking great. I added some thoughts.

zavod/docs/best_practices/curated_datasets.md
```diff
@@ -0,0 +1,88 @@
...
+
+One row per entity has nowhere to record the date a single designation ended, so end dates and removals cannot be expressed at all.
+
+### Every row cites its official source
+
+The source URL is mandatory and never blank. A row without one cannot be checked against the document it was extracted from, and there is nothing to re-fetch when the extraction is questioned.
+
+For event sources the row carries the notice identity as well: the issuing authority and its decree or announcement number. The URL locates the notice on the current site, while the decree number still identifies it after a site redesign.
+
+### Name columns in snake_case, after the FtM property
```
pudo: Hmm, I'm a bit undecided on this. I wonder if the better policy would be to have the columns titled close to the source semantics (if they have `notes` under `explanation`, use `explanation`) and then have the crawler code do the mapping. That way the transcription doesn't need to onboard FtM semantics and can focus on creating a faithful repr of the source.

zavod/docs/best_practices/curated_datasets.md
```diff
@@ -0,0 +1,88 @@
...
+
+Use the FollowTheMoney property name wherever one exists, and descriptive snake_case English for columns that are not properties. `datasets/se/soe/leadership.csv` heads its columns `name`, `alias`, `position`, `company`, `report` and `source_url`.
+
+A column named after the property it populates states what the crawler does with the cell, and one spelling across datasets lets a reviewer read any curated file without first reading its crawler.
+
+### Separate multiple values with a semicolon
+
+Pack multiple values into one cell with `;` and split them in the crawler. The delimiter is the same in every curated dataset: a per-dataset delimiter means the splitting rule has to be read out of the crawler before a cell can be interpreted.
+
+### Force-add the file: `*.csv` is gitignored
```
pudo: Should we get rid of this?

zavod/docs/best_practices/curated_datasets.md
```diff
@@ -0,0 +1,88 @@
...
+
+Set the end date rather than deleting the row. Deleting discards the record of which notice designated the entity and when, which both the next extraction and the next review read.
+
+### Write a parser program when the extraction recurs
+
+The default is no program: read the document, write the rows.
+
+Write one when the extraction recurs against the same document shape — each new edition of a report, each new version of a regulation. Recurrence is the trigger, not size: no row count makes a parser necessary, so a one-off transcription of a several-hundred-entry PDF gets none, while a twenty-row annex reissued every few months does.
+
+Keep the program next to the crawler and name it for what it parses, so the next edition is a change to reviewable code rather than a fresh extraction. `datasets/se/soe/extract_edition_2024.py` is one module per annual report edition.
```
pudo: Worried about this - will it force the crawler into trying to build complex code instead of inference in sources where that is totally impossible? 

What's the best case here? I wonder if an alternative formulation could be "if scripts can be helpful in the extraction, you can create small, re-usable utility scripts that assist with extraction in the crawler directory"?

pudo (commented):
Nice and short. Some suggestions inline!

PR #5696, opened 2026-09-17
helpers: add make_risk for listings that impose no measures

nvmbrasserie:
Some sources name entities without designating them: advisory registers, watch lists of persons of interest, or reports that identify people while stating that inclusion is not a determination. Recording those with make_sanction asserts a restriction the source does not impose.

make_risk mirrors make_sanction but builds a Risk instead of a Sanction, so the program, authority, dates and source stay on a related object rather than only on the listed entity. It takes the same arguments and mirrors the resolved programId onto the entity in the same way.

make_sanction, is_active and lookup_sanction_program_key are unchanged.


pudo (approved):
Well, this is very complicated and totally trivial at the same time :)

pudo:
If you wanted to win prized, you could make the two `make_` functions call a common internal helper for the second half of their functionality

PR #5708, opened 2026-09-21
update topics/tags

brrttwrks:
Added topic:
  - reg.action
and dataset tags:
  - issuer.west
  - sector.financial
  - list.enforcement

to more accurately describe the dataset and the entities contained in the [BIS Antiboycott](https://www.opensanctions.org/datasets/us_bis_antiboycott/) list.

These changes are additive only and do not change the code, only the metadata.


pudo (approved):
Perfect, makes sense

PR #5711, opened 2026-09-21
Make the export CLI version optional

leonhandreke:
Make the `export` version optional, matching `crawl` and `run`.

- Collection: `export` mints a new version.
- Leaf: `export` falls back to the newest local crawl, with a warning. It errors if no local crawl exists.
- A version passed with `-v` keeps the previous behavior.

`crawl` and `export` are CLI development commands. Production always uses `run`, which mints one version and threads it through crawl, export and publish. So this fallback is a development convenience.

🤖 Generated with [Claude Code](https://claude.com/claude-code) using Claude Opus 4.8


pudo (approved):
Maybe can we shorten down the Claude comment in etl.py a bit?

leonhandreke:
Done, thanks for the review!

